PR #63, opened 2015-01-31
Struct default constructor

lstratman:
A struct's default constructor can now be called to create an object instance (via Activator.CreateInstance()) instead of having to be forced to implement an explicit constructor on the struct.


lstratman:
I'll be updating the tests to reflect this change and resubmitting the pull request.

Issue #82, opened 2015-07-09
exception message thrown from python side is not transferrred to .net 


denfromufa:
If I have exception thrown in Python, e.g.

```
raise LookupError('the message is not passed to CSHARP')
```

then on .NET side it only shows:

((Python.Runtime.PythonException)$exception).Message
"LookupError : "
string

denfromufa:
I found the problem in the code: IntPtr has to be converted to PyObject_Str() pointer instead of passing the original pointer to GetManagedString().

IntPtr _pyValueStr = PyObject_Str(_pyValue);
string message = Runtime.GetManagedString(_pyValueStr);

https://github.com/pythonnet/pythonnet/blob/master/src/runtime/pythonexception.cs#L38

I found the solution here:

http://stackoverflow.com/a/15907460/2230844

PR #84, opened 2015-07-16
exception message passed from Python to .NET correctly

denfromufa:
https://github.com/pythonnet/pythonnet/issues/82

### DIFF
--- a/src/runtime/pythonexception.cs
+++ b/src/runtime/pythonexception.cs
@@ -35,7 +35,9 @@ namespace Python.Runtime {
         if ((_pyType != IntPtr.Zero) && (_pyValue != IntPtr.Zero))
         {
             string type = new PyObject(_pyType).GetAttr("__name__").ToString();
-            string message = Runtime.GetManagedString(_pyValue);
+            IntPtr _PyValueStr = Runtime.PyObject_Str(_pyValue);
+            string message = Runtime.GetManagedString(_PyValueStr);
+            //string message = Runtime.GetManagedString(_pyValue);
             _message = type + " : " + message;
         }
         if (_pyTB != IntPtr.Zero)


tonyroberts:
Hi,

you need to be careful with reference counting when creating Python objects like this. Here you've create a new Python string object (_PyValueStr) but its reference count doesn't get decremented and so that object will never be deallocated.

If you use the PyObject .net class instead, in a using block, that will ensure it gets cleaned up correctly - for example the following should solve your problem without leaking any objects

```
        if ((_pyType != IntPtr.Zero) && (_pyValue != IntPtr.Zero))
        {
            using (PyObject pyType = new PyObject(_pyType))
            using (PyObject pyTypeName = pyType.GetAttr("__name__"))
            using (PyObject pyValue = new PyObject(_pyValue))
            {
                    string type = pyTypeName.ToString();
                    string value = pyValue.ToString();
                    _message = type + " : " + value;
            }
        }
```

The ToString method on PyObject does the string conversion, whereas Runtime.GetManagedString will only return a string of the Python object is a Python string.

Could be worth adding a unit test too, since now we have CI builds set up for most of the supported configurations.

cheers,
Tony

denfromufa:
Is this unit test going to?

https://github.com/pythonnet/pythonnet/blob/master/src/tests/test_exceptions.py

tonyroberts:
Hi,

yes that seems like a good place to put it,

cheers,
Tony

tonyroberts:
This has been implemented in https://github.com/pythonnet/pythonnet/pull/88, so I think this one can be closed now. Please feel free to re-open if there's still an issue.

thanks!
Tony

denfromufa:
Did the unit test get implemented too?

tonyroberts:
Nope :( If you have one left over from your earlier work please feel free to open another PR.

thanks,
Tony

PR #87, opened 2015-08-02
Method Default Arguments

fdanny:
In my local code I ran into a issue with default arguments. The code in the methodbinder.bind check if the arguments given is less than the arguments needed. If it is true check if there are any default arguments if so pull the default values.

The one exception is the indexer where setter is mapped to the arguments are passed along with the value. An example  T[k1, k2] = v would be passed down as a tuple like (k1,k2,v). I had to check for the default arguments in the mp_ass_subscript method before calling the SetItem. 

I've included simple unit tests that tests the default argument for method calls and indexers.


tonyroberts:
Looks good! Thanks

PR #88, opened 2015-08-12
access to Exception Message and expose TraceBack property

fdanny:


fdanny:
I added Runtime.Incref before the using statement and removed ;

tonyroberts:
Great, looks good. Thanks!

denfromufa:
why is this merged into develop branch, not master?

tonyroberts:
Hi,

new development should go into either a feature branch, or directly into the develop branch (as in this case). Work done on feature branches would usually be a more sizable chunk  of work than this PR and would be merged into the develop branch when complete. Currently we're aiming towards a 2.0 release (see the pre-release on pypi), which is based off what's in the develop branch now.

Once the 2.0 release is considered ready for release it will be merged into the master branch (either indirectly via a new release branch or directly from the develop branch*) and tagged as the 2.0 release. (*ideally for released we would create a release branch from develop and only integrate bugfixes for that release into that release branch before merging into the master branch, but since there are so few contributors/changes that step could be skipped).

Bugfixes to the 1.x releases would be done in a release branch (doesn't currently exist, but could be created). Since 2.0 is stable and in pre-release I would advise just using the develop branch for now.

I wouldn't usually expect direct changes to the master branch, only merges from other branches.

I hope this answers your question?

Have a look at this post for some more background about branching models in git.
http://nvie.com/posts/a-successful-git-branching-model/

Tony

PR #104, opened 2015-09-20
Added Python 3.5 Support

BrisingrAerowing:
I added preliminary Python 3.5 support. A quick test (getting some properties from the System.Environment class) worked perfectly fine on Windows 10 x64 with Python 3.5.0 amd64.

### DIFF
--- a/src/runtime/Python.Runtime.csproj
+++ b/src/runtime/Python.Runtime.csproj
@@ -9,7 +9,7 @@
     <AssemblyName>Python.Runtime</AssemblyName>
     <RootNamespace>Python.Runtime</RootNamespace>
     <SolutionDir Condition="$(SolutionDir) == '' Or $(SolutionDir) == '*Undefined*'">..\..\</SolutionDir>
-  	<PythonBuildDir Condition=" '$(PythonBuildDir)' == '' ">$(SolutionDir)</PythonBuildDir>
+    <PythonBuildDir Condition=" '$(PythonBuildDir)' == '' ">$(SolutionDir)</PythonBuildDir>
   </PropertyGroup>
   <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'ReleaseMono|x86'">
     <OutputPath>bin\x86\ReleaseMono\</OutputPath>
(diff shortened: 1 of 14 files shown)


src/runtime/Python.Runtime.csproj
```diff
@@ -51,6 +51,7 @@
     <PlatformTarget>x64</PlatformTarget>
     <CodeAnalysisIgnoreBuiltInRuleSets>false</CodeAnalysisIgnoreBuiltInRuleSets>
     <CodeAnalysisIgnoreBuiltInRules>true</CodeAnalysisIgnoreBuiltInRules>
+    <DefineConstants>PYTHON35, UCS2</DefineConstants>
```
tonyroberts: This needs to be guarded with a condition on DefineConstants not being already set. Otherwise the ReleaseWin|x64 target will always try and build the Python 3.5 build, regardless of the settings passed in via setup.py.

This line can probably just be removed (see line 47 just above here).

src/runtime/Python.Runtime.csproj
```diff
@@ -94,27 +95,27 @@
...
+        <Reference Include="Mono.Posix, Version=2.0.0.0, Culture=neutral, PublicKeyToken=0738eb9f132ed756, processorArchitecture=MSIL">
+          <SpecificVersion>False</SpecificVersion>
+          <HintPath>..\..\packages\MonoGAC\Mono.Posix\4.0.0.0__0738eb9f132ed756\Mono.Posix.dll</HintPath>
+        </Reference>
+      </ItemGroup>
+    </When>
+  </Choose>
+  <ItemGroup>
+    <Reference Include="System" />
+  </ItemGroup>
```
tonyroberts: Try to avoid just white space differences when making feature changes. It makes it harder to see what the real diffs are.

src/runtime/runtime.cs
```diff
@@ -173,6 +177,9 @@ public class Runtime {
 #if (PYTHON34)
         internal const string dllBase = "python3.4";
 #endif
+#if (PYTHON35)
+        internal const string dllBase = "python3.4";
```
tonyroberts: Should be python3.5

tonyroberts:
@BrisingrAerowing thanks very much for doing this. I've added a few comments to the code for you to look at. In addition, I think it would be good to add 3.5 to the travis and appveyor CI builds and run the unit tests before we integrate this.

filmor:
Why was this closed?

denfromufa:
How can this unknown branch get forked to continue Python 3.5 work?

Issue #98, opened 2015-09-08
ipython notebook & script difference - loading DLLs


denfromufa:
Did anyone notice the difference in loading .NET dlls between notebook and normal python scripts.

This works in notebook:

```
import clr
clr.AddReference(r"C:\path2dll\dotnetdll")
```

But in regular python scripts, the sys.path has to be appended with the path to DLL:

```
import clr
import sys
sys.path.append(r"C:\path2dll")
clr.AddReference("dotnetdll")
```

denfromufa:
We compared sys.path, .NET version, clr.ListAssemblies and everything seems to be seem. What is so special about the notebook?

denfromufa:
posted on http://stackoverflow.com/questions/32465864

denfromufa:
the problem is resolved, pull request pending.

in interactive mode python interpreter appends emtry string `""` to `sys.path`. As a side effect of this the buggy `AssemblyManager` was loading assemblies from full path only in interpreter mode.

https://github.com/pythonnet/pythonnet/pull/102

PR #106, opened 2015-10-12
resolve https://github.com/pythonnet/pythonnet/issues/98

fdanny:
resolve issue https://github.com/pythonnet/pythonnet/issues/98 based from previous pull request

### DIFF
--- a/src/runtime/assemblymanager.cs
+++ b/src/runtime/assemblymanager.cs
@@ -208,6 +208,24 @@ namespace Python.Runtime {
             return assembly;
         }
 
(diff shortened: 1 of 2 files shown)


fdanny:
I looked at the issue, it looks like it couldn't down load python-3.2.3.amd64.msi. Here is the error
 Exception calling "DownloadFile" with "2" argument(s): "Unable to connect to the remote server"

tonyroberts:
Looks like it's just appveyor timing out trying to download Python - probably nothing to worry about.

PR #114, opened 2015-12-22
Fix for 3dsMax Python

dgsantana:
Importing clr in 3dsMax python, causes the application to crash on
GetTypes from certain assemblies.

PR #115, opened 2016-01-07
Fix for 3dsMax Python

tonyroberts:
moved from #114


dgsantana:
So this is good for merging Tony?

tonyroberts:
@dgsantana yep, thanks for the PR.

Issue #131, opened 2016-02-04
Wrong overloading selected


denfromufa:
This bug is actually very bad:

```
>>> import System
>>> System.Math.Abs(42.5)
42L
>>> System.Math.Abs(42.55)
42L
```
submitted by @antocuni on:

http://sourceforge.net/p/pythonnet/bugs/5/

denfromufa:
the patch #137 works with following expressions in Python:

```
System.Math.Abs(-50.5) # 50.5
System.Math.Max(50.5,50.1) # 50.5
System.Math.Max(System.Double(10.5),System.Double(15.5)) # 15.5
System.Math.Max(System.Double(50.5),50.1) # 50.5
```

However it throws exception on these expressions allowed in .NET, which is still better than incorrect overload:

```
System.Math.Max(50.5,50)
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-8-c097754dae1b> in <module>()
----> 1 System.Math.Max(50.5,50)

TypeError: No method matches given arguments

System.Math.Max(System.Double(50.5),System.Int32(50))

---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-9-8a9fdf54eba5> in <module>()
----> 1 System.Math.Max(System.Double(50.5),System.Int32(50))

TypeError: No method matches given arguments

```

denfromufa:
@tonyroberts can we keep this issue open since some overloading cases throw exceptions, but should just work?

Also unit tests are completely missing for overloading.

PR #137, opened 2016-02-08
Update methodbinder.cs

denfromufa:
this fixes issue https://github.com/pythonnet/pythonnet/issues/131

I tested on System.Math.Abs(50.5) & System.Math.Max(50.5,50.1)

### DIFF
--- a/src/runtime/methodbinder.cs
+++ b/src/runtime/methodbinder.cs
@@ -291,15 +291,15 @@ namespace Python.Runtime {
                             else {
                                 op = Runtime.PyTuple_GetItem(args, n);
                             }
+
+                            // this logic below handles cases when multiple overloading methods
+                            // are ambiguous, hence comparison between Python and CLR types
+                            // is necessary
                             type = null;
-                            if  (_methods.Length>1) { 
-                                IntPtr pyoptype = IntPtr.Zero;
-                                //try {
-                                    pyoptype = Runtime.PyObject_Type(op);
-                                /*}
-                                catch (System.Exception) {
                             
-                                }*/
+                            if (_methods.Length>1) { 
+                                IntPtr pyoptype = IntPtr.Zero;
+                                pyoptype = Runtime.PyObject_Type(op);
                                 Exceptions.Clear();
                                 if (pyoptype != IntPtr.Zero) { }
                                     type = Converter.GetTypeByAlias(pyoptype);
@@ -309,7 +309,6 @@ namespace Python.Runtime {
 
                             if (type != null) {
                                 if (pi[n].ParameterType != type) {
-                                    //Exceptions.Clear();
                                     margs = null;
                                     break;
                                 }
--- a/src/runtime/methodbinder.cs
+++ b/src/runtime/methodbinder.cs
@@ -282,24 +282,32 @@ namespace Python.Runtime {
 
                     for (int n = 0; n < clrnargs; n++) {
                         IntPtr op;
-                        if (n < pynargs)
-                        {
-                            if (arrayStart == n)
-                            {
+                        if (n < pynargs) {
+                            if (arrayStart == n) {
                                 // map remaining Python arguments to a tuple since
                                 // the managed function accepts it - hopefully :]
                                 op = Runtime.PyTuple_GetSlice(args, arrayStart, pynargs);
                             }
-                            else
-                            {
+                            else {
                                 op = Runtime.PyTuple_GetItem(args, n);
                             }
-
-                            type = Converter.GetTypeByAlias(Runtime.PyObject_Type(op));
+                            type = null;
+                            if  (_methods.Length>1) { 
+                                IntPtr pyoptype = IntPtr.Zero;
+                                try {
+                                    pyoptype = Runtime.PyObject_Type(op);
+                                }
+                                catch (System.Exception e) {
+                            
+                                }
+                                if (pyoptype != IntPtr.Zero) { }
+                                    type = Converter.GetTypeByAlias(pyoptype);
+                                }
+                            //Runtime.Decref(pyoptype);
 
                             if (type != null) {
                                 if (pi[n].ParameterType != type) {
-                                    Exceptions.Clear();
+                                    //Exceptions.Clear();
                                     margs = null;
                                     break;
                                 }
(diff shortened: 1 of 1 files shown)


denfromufa:
@tonyroberts tests FAILED (errors=36), any idea how to fix this?

denfromufa:
@tonyroberts please let me know if you can merge this PR? I still need to write some tests for cases described in this issue #131.

src/runtime/methodbinder.cs
```diff
@@ -275,19 +282,40 @@ internal static MethodInfo MatchSignatureAndParameters(MethodInfo[] mi, Type[] g
...
+                                try {
+                                    pyoptype = Runtime.PyObject_Type(op);
+                                }
+                                catch (System.Exception e) {
+                            
+                                }
+                                if (pyoptype != IntPtr.Zero) { }
+                                    type = Converter.GetTypeByAlias(pyoptype);
+                                }
+                            //Runtime.Decref(pyoptype);
```
tonyroberts: Is pyoptype being leaked here? PyObject_Type returns a new reference.

src/runtime/methodbinder.cs
```diff
@@ -275,19 +282,40 @@ internal static MethodInfo MatchSignatureAndParameters(MethodInfo[] mi, Type[] g
...
                                 op = Runtime.PyTuple_GetItem(args, n);
                             }
-                            Type type = pi[n].ParameterType;
+                            type = null;
+                            if  (_methods.Length>1) { 
+                                IntPtr pyoptype = IntPtr.Zero;
+                                try {
+                                    pyoptype = Runtime.PyObject_Type(op);
+                                }
+                                catch (System.Exception e) {
```
tonyroberts: Does PyObject_Type ever throw an Exception? Doesn't it just return NULL on failure?

denfromufa: @tonyroberts yes, PyObject_Type can throw SystemError,  this is from Python docs:

PyObject* PyObject_Type(PyObject *o)
Return value: New reference.
When o is non-NULL, returns a type object corresponding to the object type of object o. On failure, raises SystemError and returns NULL. This is equivalent to the Python expression type(o). This function increments the reference count of the return value. There’s really no reason to use this function instead of the common expression o->ob_type, which returns a pointer of type PyTypeObject*, except when the incremented reference count is needed.

src/runtime/methodbinder.cs
denfromufa: @tonyroberts I'm not sure here, I actually wanted to ask you how to clean this up, especially if there is exception thrown? That's why I have `Runtime.Decref(pyoptype)` commented out :)

src/runtime/methodbinder.cs
tonyroberts: I think you misunderstand. When the python docs say that a function 'raises' an exception it means that the function returns NULL and the python exception is set (see PyErr_SetString/PyErr_SetObject). That's what Exceptions.Clear() is clearing below. It doesn't mean it throws a C# exception (and if you think about it, how could it?)

denfromufa: I thought that Python exceptions trigger C# exceptions due to pythonnet hooks, but was wrong.

src/runtime/methodbinder.cs
```diff
@@ -275,19 +282,42 @@ internal static MethodInfo MatchSignatureAndParameters(MethodInfo[] mi, Type[] g
...
                                 op = Runtime.PyTuple_GetItem(args, n);
                             }
-                            Type type = pi[n].ParameterType;
+                            type = null;
+                            if  (_methods.Length>1) { 
+                                IntPtr pyoptype = IntPtr.Zero;
+                                //try {
+                                    pyoptype = Runtime.PyObject_Type(op);
+                                /*}
+                                catch (System.Exception) {
```
tonyroberts: Please don't leave commented code - you commented it out because it was incorrect; it should be deleted instead.

src/runtime/methodbinder.cs
```diff
@@ -275,19 +282,42 @@ internal static MethodInfo MatchSignatureAndParameters(MethodInfo[] mi, Type[] g
...
+                                Exceptions.Clear();
+                                if (pyoptype != IntPtr.Zero) { }
+                                    type = Converter.GetTypeByAlias(pyoptype);
+                                    Runtime.Decref(pyoptype);
+                                }
+
+
+                            if (type != null) {
+                                if (pi[n].ParameterType != type) {
+                                    //Exceptions.Clear();
```
tonyroberts: Same here, please delete this line if it's not required.

src/runtime/methodbinder.cs
```diff
@@ -92,51 +95,54 @@ internal class MethodBinder {
...
-							 Type[] typeArgs = match.GetGenericArguments();
-							 return match.MakeGenericMethod(genericTp);
-						 }
-						 return match;
-					 }
-				 }
-			 }
-			 return null;
-		 }
+         //====================================================================
```
tonyroberts: I assume nothing's actually changed here? Just whitespace? It's easier to manage white space changes in their own PR (just for future reference, don't worry about changing it back)

tonyroberts:
Hi @denfromufa, I've just left a couple more comments. Once you've had a chance to take a look at those you should squash these commits together into a single commit (see git rebase -i).

thanks,
Tony

denfromufa:
@tonyroberts I use web interface in github without git - so here I'm using the option "Add more commits by pushing to the patch-2 branch on denfromufa/pythonnet":

https://github.com/denfromufa/pythonnet/tree/patch-2

I have a different source control locally.

src/runtime/methodbinder.cs
denfromufa: I installed VS Power Tools in between these commits, that's why whitespace/tab fixes.

src/runtime/methodbinder.cs
denfromufa: done, no more commented out code

src/runtime/methodbinder.cs
denfromufa: deleted

tonyroberts:
I've rebased the commits and updated the commit comment to something more meaningful. See PR #151.

I don't really have the time to fix up your pull requests, so please take more care in the future and learn how to use git rebase.

PR #153, opened 2016-02-17
python 3.5 support

denfromufa:
@tonyroberts @stonebig @vmuriart @sdpython 

1. correct setting for tp_compare, tp_reserved, and tp_as_async based on python version
2. one weird test case is disabled
3. extended and faster test matrix on appveyor

Issue #122, opened 2016-01-28
failing tests from PTVS GUI


denfromufa:
Test Name:  testImportModule
Test FullName:  test_engine.py::EngineTests::testImportModule
Test Source:    pythonnet\src\tests\test_engine.py : line 24

```
Python 2.7.9 |Continuum Analytics, Inc.| (default, Dec 12 2014, 14:56:54) [MSC v
.1500 32 bit (Intel)] on win32
Type "help", "copyright", "credits" or "license" for more information.
Anaconda is brought to you by Continuum Analytics.
Please check out: http://continuum.io/thanks and https://binstar.org
>>> import clr
>>> from Python.Runtime import PythonEngine
>>> m = PythonEngine.ImportModule("sys")

Unhandled Exception: System.Reflection.TargetInvocationException: Exception has
been thrown by the target of an invocation. ---> System.AccessViolationException
: Attempted to read or write protected memory. This is often an indication that
other memory is corrupt.
   at Python.Runtime.Runtime.PyImport_ImportModule(String name)
   at Python.Runtime.PythonEngine.ImportModule(String name)
   --- End of inner exception stack trace ---
   at System.RuntimeMethodHandle.InvokeMethod(Object target, Object[] arguments,
 Signature sig, Boolean constructor)
   at System.Reflection.RuntimeMethodInfo.UnsafeInvokeInternal(Object obj, Objec
t[] parameters, Object[] arguments)
   at System.Reflection.RuntimeMethodInfo.Invoke(Object obj, BindingFlags invoke
Attr, Binder binder, Object[] parameters, CultureInfo culture)
   at Python.Runtime.MethodBinder.Invoke(IntPtr inst, IntPtr args, IntPtr kw, Me
thodBase info, MethodInfo[] methodinfo)
   at Python.Runtime.MethodObject.Invoke(IntPtr target, IntPtr args, IntPtr kw,
MethodBase info)
   at Python.Runtime.MethodBinding.tp_call(IntPtr ob, IntPtr args, IntPtr kw)
```

denfromufa:
```
def testArrayContains(self): ##FAILING
        """Test array support for __contains__."""

        object = Test.Int32ArrayTest()
        items = object.items

        self.assertTrue(0 in items)
        self.assertTrue(1 in items)
        self.assertTrue(2 in items)
        self.assertTrue(3 in items)
        self.assertTrue(4 in items)

[...]

PR #154, opened 2016-02-17
Rebased

denfromufa:
rebased from https://github.com/pythonnet/pythonnet/pull/153

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,9 +3,9 @@ pythonnet
 
 Python for .NET is a package that gives Python programmers nearly seamless integration with the .NET Common Language Runtime (CLR) and provides a powerful application scripting tool for .NET developers. It allows Python code to interact with the CLR, and may also be used to embed Python into a .NET application.
 
-[![Build Status](https://travis-ci.org/pythonnet/pythonnet.png?branch=master)](https://travis-ci.org/pythonnet/pythonnet)
+[![Build Status](https://travis-ci.org/denfromufa/pythonnet.png?branch=develop)](https://travis-ci.org/denfromufa/pythonnet)
 
-[![Build status](https://ci.appveyor.com/api/projects/status/c8k0miljb3n1c7be/branch/master)](https://ci.appveyor.com/project/TonyRoberts/pythonnet-480xs)
+[![Build status](https://ci.appveyor.com/api/projects/status/c8k0miljb3n1c7be/branch/master)](https://ci.appveyor.com/project/denfromufa/pythonnet)
 
 **Calling .NET code from Python**
 
--- a/setup.py
+++ b/setup.py
@@ -2,6 +2,12 @@
 Setup script for building clr.pyd and dependencies using mono and into
 an egg or wheel.
 """
+#import ptvsd
+#ptvsd.enable_attach('pythonnet')
+#while not ptvsd.is_attached():
+#    from time import sleep
+#    sleep(1)
+
 from setuptools import setup, Extension
 from distutils.command.build_ext import build_ext
 from distutils.command.install_lib import install_lib
@@ -28,6 +34,18 @@ def _find_msbuild_tool(tool="msbuild.exe", use_windows_sdk=False):
         import winreg as _winreg
 
     if use_windows_sdk:
+        if sys.version_info[:2] == (2,7):
+            locappdir = os.environ["LOCALAPPDATA"]
+            vcpy27 = (r"Programs\Common\Microsoft"
+                r"\Visual C++ for Python\9.0\WinSDK\Bin")
+            if PLATFORM == "x86":
+                mtpath = os.path.join(
+                locappdir, vcpy27, r"mt.exe")
+            elif PLATFORM == "x64":
+                mtpath = os.path.join(
+                locappdir, vcpy27, r"x64\mt.exe")
+            if os.path.exists(mtpath):
+                return mtpath      
         value_name = "InstallationFolder"
         sdk_name = "Windows SDK"
         keys_to_check = [
@@ -41,6 +59,7 @@ def _find_msbuild_tool(tool="msbuild.exe", use_windows_sdk=False):
         value_name = "MSBuildToolsPath"
         sdk_name = "MSBuild"
         keys_to_check = [
+            r"SOFTWARE\Microsoft\MSBuild\ToolsVersions\14.0",
             r"SOFTWARE\Microsoft\MSBuild\ToolsVersions\12.0",
             r"SOFTWARE\Microsoft\MSBuild\ToolsVersions\4.0",
             r"SOFTWARE\Microsoft\MSBuild\ToolsVersions\3.5",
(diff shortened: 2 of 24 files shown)


README.md
```diff
@@ -3,9 +3,9 @@ pythonnet
 
 Python for .NET is a package that gives Python programmers nearly seamless integration with the .NET Common Language Runtime (CLR) and provides a powerful application scripting tool for .NET developers. It allows Python code to interact with the CLR, and may also be used to embed Python into a .NET application.
 
-[![Build Status](https://travis-ci.org/pythonnet/pythonnet.png?branch=master)](https://travis-ci.org/pythonnet/pythonnet)
+[![Build Status](https://travis-ci.org/denfromufa/pythonnet.png?branch=develop)](https://travis-ci.org/denfromufa/pythonnet)
```
filmor: Setting branch to `develop` might make sense, but if you're pushing to the main repo you should keep the reference as it is ...

tonyroberts: Agreed. I think we should merge into master and do a 2.1 release as soon as
the 3.5 stuff is all working and tested, so the CI build statuses should be
left as they were.

denfromufa: @tonyroberts you can see on my branch that py35 is working and tested

filmor: When this is merged the same will hold for the original links :)

tonyroberts:
@denfromufa there are too many closely related commits that haven't been squashed, and a lot of commits with no meaningful commit message. I am also suspicious of the test that you have commented out, but I have not had time yet to check whether it was a valid test or not.

tonyroberts:
As I see it, there should be at most 4 commits.
 - two from sdpython's branch
 - one for your changes to interop.cs / assemblymanager.cs (with a meaningful commit message)
 - one for the changes to the travis and appveyor files

Everything else will drop out once this is rebased correctly.

setup.py
```diff
@@ -40,6 +40,18 @@ def _find_msbuild_tool(tool="msbuild.exe", use_windows_sdk=False):
                 locappdir, vcpy27, r"x64\mt.exe")
             if os.path.exists(mtpath):
                 return mtpath      
+        if sys.version_info[:2] == (3,5):
```
tonyroberts: There shouldn't be any need for hardcoded paths. Not everyone will have installed the windows SDK on the C drive.

src/runtime/interop.cs
```diff
@@ -326,37 +343,57 @@ internal class TypeOffset {
...
+#if (PYTHON35)
+        public static int was_sq_ass_slice = 0;
+#else
         public static int sq_ass_slice = 0;
+#endif
         public static int sq_contains = 0;
         /* Added in release 2.0 */
         public static int sq_inplace_concat = 0;
         public static int sq_inplace_repeat = 0;
-//} PySequenceMethods;
```
tonyroberts: Mapping functions come before sequence functions in the type struct don't they? Looks like this change has flipped the order. What's the reason for that?

src/tests/test_array.py
```diff
@@ -143,7 +143,7 @@ def testArrayContains(self):
 
         self.assertFalse(5 in items)    # "H:\Python27\Lib\unittest\case.py", line 592, in deprecated_func,
         self.assertFalse(-1 in items)   #TypeError: int() argument must be a string or a number, not 'NoneType'
-        self.assertFalse(None in items) # which threw ^ here which is a little odd.
```
tonyroberts: What's changed with the result of this test? This test seems valid and so should still pass as far as I can tell.

I can see that someone would argue that it should raise a TypeError. If that's what's happening we either need to a) preserve the existing behaviour so as not to break existing code or b) make a decision to change it and update the test to reflect that.

setup.py
denfromufa: @tonyroberts I agree, but I did not find in registry this path. also not sure if windows sdk can be installed to non-default paths?! maybe @zooba can point how this path can be mapped in more generalized way? (r"C:\Program Files (x86)\Windows Kits\8.1\bin")

src/runtime/interop.cs
denfromufa: @tonyroberts according to C-API for Python 2.7 & 3.5 mapping functions come after sequence functions:

https://docs.python.org/3/c-api/typeobj.html
https://docs.python.org/2/c-api/typeobj.html

```c
/* Method suites for standard classes */

    PyNumberMethods *tp_as_number;
    PySequenceMethods *tp_as_sequence;
    PyMappingMethods *tp_as_mapping;
```

tonyroberts: That's not how the *object* is laid out in memory though, I think. Check object.h in the python source code.

https://github.com/python/cpython/blob/3e91c558fed0e803369f8d2324f102a042cf04e3/Include/object.h

Take a look around line 465.

Easiest thing would be to check the value in a debugger both on the C side and the C# side (you would need to have a debug build of Python and enable native code debugging in order to do that).

However, given that there were no reported problems previously, I would leave it as it was, unless you have a really good understanding of this code and can demonstrate that it was previously incorrect.

tonyroberts: p.s. apologies that my first comment was misleading as I said the type struct - what actually matters here is the layout of the actual object, not the type.

denfromufa: @tonyroberts now I agree with you. Note that I did not make this change, most likely this is from @sdpython, but it slipped through my review. this is quite surprising that the order is switched intentionally in cpython due to conflicting situations, according to the comment in cpython source code:

https://github.com/python/cpython/blob/3e91c558fed0e803369f8d2324f102a042cf04e3/Include/object.h#L465

src/tests/test_array.py
denfromufa: looks like this test used to break at some point. I also broke it in PTVS before these merges:

https://github.com/pythonnet/pythonnet/issues/122

denfromufa:
@tonyroberts @vmuriart @filmor I'm quite busy with my day job until end of this week. if anyone wants to rebase properly or get a clean fork, please go ahead before that, even if it does not include my commits. I also like @tonyroberts idea testing with debug build of CPython, especially since you can step through C#/C/Python code in PTVS if debugging is needed!

vmuriart:
@denfromufa @tonyroberts I went ahead and cleaned up the git commits and squashed them into more unit like commits on [my fork](https://github.com/vmuriart/pythonnet/commits/py35). I hate to have to close this PR to open a new one again... Selavi. 

@tonyroberts i consolidated some of the strange changes of code (moving the comments you mention, some debug statements). I tried troubleshooting that commented out test as well but haven't been able to get it passing. The error that we get is the same one that was pointed out in the comment; it fails with `int` not taking `noneType`

src/runtime/interop.cs
vmuriart: fixed on rebase

setup.py
vmuriart: removing from rebase. agree with @tonyroberts 
@denfromufa may need to add it to the registry.

vmuriart:
pr #160 is cleaned up and addresses the comments above.

setup.py
zooba: `HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows Kits\Installed Roots` is where you'll find the install directories of the Windows dev kits. (In 64-bit processes on 64-bit systems it'll be `HKEY_LOCAL_MACHINE\SOFTWARE\WOW6432Node\Microsoft\Windows Kits\Installed Roots` - easiest to try the other one first and then fall back to the longer path.)

PR #160, opened 2016-02-23
Python 3.5

vmuriart:


tonyroberts:
The more i think about this the more i think this is the wrong approach. Ifdef-ing for each python version is messy and error-prone. It should be simple enough to parse the python header files and generate this automatically as part of the build.

I'll put something together when I get a chance, but for now let's put trying to integrate this 3.5 branch on hold - at least until I've investigated the alternatives.

vmuriart:
@tonyroberts, had similar thoughts on the matter. `PyWin32` has the same issue from what I've worked on so far on it. Though they simple specify a minimum required version for most of their code instead of specifying each version. 

Would it be possible to address this as part of post 2.1 release? That is to merge this branch, and release 2.1 with all the available changes. Post 2.1 could focus on leaning out the code and speed improvement.

denfromufa:
There is swig way, and c++/cli + cppsharp. @tonyroberts are you thinking in
this direction?

I tried the latter with calling into sympy and got flooded with memory
leaks, if anyone is interested I can share the source code.

There is also cython and cffi if we wanted to build the bridge on python
side.

vmuriart:
@denfromufa I would...

src/runtime/interop.cs
```diff
@@ -348,15 +361,23 @@ internal class TypeOffset {
...
         public static int bf_getsegcount = 0;
         public static int bf_getcharbuffer = 0;
 #endif
-#if (PYTHON26 || PYTHON27 || PYTHON32 || PYTHON33 || PYTHON34)
+#if (PYTHON35)
+        //typedef struct {
+        public static int am_await = 0;
+        public static int am_aiter = 0;
+        public static int am_anext = 0;
+        // } PyAsyncMethods
```
tonyroberts: these are in the wrong place. They should be between ht_type and as_number.

tonyroberts:
@vmuriart @denfromufa I'm sorry but at the moment I have very low confidence in this PR. I've spotted a few flaws just from scanning through it, without really properly comparing it to the 3.5 memory layout. It may well work just because the few bits of it we use happen to be in the right place by virtue of the right number of items being added, but that's not really good enough. Details like this are very important to get right, otherwise there will be countless random crashes later that will be almost impossible to track down.

vmuriart:
@tonyroberts no problem. what are you thinking as far as next steps?

tonyroberts:
@vmuriart I'll check some code in later today or tomorrow.

PR #161, opened 2016-02-23
Generate the TypeOffset class automatically.

tonyroberts:
Rather than manually tweak the TypeOffset class for each Python release it's easier and more robust to generate it directly from the Python headers.

This will make the 3.5 upgrade significantly easier.


tonyroberts:
@filmor ah ok, that makes sense. I had done it so it's conditionally included, but I see what you're saying - having to define yet another thing in visual studio could be nicely avoided as you suggest.

tonyroberts:
@filmor hmm the only downside to that is when building with different python configs (e.g. debug or whatever) the generated file names are different as they include the ABI flags... maybe a conditional include will still be needed for those, but for the most usual cases (i.e. the ones I checked in pre-generated files for) they can all be included and #if/defed out.

Issue #157, opened 2016-02-18
Unable to use Generics in CPython with Python.NET 


curzona:
When I use the subscript syntax from http://pythonnet.sourceforge.net/readme.html#generics I get the following error: `TypeError: unsubscriptable object`

With Python 2.7.11 + pythonnet==2.1.0.dev1
```
>python.exe
Python 2.7.11 (v2.7.11:6d1b6a68f775, Dec  5 2015, 20:32:19) [MSC v.1500 32 bit (Intel)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> import clr
>>> from System import EventHandler
>>> from System import EventArgs
>>> EventHandler[EventArgs]
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: unsubscriptable object
```

I've also tried pythonnet==2.0.0 and building form github ca15fe8 with ReleaseWin x86 and got the same error.

With IronPython-2.7.5 subscript syntax works:
```
>ipy.exe
IronPython 2.7.5 (2.7.5.0) on .NET 4.0.30319.0 (32-bit)
Type "help", "copyright", "credits" or "license" for more information.
>>> import clr
>>> from System import EventHandler
>>> from System import EventArgs
>>> EventHandler[EventArgs]
<type 'EventHandler[EventArgs]'>
```

denfromufa:
pull requests are welcome :) https://github.com/pythonnet/pythonnet/search?l=csharp&q=unsubscriptable+object&utf8=%E2%9C%93

In both cases I see exceptions, so do you expect to see exactly same message?

curzona:
@denfromufa I've updated my examples.

With IronPython the subscript syntax instantiates the EventHandler as expected but with CPython+Python.NET it returns an error.

denfromufa:
@curzona now I see the problem, this looks like just not implemented in pythonnet, not sure if this workaround would work for your case:

http://stackoverflow.com/questions/8668083/custom-eventhandler-vs-eventhandlereventargs

filmor:
It's definitely implemented, check for example the classes from `System.Collections.Generic`. I guess the problem here is that there are both a generic `EventHandler<ArgsType>` and a non-generic class `EventHandler`, which Python.NET doesn't resolve properly.

You can get the generic class by using its explicit name:

    EventHandler = getattr(System, 'EventHandler`1')
    EventHandler[EventArgs]

denfromufa:
[...]

Issue #128, opened 2016-01-31
Action delegate - type overload resolution not implemented


denfromufa:
https://mail.python.org/pipermail/pythondotnet/2015-July/001663.html

denfromufa:
http://stackoverflow.com/questions/30659933/python3-pythonnet-generic-delegates

filmor:
Could you please formulate this issue out? This is btw the exact same issue as #157, the actual generic classes have a slightly different name as there doesn't seem to be a "type overload resolution" implemented in the same manner as it is done for methods. Just use

    Action = getattr(System, 'Action`1')

tonyroberts:
Fixed in #164

PR #164, opened 2016-02-24
Move the type subscript implementation into ClassBase.

filmor:
This fixes issues #157 and #128 by allowing the use of the type
subscript operator on all types instead of just generic types,
preventing the shadowing of generic types by a non-generic one like
System.Action or System.EventHandler.


filmor:
Note, that the Travis error is not related to this particular PR: https://travis-ci.org/pythonnet/pythonnet/jobs/111424039

tonyroberts:
@filmor thanks, I've restarted the travis builds - sometimes the nuget restore times out and running again should fix it.

PR #167, opened 2016-02-26
[WIP] Implement __instancecheck__.

filmor:
This implements the `__instancecheck__` method on the CLR metaclass which results in `isinstance` working for interfaces. I'm not completely sure whether everything still works and this is missing the `__subclasscheck__` to make `issubclass` work accordingly.

Also, I'd like to simplify the method def generation in `MethodWrapper` and this particular implementation as a prerequisite to implementing `__enter__` and `__exit__` for `IDisposable` objects.

### DIFF
--- a/src/runtime/converter.cs
+++ b/src/runtime/converter.cs
@@ -80,6 +80,10 @@ namespace Python.Runtime {
         // This always returns a new reference. Note that the System.Decimal
         // type has no Python equivalent and converts to a managed instance.
         //====================================================================
+        internal static IntPtr ToPython<T>(T value)
+        {
+            return ToPython(value, typeof(T));
+        }
 
         internal static IntPtr ToPython(Object value, Type type) {
             IntPtr result = IntPtr.Zero;
--- a/src/runtime/importhook.cs
+++ b/src/runtime/importhook.cs
@@ -45,8 +45,8 @@ namespace Python.Runtime {
             IntPtr mod = Runtime.PyDict_GetItemString(dict, "__builtin__");
             py_import = Runtime.PyObject_GetAttrString(mod, "__import__");
 #endif
-              hook = new MethodWrapper(typeof(ImportHook), "__import__");
-              Runtime.PyObject_SetAttrString(mod, "__import__", hook.ptr);
+            hook = new MethodWrapper(typeof(ImportHook), "__import__", "TernaryFunc");
+            Runtime.PyObject_SetAttrString(mod, "__import__", hook.ptr);
             Runtime.Decref(hook.ptr);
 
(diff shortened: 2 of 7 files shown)


filmor:
Apparently the `testExceptionIsInstanceOfSystemObject` test fails. Why is this so? Do we actually want this or is this just an artifact of the "C# Exceptions -> Python Exceptions" mapping?

tonyroberts:
I'm not 100% sure, but it seems like it could be because Exceptions are a special case because they weren't new style classes until Python 2.5 (I think?) and so having them inherit from a new style class (like System.Object) would cause problems. This definitely isn't an issue in Python 3 (and the code for wrapping Exceptions is a lot simpler), but I guess it still could be an issue for Python 2.4 and below.

It might be worth trying Python 2.4 or 2.3 to see if this change breaks anything there (even though I know we don't have CI builds or distribute wheels for older versions of Python, but if we can maintain backwards compatibility without too much trouble then it would be nice to).

filmor:
Is there a specific reason that the Marshal structs aren't used directly but only to get the correct offsets?

tonyroberts:
@filmor I don't know I'm afraid. This code is pretty old, so possibly that was the only way to do it at the time it was originally written.

tonyroberts:
@filmor wrt writing the whole null struct, I was just thinking about how it would look if you were to write it in C, i.e.

```
    PyMethodDef methods[] = {
        {name, func, flags, doc},
        {NULL}
    }
````

I don't think it explicitly says which member is checked for null-ness in the python docs. The way you had it is probably fine, I was just thinking about how most extensions look and what potential problems could be caused by having is slightly different.

filmor:
I think you are right there and I've already changed that. I don't know why, but for some reason I though it looked like `PyMethodDef methods[] = {something, NULL};` which is obviously wrong ;)

Now, if we can decide on what to do with the exception test I'd implement `__subclasscheck__` in the same way, will add a few tests and leave the "make simple methods nice"-part for a separate PR.

tonyroberts:
@filmor for the unit test, an option would be to check that Exception *is* an instance of Object for Python >= 2.6 (since that's when __instancecheck__ was added) and leave it as it is for previous versions of Python. That would check the isinstance change is working correctly, and if anyone ever did want to use an earlier versions of Python then the previous test would still be correct (and this change won't break anything in Python 2.3 since it's not actually changing the base class of anything).

filmor:
Now, this is only missing a few tests for this specific behaviour, but apart from that it would be good if you could review the changes.

tonyroberts:
@filmor looks good to me, thanks!

PR #168, opened 2016-02-27
include windows kits path from VS 2015 to search for mt.exe

denfromufa:


tonyroberts:
Is there something special about this version of mt.exe that means it can only be used with Python 3.5?

denfromufa:
Actually I have not tested this mt.exe with Python versions other than Python 3.5. Not really sure if it is version independent, but it comes with VS 2015 installation and my understanding is that Python 3.5 compiles with VS 2015.

tonyroberts:
Your understanding is only partially correct. Python can be built with pretty much any C compiler, it just happens that the windows maintainers choose to use VS 2015 as the default compiler and the one they use to build the version of Python they release. Think about it, if what you're saying was true why would the currently 3.5 version of pythonnet be working at all when it's built using the same version of visual studio used to build it for all other python versions?

As a general rule, I really don't like this sort of messy versions dependant code where it can be avoided, and I think in this case it probably can be.

I just had a look at this file, and it seems the way the 2.7 tools have been done is pretty awful too. I guess I will have to spend some time sorting this out. Adding more and more code isn't the answer, and so I won't be accepting this PR as it stands.

denfromufa:
@zooba, is there better way to detect possible paths for mt.exe, is there
anything in python standard library to do this?

Also is mt.exe for special PY27 compiler can be found in registry?

tonyroberts:
@denfromufa nothing wrong with where you were getting the path from in the registry, just with how it was being done. This setup.py is complicated enough already, and adding new cases for every possible combination of python version and windows kit/sdk that might or might not be installed makes the code much harder to read and maintain. See PR #172 for a different approach that takes advantage of the fact that any mt.exe may be used (they all write exactly the same COFF file format) to make the code less complex.

Issue #103, opened 2015-09-17
Python 3.5 support


fdanny:
Python 3.5 support pythonnet. It supports up to 3.4.

BrisingrAerowing:
I'd like this as well.

BrisingrAerowing:
I just submitted a pull request with initial Python 3.5 support.

fdanny:
Did the pull request https://github.com/pythonnet/pythonnet/pull/104 go thru? Can we close this issue?

vmuriart:
The pull request seems to have been closed without merging.

mahibra:
Any news on the python 3.5 support? I would love to use Python.NET with the current Python version.

denfromufa:
Very weird - that pull request points to an unknown branch

denfromufa:
There is a blog post about 3.5:
http://www.xavierdupre.fr/app/pymyinstall/helpsphinx/blog/2016/2016-01-03_pythonnet.html

denfromufa:
Looks like this branch can be merged:

https://github.com/sdpython/pythonnet3

vmuriart:
From reading the blog it seems that it still has a few issues to resolve

vmuriart:
@denfromufa I'd like to withdraw my earlier statement, I reread the blog and looks like I misunderstood the issue. He had a built issue that he mentioned. 

I took a look at the [src](https://github.com/sdpython/pythonnet/commit/5484084d9732a3e72ca65874a9735a89f9cc048a) and compared it to the earlier [pr](https://github.com/pythonnet/pythonnet/pull/104/files) that was submitted. sdpython's looks alot more complete as it integrates the new 3.5 features. 

@tonyroberts any reason this can't be merged (other than no existing pull request :smile_cat: )

tonyroberts:
@vmuriart would be great if you could test and create a PR. Please make sure you add 3.5 builds to the travis and appveyor CI builds.

denfromufa:
@fdanny @mahibra @BrisingrAerowing try this branch until it is merged: 

https://github.com/denfromufa/pythonnet/tree/develop

only one very weird test case is failing on travis ci, hence I commented out it for now.

appveyor needs some updates for python 3.5, but I tested locally on my windows 7 machine.

mahibra:
Hi,
I just tried to install pythonnet from your develop branch. I used Anaconda
2.5 (Python 3.5.1) and Visual Studio Express 2015.

[...]

PR #170, opened 2016-02-27
overloading - more fixes for wordpad.py

denfromufa:
System.Drawing.Font("Tahoma", 10.0) supposed to use Single, which does not directly map to Python. Hence I added comparison in both directions for Double/Single, Int32/Int16.

This fixes issue observed by @mahibra that I reproduced today:

https://github.com/pythonnet/pythonnet/issues/103#issuecomment-185384611

### DIFF
--- a/src/runtime/methodbinder.cs
+++ b/src/runtime/methodbinder.cs
@@ -245,7 +245,7 @@ namespace Python.Runtime {
             else {
                 _methods = GetMethods();
             }
-            Type type;
+            Type clrtype;
             for (int i = 0; i < _methods.Length; i++) {
                 MethodBase mi = _methods[i];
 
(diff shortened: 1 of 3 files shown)


src/runtime/methodbinder.cs
```diff
@@ -295,34 +295,53 @@ internal static MethodInfo MatchSignatureAndParameters(MethodInfo[] mi, Type[] g
...
 
 
-                            if (type != null) {
-                                if (pi[n].ParameterType != type) {
-                                    margs = null;
-                                    break;
+                            if (clrtype != null) {
+                                if (pi[n].ParameterType != clrtype) {
+                                    IntPtr pytype = Converter.GetPythonTypeByAlias(pi[n].ParameterType);
+                                    // does pytype require Decref?
```
tonyroberts: Why is this a question? it's your code, you should understand it. Comments should help the people that come after you to understand the code, not raise questions about its correctness.

tonyroberts:
See contributing.md about grouping commits into logical units (which you've been told about numerous times already).

I think this change warrants a new unit test to demonstrate the exact problem. In the meantime, if a user is having a problem a simply workaround is to construct the exact C# type an call the method with that (not that I'm against having this fixed in the main code).

mahibra:
@denfromufa: With the modified version I am able to run the wordpad.py example. However, the commercial API still crashes when calling the method ReadNextResult with the following arguments:
```
batchRayTrace.ReadNextResult((0, 0, 0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0))
  File "C:\Anaconda3\lib\site-packages\IPython\core\interactiveshell.py", line 3066, in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)
  File "<ipython-input-6-41428a5b8fa5>", line 1, in <module>
    batchRayTrace.ReadNextResult((0, 0, 0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0))
TypeError: No method matches given arguments
```
with
```
batchRayTrace.ReadNextResult?
Type:        MethodBinding
String form: <bound method 'ReadNextResult'>
File:        c:\anaconda3\lib\site-packages\clr.pyd
Docstring:
Boolean ReadNextResult(Int32 ByRef, Int32 ByRef, Int32 ByRef, Int32 ByRef)
Boolean ReadNextResult(Int32 ByRef, Int32 ByRef, Double ByRef, Double ByRef, Double ByRef, Double ByRef, Double ByRef, Double ByRef, Double ByRef)
Boolean ReadNextResult(Int32 ByRef, Int32 ByRef, Int32 ByRef, Double ByRef, Double ByRef, Double ByRef, Double ByRef, Double ByRef, Double ByRef, Double ByRef)
Boolean ReadNextResult(Int32 ByRef, Int32 ByRef, Int32 ByRef, Double ByRef, Double ByRef, Double ByRef, Double ByRef, Double ByRef, Double ByRef, Double ByRef, Double ByRef, Double ByRef, Double ByRef)
Boolean ReadNextResult(Int32 ByRef, Int32 ByRef, Int32 ByRef, Double ByRef, Double ByRef, Double ByRef, Double ByRef, Double ByRef, Double ByRef, Double ByRef, Double ByRef, Double ByRef, Double ByRef, Double ByRef) 
```
However, it works with the methodbinder.cs version prior to patch [#151](https://github.com/pythonnet/pythonnet/pull/151)

denfromufa:
@mahibra I found out the problem last night while writing tests for improved overloading resolution in pythonnet, but have not finished my improved patch yet. Note that before it was working by accident, i.e. no type checking was done which could give you wrong results like this:

https://github.com/pythonnet/pythonnet/issues/131

The problem stems from Python 2 to 3 change of `int` to always `long` implementation and no equivalent in .NET among Int16, Int32, Int64. Right now Int64 maps to Python Long, but your overloaded methods require Int32. So try explicit types such as Int32(0) and let us know if it works that way.

mahibra:
The following command works:
```
from System import Int32, Double
batchRayTrace.ReadNextResult(Int32(0), Int32(0), Int32(0), Double(0.0), Double(0.0), Double(0.0), Double(0.0), Double(0.0), Double(0.0), Double(0.0), Double(0.0), Double(0.0), Double(0.0), Double(0.0))
```
I have to convert both the Int and the Double values. Otherwise I get the error
`TypeError: No method matches given arguments`

denfromufa:
closing this in favor of https://github.com/pythonnet/pythonnet/pull/175

PR #171, opened 2016-02-29
Include all interop files in solution.

tonyroberts:
Use defines to ensure only the relevant code is included, but add
all files to the project when PythonInteropFile isn't defined.

When building from setup.py explicitly use the interop file that
matches the python.exe in use.


denfromufa:
@tonyroberts I looked more closer at abi flags you mentioned. So now in VS solution, defining compilation symbol Py_Debug is not supported anymore? geninterop.py has to be run outside of VS?

PR #175, opened 2016-03-03
fixes overloading of integer types which is py2/3 version dependent

denfromufa:
…added tests for checking overloading behavior for main value types


tonyroberts:
@denfromufa looks good!

PR #176, opened 2016-03-07
method overloading fix for enum argument

denfromufa:


src/runtime/methodbinder.cs
```diff
@@ -309,26 +309,38 @@ internal static MethodInfo MatchSignatureAndParameters(MethodInfo[] mi, Type[] g
...
                                         }
                                     }
-                                    else {
-                                        Runtime.Decref(pyoptype);
+                                    if (!typematch) {
+                                        // this takes care of enum values
+                                        TypeCode argtypecode = Type.GetTypeCode(pi[n].ParameterType);
+                                        Type typearg = Type.GetType("System." + argtypecode.ToString());
+                                        if (typearg == clrtype) {
+                                            typematch = true;
```
tonyroberts: Would comparing the type code of the parameter type and the clrtype be more appropriate here?

denfromufa: @tonyroberts done! thanks for pointing out cleaner solution. I also added this "regression" test.

tonyroberts:
@denfromufa great, thanks.

stonebig:
a "dev2" release this wek-end ?

PR #182, opened 2016-03-09
Fix crash "Name must not be empty!"

gatapia:
Not sure why PythonNet was getting this exception (only on production server not dev machine).  Running my module with `python -m mymodule` would work fine.  Anyways, I debugged a bit and the attached change fixed my issue.

Please note this is a quick hack, just to keep my project going and it worked.  I have no idea of the underlying cause and do not want to investigate but if you think my change is not risky you may want to merge as it does fix this issue (or at least hides the issue).

A quick [Google](https://www.google.com.au/search?num=100&q="Name+must+not+be+empty%21"+PythonNet) shows that I'm not the first to get this exception.

Exception Details:
```
System.ArgumentException: Name must not be empty!
   at Python.Runtime.ModuleObject..ctor(String name) in c:\dev\libs\pythonnet\src\runtime\moduleobject.cs:line 36
   at Python.Runtime.ModuleObject.GetAttribute(String name, Boolean guess) in c:\dev\libs\pythonnet\src\runtime\moduleobject.cs:line 104
   at Python.Runtime.ImportHook.__import__(IntPtr self, IntPtr args, IntPtr kw) in c:\dev\libs\pythonnet\src\runtime\importhook.cs:line 301
   at Python.Runtime.Runtime.PyObject_Call(IntPtr pointer, IntPtr args, IntPtr kw)
   at Python.Runtime.ImportHook.__import__(IntPtr self, IntPtr args, IntPtr kw) in c:\dev\libs\pythonnet\src\runtime\importhook.cs:line 230
   at Python.Runtime.Runtime.PyObject_Call(IntPtr pointer, IntPtr args, IntPtr kw)
   at Python.Runtime.ImportHook.__import__(IntPtr self, IntPtr args, IntPtr kw) in c:\dev\libs\pythonnet\src\runtime\importhook.cs:line 230
   at Python.Runtime.Runtime.PyObject_Call(IntPtr pointer, IntPtr args, IntPtr kw)
   at Python.Runtime.ImportHook.__import__(IntPtr self, IntPtr args, IntPtr kw) in c:\dev\libs\pythonnet\src\runtime\importhook.cs:line 230
   at Python.Runtime.Runtime.PyObject_Call(IntPtr pointer, IntPtr args, IntPtr kw)
   at Python.Runtime.ImportHook.__import__(IntPtr self, IntPtr args, IntPtr kw) in c:\dev\libs\pythonnet\src\runtime\importhook.cs:line 230
   at Python.Runtime.Runtime.PyObject_Call(IntPtr pointer, IntPtr args, IntPtr kw)
   at Python.Runtime.ImportHook.__import__(IntPtr self, IntPtr args, IntPtr kw) in c:\dev\libs\pythonnet\src\runtime\importhook.cs:line 230
   at Python.Runtime.Runtime.PyObject_Call(IntPtr pointer, IntPtr args, IntPtr kw)
   at Python.Runtime.ImportHook.__import__(IntPtr self, IntPtr args, IntPtr kw) in c:\dev\libs\pythonnet\src\runtime\importhook.cs:line 230
   at Python.Runtime.Runtime.PyImport_ImportModule(String name)
   at Python.Runtime.PythonEngine.ImportModule(String name) in c:\dev\libs\pythonnet\src\runtime\pythonengine.cs:line 361
```

### DIFF
--- a/src/runtime/assemblymanager.cs
+++ b/src/runtime/assemblymanager.cs
@@ -357,7 +357,7 @@ namespace Python.Runtime {
         //===================================================================
 
         public static bool IsValidNamespace(string name) {
-            return namespaces.ContainsKey(name);
+            return !String.IsNullOrEmpty(name) && namespaces.ContainsKey(name);
         }
 
         //===================================================================


denfromufa:
@gatapia I see that google search points to issue that was happening due to broken python dependency package (not upgraded properly). So if it does not fail during import - this is worse than failing explicitly. What is your use case? Can you provide a test?

gatapia:
Sorry @denfromufa I tried to make this reproducible but could not.  It happened on one machine but not another. I know this little hack fixed the broken machine tho.  Unfortunately tracking down this "environment" issues is beyond the time I have to put towards this.

tonyroberts:
@gatapia do you have any more context, like what 'name' is that's being passed to ImportModule? It feels like there's probably another error that should get caught sooner.

gatapia:
@tonyroberts sorry I don't know; the code I am running imports quite a few modules mostly within the sklearn, pandas, numpy ecosystem.  I did not debug this while running I just hacked this fix so I have very little "root cause" information.

filmor:
Could this be related to this part of the code: https://github.com/pythonnet/pythonnet/blob/master/src/runtime/assemblymanager.cs#L320 ?

I have a patch in my local branch in place that changes this line to not use an empty string but instead `"_clr_internal"`. I'll investigate a bit.

denfromufa:
@filmor @gatapia @tonyroberts it may be possible to reproduce this error, if you have partially installed packages like in this case I had in the past:

http://stackoverflow.com/questions/28588776/python-net-name-must-not-be-empty

gatapia:
@denfromufa I tried to reproduce the error on my server after I found that SO question but even then I could not.  I reinstalled Anaconda and all my packages and my server still had the error.  I could never get a reproduceable case.  Again on all dev machines it worked, just on my prod server did it not.

dandersson:
I got a similar error in a Django-related application (Windows 7, Python 2.7.11, pythonnet 2.1.0, Django 1.9.5):

     Unhandled Exception: System.ArgumentException: Name must not be empty!
        at Python.Runtime.ModuleObject..ctor(String name)
           at Python.Runtime.ModuleObject.GetAttribute(String name, Boolean guess)
           at Python.Runtime.ImportHook.__import__(IntPtr self, IntPtr args, IntPtr kw)
           at Python.Runtime.Runtime.PyObject_Call(IntPtr pointer, IntPtr args, IntPtr kw)
           at Python.Runtime.ImportHook.__import__(IntPtr self, IntPtr args, IntPtr kw)
           at Python.Runtime.Runtime.PyObject_Call(IntPtr pointer, IntPtr args, IntPtr kw)
           at Python.Runtime.ImportHook.__import__(IntPtr self, IntPtr args, IntPtr kw)

I recreated my virtual environment from scratch in an attempt to rule out incomplete installations, as was mentioned in the earlier SO link. It did not have any effect in my case.

After downgrading to pythonnet 2.0.0.dev1, I was shown a completely different error message originating from Django (an internal server error due to a recently introduced `ImportError` in my Django configuration). When I fixed this Django error, I could run my application with pythonnet 2.0.0.dev1. After once again upgrading to pythonnet 2.1.0, not touching anything else, everything once again ran cleanly.

So, it seems like the .Net exception thrown through pythonnet in my case was not really the root error in my application, but a secondary error due to some part of the .Net stack interacting badly with in my case Django's runtime exceptions.

denfromufa:
@dandersson can you please try pythonnet recompiled with this patch from @gatapia and see if it helps finding this `ImportError` in your old Django configuration? At least it should not hide this `ImportError` and not fail silently.

Also it would be nice if you can turn it into minimal reproducible example, that is what we are really missing.

@gatapia try running your failing "prod server" using pythonnet 2.0.0.dev1 like @dandersson did - you may find some dependencies silently failing.

denfromufa:
also @dandersson and @gatapia can you try applying the patch from @filmor above?

matthid:
Hi all, I encountered the same problem and could figure out that this happens when someone uses relative imports to check if a specific module is available.

For example `site-packages\bs4\builder\__init__.py` uses:

```python
# Builders are registered in reverse order of priority, so that custom
# builder registrations will take precedence. In general, we want lxml
# to take precedence over html5lib, because it's faster. And we only
# want to use HTMLParser as a last result.
from . import _htmlparser
register_treebuilders_from(_htmlparser)
try:
    from . import _html5lib # this line reproduces this issue
    register_treebuilders_from(_html5lib)
except ImportError:
    # They don't have html5lib installed.
    pass
```

I used https://github.com/matthid/pythonnet/commit/28239416e4d30edc0ddd24e77e4181b1c4646c3b to fix the issue. See https://github.com/pythonnet/pythonnet/pull/219

filmor:
Any chance of this getting merged?

denfromufa:
@tonyroberts, can you provide Benedikt Reinartz with merging rights? I find
his contributions outstanding!

matthid:
@filmor I don't think it is the correct fix as it will lead to `Runtime.PyObject_Call` being called twice instead of properly detecting the error condition earlier.

filmor:
I'm fine with other fixes, as long as *something* gets merged. I don't like having to patch manually to keep this from failing in production :)

matthid:
Yeah its open way to long :) I hope I can provide a test case for this in my PR this weekend

PR #185, opened 2016-03-13
WPF DynamicGrid python and XAML layout files

denfromufa:
based on @tonyroberts example from mailing list:

https://mail.python.org/pipermail/pythondotnet/2015-June/001653.html

This version allows to design XAML in Visual Studio using ironpython project and run it in both ironpython and pythonnet later.

### DIFF
--- a/demo/DynamicGrid.py
+++ b/demo/DynamicGrid.py
@@ -1,0 +1,23 @@
+import clr
+import sys
+if sys.platform.lower() not in ['cli','win32']:
+    print("only windows is supported for wpf")
+clr.AddReference(r"wpf\PresentationFramework")
+from System.IO import StreamReader
+from System.Windows.Markup import XamlReader
+from System.Threading import Thread, ThreadStart, ApartmentState
+from System.Windows import Application, Window
+
+
+class MyWindow(Window):
+    def __init__(self):
+        stream = StreamReader("DynamicGrid.xaml")
+        window = XamlReader.Load(stream.BaseStream)
+        Application().Run(window)
+            
+
+if __name__ == '__main__':
+    thread = Thread(ThreadStart(MyWindow))
+    thread.SetApartmentState(ApartmentState.STA)
+    thread.Start()
+    thread.Join()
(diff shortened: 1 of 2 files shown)


tonyroberts:
I think it would be better to get rid of the IronPython references as it's a pythonnet demo. Having it work with both might be nice, but I think it could be seen as a bit confusing.

denfromufa:
The reason why ironpython is added - to make sure that Visual Studio XAML
designer can be used to design and check for errors interactively and with
intellisense using PTVS.

@tonyroberts, PTVS is officially supported by Python team at Microsoft.

So is there any alternative to not reference ironpython, e.g. to design
XAML without project? apparently this was possible in older versions of VS:

http://stackoverflow.com/questions/28375682/edit-standalone-xaml-with-visual-studio-intellisense

tonyroberts:
XAML is just a text file, so you can edit it using whatever text editor you like - VS is not required. I've used http://www.kaxaml.com/ in the past.

denfromufa:
@tonyroberts I only agree partially with you. XAML is not just text file - it is rendered as GUI and for newcomers this visual part of designing in VS is essential. In VS 2015 you can even debug XAML! People may have success with kaxaml, but it has not been developed for 5 years - no issues, no pull requests! if there is any other alternative to VS WPF XAML designer, please let me know?

https://github.com/thinkpixellab/kaxaml

tonyroberts:
vim? notepad? My point is that these demos should be what you minimally need to demonstrate a feature. If you want to write blog posts or whatever about how to debug xaml using visual studio then that would be great, but cluttering code with reference to ironpython when it's not strictly necessary (and with no explanation of why it could be desirable) isn't the best way to highlight features.

denfromufa:
@tonyroberts apparently it is possible to switch the interpreter in PTVS, so no issues with removing ironpython refs.

denfromufa:
@tonyroberts I removed ironpython refs long ago

BTW, here is another WPF demo:

https://github.com/hhspiny/WPFPy

tonyroberts:
I think a WPF demo should include some data binding.

denfromufa:
According to contributing guidelines this pull has to be open for 2 weeks
for addressing comments, before closing.

tonyroberts:
It was closed as a side effect of the develop branch being deleted. Please feel free to re-open against the master branch.

tonyroberts:
Which reminds me... I need to update the contributing file to say that PRs against master are now preferred!

denfromufa:
Thanks for clarifying!

Issue #179, opened 2016-03-08
Incorrect marshalling of string in Py_SetPythonHome


leomeuk:
I noticed that when `PythonEngine.PythonHome` was set and a subsequent `Initialize()` call was made my .Net application would close unexpectedly.

It looks like the call to `Py_SetPythonHome` made in runtime.cs is implicitly marshalling the string to a char array for the unmanaged code. My suspicion is that the Python sourcecode is keeping a record of the pointer rather than making a copy of the data. The managed runtime will free the memory allocated to char array as soon as the function finishes executing leaving a dangling pointer. This then leads to the unexpected `Initialize()` behaviour.

```
[DllImport(Runtime.dll, CallingConvention = CallingConvention.Cdecl,
    ExactSpelling = true, CharSet = CharSet.Ansi)]
internal unsafe static extern void
Py_SetPythonHome(string home);
```

The solution appears to be to explicitly marshal the string and hang onto the pointer until it is no longer needed, i.e.

```
[DllImport(Runtime.dll, CallingConvention = CallingConvention.Cdecl,
     ExactSpelling = true, CharSet = CharSet.Ansi)]
internal unsafe static extern void
Py_SetPythonHome(IntPtr home);
```

```
IntPtr str = Marshal.StringToHGlobalAnsi(@"C:\Python27");
Py_SetPythonHome(str);
// do stuff
Marshal.FreeHGlobal(str);
```

Running Win7, .Net4.5 and Python 2.7.11.

denfromufa:
@leomeuk looks like you are correct - I tested this crash in `csi`:

```
Microsoft (R) Roslyn C# Compiler version 1.1.0.51204
Loading context from 'CSharpInteractive.rsp'.
Type "#help" for more information.
> #r "C:\Python\Python27\Lib\site-packages\Python.Runtime.dll"
> using Python.Runtime;
> PythonEngine.PythonHome = @"C:\Python\Python27";
> PythonEngine.Initialize();
> PythonEngine.PythonHome
Hosting process exited with exit code -1073740940.
Loading context from 'CSharpInteractive.rsp'.
```

denfromufa:
@leomeuk can you submit a patch and corresponding unit test?

leomeuk:
@denfromufa I'll look at doing so this weekend.

PR #186, opened 2016-03-13
Explicit marshalling of Py_SetPythonHome string.

leomeuk:
Updates as described in https://github.com/pythonnet/pythonnet/issues/179.

I did look at calling `Marshal.FreeHGlobal()` in the `PythonEngine.Shutdown()` method after the call to shutdown Python is made, however it looks as though Python is doing the required memory tidy up.

Unit test has been created but is written to assume Python is installed at C:\Python27\.

### DIFF
--- a/src/embed_tests/pysethome.cs
+++ b/src/embed_tests/pysethome.cs
@@ -1,0 +1,32 @@
+﻿using System;
+using System.Reflection;
+using System.Collections.Generic;
+using NUnit.Framework;
+using Python.Runtime;
+
+
+namespace Python.EmbeddingTest
+{
+    public class PySetHomeSet
+    {
+        [SetUp]
+        public void SetUp()
+        {
+        }
+
+        [TearDown]
+        public void TearDown()
+        {
+            PythonEngine.Shutdown();
+        }
+
+        [Test]
+        public void TestSetHome()
+        {
+            string homePath = @"C:\Python27\";
+            PythonEngine.PythonHome = homePath;
+            PythonEngine.Initialize();
+            Assert.AreEqual(PythonEngine.PythonHome, homePath);
+        }
+    }
+}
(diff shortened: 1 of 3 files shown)


src/embed_tests/pysethome.cs
```diff
@@ -0,0 +1,32 @@
...
+        [Test]
+        public void TestSetHome()
+        {
+            string homePath = @"C:\Python27\";
+            PythonEngine.PythonHome = homePath;
+            PythonEngine.Initialize();
+            Assert.AreEqual(PythonEngine.PythonHome, homePath);
+        }
+    }
+}
```
denfromufa: @leomeuk you can get python path using this c# code which requires python interpreter:

```csharp
dynamic pyos = Py.Import("os");
string homepath = pyos.path.dirname(pyos.path.dirname(pyos.__file__));
```

Note that `PythonEngine.PythonPath` is also "broken" like `PythonEngine.PythonHome`:

```
> #r "C:\Python\Python27\Lib\site-packages\Python.Runtime.dll"
> using Python.Runtime;
> PythonEngine.Initialize()
> PythonEngine.PythonPath
Hosting process exited with exit code -1073740940.
Loading context from 'CSharpInteractive.rsp'.
```

denfromufa:
@leomeuk I added the comment inline in code, once your commit is ready please git squash it into one logical unit (commit) per contributing guidelines. Thanks!

tonyroberts:
Looks like this would leak the allocated string if python home was set twice? Is it safe to have Python deallocate the string allcated in .net? What happens if you deallocate it in Shutdown and set pythonhome to NULL, would that stop Python from trying to also deallocate the string?

denfromufa:
@leomeuk can you please make a pull request against master branch? develop branch is deleted.

denfromufa:
@tonyroberts I'm not worried about leaking just 1 or 2 strings, this fix is more important than a small leak.

src/embed_tests/pysethome.cs
vmuriart: Updating this PR and noticed this comment. I documented the issue with `PythonEngine.PythonPath` on #414. All the `get` functions were broken on `64bit` python.

Issue #183, opened 2016-03-12
Release Pythonnet 2.1


vmuriart:
There's been alot of changes & improvements since `2.0` release, even `2.1dev1`. (adding python 3.x upto 3.5). Perhaps its time to release `2.1` and make it available in PyPi.

vmuriart:
@tonyroberts Forgot to tag you in the question. 
We mentioned it on the Python 3.5 thread that it seemed about the right time to release, anything in particular holding this back?

tonyroberts:
There are some outstanding problems with type conversion that need to be addressed first. Unfortunately these problems were introduced in an attempt to fix some other problems, but appear to be causing more problems (see PR #176 for the latest attempt to fix). There are no other blocking problems I'm aware of.

stonebig:
no progress on this ?

denfromufa:
I'm not sure what @tonyroberts is waiting on?

vmuriart:
There were a few open tickets he wanted to close first hat I think broke
previous features

stonebig:
someone told me that the wheel created during test where somehow downloadable, but I don't find where they could be

denfromufa:
@vmuriart pull request #176 that was blocking release is merged - not sure which tickets you mean? But maybe @tonyroberts is waiting for fix on #190

denfromufa:
@stonebig here is the link to sample wheels on appveyor:

https://ci.appveyor.com/project/TonyRoberts/pythonnet-480xs/build/1.0.105/job/8yo6slgn41ffo8hp/artifacts

if you navigate to jobs and select python version, then builds are located under artifacts.

stonebig:
thanks! Unfortunately it's still failing hard on wordpad demo.

denfromufa:
@stonebig what is the error you get when you launch wordpad demo from
command line? which version (2.7..3.5, 32-64bit) did you try?

stonebig:
using pythonnet-2.1.0.dev1-cp34-cp34m-win_amd64.whl from you pinpointed Merge pull request #176 from denfromufa/patch-8:
- launchiing wordpad.py from IDLEX => ok, the windows appear,
- "Help -> About" gives "TypeError: No Method matches given argument". detail error below.. cursiously it speaks about Win32
````
Consultez la fin de ce message pour plus de détails sur l'appel du débogage
juste-à-temps (JIT) à la place de cette boîte de dialogue.

************** Texte de l'exception **************
[...]

PR #197, opened 2016-04-04
Update wordpad.py

denfromufa:
fixes https://github.com/pythonnet/pythonnet/issues/183#issuecomment-205009786

### DIFF
--- a/demo/wordpad.py
+++ b/demo/wordpad.py
@@ -414,7 +414,7 @@ class AboutForm(WinForms.Form):
         self.ShowInTaskbar = False
         self.StartPosition = WinForms.FormStartPosition.CenterScreen
         self.Text = "About"
-        self.ResumeLayout(0)
+        self.ResumeLayout(False)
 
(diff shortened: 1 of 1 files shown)


denfromufa:
@stonebig @vmuriart this should fix now wordpad.py demo

tonyroberts:
@denfromufa does this demo work for you? I just tried it with the change from this PR and whenever I try to save or open a file it hangs. I've not looked into why (and I don't know if it ever worked?).

tonyroberts:
Oh ok, I see. It's just because the main thread isn't set up as an STA thread. I'll fix than in another PR.

denfromufa:
@tonyroberts I have the same problem - I never actually tried saving or
opening files before. Looks like it never worked at least since github
transition. I just tried with versions 2.0+. @stonebig did you get save or
open files to work in wordpad.py?

tonyroberts:
@denfromufa @stonebig this is fixed now (see above and PR #198).

PR #198, opened 2016-04-04
Start the wordpad demo in an STA thread.

tonyroberts:
This fixes the problem with the open and save dialogs freezing the
application.

### DIFF
--- a/demo/wordpad.py
+++ b/demo/wordpad.py
@@ -9,6 +9,7 @@
 
 import clr
 import System.Windows.Forms as WinForms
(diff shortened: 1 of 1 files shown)


stonebig:
I can no do file->open and select a .txt file .... then "wordpad.py" enters in infinite loop (I just modifiy notepad.py, maybe I should update more things ?)

stonebig:
"save" worked, just not doing a CRLF, I suppose it's standard .NET behavior

tonyroberts:
@stonebig the problem with opening files is fixed in #199.

PR #199, opened 2016-04-04
Use stream position instead of counting read bytes.

tonyroberts:
FileStream.Read returns (bytes read, buffer), not just bytes read as the
buffer is passed by reference. This is a bit unintuative, but checking the
position is clear and less error prone.

### DIFF
--- a/demo/wordpad.py
+++ b/demo/wordpad.py
@@ -316,11 +316,10 @@ class Wordpad(WinForms.Form):
 
         buff = System.Array.CreateInstance(System.Byte, 1024)
         data = []
-        read = -1
 
-        while (read != 0):
+        while stream.Position < stream.Length:
             buff.Initialize()
-            read = stream.Read(buff, 0, 1024)
+            stream.Read(buff, 0, 1024)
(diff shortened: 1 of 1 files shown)


stonebig:
so now:
- I can open, modify, save,
- all is apparently ok if I stay in ascii letter,
- if I type this:
````
the Euro Symbol is "€"
c'est la fête à Paris !
````
- I save, and when I reload (or look via notepad++:
````
the Euro Symbol is "?"
c'est la f?te ? Paris !
````

tonyroberts:
Right, well I think as a demo it's now adequately showing how Python and c#
can interact. I don't think it warrants much more time spent on it, but I
don't mind reviewing pull requests if you want to fix it up some more.

stonebig:
no problem, thanks!

denfromufa:
@stonebig @tonyroberts - this demo also has to be adapted to python 3 for
managing str/byte/unicode conversion properly using `six` module.

stonebig:
trying to get this example working in all combinations could drown the project resources.

PR #201, opened 2016-04-06
Fix and test for overload resolution when last parameter takes a non-"params" array

### LINKED ISSUE #200: Overload resolution incorrectly assumes any array parameter in last position is a params array
There is a trivial one-line fix for this, which I'll paste below. I've tested the fix and would like to have submitted a pull request, but I am unable do so due to proxy restrictions at work.

The problem arises when you have an overloading situation like the following:

    void SomeFunc(int param1, int[] param2);
    void SomeFunc(int param1, int param2, int param3);

and you make a call like SomeFunc(1,2,3).

During overload resolution, if pythonnet encounters an overload with fewer parameters than are supplied in the call, but whose last parameter is of array type, it assumes that the last parameter is a "params" array. i.e. it treats the first overload above as if it were declared:

    void SomeFunc(int param1, params int[] param2);

The result is that pythonnet calls the first overload and passes the second and third arguments as a 2-element array in param2, when it should be calling the second overload and passing all three arguments normally. (The actual situation we are experiencing is more complex, but the above example should reproduce the behaviour.)

The fix belongs in methodbinder.cs. At line 271, there is the following code:

    } else if ((pynargs > clrnargs) && (clrnargs > 0) &&
    (pi[clrnargs-1].ParameterType.IsArray)) {
    // The last argument of the mananged functions seems to
    // accept multiple arguments as a array. Hopefully it's a
    // spam(params object[] egg) style method

This needs to be changed to explicitly check for the case of a params array. The following replacement works for me:

    } else if ((pynargs > clrnargs) && (clrnargs > 0) && 
                  Attribute.IsDefined(pi[clrnargs-1], typeof(ParamArrayAttribute))) {

omnicognate:
This fixes #200

### DIFF
--- a/src/tests/test_method.py
+++ b/src/tests/test_method.py
@@ -304,6 +304,10 @@ class MethodTests(unittest.TestCase):
         self.assertTrue(result[1] == 2)
         self.assertTrue(result[2] == 3)        
 
+    def testNonParamsArrayInLastPlace(self):
+        """Test overload resolution with of non-"params" array as last parameter."""
+        result = MethodTest.TestNonParamsArrayInLastPlace(1, 2, 3)
+        self.assertTrue(result)
 
     def testStringOutParams(self):
         """Test use of string out-parameters."""
--- a/src/runtime/methodbinder.cs
+++ b/src/runtime/methodbinder.cs
@@ -269,10 +269,8 @@ namespace Python.Runtime {
                             defaultArgList.Add((object)pi[v].DefaultValue);    
                     }
                 } else if ((pynargs > clrnargs) && (clrnargs > 0) &&
-                           (pi[clrnargs-1].ParameterType.IsArray)) {
-                    // The last argument of the mananged functions seems to
-                    // accept multiple arguments as a array. Hopefully it's a
-                    // spam(params object[] egg) style method
+                           Attribute.IsDefined(pi[clrnargs-1], typeof(ParamArrayAttribute))) {
+                    // This is a spam(params object[] egg) style method
                     match = true;
                     arrayStart = clrnargs - 1;
                 }
--- a/src/testing/methodtest.cs
+++ b/src/testing/methodtest.cs
@@ -117,6 +117,14 @@ namespace Python.Test {
             return true;
         }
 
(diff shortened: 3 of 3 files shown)


src/tests/test_method.py
```diff
@@ -304,6 +304,10 @@ def testValueParamsArgs(self):
         self.assertTrue(result[1] == 2)
         self.assertTrue(result[2] == 3)        
 
+    def testNonParamsArrayInLastPlace(self):
```
denfromufa: How about tests for the cases with params args & arrays?

denfromufa:
@omnicognate can you squash (rebase) into one commit according to contributing guidelines? you can use sourcetree for GUI or plain git command-line to do this.

omnicognate:
I can do that but it will have to wait until I have something better than and xperia z2 to do it on :-)
I'll give it a go this evening, UK time.

src/tests/test_method.py
omnicognate: Do you mean the case where there is a params keyword. If so, I believe that's covered by existing tests, eg. testObjectParamsArgs

denfromufa: I mean don't you need to add in c# and check in python tests for the
following case below?

public static bool TestNonParamsArrayInLastPlace(int i1, params int[] ints)
{ return false; }

Maybe boolean return value needs to be changed to integer, since then there
are 3 overloaded cases.

denfromufa:
you can try sgit app for android

src/tests/test_method.py
omnicognate: I just tried it, but C# won't let me have both

   TestNonParamsArrayInLastPlace(int i1, params int[] ints)

and

   TestNonParamsArrayInLastPlace(int i1, int[] ints)

It says they have the same name and parameter types, error CS0111.

denfromufa: makes sense - no more questions! the answer as to why both are not allowed
is given here: http://stackoverflow.com/a/7580295/2230844

omnicognate:
That was an adventure (bit of a git n00b), but it's down to a single commit now.

Issue #184, opened 2016-03-12
Clean up whitespacing on files


vmuriart:
In particular I'm referring to blank lines that have multiple spaces on them and no content. 

One of the issues I've noticed for people submitting Pull Requests, is that their linters/text editors automatically strip them when saving changes. This obscures the changes that were done, and happens to alot of contributors.

To make it easy for everyone else contributing, we can make a single commit whose only purpose is to strip out all the lingering white-space and fix alignment so that all subsequent contributors don't have do worry about it.

@tonyroberts if you ok with this, i don't mind going through the files and cleaning them up and submitting a PR for it.

denfromufa:
@vmuriart you may also need to include a style guide for curly braces, tabs, spaces and newlines, so that this is enforced after this cleanup

vmuriart:
@denfromufa if you have something in mind I'm all ears. I was going to let Sublime/Atom figure out what needed fixing/cleaning up :smiley:. 
I'm not particularly familiar if there is a "preferred" style-guide for C/C# as there is for Python.

denfromufa:
@vmuriart this looks good to me:

https://msdn.microsoft.com/en-gb/library/ff926074.aspx

except that curly braces are perhaps according to indent style of variant of Stroustrup:

https://en.wikipedia.org/wiki/Indent_style#Variant:_Stroustrup

tonyroberts:
I agree it would be nice to have it all consistent. The problem of editors changing whitespace/indentation etc won't necessarily be solved simply by doing that however. Everyone seems to have different settings wrt code style, so even if it was all one style someone else could have completely different styles set up in their editor and trash it all in the same way that happens now (and indeed I suspect that's why the code is in the state it's in).

In order to enforce consistent styles we would first need a style guide for contributors, preferably one that can be checked before the code is checked in and as part of the CI process.

Without having something that can be systematically be enforced there's not much point changing anything, as the same conversation will come up again and again as more changes with slight style variants are committed.

denfromufa:
@vmuriart @tonyroberts there are 2 tools for enforcing style:

http://stylecop.codeplex.com/

https://github.com/mono/mono-tools/tree/master/gendarme

vmuriart:
I think there is a way to add it as a "test" that github would check against to ensure its ok.
I was working on some project this weekend and realized that VS is the culprit for alot of the formatting inconsistencies.

denfromufa:
[...]

PR #205, opened 2016-04-17
Format and White-spacing

vmuriart:
To address #184. 
Settled on using VS standard formatting for `.cs` and Python pep8 for `.py`.
Did some house cleaning and removed/integrated old docs, solution files, etc.

Should **hopefully** help solve constant whitespacing issue on pull requests. 

Note, other than white-spacing and formatting **nothing in the code has been changed**.


vmuriart:
@tonyroberts any idea why `Appveyor` isn't running anymore on the `pull request`? Noticed it yesterday. I verified the build against my `Appveyor` account and its passing, so it `Travis-ci`

vmuriart:
As I side note, i don't fully agree with VS's choice of style, but I'd rather see consistent formatting.

tonyroberts:
I'm fine with cleaning up whitespace etc, but I think reorganising the file structure should be done in a different PR. In particular, I don't think moving very old and out dated documentation into the root folder is a good idea. I think it would be better if it was updated and then moved - rather than delay these whitespace changes until that's done, dropping them from this PR seems sensible.

tonyroberts:
Also the wording of commit f72c7e0775e189a23fa007ed4100939bb7a8a061 comes across as a bit derogatory towards mono (even if that wasn't the intention) and could give someone casually looking through the commit history the impression that mono is less supported than the microsoft runtime, which is not the case.

tonyroberts:
@vmuriart hmm not sure what happened to the appveyor CI - looks like there was some problem with the github settings, which is odd as I'm pretty sure I didn't change them! Anyway, hopefully it will be working again now..

vmuriart:
@tonyroberts I was having similar thoughts about the file cleanup after i submitted the `pr`. Just rebased and removed them from this latest one, I'll submit a separate one for those.

Good catch on the mono commit. I meant that in a joking manner (monkey-droppings) with the intention of removing their old format files since they now support the same files as vs. 

Looks like travis is still the only one running, atleast on this PR. I had a similar issue on my fork though (Travis being the problem in my case). To fix it I had to remove the hook for it, and re-add it from github.

vmuriart:
@tonyroberts what are your thoughts on `git fast-forward` for merges? I personally use a combination of both `git ff` and merge-commits depending on the change. Simple changes (1 or 2 small commits) I fastforward, and more complex ones get `merge-committed`.

tonyroberts:
@vmuriart cool, I'll merge this now and take a look at why the commit hooks aren't working later. I'll try just recreating it as you suggest. Thanks for doing this!

tonyroberts:
@vmuriart wrt ff, it really depends on the situation. For simple changed I prefer to ff where possible, but when adding significant new feature or a larger set of changes I prefer to use a feature branch and keep the merge commit - so sounds like we're thinking along the same lines :)

denfromufa:
@vmuriart @tonyroberts the indent(ation) style for curly braces is a mess even after merge. Here are 2 contradictory examples:

Allman style:
https://github.com/pythonnet/pythonnet/pull/205/files#diff-1b72bb3fa8d022b65bf52c6d695a6562R44

K&R style:
https://github.com/pythonnet/pythonnet/pull/205/files#diff-0ac34f0651a9febc4d8713852516ea6cR50

vmuriart:
Good catch. Not sure why the linter's didn't pick it up.

They should all be defaulting the K&R style since that's VS's default.

vmuriart:
Ah, just realized whats going on.
I only corrected the formatting on the `.cs` and `.py` files for bracketing
and such.
The whitespacing was done to all files. The file you pointed out as being
inconsistent was a `.c` file that only got the trailing-whitespace
treatment.

vmuriart:
Didn't realize that there were only 2 `.c` files. Had i known I wouldn't have skipped them...

denfromufa:
@vmuriart @tonyroberts I think this PR should include one `editconfig` settings file for other IDEs such as Atom, Sublime, VS Code, MonoDevelop, NPD++. See example here:


https://github.com/OmniSharp/omnisharp-roslyn/blob/dev/.editorconfig

vmuriart:
I forgot to add the `.editorconfig` but I agree. I can include one with the `pr` to remove some of the old files

Issue #203, opened 2016-04-17
Pythonnet 2.1 not identifying System.Object type


vmuriart:
I'm not sure how the tests wouldn't have caught this bug.
I have a `.net` class that has the function `AddQuery` with the following method signatures.

```
AddQuery
    Void AddQuery(api.QueryType)...
    Void AddQuery(System.Object)...
```

where `QueryType` is another custom object type. 

On `2.0` and `2.1dev1` the function works as expected. 
- I can pass a `QueryType` object to `AddQuery` and it would call the first method. 
- Calling `AddQuery` with a `string` or `int` would use the second method and process them.

On `2.1.0` and the earliest archived copy in `appveyor` this behavior changed. 
- Passing `QueryType` object to `AddQuery` still works. 
- Passing `string` type to `AddQuery` **errors** with `No method matches given arguments`

I've tested on `Python 2.7`, `Python 3.5` both x64. 
Looks like Pythonnet is no longer identifying `string` `int` and other types to be of type `System.Object` as well.

vmuriart:
Possibly related to #187 
Tested on windows 7 and 8.1

vmuriart:
Tried build before and after #176, both still have this bug present.

denfromufa:
@vmuriart 
try passing `System.Object("your_python_string")` or `System.Object(123)` for `string` or `int` types respectively.

this breaking change is because implicit casting can be dangerous:

https://github.com/pythonnet/pythonnet/issues/131

vmuriart:
@denfromufa just tried your suggestion and it semi-works. It passes through `pythonnet` and passes the `Object-type string` to the library I'm using, and breaks there. The underlying function is expecting a `String-type string` and fails to identify the `Object-type string` as such.

denfromufa:
did you try `System.String()` and `System.Int32()`? or maybe even `System.Object(System.String())`...

vmuriart:
I did. those to work, but have to explicitly cast each type before sending it. 
I'm taking a look at the src; I think there might be a better solution for the original bug #131 that would both fix the bug and not break legacy code. Might even try to fix #90 finally :smile:

denfromufa:
@vmuriart I'm think of context manager that would allow you to specify "unsafe" casting operations like before for overloaded methods, see discussion here (my 4th comment):

github.com/pythonnet/pythonnet/issues/174

denfromufa:
[...]

PR #217, opened 2016-05-26
Update methodbinder.cs

denfromufa:
https://github.com/pythonnet/pythonnet/issues/203

@vmuriart @tonyroberts

### DIFF
--- a/src/runtime/methodbinder.cs
+++ b/src/runtime/methodbinder.cs
@@ -344,7 +344,7 @@ namespace Python.Runtime
                             if (clrtype != null)
                             {
                                 bool typematch = false;
-                                if (pi[n].ParameterType != clrtype)
+                                if ((pi[n].ParameterType != clrtype) && !(pi[n].ParameterType.IsAssignableFrom(clrtype)))
                                 {
                                     IntPtr pytype = Converter.GetPythonTypeByAlias(pi[n].ParameterType);
                                     pyoptype = Runtime.PyObject_Type(op);


src/runtime/methodbinder.cs
```diff
@@ -344,7 +344,7 @@ internal Binding Bind(IntPtr inst, IntPtr args, IntPtr kw)
                             if (clrtype != null)
                             {
                                 bool typematch = false;
-                                if (pi[n].ParameterType != clrtype)
+                                if ((pi[n].ParameterType != clrtype) && !(pi[n].ParameterType.IsAssignableFrom(clrtype)))
```
tonyroberts: Isn't the != check redundant? It looks like checking !pi[n].ParameterType.IsAssignableFrom(clrtype) would be sufficient.

vmuriart:
Looks like appveyor is still being a rebel and not running.

src/runtime/methodbinder.cs
denfromufa: This is a very good catch! But actually the logic is more complicated,
since sometimes multiple methods can match `IsAssignableFrom` criteria for
the overloaded arguments.

So in addition to type equality really need to loop through `IsSubclassOf`
and build a list sorted by proximity to the argument type.

vmuriart:
Unless I'm somehow downloading the wrong version again; this `pr` lets my programs run again w.o any modification :+1: :smiley:

src/runtime/methodbinder.cs
tonyroberts: All I meant was that there can never be the case where pi[n].ParameterType == clrtype && !(pi[n].ParameterType.IsAssignableFrom(clrtype)), so this case is equivalent to checking that ParameterType is not assignable from clrtype.

Another way of saying this is that (ParameterType is not assignable from clrtype) implies (Parameter Type != clrtype). So, the expression as written is logically equivalent to simply (ParameterType is not assignable from clrtype).

vmuriart:
@denfromufa Should the `pr` include a improved version of the tests I wrote on https://github.com/vmuriart/pythonnet/commit/7acf93faa653b3a1d99f8546ad69322296587c6b ?

denfromufa:
I can custom merge or cherry pick the tests from your branch.

But I'm still looking at how to resolve methods when multiple overloaded
methods match by arguments for the "isinstance" check, e.g. System.Object
and System.String.

tonyroberts:
Insufficient testing and no reply to comments for some time, so closing without merging.

denfromufa:
I came across this chapter and will look closer how to resolve ambiguous cases:

http://ironpython.net/documentation/dotnet/dotnet.html#appendix-detailed-method-overload-resolution-rules

vmuriart:
@denfromufa did we ever come to a resolution for this?

denfromufa:
@vmuriart not yet, this change requires restructuring the way pythonnet searches the best match in overloaded methods. Right now there is assumption that only one match is possible and so pythonnet proceeds with the first match. However this is not how it works for C#. In short, we need to add a list with overloaded method matches and decide which one is the best. Otherwise, if more than one match is found, then throw an exception to let the user select the overloaded method precisely without any ambiguity in method resolution.

Issue #238, opened 2016-07-03
Constructor failing argument matching, silently selects default constructor


sigbjorn:
(Version 2.1)

Given .NET class:

``` C#
      public class A {

           public A() { a=1.0;}

           public A(double ax) { a=ax;}

           public double a;

     }
```

then in python:

``` python
from Xxx import A

a1 = A(2)   # surprise! default constructor is selected, silently, basically because 
            # there is no int->double type promotion in constructor binder
            # causing first bind to fail, then second bind with no args (that was a fix
            # to allow .NET objects to be super-classes (or be sub-classed) from
            # python succeeds..
a2 = A(2.0) # Ok, now we get the second constructor, because it's a perfect match.
```

I am fully aware that super-classes, argument-matching, (automatic) type-conversion,
 and bridging the gap between python and .NET (C#) , is very complex.
And providing a full implementation that 'just do the thing we want' is not trivial.

It might be that some of the issues related to sub-classing, as well as argument matching/conversion is related, - but I hope the simple example above illustrates that
some minor improvement at this stage could be quite useful.

There is a few fixes, that could improve this slightly:

(a.) Make simple type-promotion/argument conversion happen during constructor argument matching
(b.) If construction of a .NET object is attempted, and argument match fail, then always raise Type error

I will provide a PR for (a.) since it's a simple fix to verify a simple type-promotion from int->double etc.

Fixing (b.), raise error if no .NET constructor matches the arguments is easy, but will then break the sub-class mechanism. If we knew, by the time we are calling the .NET constructor, that it's a super-class, then we  _could_ either have a relaxed matching (as is to day, default to the default-ct, if fail), and then insist on match for all other cases.

After all, providing something that fixes (b.), and at the same time supports, and possibly extends super()/sub-classing mechanism seems harder to me, since we need to know the context in which the .NET object is constructed.

If anyone see a easy solution to (b.) , - as in knowing how to get the context of the .NET class construction, that would be great.

We could then provide a  fix where 

' you succeed with your intention creating a new object with the supplied parameters, and get the expected result, or we raise Type-exception notifying about the problem'

denfromufa:
@sigbjorn here is really where bad logic starts and where it should have really failed:

https://github.com/pythonnet/pythonnet/blob/master/src/runtime/constructorbinder.cs#L84

```csharp
if (binding == null)
            {
                // It is possible for __new__ to be invoked on construction
                // of a Python subclass of a managed class, so args may
[...]

PR #239, opened 2016-07-03
Constructor argument matching supports simple int to (float|double) types

sigbjorn:
Refer to Issue #238

This PR aim to improve  constructor argument matching so that if you have a .NET constructor accepting  a double, it will be selected if you pass it an int.

This corresponds to solution part (a.) in the referenced issue.

The PR involves tests, that with the prior version would fail, but with the fixes, works Ok.
All existing tests, run on python 3.5, x64, windows platform executes Ok.

I have not verified other platforms(mono, 32 bits etc), but could do that on request.

The PR does currently not support raising Type exception in case a falling back to a default constructor due to arguments not matching. However, if anyone contribute with hint's how to get information about the .NET creation object context (super-class, or just plain class), - I would be happy to extend the contents and tests to cover that important aspect as well.

### DIFF
--- a/src/runtime/methodbinder.cs
+++ b/src/runtime/methodbinder.cs
@@ -469,15 +469,16 @@ namespace Python.Runtime
         }
         /// <summary>
         /// Returns true if src type (int -type) is promotable to a higher resolution type (float|double)
+        /// <note>no sanity check for overflow is done, this is done elsewhere when performing
+        ///  the actual conversion.
+        ///  </note>
         /// </summary>
         /// <param name="src"></param>
         /// <param name="dst"></param>
         /// <returns></returns>
         private bool SimplePromotableType(Type src, Type dst)
         {
-            return (((src == typeof(int) || src == typeof(short)) && (dst == typeof(double) || dst == typeof(float))))
-                || ((src == typeof(long)) && (dst == typeof(double)))
-                || ((src ==typeof(int) || src == typeof(short)) && (dst == typeof(long)));
+            return (((src == typeof(int) || src == typeof(short) || src == typeof(long) ) && (dst == typeof(double) || dst == typeof(float))));
         }
 
         internal virtual IntPtr Invoke(IntPtr inst, IntPtr args, IntPtr kw)
--- a/src/runtime/methodbinder.cs
+++ b/src/runtime/methodbinder.cs
@@ -375,8 +375,20 @@ namespace Python.Runtime
                                     Runtime.Decref(pyoptype);
                                     if (!typematch)
                                     {
-                                        margs = null;
-                                        break;
+                                        if (SimplePromotableType(clrtype, pi[n].ParameterType))
+                                        {
+                                            if (!Converter.ToManaged(op, pi[n].ParameterType, out arg, false))
+                                            {
+                                                margs = null;// basically we are failing this match, but can't throw yet
+                                                break;
+                                            }
+                                        }
+                                        else
+                                        {
+                                            margs = null;
+                                            break;
+                                        }
+
                                     }
                                 }
                                 else
@@ -454,6 +466,18 @@ namespace Python.Runtime
                 return Bind(inst, args, kw, mi, null);
             }
             return null;
+        }
+        /// <summary>
+        /// Returns true if src type (int -type) is promotable to a higher resolution type (float|double)
+        /// </summary>
+        /// <param name="src"></param>
+        /// <param name="dst"></param>
+        /// <returns></returns>
+        private bool SimplePromotableType(Type src, Type dst)
+        {
+            return (((src == typeof(int) || src == typeof(short)) && (dst == typeof(double) || dst == typeof(float))))
+                || ((src == typeof(long)) && (dst == typeof(double)))
+                || ((src ==typeof(int) || src == typeof(short)) && (dst == typeof(long)));
         }
 
         internal virtual IntPtr Invoke(IntPtr inst, IntPtr args, IntPtr kw)
--- a/src/testing/constructortests.cs
+++ b/src/testing/constructortests.cs
@@ -7,10 +7,10 @@ namespace Python.Test
     //========================================================================
     // These classes support the CLR constructor unit tests.
     //========================================================================
-    public class AConstrucorTest
+    public class AConstructorTest
     {
         public string name;
-        public AConstrucorTest(string n) { name = n; }
+        public AConstructorTest(string n) { name = n; }
     }
     public class LinkConstructorTest
     {
@@ -18,7 +18,7 @@ namespace Python.Test
         {
             DefaultCtCalled = true;
         }
-        public LinkConstructorTest(AConstrucorTest a,double matchMe,AConstrucorTest b)
+        public LinkConstructorTest(AConstructorTest a,double matchMe,AConstructorTest b)
         {
             MatchMe = matchMe;
             a1 = a;
@@ -26,8 +26,8 @@ namespace Python.Test
         }
         public bool DefaultCtCalled;
         public double MatchMe;
-        public AConstrucorTest a1;
-        public AConstrucorTest a2;
+        public AConstructorTest a1;
+        public AConstructorTest a2;
     }
     public class EnumConstructorTest
     {
--- a/src/testing/constructortests.cs
+++ b/src/testing/constructortests.cs
@@ -71,4 +71,36 @@ namespace Python.Test
             this.value = v;
         }
     }
+    public class ToDoubleConstructorTest
+    {
+        public ToDoubleConstructorTest()
+        {
+            //Just default values
+        }
+        public ToDoubleConstructorTest(string a, double b,string c)
+        {
+            this.a = a;
+            this.b = b;
+            this.c = c;
+        }
+        public string a;
+        public double b;
+        public string c;
+    }
+    public class ToFloatConstructorTest
+    {
+        public ToFloatConstructorTest()
+        {
+            // just default values.
+        }
+        public ToFloatConstructorTest(string a, float b, string c)
+        {
+            this.a = a;
+            this.b = b;
+            this.c = c;
+        }
+        public string a;
+        public float b;
+        public string c;
+    }
 }
--- a/src/testing/constructortests.cs
+++ b/src/testing/constructortests.cs
@@ -7,7 +7,28 @@ namespace Python.Test
     //========================================================================
     // These classes support the CLR constructor unit tests.
     //========================================================================
-
+    public class AConstrucorTest
+    {
+        public string name;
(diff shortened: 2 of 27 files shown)


sigbjorn:
Build failing, Ok, I have to fix the build, reverse changes to .sln and .proj files I guess.

filmor:
Yes, please.

Also, your test is failing for Python 2.7 as the `super()` syntax (without arguments) is Python 3. Use `super(PySubclass, self).__init__ ...` (https://github.com/pythonnet/pythonnet/pull/239/files#diff-cce6857ae7057bf6caf15818e1ad16e0R60).

src/runtime/methodbinder.cs
```diff
@@ -455,6 +467,18 @@ internal Binding Bind(IntPtr inst, IntPtr args, IntPtr kw)
...
+        /// <summary>
+        /// Returns true if src type (int -type) is promotable to a higher resolution type (float|double)
+        /// </summary>
+        /// <param name="src"></param>
+        /// <param name="dst"></param>
+        /// <returns></returns>
+        private bool SimplePromotableType(Type src, Type dst)
+        {
+            return (((src == typeof(int) || src == typeof(short)) && (dst == typeof(double) || dst == typeof(float))))
+                || ((src == typeof(long)) && (dst == typeof(double)))
```
filmor: This logic is a bit arbitrary, isn't it? A 64bit integer fits in neither `double` nor `float`, so why make this distinction here?

src/testing/constructortests.cs
```diff
@@ -7,7 +7,28 @@ namespace Python.Test
     //========================================================================
     // These classes support the CLR constructor unit tests.
     //========================================================================
-
+    public class AConstrucorTest
```
filmor: Typo.

filmor:
@denfromufa Regarding the approach, I'm fine with doing conversions `int-type -> float-type`. This feels much more natural coming from Python where you'd do the constructor exactly like this (at least in Py3 where `int` and `float` behave very similar).

src/runtime/methodbinder.cs
sigbjorn: Yes, so it merely answers if its possible to do the conversion.
The real conversion is attempted later, using standard python.net conversion
that hopefully do the best effort, throws, etc. if there are numerical problems representing the one from the other.

sigbjorn:
Struggling a bit with build on arch-linux, mono 4.4.0.40, (before I apply my patch, using master..)
 it fail on Target RGieseckeDllExport:
: error : Error initializing task DllExportAppDomainIsolatedTask: Could not load file or assembly 'Microsoft.Build.Utilities, Version=2.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' or one of its dependencies.

Is this a know error/solution, or should I just try to build on a Ubuntu platform?

denfromufa:
> Struggling a bit with build on arch-linux, mono 4.4.0.40, (before I apply my patch, using master..)
> it fail on Target RGieseckeDllExport:
> : error : Error initializing task DllExportAppDomainIsolatedTask: Could not load file or assembly 'Microsoft.Build.Utilities, Version=2.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' or one of its dependencies.
> 
> Is this a know error/solution, or should I just try to build on a Ubuntu platform?


For Mono `RGieseckeDllExport` cannot be used, since this is Windows-specific feature of reverse pinvoke. Instead the configuration for Mono should be selected, which uses C-API of Mono and Python to bind each other.

sigbjorn:
Thanks for the hint, but when using setup.py (as in .travis.yml build),

( Configuration: Release Mono Platform: x64) is printed out during the build
or installing monodevelop, (Ubuntu 15.10 this time, with clang, python-dev etc.)  

selecting mono debug/release x64, then build the solution, the clrmodule is the only one failing,
  on ubuntu it fail with 'no accesss to the given key' indicating there is some more security related issues .

It also seems that the travis now fail on the RGiesecke related stuff, the build tools 2.0.0.0 not found etc.

Appveyor seems to be happy after 2.7 fixups

sigbjorn:
Ok, now all .sln, csproj files are back to original, hopefully that solve the build issues.

I finally did succeed building the mono 4.4.1 version, Ubuntu, python 3.4, python3-dev, libglib2.0-dev added, plus all the mono stuff from standard packages on ubuntu, using virtualenv to setup python part. 

But running the tests fail, when Load clr hook executes, "Error: No managed allocator , but we need one for AOT. Are you using non-standard GC options ?"

Sorry to bother you all with this, but maybe someone knows a fix (otherwise, I will just clean, rebuild, select packages specific to the travis.setup etc.., and retry.)

denfromufa:
This exception in Mono seems to be common lately starting ~4.4.1:

https://github.com/tjanczuk/edge/issues/449

minxomat:
CC [42169](https://bugzilla.xamarin.com/show_bug.cgi?id=42169) in Bugzilla for the managed alloc bug. More info there and [here for WSL](https://github.com/mono/website/issues/199#issuecomment-230333018).

sigbjorn:
Ok, now the py 2.6 is the only one that need som attention. I will fix that soon, so at least the build is straighten up, then we can see if we can improve this PR to a level where it can be accepted.

Downgrading mono etc. to do local builds work is Ok for now.

src/runtime/methodbinder.cs
denfromufa: > And yes, - given that there is both and int and a double constructor, we should select the one with closest match. (Requires rewrite, or redirect logic to some function that already do this ?)

@sigbjorn as far as I'm aware there is no logic for resolution order in pythonnet for multiple matches due to complexity it brings into code-base:

https://github.com/pythonnet/pythonnet/pull/217

sigbjorn:
Ok, now it even builds and run on arch-linux, py 3.4, mono 4.4.0.40, that is before the ~4.4.1, so now I can verify the mono version more easily as well. It was my vs2015, changing the .sln and the build that caused the mono build problems, sorry for the mess. Now that I got both win and mono up running it's easier to avoid.

sigbjorn:
Is there anything more I can contribute with so that this PR can be merged ?

denfromufa:
@sigbjorn the problem I have is with cases when multiple types meet the auto-conversion. Without handling such behavior this PR seems incomplete to me, but resolving this is a lot more effort and will open us to a lot of potential corner cases. Why not just throw exception if exact match is not found? The user should be smart enough to provide the exact type after the error is reported.

sigbjorn:
Yes, as mentioned previously, this PR was not originally created to solve all problems related to this issue.

and yes: 

A simple fix would be to throw exception if not perfect match when calling a constructor.
But unfortunately, that *simple fix* will also break the existing sub-class a .NET class feature, that rely on this flexible behaviour. ( My guess is that you would then say NO to this PR, or maybe I am wrong ?)

This problem could be resolved:
 If we from the context of calling the constructor, could know that we are in a chain of super().. call-chain, we could make a fix that is aware of the context,
 and then: in the context of super().. , allow anything (like it is now)
                 and in the other contexts (direct call to constructor), enforce exact match.

So if you/or anybody could point me in any direction that allow us to figure out if we are in a super()-call-chain, this is easy to fix. I have briefly checked the code, and found no such info in the call-chain from python to the method-binder in question dealing with constructors.

other alternatives:
Making the constructor-matching more selective, choosing the closest matching constructor can be done, - but requires a lot more changes. -And unfortunately, it does not provide exceptions for non-matching cases.

So my intention was to first provide something that I think resolves a lot of 'practical cases'. 

Then, if we could get some more knowledge about the object creation (especially super/sub-class and argument passing). With that knowledge in place, fix selective matching and proper super/sub-classing in addition to throwing exception when there are no matches.

denfromufa:
ok, the simple solution right now could be to raise a warning instead of
exception when going the route of ignoring the arguments?

I like the idea to find if we are in the super() chain! Have not figure out
yet.

minxomat:
> Downgrading mono etc. to do local builds work is Ok for now.

Update: It should be safe to use Mono again when 4.6.0 is released (or build from git if the tag updates at [mono/mono](http://github.com/mono/mono)).

sigbjorn:
Ok, - the moment we properly detect and figure out the super() call-chain (for all versions of python), its quite straight forward.
It requires a dive into the python type system and mechanism in play during construction of object.
This will require some more effort, - I could have a look into it next week.

src/runtime/methodbinder.cs
filmor: My point was that *dis*allowing the conversion `long -> float` doesn't make sense as the conversion `long -> double` is also lossy (64 vs 53). So either we allow all of those or none.

sigbjorn: Great, then we could try the approach I suggested in my initial comment, - just use the standard conversion function inside pythonnet. That's wider approach, but there is a hope it's consistently used for argument conversions, and thus we can terminate this discussion, since then the arguments to a constructor undergoes (almost) the same rules as an ordinary pythonnet function.
I am on vacation now, but can try to see if its possible to get in touch with a computer long enough to produce the change.

filmor:
@sigbjorn If you rebase and squash the PR, I'm okay with merging this. @vmuriart @denfromufa, what's your opinion?

denfromufa:
I'm ok with auto-conversion only when a single match is found. This likely
means the number of arguments has to be one and also requires additional
logic to iterate over all possible arguments to set this "single match is
found" flag.

vmuriart:
I'll run some tests tomorrow against my applications and check for no surprises :)

sigbjorn:
I could have a look if it's possible to transform the existing ct looping/matching (seems ad-hoc to me), into something that find the closest match. 
If I where to invest more hours into this, I think figuring out how to separate sub-class/class initialization and raise exception if no match found would pay off most.

vmuriart:
good from my side. The only thing I would change is the formatting of the tests, but i can take care of that later.

sigbjorn:
Sorry for the delay, I will try to find some time this weekend to dive into class/subclass, and/or also align call to constructor with the 'usual/standard' argument-matching.

denfromufa:
@sigbjorn to override .NET constructor we may need to use `__new__()`, not `__init__()` method.

@tonyroberts do you recall why you hooked up constructor to `__init__()`, not `__new__()` in `test_class.py`?

The problem with initializer is that it already receives an instance of the class, while new is actually the constructor of the class.

This may be a breaking change, but would enable subclassing immutable types also:

http://stackoverflow.com/a/674369/2230844

denfromufa:
@sigbjorn do you agree to MIT license for your contribution once we transition to it?

filmor:
@sigbjorn Feel free to reopen when you find the time to address the comments (in particular agreement to the MIT license).

Issue #165, opened 2016-02-24
NuGet package creation


AlexCatarino:
In order to use Python for .NET in a .NET app, we need to put Python.Runtime.dll in the references of the Visual Studio project. Unfortunately, this task has to be done manually, because we get this dll via pip (or downloading and compiling from this GitHub repository).

It would be very helpful to [create and publish a NuGet package](https://docs.nuget.org/create/creating-and-publishing-a-package) for pythonnet. If a NuGet package is generated at the end of a successful build (as suggested in the previous link), it would be quite easy to maintain a updated NuGet package for pythonnet.

denfromufa:
How about python version?

AlexCatarino:
A NuGet package for every supported version like pip/wheels.

denfromufa:
Where would the flag be set which version of pythonnet to get from Nuget?
Automagically detect system python or build flag like in pythonnet?

Note that there is limited chocolatey package for pythonnet.

How about clr.pyd?

denfromufa:
It is possible to push to nuget directly from appveyor. @tonyroberts are you willing to set this up in virtual machine settings?

tonyroberts:
@denfromufa the clr.pyd file is not needed for embedding python, only for calling .net from python, so no need for it to be in nuget. I don't know enough about how packages should be uploaded to nuget wrt different variants of the same version (eg pythonnet 2.1 would have different packages for each different python version). If you want to investigate further and make suggestions about how the appveyor build should be set up I'm happy to make the changes.

dmitriyse:
Please check my alpha binary release. It can be packed as nuget in future. (https://github.com/dmitriyse/pythonnet/releases)

denfromufa:
@dmitriyse IMO your bundle could be useful to @bc3tech for his chocolatey package, which is currently only limited to 32 bit python 2.7.

denfromufa:
https://chocolatey.org/packages/pythonnet

dmitriyse:
I released better implementation of Python.Runtime.dll loader. Feel free to check and use.
https://github.com/dmitriyse/pythonnet/releases/tag/v2.2.0-bin-beta

dmitriyse:
Hi everybody. I started work that allows to separate  Python.Runtime.dll from python specific interop methods. This efforts will finally allows to load appropriate Python.Runtime.Interop-PyXX-UCS-X.dll by simply Assembly.LoadFromFile without any AssemblyResolve handling magic.
[...]

PR #244, opened 2016-07-22
Added configurable Python.Runtime.dll loader.

dmitriyse:
Added project that builds Python.Config.dll with different Python.Runtime.dll version stored in resources.

If you put bootstrap code:

``` C#
        static Program()
        {
            Console.WriteLine("Starting application...");
            // Required to be placed in the static constructor for Mono.
            PythonConfig.EnsureInitialized();
        }

        [STAThread]
        private static int Main(string[] args)
        {
            // Mono workaround required to fix AssemblyResolve + EntryPoint class bug.
            // Classes that was referenced from EntryPoint class cannot use assemblies resolved through "AssemblyResolve"
            Action monoWorkaround = () =>
                {
                    try
                    {
                        // You should put this initialized only if some component starting to use it before first application configuration file read attempt.
                        // So in rare cases.
                        PythonEngine.Initialize();
```

to your project and switch off "copy local" property on Python.Runtime.dll assembly, then different version of Python.Runtime.dll will be loaded.
You also should put version into configuration file:

``` XML
<?xml version="1.0" encoding="utf-8" ?>
<configuration>
  <configSections>
    <section name="pythonConfig" type="Python.Config.PythonConfigSection, Python.Config, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null"/>
  </configSections>
  <pythonConfig pythonVersion="3.5"/>
</configuration>
```

By using this approach great Python.Net library will be able to be linked thorugh nuget.
(See https://github.com/pythonnet/pythonnet/issues/165)

### DIFF
--- a/src/runtime/pyobject.cs
+++ b/src/runtime/pyobject.cs
@@ -126,17 +126,7 @@ namespace Python.Runtime
                 disposed = true;
             }
         }
-
-        public long RefCount
-        {
-            get
-            {
-                if(!disposed)
-                    return Runtime.Refcount(obj);
-                return -1;
-            }
-        }
-
+        
         public void Dispose()
         {
             Dispose(true);


tonyroberts:
I like the idea in principle, but in order for it to work in all cases you would have to embed a lot of different runtimes (I think you only added x64 builds of 2.7 and 3.5). Don't forget that especially for linux people use versions of python with all sorts of different ABI flags, so it's not only the different python versions that have to be added.

The binaries should not be checked in. I think they should be built before this new component each time instead.

Including all the binaries adds a lot of bloat that some people won't want. Have you considered creating this as a separate repo (which could be under the pythonnet organization)? You could use CI to build it, picking up the latest pythonnet binaries from AppVeyor and travis (you have have to add some new builds to get the full set of binaries if you intend to support all python variants that are out there).

dmitriyse:
Unfortunately i have no enough time to maintain "Binaries" repo project. I know that there are more than 10 actively used build variants for pythonnet. But I can maintain, update and publish only distributions that i will use in my commercial development. Probably someone else can do effort that mentioned by tonyroberts. Currently I work on Ubuntu 16 and Windows, this would be most portable platforms for the near future. I will maintain this reduced binary distribution in my fork.

denfromufa:
@dmitriyse would you like if I create a repository "pythonnetbin" in pythonnet for these bundled binaries?

I see that you keep updating your bundles.

Note that I'm also admin on chocolatey for pythonnet package and can add you there.

After updating on chocolatey we need to add binaries on nuget.

My first thought is that Linux and Windows can be separate bundles. Then we can also support manylinux wheels. I'm not sure if this conflicts with your deployment?

Another idea is in the bootstrapper logic to check for all necessary flags (OS, python version, bitness, ABI flags) before loading the corresponding bundle or throwing error about unsupported combination of flags.

I understand that you cannot spend much time on this, but just wanted to have official way of distributing these bundles, since this benefits everyone!

denfromufa:
@dmitriyse also do you agree to MIT license for your contribution once we transition to it?

dmitriyse:
Hi!. Yes, it will be cool to migrate to MIT, so sure.

dmitriyse:
I performed big research work and finally get it work : autodetection for ubuntu, ubuntu+miniconda, windows.
I can help to somebody with received knowledge and ideas. But i still too busy to support additional use cases.

Main use case of my distribution is to build-in python into some CrossPlatform C# .Net 4.5.1 application.
Current solution is well suited to create nuget package for MSBuild build system, but some additional work required.

Typical scenario will be:

1) nuget package on installation should inject Python.Runtime.dll reference (without copy to output directory) Python.Bootstrapper.dll reference into project and add Python.Runtime.zip file as content file (for copy to output directory).
2) User should add initialize and shutdown code like this:
```CSharp
namespace Python.Bootstrapper.Test
{
    using System;
    using System.IO;

    using Runtime;
    public class Program
    {
        static Program()
        {
            Console.WriteLine("Starting application...");
            
            // Required to be placed in the static constructor for Mono.
            var loadedLib = PythonRuntimeBootstrapper.DetectPythonAndRegisterRuntime();
            Console.WriteLine($"{loadedLib} runtime loaded.");
        }
        static int Main(string[] args)
        {
            // Mono workaround required to fix AssemblyResolve + EntryPoint class bug.
            // Classes that was referenced from EntryPoint class cannot use assemblies resolved through "AssemblyResolve"
            Action monoWorkaround = () =>
            {
                try
                {
                    // You should put this initialized only if some component starting to use it before first application configuration file read attempt.
                    // So in rare cases.
                    //// PythonConfig.EnsureInitialized();


                    PythonEngine.Initialize();

                    try
                    {
                        using (Py.GIL())
                        {
                            dynamic sysModule = Py.Import("sys");
                            Console.WriteLine("Python engine version:");
                            Console.WriteLine(sysModule.version);
                        }
                    }
                    finally
                    {
                        PythonEngine.Shutdown();
                    }
                }
                catch (Exception ex)
                {
                    Console.WriteLine(ex);
                }

            };

            monoWorkaround();

            return 0;
        }
    }
}
```

3) All other magic will be performed automatically in runtime.

denfromufa:
@dmitriyse can you please let me know when is this ready for testing?

dmitriyse:
I will publish RC bin distribution soon. I found floating bug in 2.2.0-Beta under Mono. Currently I am testing fixes. RC version should be mature enough for production use (at least our product will use it).

dmitriyse:
RC release ready https://github.com/dmitriyse/pythonnet/releases/tag/v2.2.0-bin-RC
You can freely use it for testing.

denfromufa:
@dmitriyse I create a pythonnet organization's repository called pythonnetbin for your binaries:

https://github.com/pythonnet/pythonnetbin

filmor:
Would it be an option to have this as a separate project and just merge the essential parts here?

PR #268, opened 2016-09-26
quick fix for old nuget can't find packages

t3476:
In Linux with Mono, NuGet leaves error message:
  Unable to find version '1.2.6' of package 'UnmanagedExports'.
It can be simply solved by updating nuget.exe itself, but that's not obvious through error message, so I modified setup.py.


denfromufa:
@t3476 do you agree to dual Zope 2.0 and MIT license for your contribution?

t3476:
Yes. I give up all rights to follow whatever license.

PR #273, opened 2016-10-18
Calling isinstance(x,t) repeatedly when x or t is not actually a CLR …

ArvidJB:
…object will leak reference counts to PyFalse, resulting in memory corruption once the reference count drops to zero and PyFalse gets freed

### DIFF
--- a/src/runtime/metatype.cs
+++ b/src/runtime/metatype.cs
@@ -281,8 +281,10 @@ namespace Python.Runtime
         {
             ClassBase cb = GetManagedObject(tp) as ClassBase;
 
-            if (cb == null)
+            if (cb == null) {
+                Runtime.Incref(Runtime.PyFalse);
                 return Runtime.PyFalse;
+            }
 
             using (PyList argsObj = new PyList(args))
             {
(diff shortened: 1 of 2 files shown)


ArvidJB:
If you run the simple test I added you will most likely get a crash. Every invocation of DoInstanceCheck will return PyFalse, from this callstack:
`	Python.Runtime.dll!Python.Runtime.MetaType.DoInstanceCheck(System.IntPtr tp, System.IntPtr args, bool checkType) Line 281	C#`
` 	Python.Runtime.dll!Python.Runtime.MetaType.__instancecheck__(System.IntPtr tp, System.IntPtr args) Line 318	C#`
` 	[Native to Managed Transition]	`
` 	python35.dll!PyCFunction_Call(_object * func, _object * args, _object * kwds) Line 145	C`
` 	python35.dll!PyObject_Call(_object * func, _object * arg, _object * kw) Line 2167	C`
` 	python35.dll!PyObject_CallFunctionObjArgs(_object * callable, ...) Line 2446	C`
` 	python35.dll!PyObject_IsInstance(_object * inst, _object * cls) Line 2640	C`
` 	python35.dll!builtin_isinstance(PyModuleDef * module, _object * args) Line 624	C`
In Python 3.5.2 which I am running PyObject_IsInstance contains this code which decrements the reference count of the returned value
`        res = PyObject_CallFunctionObjArgs(checker, inst, NULL);`
`        Py_LeaveRecursiveCall();`
`        Py_DECREF(checker);`
`        if (res != NULL) {`
`            ok = PyObject_IsTrue(res);`
`            Py_DECREF(res);`
`        }`
And if you run this often enough the reference count drops to zero and "False" get freed which almost immediately crashes my python interpreter.

denfromufa:
@ArvidJB looks good to me and I confirmed that test this crashes the python interpreter - nice catch! Do you agree to MIT license once we transition? This code appears to be from @filmor, so we will let him review this.

filmor:
Yep, looks a lot more correct than before, thanks for catching this :)

ArvidJB:
Great, thanks for the quick turnaround!
I agree with both the ZPL and the MIT license for this change (and potential future changes).

PR #274, opened 2016-10-18
Strings should be able to be passed to functions expecting Object

ArvidJB:
We discovered this problem during our Python 3 conversion. This used to work with Python 2 where we had a roundtrip of
call C# method that returns C# string -> Python unicode -> pass Python unicode to C# function which takes object
Now in Python 3 this no longer works since the C# string is converted to a Python string.

### DIFF
--- a/src/runtime/methodbinder.cs
+++ b/src/runtime/methodbinder.cs
@@ -364,9 +364,11 @@ namespace Python.Runtime
                                     if (!typematch)
                                     {
                                         // this takes care of enum values
+                                        // special case: string is assignable to object
                                         TypeCode argtypecode = Type.GetTypeCode(pi[n].ParameterType);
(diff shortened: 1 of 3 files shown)


denfromufa:
@ArvidJB i think this is related to this issue:

https://github.com/pythonnet/pythonnet/issues/203

If yes, then it is more complicated due to various method overloading cases.

ArvidJB:
@denfromufa yes, this looks like it's the same issue, should we continue discussion there?

denfromufa:
Yes, let's make that a central place for this `System.Object` issue. Also
there is meta-issue linked there, which links all method overloading issues.

Let's also discuss the plan forward before anyone starts implementing this.
I spent a lot of time thinking and reading on this.

PR #275, opened 2016-10-19
Remove printing if Decref is called with NULL. Rename Decref/Incref to XDecref/XIncref

ArvidJB:
self.repr is set in tp_repr() when called (which also increases the reference count). But if it's not populated we should not decrease the reference count

### DIFF
--- a/src/runtime/classbase.cs
+++ b/src/runtime/classbase.cs
@@ -57,7 +57,7 @@ namespace Python.Runtime
             {
                 Type t = target.MakeGenericType(types);
                 ManagedType c = (ManagedType)ClassManager.GetClass(t);
-                Runtime.Incref(c.pyHandle);
+                Runtime.XIncref(c.pyHandle);
                 return c.pyHandle;
             }
 
@@ -71,7 +71,7 @@ namespace Python.Runtime
         public static IntPtr tp_richcompare(IntPtr ob, IntPtr other, int op) {
             if (op != Runtime.Py_EQ && op != Runtime.Py_NE)
             {
-                Runtime.Incref(Runtime.PyNotImplemented);
+                Runtime.XIncref(Runtime.PyNotImplemented);
                 return Runtime.PyNotImplemented;
             }
 
@@ -86,14 +86,14 @@ namespace Python.Runtime
             }
 
             if (ob == other) {
-                Runtime.Incref(pytrue);
+                Runtime.XIncref(pytrue);
                 return pytrue;
             }
 
             CLRObject co1 = GetManagedObject(ob) as CLRObject;
             CLRObject co2 = GetManagedObject(other) as CLRObject;
 			if (null == co2) {
-				Runtime.Incref(pyfalse);
+				Runtime.XIncref(pyfalse);
 				return pyfalse;
 			}
 
@@ -101,11 +101,11 @@ namespace Python.Runtime
             Object o2 = co2.inst;
 
             if (Object.Equals(o1, o2)) {
-                Runtime.Incref(pytrue);
+                Runtime.XIncref(pytrue);
                 return pytrue;
             }
 
-            Runtime.Incref(pyfalse);
+            Runtime.XIncref(pyfalse);
             return pyfalse;
         }
 #else
@@ -237,11 +237,11 @@ namespace Python.Runtime
             IntPtr dict = Marshal.ReadIntPtr(ob, ObjectOffset.DictOffset(ob));
             if (dict != IntPtr.Zero)
             {
-                Runtime.Decref(dict);
+                Runtime.XDecref(dict);
             }
             Runtime.PyObject_GC_UnTrack(self.pyHandle);
             Runtime.PyObject_GC_Del(self.pyHandle);
-            Runtime.Decref(self.tpHandle);
+            Runtime.XDecref(self.tpHandle);
             self.gcHandle.Free();
         }
     }
--- a/src/runtime/classderived.cs
+++ b/src/runtime/classderived.cs
@@ -81,7 +81,7 @@ namespace Python.Runtime
             FieldInfo fi = obj.GetType().GetField("__pyobj__");
             CLRObject self = (CLRObject)fi.GetValue(obj);
 
-            Runtime.Incref(self.pyHandle);
+            Runtime.XIncref(self.pyHandle);
 
             // when the C# constructor creates the python object it starts as a weak
             // reference with a reference count of 0. Now we're passing this object
@@ -153,7 +153,7 @@ namespace Python.Runtime
             HashSet<string> pyProperties = new HashSet<string>();
             if (py_dict != IntPtr.Zero && Runtime.PyDict_Check(py_dict))
             {
-                Runtime.Incref(py_dict);
+                Runtime.XIncref(py_dict);
                 using (PyDict dict = new PyDict(py_dict))
                 using (PyObject keys = dict.Keys())
                 {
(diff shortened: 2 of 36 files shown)


denfromufa:
@filmor actually `Decref` only it prints `Decref(NULL)` if `IntPtr.Zero`. We have a lot of cases where `IntPtr.Zero` is checked before `Decref`. But I agree that use case here is missing.

Also there are 2 places where Runtime.Decref(self.repr) is called in the same file.

ArvidJB:
The check was the problem, actually: it calls
`DebugUtil.Print("Decref(NULL)");`
which tries to print to the Console, which sometimes throws an
Exception thrown: 'System.IO.IOException' in mscorlib.dll
Additional information: The handle is invalid.
then (for some reason I don't fully understand) threw

ArvidJB:
I found the reason for the IOException: it is due to this:
> Console class members that work normally when the underlying stream is directed to a console might throw an exception if the stream is redirected, for example, to a file. Program your application to catch System.IO.IOException exceptions if you redirect a standard stream. You can also use the IsOutputRedirected, IsInputRedirected, and IsErrorRedirected properties to determine whether a standard stream is redirected before performing an operation that throws an System.IO.IOException exception.

from https://msdn.microsoft.com/en-us/library/system.console.aspx?f=255&MSPPError=-2147217396

Even without those problems the spurious "Decref("NULL")" error messages are pretty annoying.

Any objections to this change?

denfromufa:
I agree with this change but why not change in both places?

https://github.com/pythonnet/pythonnet/blob/44973c613dc1a7e1af23a84cacfcdcb1157e6caf/src/runtime/constructorbinding.cs#L239
https://github.com/pythonnet/pythonnet/blob/44973c613dc1a7e1af23a84cacfcdcb1157e6caf/src/runtime/constructorbinding.cs#L150

filmor:
Well, isn't the right thing to do then fixing/removing the debug log and documenting that `Incref` and `Decref` are safe to call on `null`?

denfromufa:
Is it really safe to call `Decref` on `IntPtr.Zero`? Why does pythonnet
have these ~9 checks all over the code?

Find all "if.*\!\=.*IntPtr.Zero.*\n.*(\n.*){0,7}Runtime.Decref", Regular
expressions, Find Results 1, Entire Solution, ""
  C:\Python\pythonnet\pythonnet\src\runtime\classbase.cs(238):
 if (dict != IntPtr.Zero)
  C:\Python\pythonnet\pythonnet\src\runtime\classderived.cs(850):
             if (dict != IntPtr.Zero)
  C:\Python\pythonnet\pythonnet\src\runtime\converter.cs(297):
           if (p != IntPtr.Zero)
  C:\Python\pythonnet\pythonnet\src\runtime\converter.cs(307):
       if ((c != IntPtr.Zero) && (c != value))
  C:\Python\pythonnet\pythonnet\src\runtime\exceptions.cs(196):
       if (op != IntPtr.Zero)
  C:\Python\pythonnet\pythonnet\src\runtime\managedtype.cs(51):
       if ((e != IntPtr.Zero) && (e != ob))
  C:\Python\pythonnet\pythonnet\src\runtime\metatype.cs(177):            if
(init != IntPtr.Zero)
  C:\Python\pythonnet\pythonnet\src\runtime\methodbinder.cs(336):
                     if (pyoptype != IntPtr.Zero)
  C:\Python\pythonnet\pythonnet\src\runtime\methodbinding.cs(164):
                   if (baseMethod != IntPtr.Zero)
  C:\Python\pythonnet\pythonnet\src\runtime\pythonexception.cs(155):
             if (_pyTB != IntPtr.Zero)
  Matching lines: 10    Matching files: 9    Total files searched: 99


On Thu, Oct 20, 2016 at 9:16 AM, Benedikt Reinartz <notifications@github.com

ArvidJB:
@denfromufa i made the other change as well. I missed it since it looks like we were not hitting that problem in ConstructorBinding.tp_dealloc in our codebase.
@filmor I think this is actually a useful message since it basically tells you that somewhere you did some inconsistent reference counting.

filmor:
@denfromufa I don't know why those checks are in, but just look at the function, both implementations (with `Py_DEBUG` set and without) check explicitly for the pointer being 0.
@ArvidJB Not really. If our reference counting is actually off in an unsafe way we would never notice, as pointers are not set to 0 when they are freed by `Decref`.

I vote we remove the message instead and document the behaviour.

ArvidJB:
@denfromufa sounds good to me.
I will try to see how easy it is to fix the general problem of writing to the Console. Maybe it's as easy as writing to stdout directly?

denfromufa:
@filmor do you mean that this line below is the check for the pointer being zero?

https://github.com/pythonnet/pythonnet/blob/44973c613dc1a7e1af23a84cacfcdcb1157e6caf/src/runtime/runtime.cs#L579

Where is equivalent check for Py_DEBUG version? I did not find this!

@ArvidJB somehow my previous message that I sent using gmail was pushed/updated only 9 hours ago on github. But in my gmail I sent it 5 days ago, before @filmor 's last message. This is just to clarify the sequence of messages we had 5 days ago.

@tonyroberts @vmuriart do you guys have any better suggestions for this pull request?

In my opinion, we need to address 2 questions:

1. Is it safe to Decref without checking for IntPtr.Zero? Hopefully someone can write a test for this, if this is safe. And then all ~9 manual checks that I referenced above have to be removed from pythonnet codebase.

2.  Do we need a debug message when trying to decref IntPtr.Zero? If yes, how to handle it safely for redirected outputs?

tonyroberts:
in the Python C API there are two function for decrementing the reference count - Py_DECREF and Py_XDECREF. The the debug build these are exposed as functions, but in the non-debug build they are macros - hence why in pythonnet the exposed function version is only used when Py_DEBUG is set.

Py_DECREF is *not* safe to call on a null pointer, but Py_XDECREF is.

If you want to be consistent with the C API I would suggest adding a C# wrapper/implementation for Py_XDECREF and use that where you are not certain that the pointer will not be null.

It is not necessary to log anything is Py_XDECREF is called with a null pointer. Logging in Py_DECREF that a null pointer was passed is useful as this would cause a crash if it behaved the same way as the C Python implementation (or if the Py_DEBUG code was used).

tonyroberts:
Just saw filmor's earlier comment that the exported Py_DecRef function is actually Py_XDECREF.

What I would do is rename Decref to Py_XDECREF and keep the null check and remove the logging - then it's clear to everyone that it can be called safely with a null pointer.

denfromufa:
@tonyroberts thanks for pointing to @filmor 's edit above. Indeed, I did not notice it either. Especially, since I was looking at my email. So new messages about edits are generally better + optional in-place edits.

@ArvidJB now I completely agree with the latest comments from @filmor and @tonyroberts. Let me know if anything is not clear? Are you planning to continue with this pull request?

ArvidJB:
@denfromufa That makes sense, I will try to make that change. Should I open a new pull request for that or is it okay to continue on this one?

denfromufa:
We can continue on this pull request and from your branch. Once ready, we
can just squash your commits into "logical unit(s)".

ArvidJB:
I renamed Incref/Decref to XIncref/XDecref instead of Py_XINCREF/Py_XDECREF since that felt like a more C# appropriate name. I removed the additional NULL check I had added and also removed the code that prints on NULL in Decref.

denfromufa:
looks like @tonyroberts approved your commit @ArvidJB 

@filmor do you agree with these changes?

filmor:
I agree, but please rename and squash the PR before merging.

denfromufa:
@filmor now we can squash or rebase and merge pull requests with one button click (select a drop-down on merge pull request). More info in this article:

https://github.com/blog/2141-squash-your-commits

PR #277, opened 2016-10-27
Assemblymanager thread safety

ArvidJB:
We observed a spurious crash when a spawned .NET thread loaded some additional assemblies triggered the AppDomain.AssemblyLoad event handler. This modified the namespaces Dictionary and led to an exception

> System.InvalidOperationException: Collection was modified; enumeration operation may not execute.
>    at System.ThrowHelper.ThrowInvalidOperationException(ExceptionResource resource)
>    at System.Collections.Generic.Dictionary`2.KeyCollection.Enumerator.MoveNext()
>    at Python.Runtime.AssemblyManager.GetNames(String nsname) in C:\git\pythonnet\src\runtime\assemblymanager.cs:line 442

when we called import clr again from the Python thread. I was able to write a test which reproduces this (most of the time).
As a fix I propose to make namespaces a ConcurrentDictionary. The same issue should also apply to the assemblies list. Unfortunately there is no ConcurrentList, so I added a lock. All the operations under the assemblies lock seem simple enough to not cause any deadlocks by triggering the same codepath on a different thread.

### DIFF
--- a/src/tests/test_module.py
+++ b/src/tests/test_module.py
@@ -65,9 +65,11 @@ class ModuleTests(unittest.TestCase):
         import System
         self.assertEquals(type(System.__dict__), type({}))
         self.assertEquals(System.__name__, 'System')
-        # the filename can be any module from the System namespace (eg System.Data.dll or System.dll)
-        self.assertTrue(fnmatch(System.__file__, "*System*.dll"),
-                        "unexpected System.__file__" + System.__file__)
+        # the filename can be any module from the System namespace
+        # (eg System.Data.dll or System.dll, but also mscorlib.dll)
+        system_file = System.__file__
+        self.assertTrue(fnmatch(system_file, "*System*.dll") or fnmatch(system_file, "mscorlib.dll"),
+                        "unexpected System.__file__" + system_file)
         self.assertTrue(System.__doc__.startswith("Namespace containing types from the following assemblies:"))
         self.assertTrue(self.isCLRClass(System.String))
         self.assertTrue(self.isCLRClass(System.Int32))
--- a/src/tests/test_module.py
+++ b/src/tests/test_module.py
@@ -360,9 +360,14 @@ class ModuleTests(unittest.TestCase):
         # spin up .NET thread which loads assemblies and triggers AppDomain.AssemblyLoad event
         ModuleTest.RunThreads()
         time.sleep(1e-5)
-        # call import clr, which in AssemblyManager.GetNames iterates through the loaded types
         for i in range(1, 100):
+            # call import clr, which in AssemblyManager.GetNames iterates through the loaded types
             import clr
+            # import some .NET types
+            from System import DateTime
+            from System import Guid
+            from System.Collections.Generic import Dictionary
+            dict = Dictionary[Guid,DateTime]()
         ModuleTest.JoinThreads()
 
 
--- a/src/tests/test_module.py
+++ b/src/tests/test_module.py
@@ -68,8 +68,8 @@ class ModuleTests(unittest.TestCase):
         # the filename can be any module from the System namespace
         # (eg System.Data.dll or System.dll, but also mscorlib.dll)
         system_file = System.__file__
-        self.assertTrue(fnmatch(system_file, "*System*.dll") or fnmatch(system_file, "mscorlib.dll"),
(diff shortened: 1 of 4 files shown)


denfromufa:
This is failing on Travis CI (Linux) for 2 Python versions on your added assert test.

Concurrent list is not possible by design, but can't we use anything else from System.Collections.Concurrent or at least subclass list with locked methods?

http://stackoverflow.com/a/9995339/2230844

This is to prevent manual lock maintenance, especially when people would not be aware of multi-threading support.

ArvidJB:
Thanks for the suggestions: I will fix the test to be more predictable and not have any collateral impact on other tests and create a little wrapper for List<>.

denfromufa:
@tonyroberts @filmor @vmuriart this pull request looks good to me. please review!

@ArvidJB is the list with locks your own implementation or you copied it from somewhere else?

ArvidJB:
@denfromufa that List is my own implementation (although I did read through the stackoverflow answers and some related discussions to come up with it).

ArvidJB:
@denfromufa good point, I removed SupportsRecursion. I also added a read lock around _list.Count which I had missed earlier.

denfromufa:
@ArvidJB can you add more imports in the test? "import clr" is very special compared to imports of Python modules and .NET assemblies.

src/tests/test_module.py
```diff
@@ -66,7 +66,8 @@ def testModuleInterface(self):
         self.assertEquals(type(System.__dict__), type({}))
         self.assertEquals(System.__name__, 'System')
         # the filename can be any module from the System namespace (eg System.Data.dll or System.dll)
-        self.assertTrue(fnmatch(System.__file__, "*System*.dll"))
+        self.assertTrue(fnmatch(System.__file__, "*System*.dll"),
+                        "unexpected System.__file__" + System.__file__)
```
denfromufa: @ArvidJB can you add a comment why this was added in test?

ArvidJB:
@denfromufa sure, added some more imports. I only added "import clr" because that's what caused our problems with its call to GetNames(), but the other code paths potentially could also have race conditions.

src/tests/test_module.py
ArvidJB: The System namespace is spread out across a couple of DLLs: mscorlib.dll, System.dll, and all the other System.*.dlls. In ModuleObject we iterate through all the known assemblies for a namespace and pick the last one for the filename. The iteration order would be the on the hash of the assembly, so it should have always been unpredictable whether you get mscorlib.dll or one of the others.
Let me change the test such that mscorlib.dll is also accepted.

ArvidJB: PS: I added the error message because I could not reproduce this on my machine and wanted to know what value it got in the integration tests.

denfromufa:
@tonyroberts @filmor @vmuriart this pull request looks good to me. please review, at least one of you!

@ArvidJB thanks for putting this together!

ArvidJB:
@denfromufa thanks for reviewing!
@tonyroberts @filmor @vmuriart, is this good to go in?

vmuriart:
The last commit confused me, did we merge `master` into the `pr` before the `pr` is merged into `master`? Otherwise looks fine to me.

denfromufa:
I updated this branch with the master to re-run the tests.

denfromufa:
@ArvidJB I will wait couple days during release of pythonnet v2.2.0 before merging this. This also gives some time to @filmor @tonyroberts in case they would like to review this. I also need to check embedding tests, which are currently not setup in CI.

filmor:
I did review this, didn't see any glaring errors. But this threading stuff is hard :)

How about using [SynchronizedCollection<T>](https://msdn.microsoft.com/en-us/library/ms668265.aspx) instead of rolling your own implementation?

ArvidJB:
@filmor I would have preferred not to write this myself but I did not find a good match in the .NET framework. SynchronizedCollection is a weird class, it's in System.ServiceModel for some reason. And it does not provide thread-safe enumeration:
https://referencesource.microsoft.com/#System.ServiceModel/System/ServiceModel/SynchronizedCollection.cs,3a568f4594ad20f5,references
It only grabs the lock when you get the enumerator but when you iterate you are still stuck with the iterator of the underlying List which can be modified concurrently.

PR #286, opened 2016-11-08
Fix exception pickling

### LINKED ISSUE #284: Trying to pickle an exception crashes the interpreter
The following code apparently tries to write to a `null` pointer:

    import clr
    import pickle
    from System import Exception
    pickle.dumps(Exception())

I'm investigating, where it fails exactly but I'm open to hints :)

Small addition: This works fine for simple objects like `System.Object()`

filmor:
Fixes #284.

This PR contains 3 componenents:

1. Remove all of the Py<2.5 workarounds that wrap exceptions in old-style class instances
2. Fix #284 by moving the exception args into the corresponding slot in the `BaseException` object instead of faking the functionality by overloading `__getattr__`
3. Remove the `__getattr__` overload altogether since the only functionality it provides (lookup of `.message`) is deprecated in Python >= 2.6

### DIFF
--- a/src/runtime/clrobject.cs
+++ b/src/runtime/clrobject.cs
@@ -30,6 +30,9 @@ namespace Python.Runtime
             this.pyHandle = py;
             this.gcHandle = gc;
             inst = ob;
+
+            // Fix the BaseException args slot if wrapping a CLR exception
+            Exceptions.SetArgs(py);
         }
 
 
--- a/src/runtime/converter.cs
+++ b/src/runtime/converter.cs
@@ -156,20 +156,7 @@ namespace Python.Runtime
             switch (tc)
             {
                 case TypeCode.Object:
-                    result = CLRObject.GetInstHandle(value, type);
-
-                    // XXX - hack to make sure we convert new-style class based
-                    // managed exception instances to wrappers ;(
-                    if (Runtime.wrap_exceptions)
-                    {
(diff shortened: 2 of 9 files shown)


filmor:
@vmuriart @denfromufa @tonyroberts Please review.

denfromufa:
@filmor can you add tests compatible with both python 2 and 3 that use cPickle:

```
try:
   import cPickle as pickle
except:
   import pickle
```

filmor:
@denfromufa Done. Please check :)

denfromufa:
So is this a "feature" or "bug fix"? If latter, then let's merge this for
v2.2.0! If not, can we wait until I release binaries for v2.2.0?

Have you ran the embedding tests to make sure that exceptions do not break
on the other side of the bridge?

filmor:
I haven't run the embedded tests, I'll check how to do that.

This is definitely a bug fix, and a pretty critical one from my point of view (as said, this was killing my worker processes in production).

denfromufa:
Ok, please let me know if you have any luck running embedding tests and
then we can merge this for v2.2.0.

BTW, why do you pickle exceptions - really seems a wild idea to me?

filmor:
The tests work against Python 3.5 (though they have various issues, I'll prepare a PR).

I pass the exceptions back from a multiprocessing worker to show them to the user, and during that they are automatically pickled.

PR #287, opened 2016-11-10
Implement writing the __cause__ attribute on exceptions

filmor:


denfromufa:
No tests? :)

filmor:
That's why it reads "WIP" ;)

filmor:
Tests added, could someone comment on whether the `XIncref` there is correct? I take it `GetInstHandle` doesn't increase the ref-count by itself?

denfromufa:
@filmor you can always stress-test for memory growth of this exception in a for-loop.

Looks like it is NOT needed to call XIncref like in this case:

https://github.com/pythonnet/pythonnet/blob/86937bb76f48207a8c97af6d52822cbd7c33dc27/src/runtime/pyobject.cs#L79

Note that PyObject(IntPtr tp) constructor does not incref the ref. count.

filmor:
@denfromufa @vmuriart @tonyroberts Please review.

tonyroberts:
I don't think an incref is necessary as the CLRObject constructor creates a new python object (so reference count of 1) and never decrefs it.

filmor:
Yep, changed that already.

Issue #288, opened 2016-11-16
can't open file 'tools/geninterop/geninterop.py': [Errno 2] No such file or directory


denfromufa:
https://github.com/pythonnet/pythonnet/issues/241#issuecomment-233685378

PR #289, opened 2016-11-16
Update setup.py to include geninterop.py in source distribution

denfromufa:
closes https://github.com/pythonnet/pythonnet/issues/288


denfromufa:
@filmor @tonyroberts @vmuriart please review.

PR #291, opened 2016-11-16
Py3 comparison operators

ArvidJB:
Looks like <, <=, >, >= implementations were dropped when moving from Python 2 to 3.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -316,7 +316,7 @@ if __name__ == "__main__":
                 sources.append(os.path.join(root, filename))
 
     for root, dirnames, filenames in os.walk("tools"):
-        for ext in (".exe"):
+        for ext in (".exe", ".py"):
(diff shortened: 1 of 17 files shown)


ArvidJB:
Actually the implementation of <, <=, >, >= seems to be broken for Python 2: it always returns -1:
`            if (Object.Equals(o1, o2))
            {
                return 0;
            }
            return -1;
`
I've disabled the tests I added for Python 2...

filmor:
If you'd cherry-pick https://github.com/pythonnet/pythonnet/commit/b709a4be0866d7797cff7002da8e02b5766faba2 on top we will use the same code for Python 2 and 3, so Py2 will also benefit from your change. `tp_compare` has been obsoleted by `tp_richcompare` a long time ago.

Issue #292, opened 2016-11-24
Adding new attributes to .Net classes is unstable


oyse:
I have a problem with adding attributes to .Net classes being unstable. E.g.

```
from System import TimeSpan

TimeSpan.my_method = lambda self: self.TotalMinutes
ts = TimeSpan.FromHours(1)
ts.my_method()
```

This code will sometimes work and other times fail with the exception 'TimeSpan' object has no attribute 'my_method'.

The failure rate is variable and for some scripts/program it will work all the time while for others it will not work at all. It seems to fail a lot more often on Python 3.5 than on Python 3.4.

If I look at either `dir(ts)` or `ts.__class__.__dict__` the method is there even when the call to my_method fails.

I have tried to look into the PythonNet code for clues, but as far as I can see the new attribute is registered by MetaType. tp_setattro without error, but the method is still not found.

filmor:
Closed by #293

PR #293, opened 2016-11-24
Fix attribute setting

filmor:
- When modifying attributes on a class, PyType_Modified has to be called
- The result of a lookup may be non-NULL without being a descriptor (e.g. if we
  set it manually using `Class.attribute = 1`)

Fixes issue #292

### DIFF
--- a/src/runtime/metatype.cs
+++ b/src/runtime/metatype.cs
@@ -214,22 +214,30 @@ namespace Python.Runtime
             if (descr != IntPtr.Zero)
             {
                 IntPtr dt = Runtime.PyObject_TYPE(descr);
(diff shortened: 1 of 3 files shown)


filmor:
This currently fails tests like this:

    del DotNetClass.__getitem__

The tests expect an exception here. The method is not actually deleted, though, it will still work afterwards.

filmor:
@denfromufa @vmuriart @tonyroberts Please review :)

filmor:
One minor (additional) change in the last push: I changed it to call `PyType_Modified` unconditionally, the docs are very slim on whether this is needed :/

filmor:
@tonyroberts Is that a formal approval? ;)

tonyroberts:
@filmor haha I thought of it as a 'formal' approval, but I agree with the chanages :)

PR #294, opened 2016-11-25
Comparison operators as implemented by ArvidJB in #291

filmor:
I've rebased and squashed @ArvidJB's changes and removed the obsolete `tp_compare` implementation such that we get comparison operators for both Python 2 and 3.

I suggest we merge this one instead and close #291.

@tonyroberts @denfromufa @vmuriart @matthid Please review.


ArvidJB:
Thanks @filmor, LGTM

Issue #94, opened 2015-09-01
.NET OBject type error "value cannot be converted to Object"


andygraham:
Using Cpython 2.7.10 and pythonnet-2.1.0.dev1

I may be missing something as I am an IronPython user just playing to see if my Windows Forms code can run on CPython and PythonNet. 

To fully use Windows Forms with PythonNet it is convenient, even necessary at times, to treat any Python object as a .NET Object so that it can be passed as a parameter to a delegate. For example to invoke a Python function to run on a WinForms GUI thread from a different thread you need to use Control.BeginInvoke which takes an Object array for the method parameters. There is no need to access these objects from .NET code, merely to package and unpackage them as .NET Object types which IronPython has no trouble doing. An example of this failure

``` python
import clr, sys
from System import Object, Array
a = Array[Object]([sys]) ; print a[0]
TypeError: cannot convert value to Object
```

I was actually interested in passing frame objects which have the same problem.

denfromufa:
can you treat sys as Dynamic object?

denfromufa:
can you treat sys as Dynamic object?

denfromufa:
can you treat sys as Dynamic object?

denfromufa:
can you treat sys as Dynamic object?

andygraham:
I don't understand the question! This is all in Python code, not C#.

tonyroberts:
@denfromufa I think you've missed the point somewhat. When passing a Python object to a managed call it is converted to a managed object (eg a python float is converted to a managed Double). If you were writing managed code you could use a dynamic type to refer to the instance of the converted object, but that doesn't help in this case as there is no conversion available from the python object to a managed object in the first place.

A possible solution would be to change Converter::ToManagedValue (in converter.cs) to fallback to convert all Python objects without a reasonable conversion to a native managed type to a PyObject managed type. This looks simple enough, but there are a couple of issues that would raise that would need to be considered:
 - If the PyObject instance is cleaned up by the garbage collector it may block on acquiring the GIL and that may cause a deadlock as all other threads will be suspended (and one of those threads may have the GIL acquired).
 - The Python object may end up being deleted by the Python garbage collector while the managed instance still exists if there are no other Python references left to the object. This would result in the object being dereferenced by the PyObject destructor after the CPython object had already been deallocated.

Both of these problems can be resolved, but I just wanted to point them out as it's not a simple as going in and fixing ToManagedValue.

denfromufa:
[...]

PR #297, opened 2016-11-29
Fixed #296 Converter support for PyObject

### LINKED ISSUE #296: Python.Runtime.Converter support for converting from/to PyObject
I have a C# class, lets call it MyCSClass, that I intend to be inherited from a Python class and used like this:
```py
class MyPythonClass(MyCSClass):
   def __init__(self):
      self.MyAttribute = 5
      self.Callback(self)
```
When an instance of MyPythonClass gets passed to C#, it gets converted to a MyCSClass and is stripped of all the attributes. Even if the method called in CS takes it as a dynamic or PyObject. An example of this could be the __init__ method above calling into the following C# code:
```cs
public class MyCSClass {
   protected void Callback(dynamic self){
       Console.WriteLine("A dynamic attribute: {0}", self.MyAttribute);
   }
}
```

However, the instance assigned to 'self' is an MyCSClass and not a dynamic MyPythonClass. Replace ```dynamic``` with ```PyObject``` and you get an exception. If suggest the the converter should support PyObject, so that callbacks like these are made possible.

Then I would have to do this instead, which is OK for me:
```cs
public class MyCSClass {
   protected void Callback(PyObject self){
       Console.WriteLine("A dynamic attribute: {0}", ((dynamic)self).MyAttribute);
   }
}
```

Python Version: ```Python 2.7.11 (v2.7.11:6d1b6a68f775, Dec  5 2015, 20:32:19) [MSC v.1500 32 bit (Intel)] on win32```
OS: Windows 7 Enterprise

rmadsen-ks:
Work in progress.

Note, testing has not been done, except for exploratory style. How is testing supposed to be done?


denfromufa:
@rmadsen-ks you can look at callbacktest.cs and test_callback.py for testing this.

denfromufa:
@rmadsen-ks do you agree to transition from Zope to MIT license?

rmadsen-ks:
@denfromufa, Test passed. I have added a test verifying that the fix does what it should. 

I agree to transition to MIT license.

src/tests/test_class.py
```diff
@@ -6,7 +6,7 @@
 import System
 import six
 from Python.Test import ClassTest
-from System.Collections import Hashtable
+from System.Collections import Hashtable
```
denfromufa: where is this needed?

rmadsen-ks: It was not needed by me, but it seems I had added a windows line-ending. Its fixed now.

denfromufa:
how about garbage collection, is CLR or Python responsible? If Python, then I do not see Decref, so it would leak memory?

rmadsen-ks:
I have tested callbacks rapidly and the ref count does not increase permanently. Neither does it seem to leak when many new objects are made. I have tried creating 50M objects with no noticable leak.

 I should not need to Decref for the following reason:

As the comment in pyobject.cs says, the PyObject assumes ownership. So if we want to convert a python object to a PyObject using ToManagedValue, we must increase the refcount, because we are giving it to the constructor of PyObject. So I implemented that part. The PyObject decreases the ref count when the garbage collector cleans it up, which happens eventually. So when a new C# PyObject has been created from a python object, it will eventually be cleaned up by the .NET GC. Is some cases ```using(){}``` can also be used, to give more predictable memory performance. 

In the same way, if we want to convert C# PyObject to a python object, we increase the ref count, because now Python has an extra reference to it, so it needs ownership. Python will clean that up on its own when it goes out of scope.

denfromufa:
@tonyroberts @vmuriart @filmor please review!

src/testing/callbacktest.cs
```diff
@@ -27,5 +27,19 @@ public string Call_simpleDefaultArg_WithEmptyArgs(string moduleName)
                 return module.simpleDefaultArg();
             }
         }
+    }
+
+    //==========================================================================
+    // Tests calling from Python into C# and back into Python using a PyObject.
+    // SelfCallbackTest should be inherited from a Python class.
```
rmadsen-ks: Comment has been changed.

denfromufa:
@rmadsen-ks this pull request does NOT resolve the issue described in #94:

```
dta@dta-Inspiron-N5050 ~/Downloads $ sudo pip install pythonnet-fix-develop-296-convert-PyObject.zip -U --force
Processing ./pythonnet-fix-develop-296-convert-PyObject.zip
Installing collected packages: pythonnet
  Found existing installation: pythonnet 2.2.0
    Uninstalling pythonnet-2.2.0:
      Successfully uninstalled pythonnet-2.2.0
  Running setup.py install for pythonnet ... done
Successfully installed pythonnet-2.2.0
You are using pip version 8.1.1, however version 9.0.1 is available.
You should consider upgrading via the 'pip install --upgrade pip' command.
dta@dta-Inspiron-N5050 ~/Downloads $ python
Python 2.7.6 (default, Jun 22 2015, 17:58:13) 
[GCC 4.8.2] on linux2
>>> import clr
>>> from System import Object, Array
>>> import sys
>>> a = Array[Object]([sys]) ; print a[0]
TypeError: value cannot be converted to Object
```

rmadsen-ks:
Hi, I'm not suggesting that it does. You want #94 fixed as part of this?

rmadsen-ks:
@denfromufa, is there anything I need to do to move this further or do I need to find another solution?

denfromufa:
@rmadsen-ks your pull request is completely valid.

filmor:
To me this also looks fine, but why do you introduce a new `RefCount` property? Either remove it or add tests for it, after that we can merge.

rmadsen-ks:
@filmor, I found it useful for verifying that I was not leaking reference counts. Without it, it is hard to get out the reference cound from C# as the NativeMethods are internal. It is not critical for me, so we can remove it.

denfromufa:
@rmadsen-ks can you please open a new PR? I made a mistake and merged this pull request without seeing @filmor concern.

rmadsen-ks:
It is here #306

Issue #119, opened 2016-01-21
Import clr fails on Linux64/Mono


rihardsk:
Running `python -c 'import clr' 2>&1` gives

```
*** Error in `python': double free or corruption (out): 0x00007ff2a2b41180 ***
Stacktrace:

  at <unknown> <0xffffffff>
  at (wrapper managed-to-native) Python.Runtime.Runtime.Py_Initialize () <0xffffffff>
  at Python.Runtime.Runtime.Initialize () <0x0003f>
  at Python.Runtime.PythonEngine.Initialize () <0x00047>
  at Python.Runtime.PythonEngine.InitExt () <0x0000f>
  at (wrapper runtime-invoke) object.runtime_invoke_void (object,intptr,intptr,intptr) <0xffffffff>

Native stacktrace:

        /usr/lib/libmonoboehm-2.0.so.1(+0xcce3a) [0x7ff2a417de3a]
        /lib/x86_64-linux-gnu/libpthread.so.0(+0x10d10) [0x7ff2a56ffd10]
        /lib/x86_64-linux-gnu/libc.so.6(gsignal+0x37) [0x7ff2a535a267]
        /lib/x86_64-linux-gnu/libc.so.6(abort+0x16a) [0x7ff2a535beca]
        /lib/x86_64-linux-gnu/libc.so.6(+0x78c53) [0x7ff2a539dc53]
        /lib/x86_64-linux-gnu/libc.so.6(+0x80c69) [0x7ff2a53a5c69]
        /lib/x86_64-linux-gnu/libc.so.6(cfree+0x4c) [0x7ff2a53a989c]
        /usr/lib/x86_64-linux-gnu/libpython2.7.so(PyString_InternInPlace+0x90) [0x7ff2a2c65ae0]
        /usr/lib/x86_64-linux-gnu/libpython2.7.so(PyString_InternFromString+0x2a) [0x7ff2a2c65b6a]
        /usr/lib/x86_64-linux-gnu/libpython2.7.so(PyType_Ready+0x1b25) [0x7ff2a2c6b6f5]
        /usr/lib/x86_64-linux-gnu/libpython2.7.so(_Py_ReadyTypes+0xb0) [0x7ff2a2c59360]
        /usr/lib/x86_64-linux-gnu/libpython2.7.so(Py_InitializeEx+0x76) [0x7ff2a2ce2106]
        [0x4181c411]

Debug info from gdb:


=================================================================
[...]

PR #300, opened 2016-12-04
Work around a broken shared detection on some platforms

denfromufa:
https://github.com/pythonnet/pythonnet/issues/119

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -150,11 +150,12 @@ class PythonNET_BuildExt(build_ext):
                 defines.append("MONO_OSX")
             else:
                 defines.append("MONO_LINUX")
-
+            
+            # Disable this check due to issues on Ubuntu
             # Check if --enable-shared was set when Python was built
-            enable_shared = get_config_var("Py_ENABLE_SHARED")
-            if enable_shared == 0:
-                defines.append("PYTHON_WITHOUT_ENABLE_SHARED")
+            # enable_shared = get_config_var("Py_ENABLE_SHARED")
+            # if enable_shared == 0:
+            defines.append("PYTHON_WITHOUT_ENABLE_SHARED")
 
         if hasattr(sys, "abiflags"):
             if "d" in sys.abiflags:
--- a/setup.py
+++ b/setup.py
@@ -10,7 +10,7 @@ from distutils.sysconfig import get_config_var
 from distutils.spawn import find_executable
 from distutils import log
 from platform import architecture
-from subprocess import Popen, CalledProcessError, PIPE, check_call
+from subprocess import Popen, CalledProcessError, PIPE, check_call, check_output
 from glob import glob
 import fnmatch
 import sys
--- a/setup.py
+++ b/setup.py
@@ -156,7 +156,7 @@ class PythonNET_BuildExt(build_ext):
             if enable_shared == 1:
                 # Double-check if libpython is linked dynamically with python
                 lddout = check_output(["ldd", sys.executable])
-                if 'libpython' not in lddout:
+                if b'libpython' not in lddout:
                     enable_shared = 0
             if enable_shared == 0:
                 defines.append("PYTHON_WITHOUT_ENABLE_SHARED")
--- a/setup.py
+++ b/setup.py
@@ -151,11 +151,15 @@ class PythonNET_BuildExt(build_ext):
             else:
                 defines.append("MONO_LINUX")
             
-            # Disable this check due to issues on Ubuntu
             # Check if --enable-shared was set when Python was built
-            # enable_shared = get_config_var("Py_ENABLE_SHARED")
-            # if enable_shared == 0:
-            defines.append("PYTHON_WITHOUT_ENABLE_SHARED")
+            enable_shared = get_config_var("Py_ENABLE_SHARED")
+            if enable_shared == 1:
+                # Double-check if libpython is linked dynamically with python
+                lddout = check_output(["ldd", sys.executable])
+                if 'libpython' not in lddout:
+                    enable_shared = 0
+            if enable_shared == 0:
+                defines.append("PYTHON_WITHOUT_ENABLE_SHARED")
 
         if hasattr(sys, "abiflags"):
             if "d" in sys.abiflags:
--- a/setup.py
+++ b/setup.py
@@ -150,15 +150,16 @@ class PythonNET_BuildExt(build_ext):
                 defines.append("MONO_OSX")
             else:
                 defines.append("MONO_LINUX")
-            
+
             # Check if --enable-shared was set when Python was built
             enable_shared = get_config_var("Py_ENABLE_SHARED")
-            if enable_shared == 1:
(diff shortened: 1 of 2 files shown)


mikofski:
This seems like an issue with xamarin mono project, _not_ .NET

Seems like a shame to have it always turned off, for a small percentage of PythonNET users

Why not do some sort of check for Linux or darwin, then disable, but if win32 use flag?

denfromufa:
@mikofski can you please explain how Mono is too blame in more details, especially if you can point to specific code?

denfromufa:
@tonyroberts @filmor @vmuriart please review. Note that I marked this pull request as release-blocking.

tonyroberts:
This change may fix a problem for some cases, but it will break it for others.

If that flag isn't reliable because of a bug in some distributions, why not check another way instead of ignoring it? Checking the output of ldd on the Python executable would tell you if it's been linked dynamically or not.

denfromufa:
@tonyroberts @filmor this was a bit of learning for me. See analysis below.

On Linux Mint it shows that python 2/3 are dynamically linked:

```
Python 2.7.6 (default, Jun 22 2015, 17:58:13) 
[GCC 4.8.2] on linux2
>>> from distutils.sysconfig import get_config_var
>>> get_config_var("Py_ENABLE_SHARED")
1

Python 3.4.3 (default, Sep 14 2016, 12:36:27) 
[GCC 4.8.4] on linux
>>> from distutils.sysconfig import get_config_var
>>> get_config_var("Py_ENABLE_SHARED")
1
```

here is the output of ldd on Linux Mint with python 2/3:
```
dta@dta-Inspiron-N5050 ~ $ ldd `which python`
	linux-vdso.so.1 =>  (0x00007fff7e9ec000)
	libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007f60536ed000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f6053328000)
	libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007f6053123000)
	libutil.so.1 => /lib/x86_64-linux-gnu/libutil.so.1 (0x00007f6052f20000)
	libz.so.1 => /lib/x86_64-linux-gnu/libz.so.1 (0x00007f6052d07000)
	libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007f6052a00000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f605392e000)
dta@dta-Inspiron-N5050 ~ $ ldd `which python3`
	linux-vdso.so.1 =>  (0x00007fff1a6e3000)
	libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007fd057cb8000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007fd0578f3000)
	libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007fd0576ee000)
	libutil.so.1 => /lib/x86_64-linux-gnu/libutil.so.1 (0x00007fd0574eb000)
	libexpat.so.1 => /lib/x86_64-linux-gnu/libexpat.so.1 (0x00007fd0572c1000)
	libz.so.1 => /lib/x86_64-linux-gnu/libz.so.1 (0x00007fd0570a7000)
	libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007fd056da1000)
	/lib64/ld-linux-x86-64.so.2 (0x00007fd057ef9000)
```

According to this wiki this most likely means that python2/python3 were build dynamically:

https://wiki.python.org/moin/BuildStatically

I tried ldd with statically build python from source and indeed it shows the correct message:

```
cd ./Python-2.7.12/
./configure --disable-shared LDFLAGS="-static -static-libgcc" CPPFLAGS="-static"
make LDFLAGS="-static" LINKFORSHARED=" "
ldd ./python 
```

with output:

```
not a dynamic executable
```

The problem is that even when python is dynamically linked, `import clr` fails! Most likely this is because the dynamic resolution is different from the expected. But forcing PYTHON_WITHOUT_ENABLE_SHARED works great. Supposedly there is a fix with setting LD_LIBRARY_PATH, but this changes Linux settings system-wide.

@tonyroberts so can you give examples on which Linux systems and how this would still fail?

tonyroberts:
@denfromufa Some versions of the Python executable are built with the python runtime embedded in them, and others are built with it as an external shared object (.so) file. For the ones that use a separate .so file, extensions like clr must also be built to use the same shared object. Ones that embed that code into the python executable must have extension modules built slightly differently (hence the Py_ENABLED_SHARED option, which should be set if a shared object is used).

ldd shows you the shared libraries an executable depends on. If Python has been built to use a shared object version of the python runtime will see libpythonXX.so in the list of shared object dependencies.

To build a version of Python that uses this shared object you should use the "--enable-shared" option when configuring. If successful, when you do ldd on python you will see libpythonXX.so in the output, and if you try the tests with clr built with PYTHON_WITHOUT_ENABLE_SHARED defined it will fail.

Building Python for static linking is something different. This is only useful for embedding Python into another executable without having dynamic link dependencies.

denfromufa:
@tonyroberts great explanation, I will try to build with "--enable-shared" and do some testing.

setup.py
```diff
@@ -151,11 +151,15 @@ def build_extension(self, ext):
             else:
                 defines.append("MONO_LINUX")
             
-            # Disable this check due to issues on Ubuntu
             # Check if --enable-shared was set when Python was built
-            # enable_shared = get_config_var("Py_ENABLE_SHARED")
-            # if enable_shared == 0:
-            defines.append("PYTHON_WITHOUT_ENABLE_SHARED")
+            enable_shared = get_config_var("Py_ENABLE_SHARED")
```
mikofski: capital "y" as in `PY_ENABLE_SHARED`?

denfromufa: https://github.com/python/cpython/search?utf8=%E2%9C%93&q=Py_ENABLE_SHARED

denfromufa:
@tonyroberts @filmor please review

setup.py
```diff
@@ -10,7 +10,7 @@
 from distutils.spawn import find_executable
 from distutils import log
 from platform import architecture
-from subprocess import Popen, CalledProcessError, PIPE, check_call
+from subprocess import Popen, CalledProcessError, PIPE, check_call, check_output
```
vmuriart: @denfromufa Did you try using [`_check_output`](https://github.com/pythonnet/pythonnet/blob/573f90367137b30519a6b204bed7fafc7b6877ef/setup.py#L282) instead?

denfromufa: No, but what is the difference?

vmuriart: It's the exact same thing 😄. Atleast it sounds like `_check_output` was copied from py27 so that earlier version (py26) could use it. I'm reviewing the changes we can make after dropping old python versions and noticed your code.

Issue #299, opened 2016-12-02
Py.Import returns null when importing tensorflow 


snarb:
Py.Import returns null when importing tensorflow at the same time python from command line or from IDE works with that import without any problems.

denfromufa:
Can you share your environment and exact code?

filmor:
It could help debugging to do a `throw new PythonException()` immediately after the `Py.Import`. In fact, I think this should be done in any case, instead of returning `null` here.

pkonovalov-softheme:
That is anaconda with python 3.5 + tensorflow. I can give remote access with team viewer to that env with installed VS and pythonnet sources. I can debug and see that looks like Py.Import return some error. But it structure is very unclear for new guy in this topic, like me.

denfromufa:
Using importlib I found the exception thrown in `tensorflow.python.platform.flags`
```

using System;
using Python.Runtime;

namespace tensorflowinclr
{
    class Program
    {
        static void Main(string[] args)
        {
            dynamic tf;
            dynamic gpb;
            dynamic np;
            using (Py.GIL()) {
                np = Py.Import("numpy");
                gpb = Py.Import("google.protobuf");
                tf = Py.Import("tensorflow");
                //tf = Py.Import("tensorflow.python.platform.flags");
                tf = Py.Import("importlib");
                tf = tf.import_module("tensorflow");
                if (tf == null) {
                    Console.WriteLine("bad import");
                }
            }
        }
    }
}
```

```
Unhandled Exception: Python.Runtime.PythonException: AttributeError : module 'sys' has no attribute 'argv'
   at Python.Runtime.PyObject.Invoke(PyTuple args, PyDict kw) in C:\Users\denfromufa\Documents\pythonnet\src\runtime\pyobject.cs:line 688
   at Python.Runtime.PyObject.InvokeMethod(String name, PyTuple args, PyDict kw) in C:\Users\denfromufa\Documents\pythonnet\src\runtime\pyobject.cs:line 756
   at Python.Runtime.PyObject.TryInvokeMember(InvokeMemberBinder binder, Object[] args, Object& result) in C:\Users\denfromufa\Documents\pythonnet\src\runtime\pyobject.cs:line 1002
   at CallSite.Target(Closure , CallSite , Object , String )
   at System.Dynamic.UpdateDelegates.UpdateAndExecute2[T0,T1,TRet](CallSite site, T0 arg0, T1 arg1)
[...]

PR #301, opened 2016-12-06
Ensure that sys.argv is set

denfromufa:
https://github.com/pythonnet/pythonnet/issues/299

### DIFF
--- a/src/embed_tests/pyimport.cs
+++ b/src/embed_tests/pyimport.cs
@@ -52,8 +52,10 @@ namespace Python.EmbeddingTest
         [Test]
         public void TestDottedName()
         {
-            PyObject module = PythonEngine.ImportModule("PyImportTest.test.one");
-            Assert.IsNotNull(module, ">>>  import PyImportTest.test.one  # FAILED");
+            PyObject module1 = PythonEngine.ImportModule("PyImportTest.test.one");
+            Assert.IsNotNull(module1, ">>>  import PyImportTest.test.one  # FAILED");
+            PyObject module2 = PythonEngine.ImportModule("PyImportTest.sysargv");
+            Assert.IsNotNull(module2, ">>>  import PyImportTest.sysargv  # FAILED");
         }
     }
 }
--- a/src/embed_tests/pyimport.cs
+++ b/src/embed_tests/pyimport.cs
@@ -1,9 +1,7 @@
 using System;
-using System.Reflection;
-using System.Collections.Generic;
+using System.IO;
 using NUnit.Framework;
 using Python.Runtime;
-using System.IO;
 
 namespace Python.EmbeddingTest
 {
@@ -52,10 +50,18 @@ namespace Python.EmbeddingTest
         [Test]
         public void TestDottedName()
         {
-            PyObject module1 = PythonEngine.ImportModule("PyImportTest.test.one");
-            Assert.IsNotNull(module1, ">>>  import PyImportTest.test.one  # FAILED");
-            PyObject module2 = PythonEngine.ImportModule("PyImportTest.sysargv");
-            Assert.IsNotNull(module2, ">>>  import PyImportTest.sysargv  # FAILED");
+            PyObject module = PythonEngine.ImportModule("PyImportTest.test.one");
+            Assert.IsNotNull(module, ">>>  import PyImportTest.test.one  # FAILED");
+        }
+
+        /// <summary>
+        /// Tests that sys.args is set. If it wasn't exception would be raised.
+        /// </summary>
+        [Test]
+        public void TestSysArgsImportException()
+        {
+            PyObject module = PythonEngine.ImportModule("PyImportTest.sysargv");
+            Assert.IsNotNull(module, ">>>  import PyImportTest.sysargv  # FAILED");
         }
     }
 }
--- a/src/tests/test_sysargv.py
+++ b/src/tests/test_sysargv.py
@@ -1,0 +1,17 @@
+# -*- coding: utf-8 -*-
+
+import unittest
+import sys
+
+class SysArgvTests(unittest.TestCase):
+    """Test sys.argv state."""
+
+    def test_sys_argv_state(self):
+        """Test sys.argv state doesn't change after clr import."""
+        argv = sys.argv
+        import clr
+        self.assertTrue(argv == sys.argv)
+
+
+def test_suite():
+    return unittest.makeSuite(SysArgvTests)
--- a/src/tests/PyImportTest/sysargv.py
+++ b/src/tests/PyImportTest/sysargv.py
@@ -1,0 +1,5 @@
+# -*- coding: utf-8 -*-
+
+import sys
+# if argv is available, as expected, then no exception
+num_args = len(sys.argv)
--- a/src/tests/runtests.py
+++ b/src/tests/runtests.py
@@ -23,6 +23,9 @@ except ImportError:
     clr.AddReference("System.Management")
 
 test_modules = (
+    # has to be first test before other module import clr
+	'test_sysargv',
+	
     # test_module passes on its own, but not here if
     # other test modules that import System.Windows.Forms
     # run first. They must not do module level import/AddReference()
--- a/src/tests/tests.pyproj
+++ b/src/tests/tests.pyproj
@@ -29,12 +29,14 @@
   <ItemGroup>
     <Compile Include="leaktest.py" />
     <Compile Include="profile.py" />
+    <Compile Include="PyImportTest\sysargv.py" />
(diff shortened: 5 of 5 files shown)


tonyroberts:
Why not use PySys_SetArgvEx?

denfromufa:
@tonyroberts because it is very simple one-liner and we already have 3 places in pythonengine.cs, where RunString() is used similarly and it is also tested. What is the benefit of using PySys_SetArgvEx, which is not exposed in pythonnet yet?

Anyway we should probably remove PySys_SetArgv in the pythonnet API, because it has security vulnerability due to default updatepath set to 1:

https://docs.python.org/2/c-api/init.html#c.PySys_SetArgvEx

filmor:
Since this is the initial setup we wouldn't pass in `updatepath = 1` anyhow. The function pointers aren't available to users anyhow as all of them are marked as `internal`.

I'm with @tonyroberts here, we should use Python's API directly whenever available. I also don't really agree with the values you're setting here, doesn't it make more sense to

1. Set them by default to the `argv` of the .NET application if available
2. Allow the user to override this by calling `Py.SetArgv(someArray)`?

denfromufa:
While writing the tests I realized that with this change now even "import clr" resets `sys.argv`. Hence need to check in advance if sys.argv is already set. If not, then set it to list with empty string.

@filmor how would I know .NET applications `args` from `public static int Main(string[] args)` call without actually being in the scope of this main method? I think this is not available to Python.Runtime.DLL.

vmuriart:
@denfromufa I got stuck on the same predicament about finding the `args`. I think that part should be skipped for now.

filmor:
@denfromufa `System.Environment.GetCommandLineArgs()` should do it. But I got stuck with the "reinitialise" issue before, so I didn't finish my alternative PR. Are you suggesting we should skip this for the 2.2.0 release? Let's discuss this over there then.

.gitignore
```diff
@@ -54,3 +54,4 @@ _UpgradeReport_Files/
 Backup*/
 UpgradeLog*.XML
 UpgradeLog*.htm
+/NUnit-3.5.0
```
vmuriart: `**/packages/*` should cover this already if you are using the project default version.

src/tests/tests.sln
```diff
@@ -0,0 +1,20 @@
...
+		Release|Any CPU = Release|Any CPU
+	EndGlobalSection
+	GlobalSection(ProjectConfigurationPlatforms) = postSolution
+		{250C535C-C060-4F0C-BD80-41F2BF373565}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
+		{250C535C-C060-4F0C-BD80-41F2BF373565}.Release|Any CPU.ActiveCfg = Release|Any CPU
+	EndGlobalSection
+	GlobalSection(SolutionProperties) = preSolution
+		HideSolutionNode = FALSE
+	EndGlobalSection
+EndGlobal
```
vmuriart: We don't need another solution. Add the pyproject to the `pythonnet.sln` instead.

denfromufa: well, I'm not sure if everyone uses PTVS by default :) How does it look with pyproj in solution, but PTVS not installed? if that's not borken, then I agree that having pyproj in solution is fine!

.gitignore
denfromufa: sorry this slipped through - I was not intending to check-in Nunit changes.

denfromufa:
@filmor @vmuriart I'm almost done with this pull request, let me add user facing API and check for System.Environment.GetCommandLineArgs().

src/tests/tests.sln
vmuriart: I can't find any references to it and I forgot I have ptvs instaled too.

vmuriart: @denfromufa just tried combining to 1 solution and running the solution on a machine without `ptvs`. It loaded the solution no problem, it asked me if i wanted to installed `ptvs` to load up the project. I declined, and it just grays out the project and everything else worked fine. It didn't ask me again if i reloaded the solution. 

The project file needs to be cleaned up though to sanitize the configurations, otherwise it pollutes the current solution. I'm ok with deleting the solution and leaving the `pyproj` without adding it to the current solution on this `pr`.

denfromufa: good finding, perhaps we should get some help from PTVS MS team for promoting their work :)

denfromufa:
@vmuriart @filmor I added more changes to grab CLR arguments if available (may need some tests). please review. I do not intend to add user facing API at this moment.

vmuriart:
The more I think about it, the less I like the reading the .net env `args`. The behavior might not be what the user expects, and we may be forcing it onto them. I'm ok with the `pr` with the exception of the very last commit, my gut feeling is that its going to cause headaches.

filmor:
Obsoleted by #347.

vmuriart:
@filmor @denfromufa did we want to keep some of the tests?

denfromufa:
@vmuriart does the coverage increase if you add these tests? :)

filmor:
Ooh, sorry missed those. Yeah, we should probably add them.

denfromufa:
@filmor can you please reopen this one to add the remaining tests?

I cannot do that because the branch is deleted.

vmuriart:
@denfromufa I just restored your branch. (or did you mean @filmor's branch from his work)

denfromufa:
OK, let me add the remaining tests, if any. Thanks @vmuriart

vmuriart:
might be easiest to create a new branch/pr and copy the tests from this one over to avoid dealing w the merge conflicts and reverting/cleaning other changes.

denfromufa:
@filmor @tonyroberts @vmuriart please review the added unit tests, CI is passing

src/embed_tests/pyimport.cs
```diff
@@ -52,8 +52,10 @@ public void TearDown()
         [Test]
         public void TestDottedName()
         {
-            PyObject module = PythonEngine.ImportModule("PyImportTest.test.one");
-            Assert.IsNotNull(module, ">>>  import PyImportTest.test.one  # FAILED");
+            PyObject module1 = PythonEngine.ImportModule("PyImportTest.test.one");
+            Assert.IsNotNull(module1, ">>>  import PyImportTest.test.one  # FAILED");
+            PyObject module2 = PythonEngine.ImportModule("PyImportTest.sysargv");
```
vmuriart: this should be its own test, not part of the previous one.

src/tests/test_sysargv.py
```diff
@@ -0,0 +1,23 @@
...
+import sys
+
+
+class SysArgvTests(unittest.TestCase):
+    """Test sys.argv state."""
+    def testSysArgvState(self):
+        """Test sys.argv state."""
+        argv = list(sys.argv) # need to make a copy of list
+        import clr
+        self.assertTrue(argv == sys.argv)
```
vmuriart: this doesn't make sense. Did importing `clr` previous cause `sys.argv` to change? Other tests `import clr` before this test, that may change being too.

denfromufa: this is to ensure that importing clr does not mess sys.argv, when it is not empty. what do you mean by "that may change being too."?

src/embed_tests/pyimport.cs
denfromufa: agreed, let me move it

src/tests/test_sysargv.py
vmuriart: I have no idea what I was trying to write... I meant to say, that since other tests are doing `import clr` before this test, the library is already loaded into memory and won't be reimported and the `import clr` statement may do nothing.

denfromufa: oh, that is a great point! then this needs to be the first test in the test runner before clr is loaded.

vmuriart: `clr` is loaded before all tests being. its the 5th line on `run_tests.py` i think.

vmuriart:
Just applied the changes needed and squash merged since the history was a bit messy.

src/tests/test_sysargv.py
denfromufa: this was missed, still importing clr before setargv test

src/tests/test_sysargv.py
```diff
@@ -0,0 +1,17 @@
...
+
+import unittest
+import sys
+
+class SysArgvTests(unittest.TestCase):
+    """Test sys.argv state."""
+
+    def test_sys_argv_state(self):
+        """Test sys.argv state doesn't change after clr import."""
+        argv = sys.argv
```
denfromufa: should be argv = list(sys.argv) - need to make a copy

PR #304, opened 2016-12-09
Support for System.Decimal

AlexCatarino:
Implements conversion from `System.Decimal` to `decimal.Decimal`

### DIFF
--- a/src/runtime/converter.cs
+++ b/src/runtime/converter.cs
@@ -101,7 +101,8 @@ namespace Python.Runtime
                 return Runtime.PyLongType;
             }
             else if ((op == doubleType) ||
-                     (op == singleType))
+                     (op == singleType) ||
+                     (op == decimalType))
             {
                 return Runtime.PyFloatType;
             }
@@ -195,6 +196,9 @@ namespace Python.Runtime
 
                 case TypeCode.Double:
                     return Runtime.PyFloat_FromDouble((double)value);
+
+                case TypeCode.Decimal:
+                    return Runtime.PyFloat_FromDouble(Convert.ToDouble((decimal)value));
 
                 case TypeCode.SByte:
                     return Runtime.PyInt_FromInt32((int)((sbyte)value));
@@ -788,7 +792,7 @@ namespace Python.Runtime
             }
 
 
-            type_error:
+        type_error:
 
             if (setError)
             {
@@ -800,7 +804,7 @@ namespace Python.Runtime
 
             return false;
 
-            overflow:
+        overflow:
 
             if (setError)
             {
--- a/src/runtime/converter.cs
+++ b/src/runtime/converter.cs
@@ -797,10 +797,22 @@ namespace Python.Runtime
                     }
                     result = d;
                     return true;
+
+                case TypeCode.Decimal:
+                    op = Runtime.PyObject_Str(value);
+                    decimal m;
+                    string sm = Runtime.GetManagedString(op);
+                    if (!Decimal.TryParse(sm, NumberStyles.Number, nfi, out m))
+                    {
+                        goto type_error;
+                    }
+                    Runtime.XDecref(op);
+                    result = m;
+                    return true;
             }
 
 
-        type_error:
+            type_error:
 
             if (setError)
             {
@@ -812,7 +824,7 @@ namespace Python.Runtime
 
             return false;
 
-        overflow:
+            overflow:
 
             if (setError)
             {
--- a/src/runtime/converter.cs
+++ b/src/runtime/converter.cs
@@ -197,9 +197,6 @@ namespace Python.Runtime
                 case TypeCode.Double:
                     return Runtime.PyFloat_FromDouble((double)value);
 
-                case TypeCode.Decimal:
-                    return Runtime.PyFloat_FromDouble(Convert.ToDouble((decimal)value));
-
                 case TypeCode.SByte:
                     return Runtime.PyInt_FromInt32((int)((sbyte)value));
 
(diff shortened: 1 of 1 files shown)


filmor:
I'm strongly against this. If anything, we should convert `System.Decimal` to `decimal.Decimal`.

denfromufa:
@filmor @AlexCatarino I agree that decimal to float is very bad idea and defeats purpose of using decimal. Your users would not appreciate this, especially for financial applications. I was bitten before by precision loss, so let's be careful.

Here is excerpt that I send to before privately:

There is no implicit conversion between floats and decimals neither in .NET nor in Python.

Conversion between `System.Decimal` and `decimal.Decimal` requires a patch in `converter.cs`.

Then need to create a decimal type using Python C-API in `runtime.cs` using this method:

http://stackoverflow.com/questions/38488205/how-do-i-create-a-python-decimal-object-from-c

Here is a quick around by using strings to avoid precision loss:

```
C:\Users\denis.akhiyarov>ptipython
Python 2.7.12 |Anaconda custom (32-bit)| (default, Jun 29 2016, 11:42:13) [MSC
.1500 32 bit (Intel)]
Type "copyright", "credits" or "license" for more information.

IPython 5.1.0 -- An enhanced Interactive Python.
?         -> Introduction and overview of IPython's features.
%quickref -> Quick reference.
help      -> Python's own help system.
object?   -> Details about 'object', use 'object??' for extra details.


In [1]: import clr


In [2]: import System


In [3]: numstr='1.'+'0'*20+'1'


In [4]: numstr
Out[4]: '1.000000000000000000001'


In [5]: float(numstr)
Out[5]: 1.0


In [6]: dec_clr=System.Decimal(numstr)
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-6-4e81bfdf9b04> in <module>()
----> 1 dec_clr=System.Decimal(numstr)

TypeError: no constructor matches given arguments


In [7]: dec_clr=System.Decimal(System.String(numstr))
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-7-dd552bfa0ced> in <module>()
----> 1 dec_clr=System.Decimal(System.String(numstr))

TypeError: no constructor matches given arguments


In [8]: System.Decimal.__overloads__
Out[8]:
System.Decimal(Int32[])
System.Decimal(UInt64)
System.Decimal(UInt32)
System.Decimal(Int64)
System.Decimal(Int32)
System.Decimal(Single)
System.Decimal(Double)
System.Decimal(Int32, Int32, Int32, Boolean, Byte)


In [9]: dec_clr=System.Decimal.Parse(numstr)


In [10]: dec_clr
Out[10]: <System.Decimal at 0x3df4230>


In [11]: float(dec_clr)
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-11-bc64575cb33e> in <module>()
----> 1 float(dec_clr)

TypeError: float() argument must be a string or a number


In [12]: dec_clr.ToString
Out[12]: <bound method 'ToString'>


In [13]: dec_clr.ToString()
Out[13]: u'1.000000000000000000001'


In [14]: from decimal import Decimal


In [15]: dec_py=Decimal(numstr)


In [16]: dec_py
Out[16]: Decimal('1.000000000000000000001')


In [17]: float(dec_py)
Out[17]: 1.0


In [18]: Decimal(dec_clr.ToString())*1.0
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-18-e8356c5f61db> in <module>()
----> 1 Decimal(dec_clr.ToString())*1.0

TypeError: unsupported operand type(s) for *: 'Decimal' and 'float'


In [19]: Decimal(dec_clr.ToString())*Decimal(1.0)
Out[19]: Decimal('1.000000000000000000001')


In [20]: System.Double(1.0)
Out[20]: <System.Double at 0x849da90>


In [21]: System.Double(1.0)*dec_clr
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-21-e2ae5724d14f> in <module>()
----> 1 System.Double(1.0)*dec_clr

TypeError: unsupported operand type(s) for *: 'Double' and 'Decimal'
```

filmor:
The failing tests show that there are still problems with this solution. Passing them is prerequisite to merging this. Since we are handling `System.String` in a similar way, I think I'm fine with the automatic conversion.

@tonyroberts @matthid @vmuriart Do you see any problems with this?

src/runtime/converter.cs
```diff
@@ -208,6 +209,17 @@ internal static IntPtr ToPython(Object value, Type type)
                 case TypeCode.UInt64:
                     return Runtime.PyLong_FromUnsignedLongLong((ulong)value);
 
+                case TypeCode.Decimal:
+                    IntPtr mod = Runtime.PyImport_ImportModule("decimal");
```
tonyroberts: Need to check mod and ctor aren't null

tonyroberts:
@filmor do you mean automatic conversion from decimal.Decimal to System.Decimal? If so, yes I agree that would make sense.

denfromufa:
> @filmor do you mean automatic conversion from decimal.Decimal to System.Decimal? If so, yes I agree that would make sense.

@filmor @tonyroberts note that automatic conversion from Python to CLR decimals should only happen with Python default of 28 digit precision and when CLR precision is 28 digit as well.
CLR precision is between 28 to 29 digits to fit into 128-bits.
Python precision is user-adjustable.

denfromufa:
from our docs, type conversion section:

Types that do not have a logical equivalent in Python are exposed as instances of managed classes or structs (System.Decimal is an example).

This will need to be updated.

denfromufa:
@vmuriart do you resolve conflicts using web editor or in command-line git? this pull request has a conflict.

vmuriart:
command-line/gui. I use `gitextensions` and `beyondcompare`.

filmor:
The conflicts are not the problem here, though, there are failing tests.

filmor:
Since there hasn't been any activity on this for quite some time and we have failing tests, I will close this PR for now. @AlexCatarino (or anyone else) feel free to reopen it when you find the time to fix the remaining issues.

PR #306, opened 2016-12-19
#296 Convert PyObject (cleanup)

rmadsen-ks:
Removed RefCount


rmadsen-ks:
Thanks all for your collaboration, it is much appreciated! :cake:

PR #307, opened 2016-12-25
python-3.6 test

stonebig:
first try a https://github.com/pythonnet/pythonnet/issues/272

Issue #272, opened 2016-10-14
add Python 3.6


stonebig:
Would it be possible to get Python 3.6 on AppVeyor ?

denfromufa:
there is no official support for python 3.6 on appveyor yet. pandas, numpy,
matplotlib are not testing py3.6 yet.

stonebig:
cgohlke did

denfromufa:
that is great! But I cannot see @cgohlke build configuration in the open, since only binaries for download are available from his website.

denfromufa:
Is there WinPython on Appveyor? E.g. they support Miniconda:

https://www.appveyor.com/docs/installed-software/#python

stonebig:
The answer is in your link, isn't it ?

stonebig:
but maybe you can download a WinPythonZero from github, then unzip somewhere and you're all set in AppVeyor ?

stonebig:
winpython-2.6.0.0Preview is out, in case it can help

stonebig:
python-3.6 final is out

PR #310, opened 2017-01-05
Support Python 3.6

### LINKED ISSUE #309: support python-3.6
this requires: (https://github.com/pythonnet/pythonnet/pull/307#issuecomment-270578222 )
- changing:
  - **|| PYTHON35**
  - per **|| PYTHON35 || PYTHON36**
- in:
  - clrmodule/ClrModule.cs
  - src/runtime/converter.cs
  - src/runtime/delegateobject.cs
  - src/runtime/exceptions.cs
  - src/runtime/importhook.cs
  - src/runtime/interop.cs
  - src/runtime/interop35.cs
  - src/runtime/pythonengine.cs
  - src/runtime/runtime.cs
  - src/runtime/typemanager.cs

also a few trick in runtime.cs (see in same comment)
not sure if we need also to generate with a clang a interop36.cs

cgohlke:
Fixes #309 and #272 . Replaces #307.

Tested with Python 3.6 64-bit and msvc14 on Windows 10.


cgohlke:
> @cgohlke do you agree to transition from Zope to MIT license?

Yes, I agree.

src/runtime/interop35.cs
```diff
@@ -3,7 +3,7 @@
 
 
 
-#if (PYTHON35)
+#if (PYTHON35 || PYTHON36)
```
filmor: Is this intentional?

cgohlke: This is a leftover from trying to reuse the file for Python 3.6. It is no longer needed. I'll remove this change. Thank you.

Issue #218, opened 2016-05-26
Add Pythonnet __version__ variable


vmuriart:
It may be user error, but I was troubleshooting of my co-worker's machines and couldn't figure out how to send him a script that would return the version number of `Pythonnet` they were using from within Python. 

Im thinking something in the lines of

```python
import clr
print(clr.__version__)
```

denfromufa:
https://github.com/pythonnet/pythonnet/issues/67

vmuriart:
closing to avoid repetition.

PR #312, opened 2017-01-07
Add __version__

### LINKED ISSUE #67: version/build number
setup version/build number

vmuriart:
Closes #67, #218.


vmuriart:
For reference, below is the alternative method I been using to detect the currently installed version.

```python
def pythonnet_version():
    # Based from: http://stackoverflow.com/a/23885252
    import pip

    for x in pip.get_installed_distributions():
        if x.key == "pythonnet":
            return x.version
```

tonyroberts:
Would it be simpler to add __version__ to clr.py (in src/runtime/resources)? Everything in that file gets added to the clr module in Initialize. A small change would be needed as currently Initialize ignores anything starting with an underscore.

vmuriart:
That would make more sense. Let me review and try that out. thanks!

vmuriart:
@tonyroberts thanks for the feedback, it worked

filmor:
It would be a lot nicer (and more future proof) if this information was taken during the build process from either a version file or the git repository (e.g. using https://github.com/warner/python-versioneer).

denfromufa:
I agree with @filmor - we need automated way of updating versions. One trigger (e.g. Github release tag) should make changes in setup.py and `__version__`.

vmuriart:
I agree, I was thinking along the same lines and was inclining towards [bumpversion](https://github.com/peritus/bumpversion). 

In the interest of separations of concerns though, can we merge this `pr` and I'll open a ticket to add the automation.

denfromufa:
If this PR does not conflict with either of these automated versioning
tools, then I agree - we can merge.

PR #320, opened 2017-01-09
Project files tune-up

vmuriart:
- Maintenance to the project to remove many unneeded files. Commit history references when/why its safe to remove the files.
- Added a CHANGELOG
- Cleaned up project references to files
- Cleaned up a couple warnings and tests

Overall it makes the project look alot less intimidating


vmuriart:
@tonyroberts what's the purpose of [`clrmodule.il`](https://github.com/pythonnet/pythonnet/blob/master/src/runtime/clrmodule.il)?

It looks like nothing references it anymore, and everything still passed after I removed it, but I didn't want to remove it since I couldn't understand its purpose.

filmor:
I like this PR a lot, keep going :)

tonyroberts:
@vmuriart re: clrmodule.il, I think this and related files (clrmodule.pp.il, oldmodule.il, buildclrmodule.bat) can be safely deleted. They're from before I started looking at this project, but from what I can gather they're legacy code that has since been replaced by clrmodule.cs.

vmuriart:
@tonyroberts [you're right](https://github.com/pythonnet/pythonnet/blob/master/src/clrmodule/ClrModule.cs#L2-L3).

vmuriart:
I think this `pr` is ready for merging. There's still a bit more to clean-up within the src itself, but I don't want to mix those in with this `pr`. 

Let me know if there's any objections with merging it in. For this particular `pr` I'd prefer we do a `merge commit` to a copy of commit messages since they explain the reasoning for each file deletion.

vmuriart:
@denfromufa @filmor @tonyroberts any objections to merging this in?

filmor:
Is it fine to merge this squashed?

PR #324, opened 2017-01-12
Refactor directives

vmuriart:
Refactors all those pesky #if to simplify process of supporting newer python versions.

Theres a section in runtime I left untouched as it has many embbded directives.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -138,7 +138,8 @@ class PythonNET_BuildExt(build_ext):
             unicode_width = ctypes.sizeof(ctypes.c_wchar)
 
         defines = [
-            "PYTHON%d%s" % (sys.version_info[:2]),
+            "PYTHON%d%d" % (sys.version_info[:2]),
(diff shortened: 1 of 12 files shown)


denfromufa:
@vmuriart so now we need to use both python3 and python3x flags?

filmor:
Yes. And .NET doesn't really have a way around that, AFAICT.

denfromufa:
@vmuriart is it also possible to do this at the top of the file:
```
#if (PYTHON32 || PYTHON33 || PYTHON34 || PYTHON35)
#define PYTHON3
#endif
```

vmuriart:
It would be needed across every file that needs that directive.  http://stackoverflow.com/questions/30350738/where-to-specify-preprocessor-directives-in-visual-studio

The project file is updated to have PYTHON2 set on it though, you would just need to update that on your build config to PYTHON3

PR #326, opened 2017-01-14
Add ISSUE_TEMPLATE and PULL_REQUEST_TEMPLATE

vmuriart:
`issues`: Ensures that issues contain the basic necessary information
`pr`: Reminds author to add changes to changelog, authors, and add tests


filmor:
Very good idea.

PR #329, opened 2017-01-22
Clean-up Tests

vmuriart:
- Remove dependency on `six`
- Fixes broken tests
- Improves Py2/Py3 testing
- Adds unused modules
- Comments broken modules
- Simplifies tests
- Renames tests
- Many pep8 clean-ups

No

### Any other comments?
Recommend `squash&merge`.  Left as separate commits for anyone that wants to see the logic for some of the changes.

### Checklist

-   [N/A ] Updated the [`CHANGELOG`](../blob/master/CHANGELOG.md)


denfromufa:
any reasoning for removing dependency on six?

vmuriart:
It wasn't necessary. This way the compatibility module is *frozen* for our tests. ie, won't be surprised if `six` introduces a bug in the future.

denfromufa:
@vmuriart i'm big minus one on removing dependency from six, it is third most dependent package on pypi:

http://kgullikson88.github.io/blog/pypi-analysis.html

it is very unlikely that we are smarter than all these folks using six.

i do not want to maintain our own subset of six, unless there is unresolved issue with six affecting us.

vmuriart:
@denfromufa. We [barely used](https://github.com/pythonnet/pythonnet/blob/cf5035c176ce2b8009f94e601ff0dad99feba552/src/tests/test_generic.py#L11-L14) `six`. Mostly we used `six` to check if we were in PY2 or PY3 and then created our own compatibility variable. Since we don't support py25, py32, we don't need `six.u(...)`. The only remaining usage for `six` were two calls to `indexbytes`.

I [refactored all the checks](https://github.com/pythonnet/pythonnet/pull/329/commits/d35a4725e3b06b01e36fd8c1c4df8c315a3fdd22) into the [compat module](http://lucumr.pocoo.org/2013/5/21/porting-to-python-3-redux/). I dont think its worth to keep six only for `indexbytes`.

denfromufa:
@vmuriart I agree, good catch on six. let's get rid of it. give me some time to review this PR after 2.2 release.

vmuriart:
@denfromufa no worries. I want to merge #301 first, and then work on this one. There's going to be a merge conflict after 2.2.0, but i should be able to fix it quickly.

vmuriart:
@filmor did you get a chance to check this out?

filmor:
Yep, seems fine on my machine ;)

filmor:
You probably should have squashed this before merging :)

vmuriart:
I was debating it, there are some benefits to squashing (compress history), but at the same time there are some losses when going back threw commit history and trying to understand why a file was added/removed.

Personally I clean-up the history before submitting the pull request. That way it has my logical units of change, without having my 10,000 commits saying "FIXUP! typo" 👍

Issue #336, opened 2017-01-28
pythonnet 2.2.1 on linux fails in python 3.4 and 3.5 (but not 2.7) - pycparser error


denfromufa:
<details>
  <summary>anaconda python 3.5 (click to expand) </summary>

```
(root) dta@dta-Inspiron-N5050 ~ $ sudo /home/dta/anaconda3/bin/pip install pythonnet -U --force
The directory '/home/dta/.cache/pip/http' or its parent directory is not owned by the current user and the cache has been disabled. Please check the permissions and owner of that directory. If executing pip with sudo, you may want sudo's -H flag.
The directory '/home/dta/.cache/pip' or its parent directory is not owned by the current user and caching wheels has been disabled. check the permissions and owner of that directory. If executing pip with sudo, you may want sudo's -H flag.
Collecting pythonnet
  Downloading pythonnet-2.2.1.tar.gz (1.5MB)
    100% |████████████████████████████████| 1.5MB 486kB/s 
Installing collected packages: pythonnet
  Found existing installation: pythonnet 2.2.0
    Uninstalling pythonnet-2.2.0:
      Successfully uninstalled pythonnet-2.2.0
  Running setup.py install for pythonnet ... error
    Complete output from command /home/dta/anaconda3/bin/python -u -c "import setuptools, tokenize;__file__='/tmp/pip-build-2mxcht23/pythonnet/setup.py';exec(compile(getattr(tokenize, 'open', open)(__file__).read().replace('\r\n', '\n'), __file__, 'exec'))" install --record /tmp/pip-xa22cnng-record/install-record.txt --single-version-externally-managed --compile:
    running install
    running build
    running build_ext
    Checking for updates from https://www.nuget.org/api/v2/.
    Currently running NuGet.exe 3.5.0.
    NuGet.exe is up to date.
    MSBuild auto-detection: using msbuild version '14.0' from '/usr/lib/mono/xbuild/14.0/bin'.
    Restoring NuGet package NUnit.2.6.2.
    Restoring NuGet package UnmanagedExports.1.2.7.
    Adding package 'UnmanagedExports.1.2.7' to folder '/tmp/pip-build-2mxcht23/pythonnet/packages'
    Adding package 'NUnit.2.6.2' to folder '/tmp/pip-build-2mxcht23/pythonnet/packages'
    Added package 'NUnit.2.6.2' to folder '/tmp/pip-build-2mxcht23/pythonnet/packages'
[...]

PR #335, opened 2017-01-27
Coverage & CI Setup Improvement

vmuriart:
### What does this implement/fix? Explain your changes.
- Add coverage (cs code limited to windows, no mature coverage tools for mono)
- Cleans up CI configurations
- Speeds up Travis ~30% by using containers
- Simplifies and unfreezes mono setup
- Speed up AppVeyor by 50% by not downloading and building with miniconda. 

### Does this close any currently open issues?

#237 
#334

### Any other comments?

Sample [coverage](https://codecov.io/gh/vmuriart/pythonnet/tree/1cfe592ae8806847bfb9c2581cef5ebac771c7c3) report.

### Checklist

Check all those that are applicable and complete.

-   [ ] Updated the [`CHANGELOG`](../blob/master/CHANGELOG.md)

PR #338, opened 2017-01-29
include fake c libs for pycparser to work

denfromufa:
https://github.com/pythonnet/pythonnet/issues/336

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -327,7 +327,7 @@ if __name__ == "__main__":
                 sources.append(os.path.join(root, filename))
 
     for root, dirnames, filenames in os.walk("tools"):
-        for ext in (".exe", ".py"):
+        for ext in (".exe", ".py", ".c", ".h"):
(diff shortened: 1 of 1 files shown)


vmuriart:
It creates merge conflict with #337, but I'll rebase #377 after this is merged.

denfromufa:
@vmuriart let me include 2.2.2 tag in setup.py and __version__ in this pull request?

vmuriart:
Not necessary, I'll add them when updating the changelog. Otherwise its going to trigger two more builds on `AppVeyor` which takes along time...

vmuriart:
Actually, if you can take a look at #335 while we wait for these to finish that would be great.
It will make getting the final built whl from appveyor that much faster.

denfromufa:
@vmuriart ok, then please add release tag, update changelog, release notes, setup.py, and `__version__` after this is merged.

Issue #262, opened 2016-09-07
MSTests failing when importing a module with imports


gatapia:
Very simple to reproduce:

In the Python.EmbeddingTests project (I am using master branch):
- Add reference to MSTest (Microsoft.VisualStudio.QualityTools.UnitTestFramework.dll 10.0.0.0)
- Convert pyimport from NUnit to MSTest:
  - remove NUnit Import, add using Microsoft.VisualStudio.TestTools.UnitTesting
  - replace attributes with MStest attributes (TestClass, TestInitialise, TestCleanup, TestMethod)
- Running TestDottedName works as expected.
- Edit one.py and add `import numpy as np` as first line
- Run TestDottedName should work first time
- Run TestDottedName again, now it fails with:

```

Test method Python.EmbeddingTest.PyImportTest.TestDottedName threw exception: 
System.AccessViolationException: Attempted to read or write protected memory. This is often an indication that other memory is corrupt.
```

This error is also being shown in my ASP.Net Web API application on restart so I hope a solution will solve both MSTest and ASP.Net.

Further Information: Full version of pyimport.cs after edits is:

```
using System;
using System.Reflection;
using Microsoft.VisualStudio.TestTools.UnitTesting;
using Python.Runtime;

namespace Python.EmbeddingTest
{
    [TestClass]
    public class PyImportTest
    {
        private IntPtr gs;

        [TestInitialize]
        public void SetUp()
        {
            string path = @"c:\dev\libs\Anaconda3\envs\python2;" + Environment.GetEnvironmentVariable("PATH", EnvironmentVariableTarget.Machine);
            Environment.SetEnvironmentVariable("PATH", path, EnvironmentVariableTarget.Process);
            Environment.SetEnvironmentVariable("PYTHONHOME", @"c:\dev\libs\Anaconda3\envs\python2", EnvironmentVariableTarget.Process);
            Environment.SetEnvironmentVariable("PYTHONPATH ", @"c:\dev\libs\Anaconda3\envs\python2\Lib", EnvironmentVariableTarget.Process);

            PythonEngine.Initialize();
            gs = PythonEngine.AcquireLock();

            //string here = Environment.CurrentDirectory;
            //trunk\pythonnet\src\embed_tests\bin\x86\DebugWin

            /*
             * Append the tests directory to sys.path
             * using reflection to circumvent the private modifires placed on most Runtime methods.
             */
            const string s = @"../../../../tests";

[...]

PR #341, opened 2017-01-30
Improve tests

filmor:
### What does this implement/fix? Explain your changes.

- Simplifies testing by giving the test DLL access to the internals of `Python.Runtime`
- Update NUnit to 3.5

### Does this close any currently open issues?

Nope.

### Any other comments?

Can be squashed.

### Checklist

Check all those that are applicable and complete.

-   [x] Make sure to include one or more tests for your change
-   [x] If an enhancement PR, please create docs and at best an example
-   [x] Add yourself to [`AUTHORS`](../blob/master/AUTHORS.md)
-   [x] Updated the [`CHANGELOG`](../blob/master/CHANGELOG.md)

All N/A.

PR #343, opened 2017-01-30
Fix shutdown

filmor:

Fixes issue #262 by countering two memory corruption issues.

#262.

### Any other comments?

Is currently based on my other PR, but doesn't strictly need it, I could rebase.

### DIFF
--- a/src/runtime/importhook.cs
+++ b/src/runtime/importhook.cs
@@ -16,7 +16,13 @@ namespace Python.Runtime
 
 #if PYTHON3
         static IntPtr py_clr_module;
-        static IntPtr module_def;
+        static IntPtr module_def = IntPtr.Zero;
+
+        internal static void InitializeModuleDef()
+        {
+            if (module_def == IntPtr.Zero)
+                module_def = ModuleDefOffset.AllocModuleDef("clr");
+        }
 #endif
 
         //===================================================================
@@ -44,8 +50,8 @@ namespace Python.Runtime
             root = new CLRModule();
 
 #if PYTHON3
-    // create a python module with the same methods as the clr module-like object
-            module_def = ModuleDefOffset.AllocModuleDef("clr");
+            // create a python module with the same methods as the clr module-like object
+            InitializeModuleDef();
             py_clr_module = Runtime.PyModule_Create2(module_def, 3);
 
             // both dicts are borrowed references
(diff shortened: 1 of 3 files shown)


vmuriart:
Is this branch contain the same changes from #341? I'm reading it from my phone so hard to tell.

vmuriart:
To answer my own question, it is. I'm testing this now but its looking good!

vmuriart:
Failing on Appveyor Py33

https://ci.appveyor.com/project/vmuriart/pythonnet/build/fix-shutdown-517/job/adnsiyd4278ypy1m

vmuriart:
@filmor you ok with me rebasing your work ontop of the master branch and push it here?

src/runtime/pythonengine.cs
```diff
@@ -411,33 +411,54 @@ public static PyObject ModuleFromString(string name, string code)
...
+                IntPtr result = Runtime.PyRun_String(
+                    code, flag, globals.Value, locals.Value
+                    );
+
+                if (Runtime.PyErr_Occurred() != 0)
+                {
+                    throw new PythonException();
+                }
+
+                return new PyObject(result);
```
vmuriart: The return within the `try` statement can get tricky. 
http://stackoverflow.com/questions/3216046/does-the-c-sharp-finally-block-always-execute

filmor: I don't really see the issue here.

filmor: To elaborate, the link that you posted warns against relying on `try {} finally {}` as the `finally` may not be executed, but in such a case the problem has either exited already or is in the process of doing so, so we don't have to worry about dereferencing the Python resources anyhow.

vmuriart: It was more of a caveat, return statements within `try` statements confuse me at times.

src/runtime/importhook.cs
```diff
@@ -70,25 +76,25 @@ internal static void Initialize()
...
-            }
 #endif
-            if (0 != Runtime.Py_IsInitialized())
-            {
+                Runtime.XDecref(root.pyHandle);
                 Runtime.XDecref(py_import);
             }
         }
 
+        internal static void Cleanup()
```
vmuriart: Is there a call to `Cleanup` missing?

filmor: That part is not used anymore as there is only a single ModuleDef instance that is kept for the whole run time, I'll remove it.

vmuriart: Cool, after it's removed I'm ok merging it. This fixes the issue for Python2.7, but it still occassionally fails on Python3 but it seems to be due to a different section of code

filmor: Yeah, I noticed that too, I'll try to reproduce the issue, for me it happens (after these patches) not anymore on Shutdown but only on Initialize.

vmuriart:
Cool, will merge after CI completes. Thanks!

PR #346, opened 2017-01-31
Fix Console config, Add pyproj, Add PY3 to configurationmanager

vmuriart:

Standardizes build configurations across the projects (`Console.csproj` had its own thing going on)
Adds @denfromufa 's *.pyproj with corrected configurations


.travis.yml
```diff
@@ -31,7 +31,7 @@ install:
 script:
   - export PYTHONPATH=`pwd`:$PYTHONPATH
   - python src/tests/runtests.py
-  # - nunit-console src/embed_tests/bin/x64/ReleaseMono/Python.EmbeddingTest.dll
+  # - nunit-console src/embed_tests/bin/Python.EmbeddingTest.dll
```
filmor: Why is this still commented out?

vmuriart: It runs the tests but all say `Cant find python.runtime` error. Didn't have a chance to debug it. I'm keeping the tests paths updated though

filmor: Okay, fine by me then, but I'll open an issue for this, we should really start running those tests.

vmuriart: On appveyor, when we run them on py3 they freeze and hang the testrunner until the build bot times out (30 minutes or so?).
I stress tested it on PY27 and its def fixed there. Ran 10 times no issues.

vmuriart: @filmor if we downgrade back to nunit `2.6.2` we can start running the tests. They will still fail, but nunit will gracefully fail and give us the error instead of whatever nunit3 is doing now

PR #347, opened 2017-01-31
Add pysetargv

filmor:

Implements an overload of `Initialize` and a `Py.SetArgv` function that allow the user to issue a `PySys_SetArgvEx` call. By default, this call is done on `Initialize` with a list that contains a single empty string (`['']`), using the arguments supplied to the .NET process if available.

#299.

### Any other comments?

Based on the `fix-shutdown` branch as otherwise I'm not able to run the unit-tests reliably..

### DIFF
--- a/src/embed_tests/InitializeTest.cs
+++ b/src/embed_tests/InitializeTest.cs
@@ -13,40 +13,26 @@ namespace Python.EmbeddingTest
         public static void LoadSpecificArgs()
         {
             var args = new[] { "test1", "test2" };
-            PythonEngine.Initialize(args);
-            try
+            using (new PythonEngine(args))
+            using (var argv = new PyList(Runtime.Runtime.PySys_GetObject("argv")))
             {
-                using (var argv = new PyList(Runtime.Runtime.PySys_GetObject("argv")))
-                {
-                    Assert.That(argv[0].ToString() == args[0]);
-                    Assert.That(argv[1].ToString() == args[1]);
-                }
-            }
-            finally
-            {
-                PythonEngine.Shutdown();
+                Assert.That(argv[0].ToString() == args[0]);
+                Assert.That(argv[1].ToString() == args[1]);
             }
         }
 
         [Test]
         public static void LoadDefaultArgs()
         {
-            PythonEngine.Initialize();
-            try
+            using (new PythonEngine())
+            using (var argv = new PyList(Runtime.Runtime.PySys_GetObject("argv")))
             {
-                using (var argv = new PyList(Runtime.Runtime.PySys_GetObject("argv")))
-                {
-                    Assert.That(argv.Length() != 0);
-                }
-            }
-            finally
-            {
-                PythonEngine.Shutdown();
+                Assert.That(argv.Length() != 0);
             }
         }
 
         [Test]
-        public static void Test()
+        public static void StartAndStopTwice()
         {
             PythonEngine.Initialize();
             PythonEngine.Shutdown();
--- a/src/embed_tests/InitializeTest.cs
+++ b/src/embed_tests/InitializeTest.cs
@@ -10,6 +10,42 @@ namespace Python.EmbeddingTest
     public class InitializeTest
     {
         [Test]
+        public static void LoadSpecificArgs()
+        {
+            var args = new[] { "test1", "test2" };
+            PythonEngine.Initialize(args);
+            try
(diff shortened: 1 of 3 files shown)


vmuriart:
This replaces #301?

src/runtime/runtime.cs
```diff
@@ -2027,11 +2027,26 @@ internal unsafe static extern IntPtr
         internal unsafe static extern IntPtr
             PyImport_GetModuleDict();
 
-
+#if !(PYTHON26 || PYTHON27)
```
vmuriart: Can this be `#if PYTHON3` ?

filmor: Yes, I'll change it.

filmor:
Yes, it's an alternative using directly the Python API function.

src/embed_tests/InitializeTest.cs
```diff
@@ -10,7 +10,29 @@ namespace Python.EmbeddingTest
...
+            }
+        }
+
+        [Test]
+        public static void LoadDefaultArgs()
+        {
+            using (new PythonEngine())
+            using (var argv = new PyList(Runtime.Runtime.PySys_GetObject("argv")))
+            {
+                Assert.That(argv.Length() != 0);
```
denfromufa: first need to check that it is not null

src/runtime/Python.Runtime.csproj
```diff
@@ -92,6 +92,7 @@
     <Optimize>false</Optimize>
     <DebugType>full</DebugType>
     <PlatformTarget>x64</PlatformTarget>
+    <DefineConstants>TRACE;DEBUG;PYTHON3;PYTHON35;UCS2</DefineConstants>
```
denfromufa: this gets overwritten by setup.py or specific needs of the user, so no need to update this

vmuriart: He probably added that during his development and forgot to remove it.

vmuriart: Updated #346 to include configuration for PY3 development.

denfromufa:
I'm OK with merging this version as long as comments are addressed

src/embed_tests/InitializeTest.cs
filmor: Why? If it is `null`, this will fail with a `NullReferenceException`, perfectly fine.

src/runtime/Python.Runtime.csproj
filmor: Yep, exactly that, I'll remove this line.

vmuriart:
@filmor rebased your work here

https://github.com/pythonnet/pythonnet/tree/pysetargv

denfromufa:
@filmor the coverage on pythonengine.cs reduced

vmuriart:
It's probably the refactoring he did. The coverage needs some fine tuning, its definetly not as robust as python's coverage engine.

src/runtime/pythonengine.cs
```diff
@@ -102,6 +124,11 @@ public static int RunSimpleString(string code)
 
         #endregion
 
+        public static void Initialize()
+        {
+            Initialize(Enumerable.Empty<string>());
+        }
+
```
denfromufa: i did not click submit review few weeks ago

src/runtime/pythonengine.cs
```diff
@@ -389,15 +413,9 @@ public static PyObject ReloadModule(PyObject module)
...
-            {
-                throw new PythonException();
-            }
+            Py.Throw();
             IntPtr m = Runtime.PyImport_ExecCodeModule(name, c);
-            if (m == IntPtr.Zero)
-            {
-                throw new PythonException();
-            }
+            Py.Throw();
```
filmor: Because this function either returns null and sets an exception, or it returns a module.

src/runtime/pythonengine.cs
```diff
@@ -1,18 +1,40 @@
...
+using System.Collections.Generic;
+using System.Linq;
 
 namespace Python.Runtime
 {
     /// <summary>
     /// This class provides the public interface of the Python runtime.
     /// </summary>
-    public class PythonEngine
+    public class PythonEngine : IDisposable
```
filmor: Because that's easier to use correctly than `Initialize` and `Shutdown`.

PR #353, opened 2017-02-02
Enable Embedded Unit Tests

vmuriart:

Enables embedded unit tests on AppVeyor

Shows failing tests on python3 during finalization but can't reproduce locally. It isn't consistent on AppVeyor either.

### Any other comments?

Downgrades nunit to 2.6.4. It can deal better with exceptions that occur during finalize, unit3 just freezes and holds the ci worker until AppVeyor times it out.

### Checklist


vmuriart:
Same reason it doesn't work with `nunit2` either 😄 . The difference is that `nunit2` writes out the traceback and releases control back.

PR #371, opened 2017-02-09
Upgrade NUnit to 3.6

vmuriart:

Since the embedded unit tests have been patched we can re-upgrade to `NUnit3` from @filmor .

`N/A`

### Any other comments?

On `v.2.x` of NUnit I'm noticing that the tests finish and hang for ~40 seconds before releasing control. on `v3.x` it gives some warnings but I'm not sure if thats related. 

Personally I'm indifferent to prefer `v2.x` since it released control and occasionally print the traceback that was key to figuring the old problem out. But since that looks to be solved, I don't have a strong argument.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,7 +37,7 @@ install:
 script:
   - export PYTHONPATH=`pwd`:$PYTHONPATH
   - python -m pytest
-  # - mono ./packages/NUnit.*/tools/nunit-console.exe src/embed_tests/bin/Python.EmbeddingTest.dll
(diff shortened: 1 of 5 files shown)


denfromufa:
do you mena this warning?

https://ci.appveyor.com/project/pythonnet/pythonnet/build/master-377/job/rpml92u11q1k5kl8#L484

vmuriart:
That one.
Also I just remembered, for some reason this messes with coverage. It makes OpenCover go over the unittests code itself. Probably needs to filter the namespace.

vmuriart:
Fixed the aforementioned coverage issue. @filmor did you still want to upgrade to NUnit3?

Issue #370, opened 2017-02-09
 why sys.stdout returned null


yagweb:
### Environment

-   Pythonnet version: 2.2.2 (.NET 4, x64)
-   Python version: 3.5
-   Operating System: Windows 10

### Details
Python is embedded in my .NET application using pythonnet.  I found the output of the print statement in a python script is not showing on the screen until a "sys.stdout.flush()" called. I don't known how to redirect sys.stdout in .NET.
In order to call the flush() method directly in .NET, I tried
 ```csharp
dynamic sys = PythonEngine.ImportModule("sys");
var stdout= sys.GetAttr("stdout"); // returns None
```
  and both sys.stdout and the variable stdout returned are None here.

In python, I tried 
```python
import sys
print(getattr(sys, 'stdout'))
```
and it outputs a `_io.TextIOWrapper` object.   

Why sys.stdout returned null here? How to make the output of python print statement show automatically?

Thanks.

denfromufa:
@yagweb can you try suggestions from here:

https://mail.python.org/pipermail/pythondotnet/2016-May/001777.html

http://stackoverflow.com/questions/37289082/redirect-stdout-stderr-from-python-embedded-in-net-application-to-text-box
https://mail.python.org/pipermail/pythondotnet/2016-May/001776.html

We should probably include some API to make this easier, since some people are already using this in ad-hoc ways.

denfromufa:
in your example make sure you acquire the GIL with `using (Py.GIL()) {}` and try this:
```
dynamic sys = PythonEngine.ImportModule("sys");
dynamic stdout= sys.stdout;
```

yagweb:
@denfromufa Yes my code was put in the brackets of '''using (Py.GIL()) {}'''. I have tried your code snippet, but the stdout is still null. I'm not sure whether it's a problem of CPython or pythonnet. Maybe somebody else can try it with the other versions of CPython to get more clues.

For redirecting stdout, 3 solutions are found according to these links, but the most elegant one (similar to the 3rd solution) provided by Cameron Hayne is not working because of the limitation of pythonnet. It needs “make any .NET object can be passed to CPython”. Here I give a briefly description, hope they can help with the API design. 
Solution 1: Not automatically, so not suggested.
‘’’
using (Py.GIL())
{
                PythonEngine.RunSimpleString("import sys\n"+
                "from io import StringIO\n"+
                "sys.stdout = mystdout = StringIO()\n");
               //test
[...]

PR #373, opened 2017-02-10
Passing an arbitrary .NET object as the value of an attribute of PyObject

yagweb:
This implement the function of passing an arbitrary .NET object as attribute of PyObject by dynamic type. 

Four modifications are located at file pyobject.cs, the methods for dynamic type TryGetMember and TrySetMember were modified, a new method GetAttrDynamic was added for TryGetMember, and a new method SetAttrDynamic was added for TrySetMember.

Unit tests were given in a new file "src/embed_tests/dynamic.cs".

An Example was given in the discussion of issue [#370](https://github.com/pythonnet/pythonnet/issues/370) , the Solution 4.

/***********Information below is outdated and can be removed*************/

However this doesn't fix the issue [why sys.stdout returned null #370](https://github.com/pythonnet/pythonnet/issues/370)
...
...

I don't know where to put the test, so I put it here. If needed, it can be added to Embedded tests with some work.

```
    class dynamicTest
    {
        public void Test()
        {
            var output = new Output();
            using (Py.GIL())
            {
                sys.stdout = output; //Now it's working

                //function test, print in python
                PythonEngine.RunSimpleString("print('Hello!')");

                //Check whether there are the same object
                var result1 = sys.stdout == output;
                Console.WriteLine($"Compare in .NET: {result1}");  //Output True   

                //Check whether we can get the attr of a python object 
                //when the value of attr is not a .NET object
                PythonEngine.RunSimpleString("import sys\n" +
                "from io import StringIO\n" +
                "sys.stderr = StringIO()\n");
                var bb = sys.stderr;
                bb.write("Hello!\n");
                Console.Write(bb.getvalue());

                /********Pass the .NET object in Python side*********/
                PythonEngine.RunSimpleString("import sys\n" +
                "from io import StringIO\n" +
                "sys.stderr = sys.stdout\n");
                //Compare in Python
                PythonEngine.RunSimpleString("import sys\n" +
                "print('Compare in Python: %s' % (sys.stderr is sys.stdout))\n");  //Output False

                //compare in .NET
                var result2 = sys.stdout == sys.stderr;
                Console.WriteLine($"Compare in .NET: {result2}");  //Output True

                /*********Pass the .NET object in .NET side********/
                sys.stdout = sys.stderr;
                //Compare in Python
                PythonEngine.RunSimpleString("import sys\n" +
                "print('Compare in Python: %s' % (sys.stderr is sys.stdout))\n");  //Output False

                //compare in .NET
                var result3 = sys.stdout == sys.stderr;
                Console.WriteLine($"Compare in .NET: {result3}");  //Output True    

                /*********When the .NET object was created in Python side*****************/
                PythonEngine.RunSimpleString(@"
import sys
from Test import Output
sys.stderr = Output()
");
                var aa = sys.stderr; //Return the .NET object, not a PyObject with an IntPtr
                aa.write("Hello!\n");            
            }
        }
    }
    public class Output
    {
        public void write(String str)
        {
            //str = str.Replace("\n", Environment.NewLine);
            Console.Write(str);
        }

        public void writelines(String[] str)
        {
            foreach (String line in str)
            {
                Console.Write(str);
            }
        }
        public void flush() { }
        public void close() { }
    }
```

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -29,6 +29,7 @@
 -   Sean Freitag ([@cowboygneox](https://github.com/cowboygneox))
 -   Serge Weinstock ([@sweinst](https://github.com/sweinst))
 -   Virgil Dupras ([@hsoft](https://github.com/hsoft))
+-   Wenguang Yang ([@yagweb](https://github.com/yagweb))
 -   Xavier Dupré ([@sdpython](https://github.com/sdpython))
 -   Zane Purvis ([@zanedp](https://github.com/zanedp))
 -   ([@ArvidJB](https://github.com/ArvidJB))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,8 @@ This document follows the conventions laid out in [Keep a CHANGELOG][].
 -   Added Embedded tests on Appveyor (#353)
 -   Added PY3 settings to configuration-manager (#346)
 -   Added `Slack` chat (#384)(#383)(#386)
+-   Added function of passing an arbitrary .NET object as the value
+    of an attribute of PyObject (#370)(#373)
 
 ### Changed
 
--- a/src/embed_tests/Python.EmbeddingTest.csproj
+++ b/src/embed_tests/Python.EmbeddingTest.csproj
@@ -65,6 +65,7 @@
     <DebugType>pdbonly</DebugType>
   </PropertyGroup>
   <ItemGroup>
+    <Reference Include="Microsoft.CSharp" />
     <Reference Include="nunit.framework, Version=3.6.0.0, Culture=neutral, PublicKeyToken=2638cd05610744eb, processorArchitecture=MSIL">
       <HintPath>..\..\packages\NUnit.3.6.0\lib\net40\nunit.framework.dll</HintPath>
     </Reference>
@@ -76,6 +77,7 @@
   </ItemGroup>
   <ItemGroup>
     <Compile Include="pyinitialize.cs" />
+    <Compile Include="dynamic.cs" />
     <Compile Include="pyimport.cs" />
     <Compile Include="pyiter.cs" />
     <Compile Include="pylong.cs" />
--- a/src/embed_tests/dynamic.cs
+++ b/src/embed_tests/dynamic.cs
@@ -1,0 +1,122 @@
+using System;
+using System.Text;
+using NUnit.Framework;
+using Python.Runtime;
+
+namespace Python.EmbeddingTest
+{
+    [TestFixture]
+    public class dynamicTest
+    {
+        private Py.GILState gil;
+
+        [SetUp]
+        public void SetUp()
+        {
+            gil = Py.GIL();
+        }
+
+        [TearDown]
+        public void TearDown()
+        {
+            gil.Dispose();
+        }
+
+        /// <summary>
+        /// Set the attribute of a pyobject with a .NET object.
+        /// </summary>
+        [Test]
+        public void AssignObject()
+        {
+            StringBuilder stream = new StringBuilder();
+            dynamic sys = Py.Import("sys");
+            sys.testattr = stream;
+            // Check whether there are the same object.
+            var _stream = sys.testattr.AsManagedObject(typeof(StringBuilder));
+            Assert.AreEqual(_stream, stream);
+
+            PythonEngine.RunSimpleString(
+                "import sys\n" +
+                "sys.testattr.Append('Hello!')\n");
+            Assert.AreEqual(stream.ToString(), "Hello!");
+        }
+
+        /// <summary>
+        /// Set the attribute of a pyobject to null.
+        /// </summary>
+        [Test]
+        public void AssignNone()
+        {
+            dynamic sys = Py.Import("sys");
+            sys.testattr = new StringBuilder();
+            Assert.IsNotNull(sys.testattr);
+
+            sys.testattr = null;
+            Assert.IsNull(sys.testattr);
+        }
+
+        /// <summary>
+        /// Check whether we can get the attr of a python object when the
+        /// value of attr is a PyObject.
+        /// </summary>
+        [Test]
+        public void AssignPyObject()
+        {
+            dynamic sys = Py.Import("sys");
+            dynamic io = Py.Import("io");
+            sys.testattr = io.StringIO();
+            dynamic bb = sys.testattr; //Get the PyObject
+            bb.write("Hello!");
+            Assert.AreEqual(bb.getvalue().ToString(), "Hello!");
+        }
+
+        /// <summary>
+        /// Pass the .NET object in Python side.
+        /// </summary>
+        [Test]
+        public void PassObjectInPython()
+        {
+            StringBuilder stream = new StringBuilder();
+            dynamic sys = Py.Import("sys");
+            sys.testattr1 = stream;
+
+            //Pass the .NET object in Python side
+            PythonEngine.RunSimpleString(
+                "import sys\n" +
+                "sys.testattr2 = sys.testattr1\n"
+            );
+
+            //Compare in Python
+            PythonEngine.RunSimpleString(
+                "import sys\n" +
+                "sys.testattr3 = sys.testattr1 is sys.testattr2\n"
+            );
+            Assert.AreEqual(sys.testattr3.ToString(), "True");
+
+            //Compare in .NET
+            Assert.AreEqual(sys.testattr1, sys.testattr2);
+        }
+
+        /// <summary>
+        /// Pass the PyObject in .NET side
+        /// </summary>
+        [Test]
+        public void PassPyObjectInNet()
+        {
+            StringBuilder stream = new StringBuilder();
+            dynamic sys = Py.Import("sys");
+            sys.testattr1 = stream;
+            sys.testattr2 = sys.testattr1;
+
+            //Compare in Python
+            PyObject res = PythonEngine.RunString(
+                "import sys\n" +
+                "sys.testattr3 = sys.testattr1 is sys.testattr2\n"
+            );
+            Assert.AreEqual(sys.testattr3.ToString(), "True");
+
+            //Compare in .NET
+            Assert.AreEqual(sys.testattr1, sys.testattr2);
+        }
+    }
+}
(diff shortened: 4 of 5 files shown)


denfromufa:
@yagweb i like this PR, the test looks good and can be added to embedded tests. Let us know if you need any help?

vmuriart:
Reviewing now. I just pushed some whitespace changes to the `pr` to avoid dealing with those later.
@yagweb I'm adding the unittest you included above as well.

yagweb:
@denfromufa @vmuriart Thank you for your Response. I have added the unittest and fixed a bug of assigning null to the attribute of a pyobject. I'll commit it later.

vmuriart:
@yagweb no problem. I'll stash my changes then.

yagweb:
@denfromufa  @vmuriart I am sorry the commits are a little mess, because in the start I did not create a branch for this modification and directly worked on my master branch. I will merge them later.

The AssignObjectInPython has a very strange error, but it works well when I called it in my C# code directly, so I just ignore it.

vmuriart:
@yagweb whats the error you are seeing?

vmuriart:
irt the commit history, you can `squash` and `force-push` to cleanup your history.

yagweb:
@vmuriart Thank you for your help. The commits have been squashed.

The error occurs when creating a .NET object in python, and It disappeared when I try to debug with calling it in my C# code.

The error information is,
```
Python.EmbeddingTest.dynamicTest.AssignObjectInPython:
System.OverflowException : Arithmetic operation resulted in an overflow.
```
and the traceback is,
```
at System.IntPtr.op_Explicit(IntPtr value)
at Python.Runtime.ManagedType.GetManagedObject(IntPtr ob) in D:\git\pythonnet\src\runtime\managedtype.cs:line 31
at Python.Runtime.ClassObject.tp_new(IntPtr tp, IntPtr args, IntPtr kw) in D:\git\pythonnet\src\runtime\classobject.cs:line 53
at e__NativeCall.Call_3(IntPtr , IntPtr , IntPtr , IntPtr )
at Python.Runtime.NativeCall.Call_3(IntPtr fp, IntPtr a1, IntPtr a2, IntPtr a3) in D:\git\pythonnet\src\runtime\nativecall.cs:line 128
at Python.Runtime.MetaType.tp_call(IntPtr tp, IntPtr args, IntPtr kw) in D:\git\pythonnet\src\runtime\metatype.cs:line 154
at Python.Runtime.Runtime.PyRun_SimpleString(String code)
at Python.Runtime.PythonEngine.RunSimpleString(String code) in D:\git\pythonnet\src\runtime\pythonengine.cs:line 120
at Python.EmbeddingTest.dynamicTest.AssignObjectInPython() in D:\git\pythonnet\src\embed_tests\dynamic.cs:line 68
```

vmuriart:
Looks like you found a bug for us! https://www.codeproject.com/Questions/528923/Howplustoplusfix-aplus-OverflowExceptionplusocc

Most likely the link above is the issue/fix. I'll take a look tomorrow to make sure and fix it on master.

yagweb:
@vmuriart I found a way to reproduce the bug. Just call that unit test function twice, at the second call, this exception will occur. The following code can do this. Hope this can help you to fix it.

```

namespace Test
{
    class Program
    {
        {            
            PythonEngine.Initialize();
            using (Py.GIL())
            {
                PythonEngine.RunSimpleString(@"
import sys
from Test import outstream
sys.stdout = outstream()
");
            }
            PythonEngine.Shutdown();

            PythonEngine.Initialize();
            using (Py.GIL())
            {
                PythonEngine.RunSimpleString(@"
import sys
from Test import outstream
sys.stdout = outstream()
");    //**exception will occor here！**
            }
            PythonEngine.Shutdown();
        }
    }
    
    public class outstream { }
}
```

vmuriart:
Just pushed a [`fixup!`](http://fle.github.io/git-tip-keep-your-branch-clean-with-fixup-and-autosquash.html) to the branch to merge with the first commit. Just wanted to clarify in case you hadn't seen one of those before and thought I was being rude 😄 .

vmuriart:
@yagweb got an update. I formatted/clean-up the code, and fixed the overflow bug on #376. Once its merged, I recommend `squash` all commits here and `rebase` ontop of `master`. I can do this for you and `force-push` the changes here if you want.

The test is still failing though, #376 just allowed the exception underlying exception to bubble up. This is the [error](https://ci.appveyor.com/project/vmuriart/pythonnet/build/pr373-rb-1069/job/axmv92k3vxfox0c7#L506) i'm seeing now. 

The error you are seeing above when doing `Initialize`, `Finalize`, `Initialize`, is the vain of my existence 👎. #343 and #365 are examples of similar issues that gave similar errors. They occur because there is an underlying exception that we didn't handle and it causes the 2nd `initialize` to fail.

yagweb:
@vmuriart another bug of this PR has been fixed. I will try to squash all commits and rebase ontop of master by myself first.

yagweb:
@denfromufa @vmuriart During developing the unittests, I found it's hard to evaluate a python expression and obtain the results. So I modified the PythonEngine.RunString method, add a new class called PySession. Now I can do this:

```
using (var ps = new PySession())
{
    var result = ps.Eval("1+2"); //return 3

    dynamic sys = ps.Import("sys");  //import, it will also be added in globals
    ps.SetLocal("stdout", new Output()); //declare a local variable
    ps.Exec("sys.stdout = stdout");  //it's working!
    ps.Exec("print('Hello!')");

    sys.stderr = 12;
    ps.Exec("print(sys.stderr+1)");

    ps.SetGlobal("bb", 100);//declare a global variable
    ps.SetLocal("cc", 10);//declare a local variable
    ps.Exec("aa=bb+cc+3");   
    var aa = ps.Get("aa"); //return 113
    Console.WriteLine(aa);
}
```
what do you think about? I'm not familiar with git, I'll try to create a new branch and create a new PR for this later.

denfromufa:
@yagweb regarding PySession can you please open an issue first, so that we can quickly iterate on purpose, design and API. You can open a pull request, but it may get affected by decisions in the issue. I think this  closely resembles the Scopes in IronPython.

vmuriart:
@yagweb I think you are missing a reference to `<Reference Include="Microsoft.CSharp" />` on `EmbeddingTest.csproj`. Probably got lost in the merge conflict resolution.

yagweb:
@denfromufa Thank you for your suggestion, I opened the issue [here](https://github.com/pythonnet/pythonnet/issues/380).

yagweb:
@vmuriart Thank you, I  have squashed this commit.

vmuriart:
pushed a fixup to minimize whitespace/encoding changes. 
From your other `pull_request` it sounds like `AssignObjectInPython` test is no longer needed and consequentially neither would `outstream` be needed?

yagweb:
@denfromufa @vmuriart I read some source code, and found pythonnet already had some powerful methods. So, this branch undergo a major revision. Please check it again.

1. the GetAttrDynamic and SetAttrDynamic method is deleted.
2. the TrySetMember should call set attribute method of CPython no matter it has the attribute or not. Because python object is dynamic. If not, a new attribute cannot be found in CPython.
3. Corresponding to the TrySetMember method, the TryGetMember method only get attribute from CPython.

The unit tests file also changed. The AssignObjectInPython has been removed and the outstream class has been replace with the StringBuilder class.

yagweb:
@vmuriart The method AssignObjectInPython can be removed. but taking the bug it found,  I think a new unit test method below could be to the file pyinitialize.cs.
```
[Test]
[Ignore("System.ArgumentException : Cannot pass a GCHandle across AppDomains")]
public void Main()
{
    PythonEngine.Initialize();
    using (Py.GIL())
    {
        PythonEngine.RunSimpleString(
            "import sys\n"+
            "from System.Text import StringBuilder\n"+
            "sys.attr1 = StringBuilder()\n");
    }
    PythonEngine.Shutdown();

    PythonEngine.Initialize();
    using (Py.GIL())
    {
        PythonEngine.RunSimpleString(
            "import sys\n" +
            "from System.Text import StringBuilder\n" +
            "sys.attr1 = StringBuilder()\n");
    }
    PythonEngine.Shutdown();
}
```

The code below can achieve the same functionality but with no error. It reveals that,
1. We should avoid new .NET object in Python when the host is a .NET program.
2. This may cause by CPython new .NET object in a new .NET environment, not in the current one.(unprecise)。Maybe it can help to resolve this bug.
```
[Test]
public void Main2()
{
    PythonEngine.Initialize();
    using (Py.GIL())
    {
        dynamic sys = PythonEngine.ImportModule("sys");
        sys.attr1 = new StringBuilder();
    }
    PythonEngine.Shutdown();

    PythonEngine.Initialize();
    using (Py.GIL())
    {
        dynamic sys = PythonEngine.ImportModule("sys");
        sys.attr1 = new StringBuilder();
    }
    PythonEngine.Shutdown();
}
```

vmuriart:
@yagweb thanks for taking the time to digging more into that bug. It looks to be an extension of #343. Your test is very similar to the one in [`pyinitialize`](https://github.com/pythonnet/pythonnet/blob/0f00e3a995d1d00d863ce9c170f8809722bcf1c9/src/embed_tests/pyinitialize.cs#L30-L38) so I'll probably add the test there.

Is the issue only present when using `StringBuilder()` or can it be any arbitrary object? (ie. a string?).

yagweb:
@vmuriart I tried several RunSimpleString combinations in the two Initialize. If keep the two RunSimpleString statement the same, System.Int32 and System.String can also trigger this exception. If the .NET object created in the two RunSimpleString statement are different, some can trigger the exception, some cannot.
So I think it can be any arbitrary object. I updated my last comment to make the test simpler.

vmuriart:
@yagweb sorry it took me a while to review, I kept getting distracted while reviewing this. 
Looks good. Just rebased it and will merge once all CI checks complete. 

irt the other issue. We should open a new issue to summarize the findings.

yagweb:
@vmuriart I got the merged message. I will rebase my branches. Thank you!

vmuriart:
No worries. I had to rebase it because I updated the `CHANGELOG` and introduced a merge conflict. Also gave me a chance to clean up some whitespacing :nerd_face:

PR #376, opened 2017-02-14
Fix OverflowException on 64bit due to bad cast

vmuriart:

Fix OverflowException on 64bit due to bad cast

Error message was `OverflowException occurred. Arithmetic operation resulted in an overflow.`
Link below explains in more detail the issue. This fixes the issue in both 32/64bit.

https://www.codeproject.com/Answers/899343/How-to-fix-OverflowException-occurred-Arithmetic-o#answer2

Issue found while working on #373.

### DIFF
--- a/src/runtime/clrobject.cs
+++ b/src/runtime/clrobject.cs
@@ -11,7 +11,7 @@ namespace Python.Runtime
         {
             IntPtr py = Runtime.PyType_GenericAlloc(tp, 0);
 
-            var flags = (int)Marshal.ReadIntPtr(tp, TypeOffset.tp_flags);
+            var flags = (long)Marshal.ReadIntPtr(tp, TypeOffset.tp_flags);
             if ((flags & TypeFlags.Subclass) != 0)
             {
                 IntPtr dict = Marshal.ReadIntPtr(py, ObjectOffset.DictOffset(tp));
--- a/src/runtime/debughelper.cs
+++ b/src/runtime/debughelper.cs
@@ -89,7 +89,7 @@ namespace Python.Runtime
         internal static void DumpInst(IntPtr ob)
         {
             IntPtr tp = Runtime.PyObject_TYPE(ob);
-            var sz = (int)Marshal.ReadIntPtr(tp, TypeOffset.tp_basicsize);
+            var sz = (long)Marshal.ReadIntPtr(tp, TypeOffset.tp_basicsize);
 
             for (var i = 0; i < sz; i += IntPtr.Size)
             {
--- a/src/runtime/managedtype.cs
+++ b/src/runtime/managedtype.cs
@@ -28,7 +28,7 @@ namespace Python.Runtime
                     tp = ob;
                 }
 
-                var flags = (int)Marshal.ReadIntPtr(tp, TypeOffset.tp_flags);
+                var flags = (long)Marshal.ReadIntPtr(tp, TypeOffset.tp_flags);
                 if ((flags & TypeFlags.Managed) != 0)
                 {
                     IntPtr op = tp == ob
@@ -63,7 +63,7 @@ namespace Python.Runtime
                     tp = ob;
                 }
 
-                var flags = (int)Marshal.ReadIntPtr(tp, TypeOffset.tp_flags);
+                var flags = (long)Marshal.ReadIntPtr(tp, TypeOffset.tp_flags);
                 if ((flags & TypeFlags.Managed) != 0)
                 {
                     return true;
--- a/src/runtime/metatype.cs
+++ b/src/runtime/metatype.cs
@@ -247,7 +247,7 @@ namespace Python.Runtime
         {
             // Fix this when we dont cheat on the handle for subclasses!
 
-            var flags = (int)Marshal.ReadIntPtr(tp, TypeOffset.tp_flags);
+            var flags = (long)Marshal.ReadIntPtr(tp, TypeOffset.tp_flags);
             if ((flags & TypeFlags.Subclass) == 0)
             {
                 IntPtr gc = Marshal.ReadIntPtr(tp, TypeOffset.magic());


denfromufa:
@vmuriart do you now see my review?

https://github.com/pythonnet/pythonnet/pull/376#pullrequestreview-21803849

vmuriart:
@denfromufa Yup I see it. The casting will do the right thing on both 32 and 64bit systems due to some `clr` magic (can't find a reference for this).

The issue would be that as part of this `magic` on 32bit systems, `clr` needs [to do 2 transactions to update the value](http://stackoverflow.com/questions/596065/under-c-sharp-is-int64-use-on-a-32-bit-processor-dangerous). This would be an issue if the `flags` were a shared variable, but since they aren't it should be safe to use.

denfromufa:
@vmuriart we have other places with this cast:

https://github.com/pythonnet/pythonnet/blob/master/src/runtime/debughelper.cs#L92
https://github.com/pythonnet/pythonnet/blob/master/src/runtime/clrobject.cs#L14
https://github.com/pythonnet/pythonnet/blob/master/src/runtime/metatype.cs#L250

denfromufa:
@vmuriart actually it is quite safe to cast intptr to int64 on both architectures. I don't think any clr magic is necessary.

http://stackoverflow.com/a/18173246/2230844

https://msdn.microsoft.com/en-us/library/x65fb868(v=vs.110).aspx

filmor:
Shouldn't those be `ReadInt32` calls anyhow? I just checked `tp_flags` which is declared as `long` so it's actually a 32bit integer.

vmuriart:
The problem is that the pointers themselves are 64bit integers which is what caused the issue.
> TL;DR: You're casting 64-bits IntPtr to a Int32... Which causes "overflow"
https://www.codeproject.com/Answers/899343/How-to-fix-OverflowException-occurred-Arithmetic-o#answer2

denfromufa:
Long is 64bit

https://msdn.microsoft.com/en-us/library/ctetwysk.aspx

vmuriart:
@denfromufa I made the same mistake earlier as you did. @filmor said `c long` which is indeed [32bit](https://en.wikipedia.org/wiki/C_data_types#Basic_types) vs `c# long` which is 64bit.

filmor:
@vmuriart `IntPtr` is `sizeof(int*)` which is 32bit on x86 and 64bit on x86_64.
@denfromufa  That's why I corrected to "C `long`", in C on most platforms, `long` is 32bit, even if `int*` is 64bit.

What I'm saying is we currently read `tp_flags`, which is (at least for the main x86 platforms) a 32bit integer as if it was a pointer, which strikes me as quite a bad idea.

vmuriart:
@filmor thanks for the clarification 👍. If i understood correctly, your concern is not with the `pull_request` but rather with the current implementation of `tp_flags` being a `32bit` integer correct?

filmor:
No, my concern is with this pull request. If we change something to these struct accesses we should do it right. And the right way is not to read it as pointer-sized and then blowing it up to 64bit. The right way is to find the correct declarations (which will probably always be 32bit integers) and use the respective `Marshal` function on this one.

denfromufa:
@filmor I see now, in my email it still had long, not "c long" :)

"C long" is 32-bit on Windows and 64-bit on *Nix OSes. But this is
implementation details and not guaranteed.

So ReadIntPtr should take care of architecture differences.

filmor:
It takes care of architecture differences with regard to *pointers*, not `long`.

denfromufa:
> It takes care of architecture differences with regard to pointers, not long.

@filmor you are right, but still "c long" does not have to be 32bit.

denfromufa:
@filmor @vmuriart we can use ctypes to determine this!

image

vmuriart:
I'll rewrite this after I come back from vacation. The better thing to do is split up the `&` and do the comparison to `IntPtr.Zero` to avoid casting at all.

vmuriart:
@yagweb do you still have a way to reproduce this issue we originally saw on #373?

yagweb:
@vmuriart The example I provided can reproduce this OverflowException you talk about. After applying you previous fixup with replacing int to Int64, It will raise the appdomain issue two statements later.
Using any .NET object in the test function can trigger the exception, not just StringBuilder.

I can create a PR to add the unit test if you need.

vmuriart:
Yes please. THat be great

denfromufa:
@vmuriart @filmor do you agree to work on `typeof(tp_flags)==c_long` in another pull request? Looks like this localized problem is temporarily resolved with `long`. But this code may change as part of the bigger issue. Definitely we cannot use the ctypes approach during `Py_Initialize()` and I have not found pinvoke equivalent. We can pass a compilation flag to c# in setup.py to define the size of c_long.

yagweb:
@vmuriart @denfromufa @filmor A Minimal unit test to reproduce the Overflow Exception and across AppDomains Exception was provided [here](https://github.com/pythonnet/pythonnet/pull/393).

vmuriart:
@denfromufa im ok with continuing the work for this on another PR and applying `long` for now.
For the `c_long` size, why can't we use `runtime.Is32Bit`?

denfromufa:
For c_long the size depends on platform implementation details:

http://stackoverflow.com/a/2800654/2230844

C#/IL cannot be aware of this. Like I suggested we can pass a compilation
flag from setup.py to .NET project.

filmor:
Meh, we have come so far in having a single build here, I wouldn't want to paddle back just for this one tiny detail. I guess it's easier to just bite the bullet and have `WriteLong` and `ReadLong` functions that behave like `R/WIntPtr` everywhere and like `R/WInt32` on Windows.

denfromufa:
@filmor can explain what you mean by these terms?

> R/WIntPtr everywhere and like R/WInt32 on Windows.

filmor:
@denfromufa Shortcuts for `ReadLong` and `WriteLong`.

@vmuriart I don't get what you are trying

filmor:
Sorry, fat-fingered the Close button :/

filmor:
I meant a function like this:

    long ReadLong(IntPtr ptr, int offset)
    {
        if (IsWindows)
            return (long)Marshal.ReadInt32(ptr, offset);
        return (long)Marshal.ReadIntPtr(ptr, offset);
    }

Same for write.

Issue #265, opened 2016-09-16
meta-issue: overloading improvements


denfromufa:
- [ ] ref/out https://github.com/pythonnet/pythonnet/pull/227
- [ ] constructor https://github.com/pythonnet/pythonnet/pull/239 https://github.com/pythonnet/pythonnet/issues/238
- [ ] subtypes https://github.com/pythonnet/pythonnet/issues/203
- [ ] add content manager to completely ignore type checking for overloading https://github.com/pythonnet/pythonnet/issues/174#issuecomment-192835858
- [ ] fix tests when only type-checking with only one method available (no overloading) https://github.com/pythonnet/pythonnet/issues/190
- [ ] default arguments https://github.com/pythonnet/pythonnet/issues/80
- [ ] unsigned https://github.com/pythonnet/pythonnet/issues/283
- [ ] params https://github.com/pythonnet/pythonnet/issues/331
- [ ] named/keyword arguments

ArvidJB:
How about the following idea: we could reuse the work that the DLR has already done to support correct overloading resolution (for IronPython). To use that we would first convert the Python arguments to their corresponding .NET types and then let the DLR decide which of the overloads to pick. This should add support for named arguments, generic type inference, etc.
I tinkered with this a bit and I think it would work correctly, but it would be a fairly major piece of work. Do you think this is a worthwhile route to pursue? It would add a reference to https://github.com/IronLanguages/dlr to python.net by adding this NuGet package https://www.nuget.org/packages/DynamicLanguageRuntime/

denfromufa:
@ArvidJB this is interesting suggestion. One problem I see is that conversion from Python to .NET types can still be ambiguous even before letting DLR to do the overloading resolution. For example, passing a float could be converted to a single or a double, passing an integer could be converted to int16, int32, or int64.

Here are some comments from @matthid where he suggested to use Roslyn for this:

https://github.com/pythonnet/pythonnet/issues/203#issuecomment-264690218

AlexCatarino:
@denfromufa, you haven't listed "named arguments". 
I guess it is not support by @ArvidJB's comment. 
Our experience at QuantConnect is that it is not. Could you please confirm?

denfromufa:
@AlexCatarino probably not, can you provide a failing case not working for you?

jaredbroad:
Fairly specific to our API but: it silently defaults as if the named args aren't there. 

```
class BasicTemplateAlgorithm(QCAlgorithm):

[...]

PR #377, opened 2017-02-15
Method overload object type

### LINKED ISSUE #203: Pythonnet 2.1 not identifying System.Object type

vmuriart:

Fixes Method object overloading without breaking changes introduced by #131 and #151.

I added the tests/examples from #131 to the [tests](https://github.com/pythonnet/pythonnet/blob/6afc9e684856fca988c6d5260ee2d6de3735b7cf/src/tests/test_method.py#L748-L771) to ensure that the behavior didn't get reverted.

Closes #203 

### Any other comments?

This patch fixes the issue from #203, but `MethodBinder.Bind` needs rewriting since its basically been patched over and over again to fix edge cases. I didn't want to do it as part of this `pull_request` since it would affect more than just fixing #203.

### Checklist

### DIFF
--- a/src/runtime/methodbinder.cs
+++ b/src/runtime/methodbinder.cs
@@ -375,7 +375,7 @@ namespace Python.Runtime
                             if (clrtype != null)
                             {
                                 var typematch = false;
-                                if (pi[n].ParameterType != clrtype)
+                                if ((pi[n].ParameterType != typeof(object)) && (pi[n].ParameterType != clrtype))
                                 {
                                     IntPtr pytype = Converter.GetPythonTypeByAlias(pi[n].ParameterType);
                                     pyoptype = Runtime.PyObject_Type(op);
--- a/src/testing/methodtest.cs
+++ b/src/testing/methodtest.cs
@@ -131,6 +131,56 @@ namespace Python.Test
             return "Got object";
         }
 
+        public static string TestOverloadedObjectTwo(int a, int b)
+        {
+            return "Got int-int";
+        }
+
+        public static string TestOverloadedObjectTwo(string a, string b)
+        {
+            return "Got string-string";
+        }
+
+        public static string TestOverloadedObjectTwo(string a, int b)
+        {
+            return "Got string-int";
+        }
+
+        public static string TestOverloadedObjectTwo(string a, object b)
+        {
+            return "Got string-object";
+        }
+
+        public static string TestOverloadedObjectTwo(int a, object b)
+        {
+            return "Got int-object";
+        }
+
+        public static string TestOverloadedObjectTwo(object a, int b)
+        {
+            return "Got object-int";
+        }
+
+        public static string TestOverloadedObjectTwo(object a, object b)
+        {
+            return "Got object-object";
+        }
+
+        public static string TestOverloadedObjectTwo(int a, string b)
+        {
+            return "Got int-string";
+        }
+
+        public static string TestOverloadedObjectThree(object a, int b)
+        {
+            return "Got object-int";
+        }
+
+        public static string TestOverloadedObjectThree(int a, object b)
+        {
+            return "Got int-object";
+        }
+
         public static bool TestStringOutParams(string s, out string s1)
         {
             s1 = "output string";
--- a/src/testing/methodtest.cs
+++ b/src/testing/methodtest.cs
@@ -116,6 +116,21 @@ namespace Python.Test
             return args;
         }
 
+        public static string TestOverloadedNoObject(int i)
+        {
+            return "Got int";
+        }
+
+        public static string TestOverloadedObject(int i)
+        {
+            return "Got int";
+        }
+
+        public static string TestOverloadedObject(object o)
+        {
+            return "Got object";
+        }
+
         public static bool TestStringOutParams(string s, out string s1)
         {
             s1 = "output string";
--- a/src/tests/test_method.py
+++ b/src/tests/test_method.py
@@ -790,3 +790,32 @@ def test_object_in_param():
 
     res = MethodTest.TestOverloadedObject("test")
     assert res == "Got object"
+
+
+def test_object_in_multiparam():
+    """Test method with object multiparams behaves"""
+
+    res = MethodTest.TestOverloadedObjectTwo(5, 5)
+    assert res == "Got int-int"
+
+    res = MethodTest.TestOverloadedObjectTwo(5, "foo")
+    assert res == "Got int-string"
+
+    res = MethodTest.TestOverloadedObjectTwo("foo", 7.24)
+    assert res == "Got string-object"
+
+    res = MethodTest.TestOverloadedObjectTwo("foo", "bar")
+    assert res == "Got string-string"
+
+    res = MethodTest.TestOverloadedObjectTwo("foo", 5)
+    assert res == "Got string-int"
+
+    res = MethodTest.TestOverloadedObjectTwo(7.24, 7.24)
+    assert res == "Got object-object"
+
+
+def test_object_in_multiparam_exception():
+    """Test method with object multiparams behaves"""
+
+    with pytest.raises(TypeError):
+        MethodTest.TestOverloadedObjectThree("foo", "bar")
--- a/src/tests/test_method.py
+++ b/src/tests/test_method.py
@@ -769,3 +769,25 @@ def test_wrong_overload():
     res = System.Math.Max(System.Double(50.5), 50.1)
     assert res == 50.5
     assert type(res) == float
+
+
+def test_no_object_in_param():
+    """Test that fix for #203 doesn't break behavior w/ no object overload"""
+
+    res = MethodTest.TestOverloadedNoObject(5)
+    assert res == "Got int"
+
+    with pytest.raises(TypeError):
+        MethodTest.TestOverloadedNoObject("test")
+
+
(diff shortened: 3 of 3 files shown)


denfromufa:
@vmuriart what happens with multiple arguments, when overloaded methods have `object` argument?

vmuriart:
@denfromufa good question. No idea.

`foo(int a, object b)`
`foo(object a, int b)`
`foo(object a, object b)`
`foo(int a, int b)`

These 4 functions look enough to cover the possibilities or do I need more?

denfromufa:
we need one super-loaded method which covers all these cases from this meta-issue :)

https://github.com/pythonnet/pythonnet/issues/265

vmuriart:
That sounds much broader than the scope of this `pull_request`. 
I'm planning on refactoring `methodbinder` to clean up alot of what is going in there.

denfromufa:
@vmuriart I understand, your example looks good. My suggestion was not serious anyway. But we may hit issues when dealing with sub-classes and ambiguous cases in the overload. IMO, pythonnet should be very strict by default in method resolution. See examples how C# compiler had to break method overloading in .NET 4.0:

http://csharpindepth.com/Articles/General/Overloading.aspx

vmuriart:
@denfromufa No worries. My goal is to restore similar behavior to what `pythonnet==2.0.0` had irt method objects. I had to freeze the version of pythonnet I was using to `v2.0` because of it.
I'm ok w switching to a stricter method resolution, as long as it is part of `v3.0` 😄

vmuriart:
@denfromufa Multiple arguments with objects works as expected.

denfromufa:
@vmuriart no failure cases in overloads to make sure that they fail. No strings in overloads, although you use strings as arguments.

vmuriart:
I think I understood your request correctly. I added a test to check for failure/exception.

denfromufa:
Can you add more overloaded methods such as:

```
public static string TestOverloadedObjectTwo(string a, string b){return
"Got string-string";}
```

Which one would the method resolution pick object or string? Does the order
of defined methods matter?

Wrong result would be much worse than flagging ambiguity or failure in
method resolution.

vmuriart:
@denfromufa I think you are asking for [these tests](https://github.com/pythonnet/pythonnet/blob/6afc9e684856fca988c6d5260ee2d6de3735b7cf/src/tests/test_method.py#L748-L771) I added them on a [separate commit](https://github.com/pythonnet/pythonnet/commit/ce40fb1a72e1e10ec8428bdd55bab7c08e2866ca). 

Earlier when I was working on this pull request I used those tests to make sure I wasn't reverting the behavior you are describing. On earlier implementations, I reverted the behavior so those tests kept me in check.

The string-string check is analogous to the [int-int](https://github.com/pythonnet/pythonnet/pull/377/files#diff-1eb23401deaa91056741630bfbed1db7R134) check vs [obj-obj](https://github.com/pythonnet/pythonnet/pull/377/files#diff-1eb23401deaa91056741630bfbed1db7R149) check

vmuriart:
@denfromufa `int` are also a subclass of `System.Object`.

```
System.Object
   System.ValueType
      System.Int32
```

denfromufa:
yes, int is also object subclass. However you use int in C# overloaded methods, while not using strings. Let me give some examples.

denfromufa:
@vmuriart here is what I suggested to add. I cannot test locally, due to #385 

```CSHARP
public static string TestOverloadedObject(object o)
        {
            return "Got object";
        }

        public static string TestOverloadedObjectTwo(int a, int b)
        {
            return "Got int-int";
        }

        public static string TestOverloadedObjectTwo(string a, string b)
        {
            return "Got string-string";
        }

        public static string TestOverloadedObjectTwo(string a, int b)
        {
            return "Got string-int";
        }

        public static string TestOverloadedObjectTwo(string a, object b)
        {
            return "Got string-object";
        }

        public static string TestOverloadedObjectTwo(int a, object b)
        {
            return "Got int-object";
        }

        public static string TestOverloadedObjectTwo(object a, int b)
        {
            return "Got object-int";
        }

        public static string TestOverloadedObjectTwo(object a, object b)
        {
            return "Got object-object";
        }

        public static string TestOverloadedObjectThree(object a, int b)
        {
            return "Got object-int";
        }

        public static string TestOverloadedObjectThree(int a, object b)
        {
            return "Got int-object";
        }
```

```
def test_object_in_multiparam():
    """Test method with object multiparams behaves"""

    res = MethodTest.TestOverloadedObjectTwo(5, 5)
    assert res == "Got int-int"

    res = MethodTest.TestOverloadedObjectTwo(5, "foo")
    assert res == "Got int-object"

    res = MethodTest.TestOverloadedObjectTwo("foo", 5)
    assert res == "Got object-int"

    res = MethodTest.TestOverloadedObjectTwo("foo", "bar")
    assert res == "Got object-object"

    res = MethodTest.TestOverloadedObjectTwo("foo", System.Object)
    assert res == "Got string-object"
   
    res = MethodTest.TestOverloadedObjectTwo("foo", "bar")
    assert res == "Got string-string"

    res = MethodTest.TestOverloadedObjectTwo("foo", 5)
    assert res == "Got string-int"


```

vmuriart:
ok, I'll add these tomorrow

vmuriart:
@denfromufa Added your tests, but had to modify remove a couple of the conditions I had added earlier since the function signatures had changed.

src/runtime/methodbinder.cs
```diff
@@ -375,7 +375,7 @@ internal Binding Bind(IntPtr inst, IntPtr args, IntPtr kw, MethodBase info, Meth
                             if (clrtype != null)
                             {
                                 var typematch = false;
-                                if (pi[n].ParameterType != clrtype)
+                                if (pi[n].ParameterType != typeof(object) && pi[n].ParameterType != clrtype)
```
denfromufa: can you add brackets around each logical?

denfromufa:
@vmuriart I left 2 comments inline with the code

vmuriart:
I'll review on Tuesday. Though the parenthesis seem superfluous since it
isn't consistent throughout.

denfromufa:
yes, brackets is very minor. But I feel the same way as this person:

http://stackoverflow.com/a/23816736/2230844

vmuriart:
@denfromufa Added the parenthesis. Good to squash-merge?

denfromufa:
you did not address the second comment. @vmuriart will this break if you add:
```
public static string TestOverloadedObjectTwo(int a, string b)
        {
            return "Got int-string";
        }
```

vmuriart:
There was only a single comment. 

image

And the answer is no. It simply changes the result of which method is used as it will pick the closest match it can find before it picked `object`.

denfromufa:
so can you add it? because testing closest match is important.

so you don't see this?

image

vmuriart:
It's not there. It even says on your screen shot that its `pending`.

src/tests/test_method.py
```diff
@@ -769,3 +769,50 @@ def test_wrong_overload():
...
+    assert res == "Got object"
+
+
+def test_object_in_multiparam():
+    """Test method with object multiparams behaves"""
+
+    res = MethodTest.TestOverloadedObjectTwo(5, 5)
+    assert res == "Got int-int"
+
+    res = MethodTest.TestOverloadedObjectTwo(5, "foo")
```
vmuriart: Now its there.

vmuriart:
Added test and its passing as well.

src/testing/methodtest.cs
```diff
@@ -116,6 +116,71 @@ public static int[] TestOverloadedParams(int v, int[] args)
...
+        {
+            return "Got int-object";
+        }
+
+        public static string TestOverloadedObjectTwo(object a, int b)
+        {
+            return "Got object-int";
+        }
+
+        public static string TestOverloadedObjectTwo(object a, object b)
```
denfromufa: you don't test this method below

vmuriart: because you changed the test for it above. I re-added a test for it.

vmuriart:
@denfromufa test added and passing as well.

denfromufa:
@vmuriart I'm still surprised that this does not break ;) i approve on my side. in the future we should definitely improve testing all these combinations of arguments by using hypothesis or parametrizing pytest. Let's release v2.3.0.dev1 version after this merge.

@filmor @tonyroberts please review!

vmuriart:
@denfromufa I didn't want to introduce `parametrizing` tests yet. I figured I'd slowly introduce `py.test` features instead of all at once.

PR #379, opened 2017-02-15
Update version

denfromufa:
proper tag when building from master

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -368,7 +368,7 @@ if not os.path.exists(_get_interop_filename()):
 
 setup(
     name="pythonnet",
-    version="2.2.2",
+    version="2.3.0.dev1",
     description=".Net and Mono integration for Python",
     url='https://pythonnet.github.io/',
     license='MIT',
--- a/src/runtime/resources/clr.py
+++ b/src/runtime/resources/clr.py
@@ -2,7 +2,7 @@
 Code in this module gets loaded into the main clr module.
 """
 
-__version__ = "2.2.2"
+__version__ = "2.3.0.dev1"
 
 
 class clrproperty(object):


vmuriart:
Do we want to continue using the suffix `dev1` etc for development? I was dropping it on the branch I had w automatic version update.

denfromufa:
@vmuriart to be consistent with the past - yes. if you have something better, please suggest.

vmuriart:
Any thoughts on updating the [`AssemblyVersions`](https://github.com/pythonnet/pythonnet/blob/master/src/runtime/AssemblyInfo.cs#L8-L13)? I don't know why it was bumped to `v4.0.0.2` to begin with.

vmuriart:
@denfromufa you're right, lets keep it for now. Makes it easier to distinguish when users are using one of the CI builds vs one from `pypi`. Though I honestly prefer the dual branch setup of `master`/`develop` since that makes versioning easier in my experience (especially when dealing w HotFixes). 

I'll open an issue for it later to revisit the idea later. The bad thing about the dual branch system is that it can create a `branch-merge` spaghetti-hell if not managed properly (earlier git-history of the repo is an example of it)

denfromufa:
@vmuriart no idea on assembly version, but here is intial:

https://github.com/pythonnet/pythonnet/blob/614c4b07ff5d824fb5d92de7ced0e77387ac495c/pythonnet/pythonnet/src/runtime/assemblyinfo.cs

vmuriart:
Yup. 035b7ad006225edd9a9f58601c7409ea2ffe0cfb updates it to `2.0.0.2`. and then 74e178cdc65357809d2633fb7f757ba13dfbaf1a bumped it `4.0.0.1`. The closest thing I can come up with is that on that commit they said

> # Tool and Framework versions all bumped for .NET 4.0 #

And bumped the version number to `4.0.0.1` cuz of it.
Right now there is no issue with changing it, but it may cause issues once (if?) we start strong-name signing the package.

vmuriart:
codecov's servers have been on the fritz this morning. Most likely one of the CI builds timeout when submitting its report.

Issue #380, opened 2017-02-15
Add a class to manage the context of a interaction with Python and simplify the variable exchanging


yagweb:
"eval" and "exec" are the most frequently used methods when interacting with Python.  However, this two methods implemented [here](https://github.com/pythonnet/pythonnet/pull/389) are hard to use because maintain the global variable dictionary and local variable dictionary is a tedious and repeated work.

The nature of the problem lies in the variable exchanging between .NET and Python.

I think a scope class can be added to manage the context of a interaction with Python and simplify the variable exchanging.
A Scope should act like a Python Module in .NET. In the Scope class, many python functions can be wrapped to operate with the locals and globals for easy-to-use and speeding up. 

For example, the eval example provided in that PR,
```
dynamic sys = Py.Import("sys");
sys.attr1 = 100;
PyDict locals = new PyDict();
locals.SetItem("sys", sys);
locals.SetItem("a", new PyInt(10));
 
PythonEngine.Exec("c = sys.attr1 + a + 1", null, locals.Handle);
var c = locals.GetItem("c").AsManagedObject(typeof(Int32));

Console.WriteLine("c= {0}", c);
```

can be written as,

```
var ps = Py.Session('test')
sys = ps.Import("sys");
sys.attr1 = 100;
ps.SetLocal("a", 10);
ps.Exec("c = sys.attr1 + a + 1");   //pass
var c = ps.Get<System.Int32>("c");  //get the local variable and convert it to .NET type.

ps.Dispose(); 
or
ps.Suspend(); //use it later. Py.Session('test') will return it.
```

I implement a prototype of this proposal [here](https://github.com/pythonnet/pythonnet/pull/381) to help discussion.

yagweb:
I think It's a better practice to avoid using RunString with one or two python statements in .NET code, because it is low runtime efficiency. This proposal and [the proposal](https://github.com/pythonnet/pythonnet/pull/373) can help to achieve this goal.

vmuriart:
Do you still need to `initialize` and `finalize` the engine? From your proposal it doesn't sound like it but looking at the tests you are [initializing and finalizing it.](https://github.com/yagweb/pythonnet/blob/5634ccb8a2a40851b332fee9c60b3e241677ef6a/src/embed_tests/pyscope.cs#L12-L18)

yagweb:
@vmuriart Thank you for your remind. I updated the unit test and removed the initialize and finalize methods.

[...]

Issue #330, opened 2017-01-22
pypy support in pythonnet


denfromufa:
cpyext is mature enough (numpy, cython, scipy, pandas soon) to support pythonnet in pypy. In fact pypy used to have a branch which ran on top of CLR (similar to IronPython), but this branch is not maintained.

The use of pypy in .NET is quite popular request:

http://stackoverflow.com/questions/6084697/is-it-possible-to-embed-pypy-into-a-net-application

The only limitation of pypy is 32-bit on Windows and poor support for Python 3+, although the latter is improving with funding from Mozilla.

Also this might be quite nice collaboration with pypy developers due to their level of expertise in python C-API and cffi.

denfromufa:
https://bitbucket.org/pypy/pypy/src/default/pypy/module/cpyext/

vmuriart:
```python traceback
src/monoclr/pynetinit.c: In function ‘main_thread_handler’:
src/monoclr/pynetinit.c:106:5: error: unknown type name ‘Dl_info’
src/monoclr/pynetinit.c:107:5: warning: implicit declaration of function ‘dladdr’ [-Wimplicit-function-declaration]
src/monoclr/pynetinit.c:107:22: error: ‘Py_Initialize’ undeclared (first use in this function)
src/monoclr/pynetinit.c:107:22: note: each undeclared identifier is reported only once for each function it appears in
src/monoclr/pynetinit.c:109:23: error: request for member ‘dli_fname’ in something not a structure or union
src/monoclr/pynetinit.c:110:29: error: request for member ‘dli_fname’ in something not a structure or union
```

above is the traceback from building on pypi/travis

denfromufa:
I tried on pypy with pythonnet on Win10. It builds fine, but then to find `clr` during import.

```
C:\Python\pypy2-v5.6.0-win32>bin\pip install pythonnet
Collecting pythonnet
  Downloading pythonnet-2.2.2.tar.gz (1.5MB)
    100% |################################| 1.5MB 480kB/s
Building wheels for collected packages: pythonnet
  Running setup.py bdist_wheel for pythonnet ... done
  Stored in directory: C:\Users\denis\AppData\Local\pip\Cache\wheels\14\da\65\1042e21167805ad08c6319a87c9cd6b2e6e97ff828aa37eb3d
Successfully built pythonnet
Installing collected packages: pythonnet
Successfully installed pythonnet-2.2.2

C:\Python\pypy2-v5.6.0-win32>pypy
[...]

Issue #285, opened 2016-11-07
acquire GIL by default when making PyObject calls


denfromufa:
Right now the GIL has to be acquired for each call to PyObject functions or methods. Maybe these should be with GIL acquired by default? We can include context manager to force releasing of GIL by explicit request.

denfromufa:
here is how this was done by @matthid in F#:

https://yaaf.de/blog/post/2016-05-28/Having%20Fun%20with%20Computation%20Expressions

filmor:
Can you give an example? Do you think this is actually feasible for the next release? If not, please remove the milestone.

PR #381, opened 2017-02-15
add a scope class to manage the context of interaction with Python an…

yagweb:
It‘s a prototype of this proposal [here](https://github.com/pythonnet/pythonnet/issues/380) to help its discussion.

Modifications are located at file pythonengine.cs, A class named PyScope was added for the variable management and exchanging between .NET and Python.

Several unit tests were given in a new file "src/embed_tests/pyscope.cs" which can be used as examples.

One of the example,

```
using (Py.GIL())
{
    dynamic ps = Py.CreateScope();
    ps.Set("bb", 100);         //declare a variable
    ps.cc = 10;                  //declare a variable dynamically
    int aa = ps.Eval<int>("bb+cc+3");     //113

    //add a function to the scope, and call it
    //global keyword was used here to show what pyscope can do 
    ps.aa = 0; 
    ps.Exec(
        "def func1():\n" +
        "    global aa \n" +
        "    aa = bb+cc+3 \n"            
    );
    ps.func1(); //call the function
    int aa1 = ps.Get<int>("aa");  //get variable
    int aa2 = ps.aa.As<int>();     //get variable dynamically

    //create a new scope and import the variables
    PyScope ps2 = ps.CreateScope();
    ps2.ImportModule("numpy", "np");
    var aa3 = ps2.Eval<float>("np.sin(bb)");
}
```


src/runtime/pythonengine.cs
```diff
@@ -439,7 +454,7 @@ public static PyObject ModuleFromString(string name, string code)
                 borrowedLocals = false;
             }
 
-            var flag = (IntPtr)257; /* Py_file_input */
+            var flag = (IntPtr)_flag; /* Py_file_input */
```
vmuriart: The comment here should be updated

src/embed_tests/pyscope.cs
```diff
@@ -0,0 +1,74 @@
...
+        }
+
+        [TearDown]
+        public void TearDown()
+        {
+            PythonEngine.Shutdown();
+        }
+
+        /// <summary>
+        /// Set the attribute of a pyobject to null.
```
vmuriart: Is comment right?

yagweb:
@vmuriart These two comments are updated.

denfromufa:
@yagweb can you update docs or readme with few examples?

yagweb:
@denfromufa The unit tests in "src/embed_tests/pyscope.cs" are commented and can be used as examples. 

Currently the relationships among PythonEngine, Py, PyScope, Py.GILState, PyScope.PyScope are a little messy. The APIs  may changes at any time, so I may not add these examples into docs or readme until the APIs became stable.

vmuriart:
@yagweb should be clarify the intents/differences of them before adding the feature? I agree on not adding it to the documented `API` until we hash it out though.

denfromufa:
@yagweb @vmuriart I'm still struggling to understand use cases for scopes, except for storing local state.
Maybe I'm missing something? That's why I asked to add something in the docs.

yagweb:
@vmuriart I am sorry for that, the modifications are too many last time. Last time, the PR is cleaned up by separating a new PR ‘add eval and exec’ from it, merging two classes, renaming methods, etc. I updated all my comments and forgot to inform you. In the future, I will provide detailed comments for each commit.

@denfromufa @vmuriart Storing local state is the start point. I wanted to embedding python into a C# application using pythonnet, and found the APIs for this scene are hard to use.  In my another comment [here](https://github.com/pythonnet/pythonnet/issues/380), I concluded that 

> A scope class aims at managing the context of a interaction with Python and simplify the variable exchanging. It should act like a Python Module in .NET. In the Scope class, many python functions can be wrapped to operate with the local variables and global variables for easy-to-use and speeding up, such as the ImportAs method.

I just make a new commit, now this PR implements almost all the functions of IronPython about accessing Python code from .NET code. I still don't know whether it's enough or not until it's actually used.

yagweb:
@denfromufa @vmuriart  The most complicated thing about using pythonnet in .NET application is the GIL.  This is the cause of what I said "Currently the relationships among PythonEngine, Py, PyScope, Py.GILState, PyScope.PyScope are a little messy.". Here is a related [proposal](https://github.com/pythonnet/pythonnet/issues/285).

I think there are three scenes when embedding CPython.

1. If the application is single thread, maybe it's no need to AcquireLock and ReleaseLock all the time.

2. In a multi-thread application, if several fast execution python functions need to called, we can treat the CPython as a shared resource, and the GIL as a lock in .NET. We can wrap each function with AcquireLock and ReleaseLock, this is what the pythonnet do right now.

3. In a multi-thread application,  if the python functions needs waiting IO and very time consuming. Then the GIL cannot be managed in .NET code. The python threading can be used and the .NET code should use asynchronous calls to interacting with Python. A similar use case is the Dispatcher of WPF.

In the multi-thread scenes, it's better to use IronPython if possible. In the last scene, I think acquire GIL by default is impossible.

denfromufa:
@yagweb in my view the multi-threading and asynchronous execution can be outside of this PR. GIL issues are well-known in Python community. Besides multi-processing in CPython you can look at Cython with nogil contexts, and Numba with nopython mode for numerical code. PyPy-STM is not ready for production use and only available on Linux. PyPy with cpyext is currently crashing when importing pythonnet: 

https://github.com/pythonnet/pythonnet/issues/330

vmuriart:
Even for the 1st scene the `gc` and [`Finalize` in `classderived.cs`](https://github.com/pythonnet/pythonnet/blob/master/src/runtime/classderived.cs#L887-L890) might throw a wrench into things.

Looks like might need to rebase this one since #389 was merged in.

yagweb:
@denfromufa I see, thank you. 
When users need multi-core parallel in embedded CPython, they should design their code with any means including Cython, Numba etc. It’s out of pythonnet. So the 4th scene can be deleted from this list.

@vmuriart Thank you. it seems like there no "real" single-threaded application since the GC working behind and needing to call CPython. So the 1st scene can also be deleted from this list.

@denfromufa @vmuriart May it be better to make the PyScope class independent from the GIL solution?

denfromufa:
> @denfromufa @vmuriart May it be better to make the PyScope class independent from the GIL solution?

I think it is safer that GIL is held inside PyScope by default. You can maybe pass non-default argument to release the GIL.

vmuriart:
@yagweb just finished reviewing this `pr` and makes sense what you're proposing. I think there's some refactoring we can do to clean up the code and there's a couple tests I want to run (like what happens if we create two `PyScope` instead of a `PyScope` and a `SubScope`) (does changing a variable in the subscope that exists on the parent scope modify it).

vmuriart:
Looks like there is a `gc` issue lingering somewhere. 

https://travis-ci.org/pythonnet/pythonnet/jobs/204901628#L1606

denfromufa:
@vmuriart how about adding this?

https://github.com/pythonnet/pythonnet/pull/400#issuecomment-282617815

src/runtime/pythonengine.cs
```diff
@@ -491,6 +491,452 @@ public enum RunFlagType
...
+            this.state.AcquireLock();
+        }
+
+        public virtual void Dispose()
+        {
+            if(isDisposed)
+            {
+                return;
+            }
+            Runtime.XDecref(globals);
```
vmuriart: `gc` issue is most likely here. You'll need to check that the engine is initialize and that it isn't finalizing before running these.

vmuriart:
That wouldn't fix it. The issue is that the engine get shutdown earlier at "some point" and looks like currently there is nothing checking the status of the engine before trying to run more commands on it.

vmuriart:
Rebased and updated some of the internal changes

yagweb:
create unnamed pyscope, make PyScope.GILState save

@vmuriart what do you think about the refactoring?
 I suggest use the PyScope.GILState class to replace the Py.GILState class first

yagweb:
Currently, after the SubScope is created from a PyScope, they will be totally independent from each other. That is to say, changing a variable in the SubScope that exists on the parent scope will not affect the parent scope, vice versa. 

Changing a variable in the SubScope should never affect the parent scope. But there are is a way to make the changing of parent scope affect the SubScope, I don't know whether we need it or not. In my opinion, a SubScope should be disposed immediately after it has been used. A scope can be kept and shared.

vmuriart:
Looks like the Travis build didn't trigger. May need to push a dummy commit (add a new line to the readme or something like that) for it to trigger. The scope stuff makes sense, probably a couple tests to ensure behavior. irt refactoring, `pythonengine` itself needs to be refactored as a whole and is probably outside the scope of this `pr`.

vmuriart:
Looks like the latest change propagated the `gc` issue into the python tests.

yagweb:
@denfromufa @vmuriart I tried multi-thread programming in embedded Python, both python threading and .NET Thread.  I think it's better to make the PyScope class independent from GIL. Here is the reason and more.

I think there are two ways to make a CLR class which interacts with CPython thread-safe:

1. Make it independent from the GIL, and users must acquire the GIL before visit its members.
For this scene, I propose pythonnet provide a IPyObject interface, 
```
public interface IPyObject: IDisposable {}
```
all CLR classes defined in this way should implement this interface, to tell users GIL is needed to visit its member. 
PyObject can use this interface instead of the original IDisposable interface for consistency.
Class using this design was used in my practice, a separate issue can be opened for this topic.

2. In all the properties/members of the class which interacts with CPython, wrap its code with 
```
using(Py.GIL()
{
    /.../
}
```
Thus, this class can be treated a normal CLR class.

GILState object shared between methods leads thread-unsafe. Currently, the PyScope class stores GILState object in its field, so it must be refactored using either the 1st way or the 2nd way.

Considering the local dict is shared between methods, and during a conversation with CPython usually several methods of a PyScope will be called continuously, so I think the 1st way is the only choice. 
Although it may cause users write more statements, it help make the PyScope more general and thread-safe.

If you agreed with me, I will remove the GILState usage in the PyScope class.

denfromufa:
What is the use case for subscope or nested scope? Especially due to GIL this can get tricky? What is the problem that subscope can solve?

vmuriart:
compatibility applications written for IronPython for one.

vmuriart:
There is a nice diagram on the IronPython book explaining this, but I can't find it online.

denfromufa:
Which book?

vmuriart:
didn't realize they had more than one. 
https://www.amazon.com/IronPython-Action-Michael-J-Foord/dp/1933988339

denfromufa:
Just purchased from Manning. What chapter, figure or table number did you
mean from this book? Here is the most controversial review of this book.
How the tech world has changed since then?!

http://neopythonic.blogspot.com/2009/06/ironpython-in-action-and-decline-of.html?m=1

On Sat, Mar 4, 2017, 2:25 PM Victor Uriarte <notifications@github.com>
wrote:

didn't realize they had more than one.
https://www.amazon.com/IronPython-Action-Michael-J-Foord/dp/1933988339

—
You are receiving this because you were mentioned.


Reply to this email directly, view it on GitHub
<https://github.com/pythonnet/pythonnet/pull/381#issuecomment-284178369>,
or mute the thread

vmuriart:
Page 400

denfromufa:
OK, so I still don't see nested scopes or sub-scopes in embedding
IronPython from this book. Even more confusing is that the book and the
diagram refer to Scope and ScriptScope. The latter concept was created
specifically for DLR for pre-compiling execution graphs. This allows to
compile once and execute many times. It also provides interoperability with
IronRuby, since ScriptScope is a DLR concept not IronPython.
Another point is that IronPython most likely can have multiple scopes at
the same times without GIL (I have not verified this).

I don't think that any of this apply to pythonnet.

I like that we can isolate objects (except modules) within the scope, so
this use case is probably enough.

Please correct me if I missed anything?

So back to sub-scopes, what is the benefit?

yagweb:
Maybe the name 'SubScope' is misleading.

In my last comment, I gave a design principle of the 'SubScope' method, 

> 1. Changing a variable in the SubScope should never affect the origin scope. 
> 2. Changing origin scope can affect all its SubScopes. There are is a way to implement it, but I don't think it's a good design.
> 3. After the SubScope is created from a PyScope, they will be totally independent from each other. It's similar to clone, except all the local and global variables are put into the global dict of the SubScope. 

The 3th point is the PyScope choosed currently, so 'origin scope' may be a better address here than 'parent scope'. The method SubScope can be renamed to NewScope or something else to avoid misleading.
The field 'parent' will be removed from PyScope class, so there is no nested scopes.

For this kind of 'SubScope', the link @denfromufa gave can be used an example,
[a library for embedding python with pythonnet and using numpy, matplotlib](https://github.com/Lodomir/PythonInterface)

In the doc, it said,

> ... The component has to be initialized with an existing PythonConnection instance. Internally, however, it manages it own local dictionary, so no variables are overwritten while plotting. The global dictionary (e.g. imports) is shared, but this should normally not be a big problem.

PythonConnection there works as a PyScope here, it use push and pop to realize 'no variables are overwritten while plotting'. This leads to that it can never be used in a multi-thread environment.

In general, a **named** PyScope is a custom environment by importing module, functions, classes etc, such as a scientific environment, by calling its 'SubScope' method, you can obtain a private environment to do you work. An **anonymous** PyScope acts like the Scope of IronPython.

Use static method or make the 2nd constructor public can also do this private environment creation job, I think this way is more efficient and elegant.

yagweb:
The methods Compile, Execute and Execute<T> of PyScope have implemented to simulate the function of ScriptScope in IronPython.

The Compile function compile python code into ast (corresponding to the ScriptScope), then using Execute and Execute<T>, the ast can be execute many times high efficiently.

All the methods of PyScope has renamed to be compatible with IronPython Scope. 

Except, the GetVariable method of PyScope implements the function of GetVariable and TryGetVariable  of IronPython simultaneously. They will be seperated to be compate with IronPython.

yagweb:
I insist on removing the GIL operation from PyScope. The present PyScope is thread-unsafe. 

I can provided an example to show the usage of PyScope in a multi-thread environment after the GIL removed.

yagweb:
@vmuriart @denfromufa 
GIL is removed from PyScope. Constructors and APIs are cleaned.

GetVariable methods modified, and TryGetVariable methods are added to be compatible with IronPython.

A new unit test method TestThread is added. This test shows how the new PyScope can be used in multi-thread environment safely.

yagweb:
A better explaination for the relationship among PyScopes:

> A PyScope acts like a Python Module in CLR, not a Class. So it can be shared by importing and can not be inherited. 

For the PyScope class, a new method ImportScope is added, the SubScope method is renamed to CreateScope, it means, create an anonymous PyScope, and import all the objects from the current PyScope.

yagweb:
The dict ‘locals ’ and dict ‘globals’ are merged into one single dict ‘variables’;
add some new unit tests;
extract the code of pyscope class from pythonengine.cs and put them into a new file pyscope.cs.
branch rebased;

vmuriart:
Sorry for lack of follow-up from our side. We were prepping the release of `2.3.0`.  
I'll take a look in the morning. Thanks for keeping the branch upto date w the `master` branch, makes it easier to review 👍.

denfromufa:
@yagweb I plan to review during this week. BTW, boost.python also has a concept of scopes:

http://www.boost.org/doc/libs/1_48_0/libs/python/doc/v2/scope.html

Here is a blog post about ScriptScope in ironpython, ironically called pythonnet:

http://derekwill.com/2014/08/15/ironpython-sample-application/

One thing I noticed is both boost.python and ironpython implementation is the ability to also use classes.

So can we instantiate python objects in pyscope and then get them disposed?

filmor:
Why did you merge locals and globals? They have a distinct meaning and it can be quite useful to set both.

denfromufa:
@filmor can you give example?

yagweb:
@denfromufa This PyScope has the same behavior as a Python Module, so it can use class. A new unit test TestScopeClass was added to show this feature.

using (Py.GIL())
{
    ps.bb = 100;
    ps.Exec(
        "class class1():\n" +
        "    def __init__(self, value):\n" +
        "        self.value = value\n" +
        "    def call(self, arg):\n" +
        "        return self.value + bb + arg\n" +
        "    def update(self, arg):\n" +
        "        global bb\n" +
        "        bb = self.value + arg\n");//can use scope variables
    dynamic obj1 = (ps as dynamic).class1(20);      //create an instance of the class
    var result = obj1.call(10).AsManagedObject(typeof(int));
    Assert.AreEqual(130, result);

    obj1.update(10);
    result = ps.GetVariable<int>("bb");
    Assert.AreEqual(30, result);
}

yagweb:
@filmor No need to seperate the locals and globals in PyScope, suppose an example below,
```
var ps = Py.CreateScope()
ps.aa = 10

ps.Exec(
"def func1():\n", 
"    return aa * 10\n", //read the scope variables
<globals>, <locals>)

ps.Exec(
"def func2():\n", 
"    return func1(10)\n", //call the function created
<globals>, <locals>)

ps.func2()   //
```

In order to make the function 'func1' can read the scope variables, all the scope varables should be put into the \<globals\> dict.
As we know, the function 'func1' will be put into the \<locals\> dict.
In order to make the function 'func2' can call the function 'func1', the function 'func1' should be put into the \<globals\> dict.
Combine these two needs, after call the first ps.Exec statement, function 'func1' should be put into the 、\<globals\> and \<locals\> at the same time.

So pass the \<globals\> and \<locals\> parameter with the same dict is the best design here.

More important basis for this design: the globals and locals of a python module is the same.

yagweb:
If users want to exec a code in a scope and don't want the variables produced be put into that scope, the solution recommended is use the CreateScope/ImportScope method of the PyScope.
```
var ps = Py.CreateScope()
...action on the ps

//create a new scope an import the variables from that scope
var ps1 = ps.CreateScope()
or
var ps1 = Py.CreateScope()
ps1.ImportScope('the name')  #if the scope has a name, similar to python 'import * from module'

ps1.Exec(...)  #exec the code here
ps1.GetVarible<>(...)
ps1.Dispose()  #discard the scope
```

If this solution does not meet some very special demands, then call the ps.Variables() to get the variable dict, and use the PythonEngine.Exec directly.

yagweb:
I rebased this branch, it cause some tests throw very strange exception which worked before

filmor:
This is how *you* want to use this. I do agree that `locals = globals` is a sane default but not allowing the user to use this in the way they would expect from a Python embedding is not a good idea. There are perfectly fine use-cases for separate `locals` and `globals` dictionaries.

vmuriart:
@yagweb whats the exception, and which commit was it based on before?

denfromufa:
@yagweb @vmuriart all tests are passing in both CI, only coverage is complaining.

vmuriart:
@denfromufa if its a GC issue or something alike the issue can be intermittent. I have various branches/commits I can't merge in because they cause intermittent failures.

denfromufa:
I'll close and re-open this pull request to see if CI catches the exception this time.

yagweb:
@denfromufa @filmor I'm sorry for my carelessness, its a bug of the newly added method Variables. The unit tests in the project does not cover it. I just fixed it.

yagweb:
@filmor PyScope is an additional function for embedding python. PyScope has only one variable dict, but users can still use separate locals and globals dictionaries.

```
ps = Py.CreateScope()
//actions on the scope

locals = new PyDict()
PythonEngine.Exec(code, ps.Variables(), locals)

globals = new PyDict()
PythonEngine.Exec(code, globals, ps.Variables())

PythonEngine.Exec(code, globals, locals)

PythonEngine.Exec(code, ps.Variables(), ps.Variables()) //equalent to the ps.Exec
```

src/embed_tests/TestPyScope.cs
```diff
@@ -0,0 +1,346 @@
...
+        /// <summary>
+        /// Create a class in the scope, the class can read variables in the scope.
+        /// Its methods can write the variables with the help of 'global' keyword.
+        /// </summary>
+        [Test]
+        public void TestScopeClass()
+        {
+            using (Py.GIL())
+            {
+                dynamic _ps = ps;
```
denfromufa: why is this dynamic scope necessary?

denfromufa: oh, I see below, you dynamically accessed the class

denfromufa:
@yagweb this is enormous PR, sorry for late response. I reviewed and do not see any obvious defects. The API design is clean. 

@vmuriart wanted to review at some point:
https://github.com/pythonnet/pythonnet/pull/381#issuecomment-285926427

@filmor do you agree with @yagweb responses about globals() and locals()? If not, can you please provide a counter-example showing what is not possible with the current design?

filmor:
@denfromufa I don't agree, I'm preparing a longer review. In the current form I wouldn't merge this PR.

denfromufa:
@filmor how is your review of this PR going? @vmuriart did you have a chance to review?

src/runtime/pyscope.cs
```diff
@@ -0,0 +1,427 @@
...
+    {
+    }
+
+    public class PyScope : DynamicObject, IPyObject
+    {
+        public string Name
+        {
+            get;
+            private set;
+        }
```
filmor: What is this for and why is it not optional?

src/runtime/pyscope.cs
```diff
@@ -0,0 +1,427 @@
...
+
+        public PyScope CreateScope()
+        {
+            var scope = new PyScope(null);
+            scope.ImportScope(this);
+            return scope;
+        }
+
+        public void ImportScope(string name)
```
filmor: I'm not sure about this name. I think here the distinction in global/local would help clarifying the semantics: Variables of the "imported" scope should be treated as global.

src/runtime/pyscope.cs
```diff
@@ -0,0 +1,427 @@
...
+        public PyObject ImportModule(string name, string asname)
+        {
+            Check();
+            PyObject module = PythonEngine.ImportModule(name);
+            if (asname == null)
+            {
+                asname = name;
+            }
+            SetVariable(asname, module);
+            return module;
```
filmor: Also here, imported modules are global names.

src/runtime/pyscope.cs
```diff
@@ -0,0 +1,427 @@
...
+            PyObject pyObj = Execute(script);
+            if (pyObj == null)
+            {
+                return default(T);
+            }
+            var obj = (T)ToManagedObject<T>(pyObj);
+            return obj;
+        }
+
+        public T ExecuteVariable<T>(string name)
```
filmor: For what is this needed?

src/runtime/pyscope.cs
```diff
@@ -0,0 +1,427 @@
...
+        /// Evaluate a Python expression and return the result as a PyObject
+        /// or null if an exception is raised.
+        /// </remarks>
+        public PyObject Eval(string code)
+        {
+            Check();
+            var flag = (IntPtr)Runtime.Py_eval_input;
+            IntPtr ptr = Runtime.PyRun_String(
+                code, flag, variables, variables
+            );
```
filmor: There is already code for evaluating string inputs, you shouldn't call `Runtime.` functions here.

src/runtime/pyscope.cs
```diff
@@ -0,0 +1,427 @@
...
+            Check();
+            Exec(code, variables, variables);
+        }
+
+        private void Exec(string code, IntPtr _globals, IntPtr _locals)
+        {
+            var flag = (IntPtr)Runtime.Py_file_input;
+            IntPtr ptr = Runtime.PyRun_String(
+                code, flag, _globals, _locals
+            );
```
filmor: Same as above.

src/runtime/pyscope.cs
```diff
@@ -0,0 +1,427 @@
...
+        /// Add a new variable to the variables dict if it not exists
+        /// or update its value if the variable exists.
+        /// </remarks>
+        public void SetVariable(string name, object value)
+        {
+            Check();
+            using (var pyKey = new PyString(name))
+            {
+                IntPtr _value = Converter.ToPython(value, value?.GetType());
+                int r = Runtime.PyObject_SetItem(variables, pyKey.obj, _value);
```
filmor: Same here, use `PyDict` internally and just straight assignment.

src/runtime/pyscope.cs
```diff
@@ -0,0 +1,427 @@
...
+                    }
+                    if (op == Runtime.PyNone)
+                    {
+                        return null;
+                    }
+                    return new PyObject(op);
+                }
+                else
+                {
+                    throw new PyScopeException(String.Format("'ScopeStorage' object has no attribute '{0}'", name));
```
filmor: This is a weird error message. Also, why is this not implemented in terms of `TryGetVariable`?

src/runtime/pyscope.cs
```diff
@@ -0,0 +1,427 @@
...
+            return true;
+        }
+
+        public override bool TrySetMember(SetMemberBinder binder, object value)
+        {
+            this.SetVariable(binder.Name, value);
+            return true;
+        }
+
+        private object ToManagedObject<T>(PyObject pyObj)
```
filmor: Should return `T` instead of `object` and should actually be part of `PyObject` instead of privately implemented here.

src/runtime/pyscope.cs
```diff
@@ -0,0 +1,427 @@
...
+                return pyObj;
+            }
+            return pyObj.AsManagedObject(typeof(T));
+        }
+
+        private void Check()
+        {
+            if (isDisposed)
+            {
+                throw new PyScopeException("'ScopeStorage' object has been disposed");
```
filmor: Use `nameof` here.

src/runtime/pythonengine.cs
```diff
@@ -538,6 +546,56 @@ public static GILState GIL()
...
         }
+        
+        private static Dictionary<string, PyScope> NamedScopes = new Dictionary<string, PyScope>();
+
+        public static PyScope CreateScope()
+        {
+            return CreateScope(null);
+        }
+
+        public static PyScope CreateScope(string name)
```
filmor: Should have a default parameter of `null`.

src/runtime/pythonengine.cs
```diff
@@ -538,6 +546,56 @@ public static GILState GIL()
...
+        {
+            if (name != null && NamedScopes.ContainsKey(name))
+            {
+                throw new PyScopeException(String.Format("ScopeStorage '{0}' has existed", name));
+            }
+            var scope = new PyScope(name);
+            if (name != null)
```
filmor: What about empty strings?

src/runtime/pythonengine.cs
```diff
@@ -538,6 +546,56 @@ public static GILState GIL()
 
             return new GILState();
         }
+        
+        private static Dictionary<string, PyScope> NamedScopes = new Dictionary<string, PyScope>();
```
filmor: I don't think that this is the right place for the global scope storage. Is there any particular reason for not putting this into the `PyScope` class?

src/runtime/pythonengine.cs
```diff
@@ -538,6 +546,56 @@ public static GILState GIL()
...
+            return NamedScopes.ContainsKey(name);
+        }
+
+        public static PyScope GetScope(string name)
+        {
+            if (name != null && NamedScopes.ContainsKey(name))
+            {
+                return NamedScopes[name]; 
+            }
+            throw new PyScopeException(String.Format("ScopeStorage '{0}' not exist", name));
```
filmor: Use `nameof` and format strings.

yagweb:
@denfromufa @filmor Thank you for your affirmation. 

@filmor responses for your questions,

> 1. Separation of global and local should be like in Python, i.e. imported stuff should be global.

I still think PyScope is a wrapper of Python module in .NET. According to this idea, a new field 'module' to the PyScope is added and the variables dict is the variables dict of that module. Then a new method ImportScope is added, the old 'ImportScope' method is renamed to 'ImportAllFromScope'.

For the topic of separation of global and local, I gave ways about how to use separate locals and globals dictionaries before.

Mantaining two dicts is not conform to principle of 'PyScope is a module' and will make the API complicated. 
Since you insist on directly support for this function from PyScope, instead of adding a locals dict, I add a optional PyDict parameter for all the Exec/Eval/Execute methods. Users can create any number of local dicts as they want.

Please check these changes.

> 2. A dictionary-like interface would be nice.

not quite understand.

> 3. The function names for accessing the variables are a bit too verbose for my taste, Get<T> and Set would be enough. Also, Set could have an optional boolean parameter global to put something into the global dictionary.

The method names and the Exception messages came from IronPython. 

"optional PyDict parameter" was added for Exec/Eval/Execute methods instead of "optional boolean parameter" for SetVariable method.

> 4. There is a lot of stuff implemented directly on top of Runtime., the code should reuse as much as possible already existing abstractions.

The methods you mentioned to reuse are PyDict methoeds and RunString method. Calling the CPython API directly is simple enough and they are more efficiency and flexibly. We can discuss it together.

> 5. The global scope storage doesn't belong into Py but instead into PyScope s.t. this part is self-contained.

I thought it before, we can discuss it together.

src/runtime/pyscope.cs
```diff
@@ -0,0 +1,425 @@
...
+            : base(message)
+        {
+
+        }
+    }
+
+    /// <summary>
+    /// Classes implement this interface must be used with GIL obtained.
+    /// </summary>
+    public interface IPyObject : IDisposable
```
yagweb: Replace this interface with Attritue or just remove it?

src/runtime/pyscope.cs
yagweb: This field now represents the __name__ of python module, check the code.

src/runtime/pyscope.cs
```diff
@@ -0,0 +1,427 @@
...
+            get;
+            private set;
+        }
+
+        private bool isDisposed;
+
+        internal PyScope(string name)
+        {
+            this.Name = name;
+            variables = Runtime.PyDict_New();
```
yagweb: See the responses below.

src/runtime/pyscope.cs
```diff
@@ -0,0 +1,427 @@
...
+            );
+            InitVariables();
+        }
+
+        //To compitable with the python module.
+        //Enable locals() and glocals() usage in the python script.
+        private void InitVariables()
+        {
+            SetVariable("locals", (Func<PyDict>)Variables);
+            SetVariable("globals", (Func<PyDict>)Variables);
```
yagweb: See the responses below.

src/runtime/pyscope.cs
```diff
@@ -0,0 +1,427 @@
...
+        }
+        
+        /// <summary>
+        /// the dict for scope variables
+        /// </summary>
+        internal IntPtr variables { get; private set; }
+
+        public long Refcount
```
yagweb: It's used for debug, now there are 2 IntPtr in the PyScope and the code seems stable, so it was removed.

src/runtime/pyscope.cs
```diff
@@ -0,0 +1,427 @@
...
+        {
+            get
+            {
+                return Runtime.Refcount(variables);
+            }
+        }
+
+        public event Action<PyScope> OnDispose;
```
yagweb: PyScopes are managed by Py class, when it was disposed, Py class can do some cleaning work through registering this event.

In the future, methods in Py class maybe put in an independent ScopeManager class, see the responses below.

src/runtime/pyscope.cs
yagweb: See the responses below.

src/runtime/pyscope.cs
yagweb: I thought it was removed ...  
Now it was removed.

src/runtime/pyscope.cs
yagweb: See the responses below.

src/runtime/pyscope.cs
yagweb: To be compatible with IronPython, See the responses below.

src/runtime/pyscope.cs
yagweb: Same as above.

src/runtime/pythonengine.cs
yagweb: See the responses below.

src/runtime/pythonengine.cs
yagweb: Same as above.

src/runtime/pythonengine.cs
yagweb: Empty string means the scope will not be registered. It should be used and discarded immediately, or it was save in a class field.

Named scope can be shared just like a python module.
Now we have two methods for scope sharing, ImportScope and ImportAllFromScope.

src/runtime/pyscope.cs
yagweb: I cannot agree more.
I thought the name AsManagedObject is too verbose because this method is heavily used, what's more, it can not accept object/dynamic as it parameter (it can be fixed easily).

 I had planed to replace it for a long time. Since you mentioned it now, a new As<T> method was added to the PyObject.

src/embed_tests/TestPyScope.cs
```diff
@@ -0,0 +1,346 @@
...
+                _ps.bb = 100;
+                ps.Exec(
+                    "class class1():\n" +
```
yagweb: Revised.

src/runtime/pythonengine.cs
yagweb: By contrast, I want to split the two methods CreateScope() and CreateScope(string name).

CreateScope() is unique, CreateScope(string name) will be put in the ScopeManager class.

src/runtime/pyscope.cs
filmor: Attribute would make more sense, yes. Could allow us to test this at least at runtime.

denfromufa:
@yagweb your latest changes are failing on both CI:

https://travis-ci.org/pythonnet/pythonnet/jobs/219680218#L1612

```
1) Error : Python.EmbeddingTest.PyScopeTest.TestVariables
Python.Runtime.PyScopeException : 'ScopeStorage' object has no attribute 'locals'
  at Python.Runtime.PyScope.GetVariable (System.String name) [0x0007e] in <610d7dc9d1304e97bc766cb4e052a671>:0 
  at Python.Runtime.PyScope.TryGetMember (System.Dynamic.GetMemberBinder binder, System.Object& result) [0x00008] in <610d7dc9d1304e97bc766cb4e052a671>:0 
  at (wrapper dynamic-method) System.Object:CallSite.Target (System.Runtime.CompilerServices.Closure,System.Runtime.CompilerServices.CallSite,object)
  at System.Dynamic.UpdateDelegates.UpdateAndExecute1[T0,TRet] (System.Runtime.CompilerServices.CallSite site, T0 arg0) [0x00127] in <2392cff65f724abaaed9de072f62bc4a>:0 
  at Python.EmbeddingTest.PyScopeTest.TestVariables () [0x00144] in <6cac7fa3926c454bacf83467587dd86d>:0 
  at (wrapper managed-to-native) System.Reflection.MonoMethod:InternalInvoke (System.Reflection.MonoMethod,object,object[],System.Exception&)
  at System.Reflection.MonoMethod.Invoke (System.Object obj, System.Reflection.BindingFlags invokeAttr, System.Reflection.Binder binder, System.Object[] parameters, System.Globalization.CultureInfo culture) [0x00038] in <f712f98eb8e445c8918edaf595bbe465>:0 
```

yagweb:
@denfromufa It's a bug of the unit test method, I just fixed it.
The methods in Py are moved into a new class PyScopeManager.

denfromufa:
@filmor have all your concerns been addressed by @yagweb? Someone likes PyScopes and is waiting to use this feature:

https://mail.python.org/pipermail/pythondotnet/2017-April/001870.html

JimSEOW:
Good to have feature

yagweb:
@denfromufa One of @filmor's suggestions is,

> The function names for accessing the variables are a bit too verbose for my taste, Get<T> and Set would be enough. 

My response is,

> I agreed the names are too verbose, but these names and the Exception messages all came from IronPython. Just to be compatible.

Personally I also prefer to the shorter names "Get, Set, Remove, Contains".

Compatible with IronPython or more concise, what's your opinion?

JimSEOW:
@yagweb  will the scope class address the GET, SET issues discussed [here](https://github.com/pythonnet/pythonnet/issues/410)

yagweb:
@JimSEOW I put my modifications [here](https://github.com/yagweb/PythonInterface/commit/c3ebc6165b1e151f3a36fe9daef8be7537138c5f). If your have any other questions, please talk it there.

denfromufa:
@yagweb regarding your previous [comment](https://github.com/pythonnet/pythonnet/pull/381#issuecomment-297683716) I would go for more concise, @tonyroberts @vmuriart @filmor do you agree?

JimSEOW:
@yagweb @denfromufa  etc. 
When Xamarin create .NET binding to C++ libraries, they try their best to .dotNETfy the binding using semantics that are characteristics to the .NET ecosystem. 
Perhaps we can do the same here.
=> Get and Set are more the typical Java syntax for properties (not sure how easy it is to replace them with alternatives)
=> I prefer Execute to Exec
=> We are addressing uses from both .NET and python ecosystems. It will be interesting to strike a balance. However, we are talking about Python 4 .NET here. What can we learn from Ironpython to suggest some naming suggestions here?

denfromufa:
@filmor @vmuriart I would like to use this API during my upcoming presentation on June 8th at Microsoft. Please review and comment, if not accepting!

vmuriart:
I'll review over the weekend. I don't think this one introduced any GC bugs but im seeing a bunch of warnings showing up on the Python Tests now (though I suspect its unrelated to this but want to check).

I'll review @filmor comments to make sure requests were resolved.

src/embed_tests/TestPyScope.cs
```diff
@@ -0,0 +1,372 @@
...
+                var result2 = func2().As<int>();
+                Assert.AreEqual(-10, result2);
+                scope.SetVariable("cc", 10); //rollback
+
+                ps.SetVariable("cc", 20);
+                var result3 = func2().As<int>();
+                Assert.AreEqual(10, result3);
+                ps.SetVariable("cc", 10); //rollback
+
+                scope.Dispose();
```
filmor: Same here, wrap in `using`.

src/runtime/pyscope.cs
```diff
@@ -0,0 +1,535 @@
...
+        {
+            if (Runtime.PyObject_Type(ptr) != Runtime.PyModuleType)
+            {
+                throw new PyScopeException("object is not a module");
+            }
+            if(manager == null)
+            {
+                manager = PyScopeManager.Global;
+            }
+            Manager = manager;
```
filmor: `Manager = manager ?? PyScopeManager.Global`

src/runtime/pyscope.cs
```diff
@@ -0,0 +1,535 @@
...
+        /// </summary>
+        internal readonly IntPtr variables;
+
+        private bool isDisposed;
+
+        internal readonly PyScopeManager Manager;
+
+        public event Action<PyScope> OnDispose;
+        
+        public PyScope(IntPtr ptr, PyScopeManager manager)
```
filmor: Things using `IntPtr` should not be `public`, can this be `internal` instead?

src/runtime/pyscope.cs
```diff
@@ -0,0 +1,535 @@
...
+        {
+            if (Runtime.PyObject_Type(ptr) != Runtime.PyModuleType)
+            {
+                throw new PyScopeException("object is not a module");
+            }
+            if(manager == null)
```
filmor: Formatting, just run a `Ctrl-K Ctrl-D` in Visual Studio.

src/runtime/pyscope.cs
```diff
@@ -0,0 +1,535 @@
...
+        {
+            var scope = Manager.Create();
+            scope.ImportAll(this);
+            return scope;
+        }
+
+        public dynamic Import(string name, string asname = null)
+        {
+            Check();
+            if (asname == null)
```
filmor: `string.IsNullOrWhitespace` instead.

src/runtime/pyscope.cs
```diff
@@ -0,0 +1,535 @@
...
+            {
+                asname = name;
+            }
+            var scope = Manager.TryGet(name);
```
filmor: For `Try` the usual "protocol" is to let it return a `bool` and output to an `out` variable:

     PyScope scope;
     if (Manager.TryGet(name, out scope)) ...

src/runtime/pyscope.cs
```diff
@@ -0,0 +1,535 @@
...
+            if(scope != null)
+            {
+                Import(scope, asname);
+                return scope;
+            }
+            PyObject module = PythonEngine.ImportModule(name);
```
filmor: Wrap in `else` branch to make the intention clear.

src/runtime/pyscope.cs
```diff
@@ -0,0 +1,535 @@
...
+        /// <summary>
+        /// Import Method
+        /// </summary>
+        /// <remarks>
+        /// The import .. as .. statement in Python.
+        /// Import a module,add it to the variables dict and return the resulting module object as a PyObject.
+        /// </remarks>
+        public void Import(PyObject module, string asname = null)
+        {
+            if (asname == null)
```
filmor: `string.IsNullOrWhitespace`

src/runtime/pyscope.cs
```diff
@@ -0,0 +1,535 @@
...
+            {
+                asname = module.GetAttr("__name__").ToString();
```
filmor: Use `.As<string>()` instead? `ToString` is much too forgiving here.

src/runtime/pyscope.cs
```diff
@@ -0,0 +1,535 @@
...
+            }
+        }
+
+        public void ImportAll(PyObject module)
+        {
+            if (Runtime.PyObject_Type(module.obj) != Runtime.PyModuleType)
+            {
+                throw new PyScopeException("object is not a module");
+            }
+            var module_dict = Runtime.PyModule_GetDict(module.obj);
```
filmor: This leaks a reference.

src/runtime/pyscope.cs
```diff
@@ -0,0 +1,535 @@
...
+            int result = Runtime.PyDict_Update(variables, module_dict);
+            if (result < 0)
```
filmor: Use `CheckError...` function in all similar places. Or is there a particular reason to do the error check manually here?

src/runtime/pyscope.cs
filmor: We have no compatibility with IronPython here anyhow. `ScopeStorage` doesn't exist, this error is extremely confusing to me. And I still don't get why this does not just call `TryGetVariable`, raising an exception if the result is `false`.

src/runtime/pyscope.cs
```diff
@@ -0,0 +1,535 @@
...
+            using (var pyKey = new PyString(name))
+            {
+                if (Runtime.PyMapping_HasKey(variables, pyKey.obj) != 0)
+                {
+                    IntPtr op = Runtime.PyObject_GetItem(variables, pyKey.obj);
+                    if (op == IntPtr.Zero)
+                    {
+                        throw new PythonException();
+                    }
+                    if (op == Runtime.PyNone)
```
filmor: Leaks a reference. Also, I'm not sure whether this is the right place for this conversion.

src/runtime/pyscope.cs
```diff
@@ -0,0 +1,535 @@
...
+            if (!result)
+            {
+                value = default(T);
+                return false;
+            }
+            if (pyObj == null)
+            {
+                value = default(T);
+                return true;
+            }
```
filmor: This means that if someone does `TryGetVariable<int>("something", out val)` where `something = None`, they will get as a result `val = 0` without any further indication. Seems like a very bad idea for value types.

src/runtime/pyscope.cs
```diff
@@ -0,0 +1,535 @@
...
+        [PyGIL]
+        public PyScope Create(string name)
+        {
+            if (String.IsNullOrEmpty(name))
+            {
+                throw new ArgumentNullException(nameof(name));
+            }
+            if (name != null && NamedScopes.ContainsKey(name))
+            {
+                throw new PyScopeException($"PyScope '{name}' has existed");
```
filmor: `"A scope of name '{name}' does already exist"`

src/runtime/pyscope.cs
```diff
@@ -0,0 +1,535 @@
...
+        }
+
+        [PyGIL]
+        public PyScope Create(string name)
+        {
+            if (String.IsNullOrEmpty(name))
+            {
+                throw new ArgumentNullException(nameof(name));
+            }
+            if (name != null && NamedScopes.ContainsKey(name))
```
filmor: `Contains(name)`

src/runtime/pyscope.cs
```diff
@@ -0,0 +1,535 @@
...
+        {
+            if (name == null)
+            {
+                throw new ArgumentNullException(nameof(name));
+            }
+            if (NamedScopes.ContainsKey(name))
+            {
+                return NamedScopes[name];
+            }
+            throw new PyScopeException($"PyScope '{name}' not exist");
```
filmor: Maybe `"There is no scope named '{name}' registered in this manager"`?

src/runtime/pyscope.cs
```diff
@@ -0,0 +1,535 @@
...
+        }
+
+        public PyScope TryGet(string name)
+        {
+            PyScope value;
+            NamedScopes.TryGetValue(name, out value);
+            return value;
```
filmor: As said above, just forward the semantics of `TryGetValue` directly.

filmor:
Regarding doing things like IronPython: I doubt that we actually want API parity on the .NET side, for that we are too far away. I also don't see much of a benefit, pythonnet is not built on top of the scripting architecture used by IronPython but directly on Python's API, so they can not be replacements for each other anyhow.

This is completely disconnected from the discussion how the *Python side* should look like.

src/runtime/pyscope.cs
yagweb: I just checked, the Runtime.PyModule_GetDict method does not increase the refcount of the module dict.

src/runtime/pyscope.cs
yagweb: This usage comes from the Update method of PyDict.

src/runtime/pyscope.cs
yagweb: Fixed.

src/runtime/pyscope.cs
yagweb: Sorry, I'm not quite understand?

filmor: You already defined a member function `Contains`, use that one.

src/runtime/pyscope.cs
filmor: Sorry, misread the docs :)

yagweb:
@filmor  Very thank you for your detailed and valuable comments.

@denfromufa @filmor @vmuriart The last question is, should the methods 'GetVariable', 'SetVariable' etc. be renamed to shorter ones as 'Get', 'Set' etc.?

denfromufa:
@yagweb i like the shorter names Get & Set.

yagweb:
The names 'Get & Set' etc. make sense, because variables are the only content a scope managed. 
An example of this design is the 'Add' method of the List. 

Methods are renamed and documents are added.

src/runtime/pyscope.cs
```diff
@@ -21,48 +23,45 @@ public interface IPyObject : IDisposable
...
+        internal static PyScope New(string name = null)
         {
-            this.Name = name;
-            variables = Runtime.PyDict_New();
-            if (variables == IntPtr.Zero)
+            if (name == null)
+            {
+                name = "";
+            }
+            var module = Runtime.PyModule_New(name);
```
denfromufa: i'm still pondering what did you notice here so special @filmor? :thinking:

denfromufa:
@vmuriart i'm planning to merge this on Wednesday, let us know if you have any feedback before that. @yagweb great work, thanks for enduring so many iterations of reviews.

PR #384, opened 2017-02-16
stackoverflow/slack shields

denfromufa:


vmuriart:
That is one long shield, and the `label` doesn't work because of the period in `python.net`.
`https://img.shields.io/stackexchange/stackoverflow/t/python.net.svg?label=StackOverflow`

We can use `https://img.shields.io/badge/StackOverflow-python.net-blue.svg` instead though. The original badge kept timing out on me. What do you think?

vmuriart:
Might as well add the `slack` badge too. We can use either `https://img.shields.io/badge/chat-slack-color.svg?style=social?link=http://left&link=http://right`  or make another generic one.

denfromufa:
it works on my machine:

https://github.com/pythonnet/pythonnet/blob/ef6e47d0ee5d5b02e6c50f81d86567cc10c126b8/README.md

vmuriart:
It works on mine about 25% of the time.

vmuriart:
Here's the complete slack badge `https://img.shields.io/badge/chat-slack-color.svg?style=social&link=https://pythonnet.slack.com&link=https://pythonnet.slack.com` You don't need to add an extra link ontop of it.

vmuriart:
Can you match the badge link style that the other badges are using? The image link is located at the bottom of the `README.md`

PR #389, opened 2017-02-17
Add PythonEngine Eval and Exec

yagweb:
"eval" and "exec" are two of the most frequently used methods when interacting with Python. However, I found there is not "eval" provided currently, this statement var flag = (IntPtr)257; in PythonEngine.RunString method makes this method can only be used as "exec".

Two .NET functions equivalent to the Python “evel” and “exec” are implemented by wrapping the PyRun_String function just like the CPython does. PythonEngine.RunString method became internal.

All the modifications located in pythonengine.cs.  All the places using  PythonEngine.RunString are replaced with the new PythonEngine.Exec method.

Unit tests are given in the new file "src/embed_tests/pyrunstring.cs", They can also be used as examples. List one of them below.

Eval the python expression "sys.attr1 + a + 1" and obtain its value.
```
sys.attr1 = 100;
PyDict locals = new PyDict();
locals.SetItem("sys", sys);
locals.SetItem("a", new PyInt(10));

var b = PythonEngine.Eval("sys.attr1 + a + 1", null, locals.Handle)
     .AsManagedObject(typeof(Int32));

Console.WriteLine("sys.attr1 + a + 1 = {0}", b);
```

CHANGELOG.md will be updated later.

### DIFF
--- a/src/embed_tests/pyrunstring.cs
+++ b/src/embed_tests/pyrunstring.cs
@@ -1,0 +1,61 @@
+using System;
+using NUnit.Framework;
+using Python.Runtime;
+
+namespace Python.EmbeddingTest
+{
+    public class RunStringTest
+    {
+        private Py.GILState gil;
+
+        [SetUp]
+        public void SetUp()
+        {
+            gil = Py.GIL();
+        }
+
+        [TearDown]
+        public void Dispose()
+        {
+            gil.Dispose();
+        }
+
+        [Test]
+        public void TestRunSimpleString()
+        {
+            int aa = PythonEngine.RunSimpleString("import sys");
+            Assert.AreEqual(aa, 0);
+
+            int bb = PythonEngine.RunSimpleString("import 1234");
+            Assert.AreEqual(bb, -1);
+        }
+
+        [Test]
+        public void TestEval()
+        {
+            dynamic sys = Py.Import("sys");
+            sys.attr1 = 100;
+            var locals = new PyDict();
+            locals.SetItem("sys", sys);
+            locals.SetItem("a", new PyInt(10));
+
+            object b = PythonEngine.Eval("sys.attr1 + a + 1", null, locals.Handle)
+                .AsManagedObject(typeof(Int32));
+            Assert.AreEqual(b, 111);
+        }
+
+        [Test]
+        public void TestExec()
+        {
+            dynamic sys = Py.Import("sys");
+            sys.attr1 = 100;
+            var locals = new PyDict();
+            locals.SetItem("sys", sys);
+            locals.SetItem("a", new PyInt(10));
+
+            PythonEngine.Exec("c = sys.attr1 + a + 1", null, locals.Handle);
+            object c = locals.GetItem("c").AsManagedObject(typeof(Int32));
+            Assert.AreEqual(c, 111);
+        }
+    }
+}
--- a/src/embed_tests/pyrunstring.cs
+++ b/src/embed_tests/pyrunstring.cs
@@ -24,10 +24,10 @@ namespace Python.EmbeddingTest
         public void TestRunSimpleString()
         {
             int aa = PythonEngine.RunSimpleString("import sys");
-            Assert.AreEqual(aa, 0);
+            Assert.AreEqual(0, aa);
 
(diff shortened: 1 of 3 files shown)


src/runtime/pythonengine.cs
```diff
@@ -406,15 +393,47 @@ public static PyObject ModuleFromString(string name, string code)
...
         /// RunString Method
         /// </summary>
         /// <remarks>
         /// Run a string containing Python code. Returns the result of
         /// executing the code string as a PyObject instance, or null if
         /// an exception was raised.
         /// </remarks>
-        public static PyObject RunString(
-            string code, IntPtr? globals = null, IntPtr? locals = null
+        internal static PyObject RunString(
```
vmuriart: Couldn't this stay as a public method? Not sure how much its being used at the moment, but I wouldn't like to pull it off the API without first deprecating it.

src/runtime/pythonengine.cs
```diff
@@ -406,15 +393,47 @@ public static PyObject ModuleFromString(string name, string code)
...
+        /// Eval Method
+        /// </summary>
+        /// <remarks>
+        /// Evaluate a Python expression and returns the result.
+        /// It's a subset of Python eval function.
+        /// </remarks>
+        public static PyObject Eval(string code, IntPtr? globals = null, IntPtr? locals = null)
+        {
+            PyObject result = RunString(code, globals, locals, RunFlagType.Eval);
+            return result;
```
vmuriart: Should the result be checked for exception before returning?

filmor: Not needed, `RunString` already does that via `Py.Throw`.

src/runtime/pythonengine.cs
```diff
@@ -406,15 +393,47 @@ public static PyObject ModuleFromString(string name, string code)
...
+        /// Exec Method
+        /// </summary>
+        /// <remarks>
+        /// Run a string containing Python code. 
+        /// It's a subset of Python exec function.
+        /// </remarks>
+        public static void Exec(string code, IntPtr? globals = null, IntPtr? locals = null)
+        {
+            PyObject result = RunString(code, globals, locals, RunFlagType.File);
+            if (result.obj != Runtime.PyNone)
```
vmuriart: @filmor then on the flip-side, is this check here then superfluous?

src/runtime/pythonengine.cs
vmuriart: I just re-read your statement on the pull_request, I understand this change now. I think I would still want to deprecate it first before removing it a release or two later.

vmuriart:
Just ran the tests on Travis after #391. PY27 fails with `Fatal Python error: auto-releasing thread-state, but no thread-state for this thread` which is unrelated to these changes. Looks good overall. 

Only changes would be to keep `RunString` public but add a [deprecation warnings instead.](http://stackoverflow.com/a/1759357/5208670).

One of the tests is writing to stderr in the [middle of the tests](https://travis-ci.org/vmuriart/pythonnet/jobs/204358922#L1423) not sure if we can do much on it though.

src/runtime/pythonengine.cs
yagweb: @vmuriart I agree with deprecating this method first. When this method become internal, the type of its last argument _flag can be changed from enum RunFlagType to IntPtr, and the enum RunFlagType can be removed. In the Eval and Exec methods, it can calls this RunString method using Runtime.Py_eval_input and Runtime.Py_file_input for this argument.

src/runtime/pythonengine.cs
yagweb: @vmuriart RunString will always return None when passing the last parameter Runtime.Py_file_input. So this check failed only when there is a runtime bug of CPython. It can be removed.

src/embed_tests/pyrunstring.cs
```diff
@@ -24,10 +24,10 @@ public void Dispose()
         public void TestRunSimpleString()
         {
             int aa = PythonEngine.RunSimpleString("import sys");
-            Assert.AreEqual(aa, 0);
```
vmuriart: @yagweb I think you had these backwards by mistake, its the opposite other than python. I fixed them before merge.

PR #393, opened 2017-02-19
Minimal unit test for Overflow Exception and across AppDomains Exception reproduction

yagweb:
Add a **MINIMAL** unit test function for Overflow Exception and across AppDomains Exception reproduction.

The MINIMAL steps to reproduce these exceptions can be concluded into,

1. During the first time PythonEngine Initialize and Shutdown, import **ANY** class or struct from .NET.
2. After the second time PythonEngine Initialize, import **THE SAME** class/struct from .NET.
3. Create an instance of **THE** .NET class/struct in Python, then the System.OverflowException Exception will be raised here. If replacing int with Int64, OverflowException will be replaced with AppDomain exception.

Usage:

Call this test function in the C# code with,
```
new Python.EmbeddingTest.PyInitializeTest().ReInitialize();
```

### DIFF
--- a/src/embed_tests/pyinitialize.cs
+++ b/src/embed_tests/pyinitialize.cs
@@ -36,5 +36,33 @@ namespace Python.EmbeddingTest
             PythonEngine.Initialize();
             PythonEngine.Shutdown();
         }
+
+        [Test]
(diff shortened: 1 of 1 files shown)


vmuriart:
@yagweb thanks for the contribution again!. I'll review and merge this `pr` in a bit.

PR #395, opened 2017-02-20
improve tests.pyproj for intellisense and running tests

denfromufa:
adds missing python scripts, adds search path (still need to add ..\..\build\..\ somehow)

No.

### Any other comments?
NA

### Checklist

-   [ ] Make sure to include one or more tests for your change - NA
-   [ ] If an enhancement PR, please create docs and at best an example - NA
-   [ ] Updated the [`CHANGELOG`](../blob/master/CHANGELOG.md) - NA

### DIFF
--- a/src/tests/tests.pyproj
+++ b/src/tests/tests.pyproj
@@ -6,7 +6,7 @@
     <ProjectGuid>{250c535c-c060-4f0c-bd80-41f2bf373565}</ProjectGuid>
     <ProjectHome />
     <StartupFile>runtests.py</StartupFile>
-    <SearchPath />
+    <SearchPath>..\..\</SearchPath>
     <WorkingDirectory>.</WorkingDirectory>
     <OutputPath>.</OutputPath>
     <ProjectTypeGuids>{888888a0-9f3d-457c-b088-3a5042f75d52}</ProjectTypeGuids>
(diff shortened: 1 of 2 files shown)


denfromufa:
@vmuriart should it be `from _compat import input`?

src/tests/tests.pyproj
```diff
@@ -6,7 +6,7 @@
     <ProjectGuid>{250c535c-c060-4f0c-bd80-41f2bf373565}</ProjectGuid>
     <ProjectHome />
     <StartupFile>runtests.py</StartupFile>
-    <SearchPath />
+    <SearchPath>..\..\</SearchPath>
```
vmuriart: whats this do?

vmuriart:
@denfromufa are you refering to the `run_tests.py`? 

A better solution is to change 

```python
if __name__ == '__main__':
    main()
    if '--pause' in sys.argv:
        print("Press enter to continue")
        input()
```

to 

```python
if __name__ == '__main__':
    main()
```

I doubt anyone actually uses the `--pause` flag. I would also add a deprecation comment on the file to ask users to simply run `pytest` instead.

src/tests/tests.pyproj
denfromufa: This adds root folder of pythonnet to search paths. Helps when building with --inplace option.

denfromufa:
@vmuriart pause is only helpful when debugging to attach to the python process.

My concern was not with input, but how relative import for _compat module is done.

vmuriart:
give it a try without the period. If i remember right, all other modules will fail without it. This was a while ago though, there might have been other issues that were causing it though

denfromufa:
`from ._compat import input` seems to be fine now.

PR #398, opened 2017-02-22
Bumpversion

### LINKED ISSUE #319: Automate version update
Add `bumpversion` or similar tool

vmuriart:

Automate version update.
Fixes conda-recipe version from lagging (used to look at last git-tag)
Fixes and centralizes AssemblyInfo version information

Closes #319 

### Any other comments?

Based from: https://github.com/peritus/bumpversion/issues/77#issuecomment-130696156

Example usage:

`bumpversion patch`: `0.1.0` -> `0.1.1.dev0`
`bumpversion release`: `0.1.1.dev0` -> `0.1.1`
`bumpversion minor`: `0.1.1` -> `0.2.0.dev0`
`bumpversion dev`: `0.2.0.dev0` -> `0.2.0.dev1`
`bumpversion release`: `0.2.0.dev1` -> `0.2.0`

I chose to not add auto-commit/auto-tag. For one, its broken on PY3 and also would prefer testing out before potentially causing the git-history to get extremely messy.

### Checklist


vmuriart:
conda-builds with correct version.

image

denfromufa:
@vmuriart so why not versioneer which is also internally used by conda-build? bumpversion has stalled for 2 years and it is also used 10 times less on github:

https://github.com/search?l=Python&q=bumpversion&ref=searchresults&type=Code&utf8=%E2%9C%93
https://github.com/search?l=Python&q=versioneer&type=Code&utf8=%E2%9C%93

Is it easier to use or has some features?

vmuriart:
I forgot to add my reasoning, thanks for the reminder. 

`versioneer` requires that the source code be checked-in with the repository and updated periodically for new updates. 

the `version` is calculated at runtime with `import versioneer; __version__ = versioneer.get_version()` which works great on python script/files but not as well on static files or non-python language files (such as our use case). `bumpversion` on the other hand statically updates all the files. In the end this is what tipped the scales. 

`bumpversion` has alot of quirks (which I documented for how annoying they were) but in the end was the simpler solution. I gave `versioneer` multiple tries but in the end I couldn't integrate it for our use.

denfromufa:
can you add instructions to developer docs?

https://github.com/pythonnet/pythonnet/wiki/Development-setup

vmuriart:
Added the documentation to the `wiki`.
Just reviewed the search results from above, I think they are a bit skewed since `versioneer` commits itself into each project that uses it and might show up in the results more because of it.

PR #399, opened 2017-02-22
Fix Py_Main/PySys_SetArgvEx(...) UCS4/PY3 no mem

vmuriart:

Fix Py_Main & PySys_SetArgvEx execution on PY3/UCS4 (Linux).

`N/A`

### Any other comments?

Based on @dmitriyse's work. 
https://github.com/dmitriyse/pythonnet/commit/8a70f09cb8faa9461d13480fb9ee2ee60b5b9acc

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,6 +47,7 @@ This document follows the conventions laid out in [Keep a CHANGELOG][].
 -   Fixed wrong version number in `conda.recipe` (#398)
 -   Fixed fixture location for Python tests and `Embedded_Tests`
 -   Fixed `PythonException` crash during Shutdown (#400)
+-   Fixed `Py_Main` & `PySys_SetArgvEx` no mem error on `UCS4/PY3` (#399)
 
 ### Removed
 
--- a/src/embed_tests/pyinitialize.cs
+++ b/src/embed_tests/pyinitialize.cs
@@ -33,11 +33,6 @@ namespace Python.EmbeddingTest
         [Test]
         public static void LoadSpecificArgs()
         {
-            if (Environment.GetEnvironmentVariable("TRAVIS") == "true" &&
-                Environment.GetEnvironmentVariable("TRAVIS_PYTHON_VERSION") != "2.7")
-            {
-                Assert.Ignore("FIXME: Fails on Travis/PY3+: Fatal Python error: no mem for sys.argv");
-            }
             var args = new[] { "test1", "test2" };
             using (new PythonEngine(args))
             using (var argv = new PyList(Runtime.Runtime.PySys_GetObject("argv")))
--- a/src/runtime/runtime.cs
+++ b/src/runtime/runtime.cs
@@ -667,8 +667,41 @@ namespace Python.Runtime
         public unsafe static extern int
             Py_Main(
                 int argc,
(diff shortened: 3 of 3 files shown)


vmuriart:
I didn't cherry pick this commit in this case. He had re-organized/re-structured his files and was going to be a headache to cherry pick. I gave him credit on the commit message though. @dmitriyse are you ok w this?

filmor:
You can just do a `git commit --amend --author "Something <something>"` to give credit :)

Could you or @dmitriyse explain why this fixes the issue?

vmuriart:
@filmor good call. Fixed and pushed.

vmuriart:
@filmor I'll let @dmitriyse  since he figured it out. My best educated guess is that the encoding was breaking for `UCS4` and thus why we only saw the issue on linux.

denfromufa:
This code definitely needs comments on how these unmanaged strings are manually built.

I'm surprised that it has to be so convoluted. Why can't we just use PyString_* methods?

vmuriart:
@denfromufa I have some work to refactor that logic. We have about 3 different variations of this same thing, each variation being used atleast twice. When I tried refactoring earlier, I was having issues that either the memory was being freed-up too early or never at all.  We do the same thing on `PyString_*` as this cuz of the same memory life issue.

I settled for this `pr` until I get the memory-life tuned just right.

denfromufa:
@vmuriart can you point to "3 different variations of this same thing, each variation being used atleast twice."?

vmuriart:
The last two are a bit of a stretch, but still related. 

https://github.com/pythonnet/pythonnet/blob/master/src/runtime/runtime.cs#L1511
https://github.com/pythonnet/pythonnet/blob/master/src/runtime/runtime.cs#L1664
https://github.com/pythonnet/pythonnet/blob/master/src/runtime/runtime.cs#L1731
https://github.com/pythonnet/pythonnet/blob/master/src/runtime/runtime.cs#L1799
https://github.com/pythonnet/pythonnet/blob/master/src/runtime/converter.cs#L609
https://github.com/pythonnet/pythonnet/blob/master/src/runtime/runtime.cs#L923

vmuriart:
@filmor I took a closer look at the issue and I was on the right track. Currently we don't distinguish between the two different encoding, and fails to work on `UCS4`. This implementation manually marshals the string array and applies to correct encoding passing it to python.

I'm working on an `ICustomMarshaler` to replace all instances that were are doing this, but I would prefer to have time one in the git-history so that we can fall-back to in case the `ICustomMarshaler` causes issues.

that being said, @filmor @denfromufa any objections to merging this then?

dmitriyse:
Hi! There is a problem with ICustomMarshaler - it's not supported by CoreCLR. Please consider to use two Methods solution: First Method - is interop; Second method - is interop wrapper with marshaling inside.

dmitriyse:
Ups, i am worng.
https://github.com/dotnet/corefx/issues/1346
ICustomMarshaler is coming back in NetStandard 2.0
So we can use it. Everybody will migrate to NetStandard 2.0 after this march.

vmuriart:
@dmitriyse your first comment scared me. You posted it just has I was finishing writing the 2nd `ICustomMarshaler`.

denfromufa:
@vmuriart did anyone review this pull request?

PR #400, opened 2017-02-23
Fix PythonException GC -  no thread-state for this thread

vmuriart:

Re-enable most `PyTuple` tests on PY27/Travis.
Prevent GC trying to deref values after python has been closed.

`Fails on Travis/PY27: Fatal Python error: auto-releasing thread-state, but no thread-state for this thread`
Seen mostly on PY27/Travis on PyTuple tests during GC of tests.

```
Fatal Python error: auto-releasing thread-state, but no thread-state for this thread
Stacktrace:
at (wrapper managed-to-native) Python.Runtime.Runtime.PyGILState_Release (intptr)
at Python.Runtime.PythonEngine.ReleaseLock (intptr)
at Python.Runtime.PythonException.Dispose ()
at Python.Runtime.PythonException.Finalize ()
at (wrapper runtime-invoke) object.runtime_invoke_virtual_void__this__ (object,intptr,intptr,intptr)
```

### Any other comments?

Similar fix to #365.
Since `PythonException` doesn't inherit from `PyObject`, need to reapply fix.

### Checklist

### DIFF
--- a/src/embed_tests/pytuple.cs
+++ b/src/embed_tests/pytuple.cs
@@ -7,29 +7,6 @@ namespace Python.EmbeddingTest
     public class PyTupleTest
     {
         /// <summary>
-        /// Tests set-up. Being used to skip class on Travis/PY27
-        /// </summary>
-        /// <remarks>
-        /// FIXME: Fails on Travis/PY27: All tests below (unless otherwise stated)
-        /// Fatal Python error: auto-releasing thread-state, but no thread-state for this thread
-        /// Stacktrace:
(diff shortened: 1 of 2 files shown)


denfromufa:
@vmuriart @tonyroberts @filmor shouldn't `IsFinalizing` be volatile? This is to avoid reading the old value while the new value is being written by `AtExit()`?

denfromufa:
http://stackoverflow.com/a/154803/2230844 regarding volatile

PR #403, opened 2017-02-24
Clean dllImport calls in Runtime

vmuriart:

Simplify calls to `dllImport` by removing arguments that are default, not-applicable, or redundant.

`n/a`

### Any other comments?

The clean-up exposes where CharSet is actually needed to marshal Strings. 

### Checklist

### DIFF
--- a/src/runtime/runtime.cs
+++ b/src/runtime/runtime.cs
@@ -583,297 +583,241 @@ namespace Python.Runtime
 #if Py_DEBUG
         // Py_IncRef and Py_DecRef are taking care of the extra payload
         // in Py_DEBUG builds of Python like _Py_RefTotal
-        [DllImport(Runtime.dll, CallingConvention = CallingConvention.Cdecl,
-            CharSet = CharSet.Ansi)]
+        [DllImport(Runtime.dll, CallingConvention = CallingConvention.Cdecl)]
         private unsafe static extern void
             Py_IncRef(IntPtr ob);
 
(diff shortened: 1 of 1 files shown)


denfromufa:
@vmuriart isn't explicit better than implicit? Can you at least add the docs which show the default calling conventions?

vmuriart:
Each commit message has a reference to the links.

denfromufa:
sorry, i still don't see any links. this SO answer mentions to declare cdecl:  http://stackoverflow.com/a/15664100/2230844

vmuriart:
I just realized you asked for the callingconventions one specifically, looks like i skipped that one sorry.
https://msdn.microsoft.com/en-us/library/system.runtime.interopservices.callingconvention(v=vs.110).aspx

In terms of explicit vs implicit. It's way overengineered, and makes the code less legible. I am all for including them if they are necessary but in most cases they aren't. Fixing the marshal issue became alot simpler after I completed this `pr` because it made alot more sense where we were actually needing to use it.

vmuriart:
`CallingConvention` is the only one I am still debating if we actually need to keep. It seems that for our use case it isn't necessary, but I'm ok removing that part from this `pr` and revisiting it later.

vmuriart:
http://stackoverflow.com/a/6511236/5208670 suggest that we are ok without it. None of the calls python API that I've seen have `vargs`, and those that may we can add the calling convention to them (and would distinguish them from the rest)

denfromufa:
BTW, default calling convention for pinvoke is stdcall, but cdecl is more compatible across various compilers.

PR #405, opened 2017-02-24
Fix the issue of sys.argv being cleared on import clr.

### LINKED ISSUE #404: Compatibility issue between argparse and pythonnet packages
### Environment

-   Pythonnet version: 2.3.0.dev1 (built from master on 23/02/2017)
-   Python version: 2.7.13 64bit
-   Operating System: Windows 7 64bit

### Details

-   I tested the pythonnet version from yesterday, 23/02/2017, as we are interested in the bug fix for the method overload issue #377. I noticed that we are now running into compatibility issues with the package argparse.

The package argparse now only works correctly, if the parsing of the command line arguments is done BEFORE the import statement of pythonnet (import clr).

If I revert my pythonnet version to pythonnet==2.2.2, then the problem disappears.

The following minimal test code illustrates the problem:
```
file: test.py
import argparse
# import clr # can't put it here, args.arg1 will be None

parser = argparse.ArgumentParser()
parser.add_argument('--arg1', '-a', help='argument 1', dest='arg1')
args = parser.parse_args()

import clr   
print('Argument 1 = ')
print(args.arg1)
```
Then I start the script from command line as
> python test.py --arg1 "Hello World"
```
If I run the code with the "import clr" statement after the call to parser.parse_args() then I get the desired output: 
Argument 1 =
hello world

However, if I put the "import clr" at the beginning of the file, I get the output:
Argument 1 =
None

filmor:
Closes #404.

Introduces a `setSysArgv` flag to all `Initialize` functions in `PythonEngine`, defaulting to `true` and set to `false` for `InitExt` such that `import clr` doesn't call `PySys_SetArgvEx`.

#404.

### Any other comments?

The test should already be in from another PR, from what I understood and since the change introducing this issue has not been released, yet, I don't think it need to go into the changelog.

### DIFF
--- a/src/runtime/pythonengine.cs
+++ b/src/runtime/pythonengine.cs
@@ -122,7 +122,12 @@ namespace Python.Runtime
 
         public static void Initialize()
         {
-            Initialize(Enumerable.Empty<string>());
+            Initialize(setSysArgv: true);
+        }
+
+        public static void Initialize(bool setSysArgv = true)
+        {
+            Initialize(Enumerable.Empty<string>(), setSysArgv: setSysArgv);
         }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


vmuriart:
@filmor just fixed the test for this on `master` if you want to check against it. 
I ran the test against `v2.2.2` and our current `master` to make sure it passed/xfailed respectively.

PR #407, opened 2017-02-26
Refactor Managed-Native conversion - ICustomMarshaler

vmuriart:

Refactors conversions from Managed to Native type using `ICustomMarshaler`'s to handle the various conversions scenarios and manage releasing unmanaged memory.

#408 

### Any other comments?

This doesn't need to implement Native to Managed conversions as it's not needed for the purposes of the `pr`.

### DIFF
--- a/src/runtime/CustomMarshaler.cs
+++ b/src/runtime/CustomMarshaler.cs
@@ -123,4 +123,49 @@ namespace Python.Runtime
             return Instance;
         }
     }
+
+
+    /// <summary>
+    /// Custom Marshaler to deal with Managed String to Native
+    /// conversion on UTF-8. Use on functions that expect UTF-8 encoded
+    /// strings like `PyUnicode_FromStringAndSize`
+    /// </summary>
+    /// <remarks>
+    /// If instead we used `MarshalAs(UnmanagedType.LPWStr)` the output to
+    /// `foo` would be `f\x00o\x00o\x00`.
+    /// </remarks>
+    public class Utf8Marshaler : MarshalerBase
+    {
+        private static readonly MarshalerBase Instance = new Utf8Marshaler();
+        private static readonly Encoding PyEncoding = Encoding.UTF8;
+
+        public override IntPtr MarshalManagedToNative(object managedObj)
+        {
+            var s = managedObj as string;
+
+            if (s == null)
+            {
+                return IntPtr.Zero;
+            }
+
+            byte[] bStr = PyEncoding.GetBytes(s + "\0");
+            IntPtr mem = Marshal.AllocHGlobal(bStr.Length);
+            try
+            {
+                Marshal.Copy(bStr, 0, mem, bStr.Length);
+            }
+            catch (Exception)
+            {
+                Marshal.FreeHGlobal(mem);
+                throw;
+            }
+
+            return mem;
+        }
+
+        public static ICustomMarshaler GetInstance(string cookie)
+        {
+            return Instance;
+        }
+    }
 }
--- a/src/runtime/CustomMarshaler.cs
+++ b/src/runtime/CustomMarshaler.cs
@@ -42,7 +42,7 @@ namespace Python.Runtime
     public class StrMarshaler : MarshalerBase
     {
         private static readonly MarshalerBase Instance = new StrMarshaler();
-        private static readonly Encoding PyEncoding = Runtime.UCS == 2 ? Encoding.Unicode : Encoding.UTF32;
+        private static readonly Encoding PyEncoding = Runtime.PyEncoding;
 
         public override IntPtr MarshalManagedToNative(object managedObj)
         {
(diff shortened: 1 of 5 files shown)


vmuriart:
@dmitriyse in case you wanted to review.

vmuriart:
Expanded it to include #408 as well since its related.

src/runtime/CustomMarshaler.cs
```diff
@@ -0,0 +1,168 @@
...
+            }
+
+            byte[] bStr = PyEncoding.GetBytes(s + "\0");
+            IntPtr mem = Marshal.AllocHGlobal(bStr.Length);
+            try
+            {
+                Marshal.Copy(bStr, 0, mem, bStr.Length);
+            }
+            catch (Exception)
+            {
```
denfromufa: no exception throwing here?

src/runtime/CustomMarshaler.cs
```diff
@@ -0,0 +1,168 @@
...
+                Marshal.FreeHGlobal(mem);
+            }
+
+            return mem;
+        }
+
+        public static ICustomMarshaler GetInstance(string cookie)
```
denfromufa: What is the purpose of `string cookie`?

src/runtime/CustomMarshaler.cs
vmuriart: what you mean? like re-throwing the `Exception`? That part was meant to de-allocate the memory in case something goes wrong, otherwise keep it.

src/runtime/CustomMarshaler.cs
vmuriart: Its part of the implementation of the interface. On the link below scroll down to `Implementing the GetInstance Method`. For our use case we don't need the cookie though but left it since its part of the implementation. 

https://msdn.microsoft.com/en-us/library/system.runtime.interopservices.icustommarshaler(v=vs.110).aspx#Remarks

vmuriart: Our old `monosupport` actually was an `ICustomMarshaler` too but it depended on `mono` libraries to work. It also had the same signature, except it was [called `s`](https://github.com/pythonnet/pythonnet/blob/v2.2.2/src/runtime/monosupport.cs#L16-L19)

src/runtime/CustomMarshaler.cs
denfromufa: isn't failing on Marshal.Copy need to be thrown as Exception? If not, how is this handled later?

src/runtime/CustomMarshaler.cs
denfromufa: oh, i see on Mono: // The Utf32Marshaler was written Jonathan Pryor and has been placed in the PUBLIC DOMAIN.

denfromufa: @vmuriart so is custom marshalling dependent on COM interop now?

> This static method is called by the common language runtime's COM interop layer to instantiate an instance of the custom marshaler.

vmuriart: No. I think that line is in reference that the `COM Interop` layer can call the instance. Its probably related to the whole `cookie` thing in which the instance returned depends on which `cookie` is passed.

src/runtime/CustomMarshaler.cs
vmuriart: I'll `throw` the exception again.

vmuriart: Added the throw in the latest commit

src/runtime/CustomMarshaler.cs
denfromufa: Why did previous Mono string not null-terminating like you added here?

vmuriart: because it didn't do the `marshal` at all, it passed it `mono` to do it. Mono itself then did it [here](https://github.com/mono/mono/blob/master/mcs/class/Mono.Posix/Mono.Unix/UnixMarshal.cs#L330)

denfromufa:
I'm fine with this pull request but it is a little above my knowledge.

denfromufa:
@vmuriart somehow I completely missed the whole point of this PR. Do you know why I did not see your PR description while reviewing on github?

> Add ICustomMarshaler Utf8Marshaler
> Refactor PyString_FromStringAndSize
> Link explains why `MarshalAs(UnmanagedType.LPWStr)` or `CharSet.Unicode` don't work
> 
> http://stackoverflow.com/a/25128147/5208670

vmuriart:
i dont understand you question

denfromufa:
I missed the headers of commits in this PR, e.g.:

https://github.com/vmuriart/pythonnet/commit/e487076d90ac86f15150c1b41113565364fb7850

I'm not sure if there is any better way to expose these headers during PR?

vmuriart:
Did you try the commit tab?

image

denfromufa:
Ok, now I see. So by default the descriptions are collapsed. Thank you for the suggestion.

PR #279, opened 2016-10-30
[WIP] PythonHome Path is correctly set

denfromufa:
*Blank Description*

------------------------

Pending for merge:

- [ ] Generalize/Improve Unittests (has to be able to run in Appveyor, Local (travis could wait since embdded tests arent working there)

- [ ] Rebase and fix merge conflicts. (@vmuriart)

PR #415, opened 2017-03-03
Fix set pythonhome

### LINKED ISSUE #179: Incorrect marshalling of string in Py_SetPythonHome

vmuriart:

Fixes setters for `PythonEngine.PythonHome` `PythonEngine.ProgramName` and `PythonEngine.PythonPath`. 

It extends #279/#186: 
- Fix the issue in Python 2/3 and 32/64bit
- Fix memory leak on multiple `setters`

Closes #179
Closes #279 (replaces)
Closes #186 (replaces)

### Any other comments?

There's quite a bit of repeated code on this `pr`. I'll perform the clean-up after this `pr` is merged to avoid obscuring how this is being fixed. 

`PythonEngine.PythonPath`, is misleading since it doesn't directly map to `PYTHONPATH` and should be called `Path` instead. `py_set_path` doesn't exist on `PY27` but it was erroneously defined at some point. 
Needs to be removed, but its outside the scope of this `PR`. 

### Checklist

### DIFF
--- a/src/runtime/pythonengine.cs
+++ b/src/runtime/pythonengine.cs
@@ -15,6 +15,8 @@ namespace Python.Runtime
         private static DelegateManager delegateManager;
         private static bool initialized;
         private static IntPtr _pythonHome = IntPtr.Zero;
+        private static IntPtr _programName = IntPtr.Zero;
+        private static IntPtr _pythonPath = IntPtr.Zero;
 
         public PythonEngine()
         {
@@ -65,7 +67,14 @@ namespace Python.Runtime
 
                 return result ?? "";
             }
-            set { Runtime.Py_SetProgramName(value); }
+            set
+            {
+                Marshal.FreeHGlobal(_programName);
+                _programName = Runtime.IsPython3
+                    ? UcsMarshaler.GetInstance("").MarshalManagedToNative(value)
+                    : Marshal.StringToHGlobalAnsi(value);
+                Runtime.Py_SetProgramName(_programName);
+            }
         }
 
         public static string PythonHome
@@ -81,10 +90,7 @@ namespace Python.Runtime
             }
             set
             {
-                if (_pythonHome != IntPtr.Zero)
-                {
-                    Marshal.FreeHGlobal(_pythonHome);
-                }
+                Marshal.FreeHGlobal(_pythonHome);
                 _pythonHome = Runtime.IsPython3
                     ? UcsMarshaler.GetInstance("").MarshalManagedToNative(value)
                     : Marshal.StringToHGlobalAnsi(value);
@@ -103,7 +109,14 @@ namespace Python.Runtime
 
                 return result ?? "";
             }
-            set { Runtime.Py_SetPath(value); }
+            set
+            {
+                Marshal.FreeHGlobal(_pythonPath);
+                _pythonPath = Runtime.IsPython3
+                    ? UcsMarshaler.GetInstance("").MarshalManagedToNative(value)
+                    : Marshal.StringToHGlobalAnsi(value);
+                Runtime.Py_SetPath(_pythonPath);
+            }
         }
 
         public static string Version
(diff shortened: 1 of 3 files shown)


vmuriart:
miniconda failures seem to be due to some update to the pre-installed ones. Not related to this pr.

denfromufa:
@vmuriart I don't see miniconda failures?

src/runtime/pythonengine.cs
```diff
@@ -64,7 +67,17 @@ public static string ProgramName
 
                 return result ?? "";
             }
-            set { Runtime.Py_SetProgramName(value); }
+            set
+            {
+                if (_programName != IntPtr.Zero)
```
denfromufa: the check for IntPtr.Zero is not necessary

vmuriart:
Already fixed the miniconda fails.

src/runtime/pythonengine.cs
```diff
@@ -64,7 +67,17 @@ public static string ProgramName
...
+                {
+                    Marshal.FreeHGlobal(_programName);
+                }
+                _programName = Runtime.IsPython3
+                    ? UcsMarshaler.GetInstance("").MarshalManagedToNative(value)
```
denfromufa: 2 custom marshalling articles suggest that GetInstance should be a singleton like this:

```
public static ICustomMarshaler GetInstance(string cookie)
   {
       if (marshaler == null)
       {
              marshaler = new UcsMarshaler();
       }
       return marshaler;
   }
```

src/runtime/pythonengine.cs
vmuriart: What would you replace it with? Without it, we'd end up trying to `dealloc` null

src/runtime/pythonengine.cs
denfromufa: and this SO Q/A confirms: http://stackoverflow.com/q/18498452/2230844

vmuriart: it is already. https://github.com/pythonnet/pythonnet/blob/0a478024175e52106d9d4ff2230e1f505e56d7d8/src/runtime/CustomMarshaler.cs#L166

denfromufa: oh, sorry, missed that!

vmuriart: No worries. I almost went w that implementation

src/runtime/pythonengine.cs
denfromufa: according to docs this is fine: https://msdn.microsoft.com/en-us/library/system.runtime.interopservices.marshal.freehglobal(v=vs.110).aspx#Anchor_1

vmuriart: ok makes sense. My original implementation `dealloc` twice. A better check would have been to check against `null`. Ill fix it up

denfromufa:
@vmuriart this pull request adds @leomeuk as contributor, so can you add him? @leomeuk do you agree to transition from Zope to MIT? Your original pull request was made under the old Zope license.

vmuriart:
I forgot about that. @leomeuk hasn't been active since he opened the original `pr`. Note that all his contributions survived in the end.

Issue #155, opened 2016-02-17
Trying to Integrate Python.Net in ASP.Net App Fails Error: "Attempted to access an unloaded AppDomain"


gatapia:
Hi All,

Has anyone had any success in running some python code from an ASP.Net (Web.Api) server?  I have a python module I load on application start and I have a controller that periodically communicates with this module.

I load the python module with the code taken from PyImportTest.cs file.  And I periodically call this loaded module from Controllers using module.InvokeMethod(...).AdManagedObject(...).

This all works great in Tests but under webserver conditions (threading I am assuming) this fails with a: "Attempted to access an unloaded AppDomain" error.

Any pointers?  Is this a threading issue? Can I get back onto the thread that kicked off the python module somehow before calling InvokeMethod?

Or is this a bug with PythonNet?

denfromufa:
@gatapia can you provide minimal reproducible example for both conditions: tests and webserver?

denfromufa:
@gatapia possibly related https://github.com/pythonnet/pythonnet/issues/109 may need to use `PythonEngine.BeginAllowThreads()`

gatapia:
Tried BeginAllowThreads/EndAllowThreads and did not help.  Current Symptoms:
- Appears to work sometimes?? (Perhaps after killing w3wp process)
- Second time (sometimes first also) we get AppDomain unloaded exceptions
- Other times it appears to halt (perhaps at AcquireLock)
- Once its locked once or AppDomain error occurs I have to manually kill w3wp process.
- Using log4net is not much use as once an error happens the logger stops logging so hard to get a good feel for what is happening.

Here is my code:

**Initialisation (Application Start)**
```
private bool InitEngine() {
      if (PythonEngine.IsInitialized) return true;
      PythonEngine.Initialize();                  
      pythread = PythonEngine.BeginAllowThreads();
      pylock = PythonEngine.AcquireLock();      

     Type rt = typeof(Runtime);      
      var PyString_FromString = rt.GetMethod("PyString_FromString", BindingFlags.NonPublic | BindingFlags.Static);
      var pyStrPtr = (IntPtr) PyString_FromString.Invoke(null, new object[] { dir.Replace('\\', '/') });
      var PySys_GetObject = rt.GetMethod("PySys_GetObject", BindingFlags.NonPublic | BindingFlags.Static);
      var SysDotPath = (IntPtr) PySys_GetObject.Invoke(null, new object[] {"path"});
[...]

Issue #109, opened 2015-10-23
using(Py.Gil()) hangs while calling from other than init thread


zwadar:
Hi,
I'm using python.net in C# app and mainly calling python from worker threads. However it always hangs when trying to acquirelock from different thread than thread that did first initialization.
I'm not familiar with exact workflow with GIL, but for me, adding BeginAllowThreads() and EndAllowThreads() to GILState class helped, but I'm not sure it is proper solution.
```C#
 public class GILState : IDisposable
        {
            private IntPtr state;
            private IntPtr thread_state;
            internal GILState()
            {
                thread_state = PythonEngine.BeginAllowThreads();
                state = PythonEngine.AcquireLock();
            }
            public void Dispose()
            {
                PythonEngine.ReleaseLock(state);
                GC.SuppressFinalize(this);
                PythonEngine.EndAllowThreads(thread_state);
            }
            ~GILState()
            {
                Dispose();
            }
        }
```

denfromufa:
Someone tried this before:
https://mail.python.org/pipermail/pythondotnet/2010-December/001058.html

tonyroberts:
The main thread will hold the GIL after initialization until you explicitly release it by calling PythonEngine.BeginAllowThreads() from the main thread (not from your background thread). This is how python threading works, it's not specific to pythonnet.

jakrivan:
I'd vote for reopening this issue.
The .net wrapper can handle threading gracefully while ensuring the underlying threading model of CPython. This can be nasured e.g. by holding internal register of all threads attempting to enter python (via PythonEngine.AcquireLock) engine and calling the ensure/release for them. Example solution in C++ can be found here: https://stackoverflow.com/questions/29100082/python-gil-and-threads

I'll try to find some time to create pull request.

jakrivan:
Btw. untill this is fixed, there probably should not be finalizer on PyObject. It's much better to have some memory leacks; rather than having entire finalizer thread locked up - as finalizer thread is dfefinitely guaranteed to run on different thread than the thread acquiring the engine

PR #419, opened 2017-03-06
Enable multi-thread usage in embedded python by default

yagweb:
When embedding CPython in .NET, the thread which initializes CPython will hold the GIL after initialization. This blocks other threads (both CLR Thread and Python threading) acquire the GIL until PythonEngine.BeginAllowThreads() is called in this thread. 

Since let multiple threads visit CPython is a basic need when embedding CPython, for example, the GC thread need to visit CPython at [Finalize in classderived.cs](https://github.com/pythonnet/pythonnet/blob/master/src/runtime/classderived.cs#L887-L890) pointed out by @vmuriart,

It's better to release the GIL after the CPython initialized. Then all the CLR threads (including the thread which initializes CPython) and Python threads can compete the GIL on equal ground.

Here is a related [closed issue](https://github.com/pythonnet/pythonnet/issues/109), and maybe [this one](https://github.com/pythonnet/pythonnet/issues/155).

There are three public Initialize methods, refactoring is needed to **seperate 'the Initialization from CPython' from 'the Initialization from CLR'**, for example by adding a new private bool field. Other work may benefit from this seperation, such as the sys.stdout redirection in the example below.

The BeginAllowThreads and EndAllowThreads can be changed into internal methods in the future.

Below is an example, the main thread, all the CLR threads and all the python threads can run concurrently. 
```
public static void ThreadTest()
{
    PythonEngine.Initialize();
    PythonEngine.Shutdown(); //shut down test
    PythonEngine.Initialize();
    RedirectionFinal.RedirectToConsole(); //Redirect the sys.stdout to Console.Out
    
    //Python thread
    PyDict dict;
    dynamic sys;
    using (Py.GIL())
    {
        sys = PythonEngine.ImportModule("sys");
        sys.is_run = true;
        dict = new PyDict();
        PythonEngine.Exec(
            "def target(thid):\n" +
            "    import sys\n" +
            "    from time import sleep\n" +
            "    round = 0\n" +
            "    while(sys.is_run):\n" +
            "        print('hello from python thread %s, round %d' % (thid, round))\n" +
            "        round += 1\n" +
            "        sleep(0.5)\n",
            null, dict.Handle
        );
        // test the function first, because the exception in python thread cannot be catched by CLR
        //dynamic target = dict.GetItem("target");
        //target(1);
    }
    using (Py.GIL())
    {
        for (int i = 0; i < 2; i++)
        {
            dict.SetItem("thid", new PyInt(i));
            PythonEngine.Exec(
                "from threading import Thread\n" +
                "th = Thread(target = target, args = (thid, ))\n" +
                "th.start()\n",
                null, dict.Handle
            );
        }
    }

    //CLR worker thread
    bool is_run = true;
    for (int i = 0; i < 2; i++)
    {
        Thread th = new Thread(() =>
        {
            int round = 0;
            while (is_run)
            {
                using (Py.GIL())
                {
                    string msg = String.Format("print('hello from CLR thread {0}, round {1}')", i, round);
                    PythonEngine.RunSimpleString(msg);
                    round += 1;
                    Thread.Sleep(500);
                }
            }
        });
        th.Start();
    }

    //CLR main thread
    {
        int round = 0;
        for (int i = 0; i < 3; i++)
        {
            using (Py.GIL())
            {
                string msg = String.Format("print('hello from main thread, round {0}')", round);
                PythonEngine.RunSimpleString(msg);
                round += 1;
                Thread.Sleep(500);
            }
        }
    }
    Console.Read();
    is_run = false;
    using (Py.GIL())
    {
        sys.is_run = false;
    }
}
```


denfromufa:
Inviting multi-threading expert @ArvidJB to this thread in case he has any experience with this.

@yagweb will your changes to multihtreaded execution work with demo GUI apps?

ArvidJB:
Unfortunately I only have experience in calling .NET _from_ Python, since that's the only thing we do.

That being said, this change seems to go in the right direction: no .NET thread should hold the GIL until it wants to actually call into Python and acquiring it explicitly or implicitly before any call seems correct to me.

The big question is how much existing code would be broken by this change?

yagweb:
Python threads was added to the example above. Now the main thread, all the CLR threads and all the python threads can run concurrently directly.

@denfromufa what does the demo GUI app look like?

Effection to the embedding users who don't use multi-thread,

The embedding doc asks all the calls to python should be wrapped with AcquireLock and ReleaseLock methods, including the calls from the thread who initialize the CPython. If users strictly follow the doc, nothing needs to change. Some codes forgot to wrap with Py.GIL(){...} may run in the past, now it's not working.

Effection to the embedding users who use multi-thread,

Besides the effection above, the use of BeginAllowThreads and EndAllowThreads should be removed. Because multi-thread is enabled by default.
In the future, these two methods should be changed into internal methods.

denfromufa:
@vmuriart do you know why the embedding tests break off without finishing on Appveyor - no errors. On Travis CI it is showing "1) Error : 
Tcp transport error."

vmuriart:
Its most likely the same error as Travis. AppVeyor does some weird things w the console that cause issues when the underlying program freezes like its happening on this case.

denfromufa:
@yagweb here are the GUI demos that I mentioned:

https://github.com/pythonnet/pythonnet/tree/master/demo

We can also look at pywebview demos using pythonnet.

filmor:
@yagweb This is apparently crashing quite a bit, are you still working on this branch? If not, I'd close it for now and you (or someone else) can reopen it when work is picked up again.

denfromufa:
no activity, closed for now

PR #427, opened 2017-03-11
Fix numpy array and README example

### LINKED ISSUE #249: Errors with example calling Python from .NET
When trying to run the numpy example from the Readme file I see the following error:
TypeError : long() argument must be a string or a number, not '0, Culture=neutral, PublicKeyToken=b77a5c561934e089]]'

This seems to be cause by the Py.kw function call. However, the numpy.array does not seem to be working properly.  The console output looks like this:

image

This is from using the latest master branch and Python 2.7.  Has the array handling changed recently?

vmuriart:

Generic Lists were falling through and being classified as `typecode.Object`
To solve this, adding a specific processing branch for `Generic Lists` only
to avoid breaking the changes from 471673a1fa4691b3becb4d9fee6941308535bb04

Closes #249

### Any other comments?

`n/a`

### Checklist

### DIFF
--- a/src/runtime/converter.cs
+++ b/src/runtime/converter.cs
@@ -1,5 +1,6 @@
 using System;
 using System.Collections;
+using System.Collections.Generic;
 using System.Globalization;
 using System.Reflection;
 using System.Runtime.InteropServices;
@@ -131,6 +132,16 @@ namespace Python.Runtime
                 result = Runtime.PyNone;
                 Runtime.XIncref(result);
                 return result;
+            }
+
+            if (value is IList && value.GetType().IsGenericType)
+            {
+                var resultlist = new PyList();
+                foreach (object o in (IEnumerable)value)
+                {
+                    resultlist.Append(new PyObject(ToPython(o, o.GetType())));
+                }
+                return resultlist.Handle;
             }
 
             // it the type is a python subclass of a managed type then return the
--- a/src/runtime/converter.cs
+++ b/src/runtime/converter.cs
@@ -136,12 +136,18 @@ namespace Python.Runtime
 
             if (value is IList && value.GetType().IsGenericType)
             {
-                var resultlist = new PyList();
-                foreach (object o in (IEnumerable)value)
+                using (var resultlist = new PyList())
                 {
-                    resultlist.Append(new PyObject(ToPython(o, o.GetType())));
+                    foreach (object o in (IEnumerable)value)
+                    {
+                        using (var p = new PyObject(ToPython(o, o?.GetType())))
+                        {
+                            resultlist.Append(p);
+                        }
+                    }
+                    Runtime.XIncref(resultlist.Handle);
+                    return resultlist.Handle;
                 }
-                return resultlist.Handle;
             }
 
             // it the type is a python subclass of a managed type then return the
--- a/README.md
+++ b/README.md
@@ -63,7 +63,7 @@ static void Main(string[] args)
 
         double c = np.cos(5) + sin(5);
         Console.WriteLine(c);
-        /* this block is temporarily disabled due to regression #249
+
         dynamic a = np.array(new List<float> { 1, 2, 3 });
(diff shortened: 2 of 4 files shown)


src/runtime/converter.cs
```diff
@@ -133,6 +134,16 @@ internal static IntPtr ToPython(object value, Type type)
                 return result;
             }
 
+            if (value is IList && value.GetType().IsGenericType)
+            {
+                var resultlist = new PyList();
+                foreach (object o in (IEnumerable)value)
+                {
```
denfromufa: @vmuriart can you make this consistent with this syntax in the same file? https://github.com/pythonnet/pythonnet/blob/423407a4e7ff783d8f4d9498eaa51bf8a12382ce/src/runtime/converter.cs#L214

vmuriart: Done.

src/runtime/converter.cs
```diff
@@ -133,6 +134,22 @@ internal static IntPtr ToPython(object value, Type type)
                 return result;
             }
 
+            if (value is IList && value.GetType().IsGenericType)
+            {
```
denfromufa: To be completely consistent with the converter code the IEnumerable should be replaced with IList like you did this pull request.

The commit message mentioned that strings should not be converted to `PyList`. 

https://github.com/pythonnet/pythonnet/commit/471673a1fa4691b3becb4d9fee6941308535bb04

But `IEnumerable` is not quite right when converting to `PyList` anyway. `IList` is the closest interface like you selected. 
Other example which are IEnumerable, but not IList besides strings are HashSet, maybe even HashTable.

src/runtime/converter.cs
```diff
@@ -133,6 +134,22 @@ internal static IntPtr ToPython(object value, Type type)
...
+                using (var resultlist = new PyList())
+                {
+                    foreach (object o in (IEnumerable)value)
```
vmuriart: @denfromufa are you talking about the `IEnumberable` here?

denfromufa: No, I meant this place:

https://github.com/pythonnet/pythonnet/blob/423407a4e7ff783d8f4d9498eaa51bf8a12382ce/src/runtime/converter.cs#L208

vmuriart: I left that alone on purpose as it wasn't necessary to modify for what this `pr` is trying to do. 

The correct thing there is probably to delete that section of code altogether though (or delete `typecode.Object`). The remaining typecodes are `Empty`, `DBNull`, `Decimal` and `DateTime` and i don't think any of those should be converted to `PyList`. 
Also, I can't think of a situation in which the `default` branch of the switch would be hit because we have `Typecode.Object` defined; It would always be called instead of the `default`. This [example](http://stackoverflow.com/a/7542810/5208670) is probably closer to how it should be written. 
Either way though, all those changes aren't needed for the purposes of fixing the `numpy arrays` which is why I didn't include them.

denfromufa: Agreed, I opened a separate issue for this dead code: https://github.com/pythonnet/pythonnet/issues/428

vmuriart:
ok to merge then?

denfromufa:
I'm not in the reviewers' list :) Can you mention this pull request on the original issue, if anyone wants to look at this?

vmuriart:
not sure why you were removed. I added you originally, maybe it was removed after you submitted your review and I committed the fixes?

The original issue has a reference to this `pr` as well. The mention of #249 on the first comment created a link back to this one. Same w the commit

denfromufa:
Yes, I think the review process can somehow remove the original reviewers. I can see 3 reviewers on the top.

The links you created in this pull request do not notify anyone on the original issue, because it is not a message in the issue tracker.

vmuriart:
oh, i see what you're trying to do. you want them to review this `pr` if they have input on it right? Otherwise they should get the notification once the issue closes when this merges.

Issue #432, opened 2017-03-21
Sypport typeof(T) or clr.GetClrType


vivainio:
In order to do reflection, you need to acquire the Type instance for imported class.

In C#, this is done by

```
Type t = typeof(MyClass)
```

In IronPython, this is done by

```
t = clr.GetClrType(MyClass)
```

Python.net seems to be missing this functionality.

I currently have this horrible hack, and I'm unsure how far it will work:

```python
def get_clr_type(typ):
    """ simulate typeof(T) or GetClrType from IronPython.

    Horrible hack, redo
    """
    name = typ.__module__ + "." + typ.__name__
    return Type.GetType(name)
```

For reference, code using this resides in: https://github.com/vivainio/pynetext/blob/master/pynetext.py

denfromufa:
@vivainio pull requests are welcome!

Your approach is in the right direction:

https://mail.python.org/pipermail/pythondotnet/2005-July/000371.html

vivainio:
@denfromufa Ok, I'll take a look if I can figure it out how to build & work with the codebase :)

denfromufa:
@vivainio pythonnet runs on 2 CI servers: Travis and Appveyor. So the pull request will automatically do this and also run the tests.

You can also look at these docs to see how to build locally:
https://github.com/pythonnet/pythonnet/wiki/Troubleshooting-on-Windows-and-Linux

Please report any issues back about building locally, so that we can improve the wiki.

denfromufa:
@vivainio thanks for this contribution!

KEBO80:
As an anlternative solution, I came up with the following by creating a static C# helper function:

   public static class ReflectionHelper
    {
        public static Type GetType<T>() => typeof(T);
    }

This I can import this in Python code and use it for any NET type in the folllowing way:

int32type = ReflectionHelper.GetType[System.Int32]()

intt32type is of System.RuntimeType

JoaBjo:
Could you elaborate a little bit on your alternative solution @KEBO80? That class doesn't compile since T is not defined.

filmor:
@JoaBjo Did you try to compile it? :)

This code is valid, the function `GetType<T>()` is a generic function using one type parameter `T` and no function parameters. In C# you would call it as `var int32Type = ReflectionHelper.GetType<System.Int32>()`, KEBO80 gave the corresponding Python syntax (square brackets `[]` instead of `<>` to select the type parameter).

JoaBjo:
[...]

PR #433, opened 2017-03-22
Support clr.GetClrType() - as in IronPython

vivainio:

Implements https://github.com/pythonnet/pythonnet/issues/432

Yes, #432 

### Any other comments?

...

### Checklist

### DIFF
--- a/src/tests/test_module.py
+++ b/src/tests/test_module.py
@@ -355,16 +355,22 @@ def test_clr_add_reference():
 def test_clr_get_clr_type():
     """Test clr.GetClrType()."""
     from clr import GetClrType
-    from System import String
+    import System
     from System import IComparable
     from System import ArgumentException
-
-    assert GetClrType(String).FullName == "System.String"
+    assert GetClrType(System.String).FullName == "System.String"
     comparable = GetClrType(IComparable)
     assert comparable.FullName == "System.IComparable"
     assert comparable.IsInterface
+    assert GetClrType(int).FullName == "System.Int32"
+    assert GetClrType(str).FullName == "System.String"
+    assert GetClrType(float).FullName == "System.Double"
+    dblarr = System.Array[System.Double]
+    assert GetClrType(dblarr).FullName == "System.Double[]"
 
-    with pytest.raises(ArgumentException):
+    with pytest.raises(TypeError):
+        GetClrType(1)
+    with pytest.raises(TypeError):
         GetClrType("thiswillfail")
 
 def test_assembly_load_thread_safety():
--- a/src/tests/test_module.py
+++ b/src/tests/test_module.py
@@ -352,6 +352,16 @@ def test_clr_add_reference():
     with pytest.raises(FileNotFoundException):
         AddReference("somethingtotallysilly")
 
+def test_clr_get_clr_type():
+    """Test clr.GetClrType()."""
+    from clr import GetClrType
+    from System import String
+    from System import IComparable
+
+    assert GetClrType(String).FullName == "System.String"
+    comparable = GetClrType(IComparable)
+    assert comparable.FullName == "System.IComparable"
+    assert comparable.IsInterface
 
 def test_assembly_load_thread_safety():
     from Python.Test import ModuleTest
--- a/src/tests/test_module.py
+++ b/src/tests/test_module.py
@@ -357,11 +357,15 @@ def test_clr_get_clr_type():
     from clr import GetClrType
     from System import String
     from System import IComparable
+    from System import ArgumentException
 
     assert GetClrType(String).FullName == "System.String"
     comparable = GetClrType(IComparable)
     assert comparable.FullName == "System.IComparable"
     assert comparable.IsInterface
+
+    with pytest.raises(ArgumentException):
+        GetClrType("thiswillfail")
 
 def test_assembly_load_thread_safety():
     from Python.Test import ModuleTest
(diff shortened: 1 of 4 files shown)


vivainio:
This was surprisingly easy, as the type of object passed to GetClrType was actually an instance of Type already.

The implementation just passes it back after a cast, yielding a RuntimeType instance that you can do reflection on.

I will document the feature after you agree in principle that this is the right thing to do.

denfromufa:
Can you please add a minimal test?

src/tests/test_module.py
```diff
@@ -352,6 +352,16 @@ def test_clr_add_reference():
     with pytest.raises(FileNotFoundException):
         AddReference("somethingtotallysilly")
 
+def test_clr_get_clr_type():
```
denfromufa: can you add the test to invoke the ArgumentException error?

vivainio: Done

denfromufa:
@vivainio this looks great, please add yourself to authors, this feature to changelog, and minimal doc somewhere.

filmor:
Note that this behaviour is not exactly the same as what IronPython is doing: https://github.com/IronLanguages/main/blob/ipy-2.7.7/Languages/IronPython/IronPython/Runtime/ClrModule.cs#L525

I haven't tried this out but it looks like the exception they raise here would be `TypeError` because of the method resolution failing. Couldn't you use the exact same implementation?

denfromufa:
@filmor i just tried and you are right:

```
IronPython 2.7.5 (2.7.5.0) on .NET 4.0.30319.42000 (32-bit)
>>> import clr
>>> clr.GetClrType
<built-in function GetClrType>
>>> clr.GetClrType(1)
TypeError: expected Type, got int
```

denfromufa:
I deleted my previous message, here is how it works in ironpython:

```
IronPython 2.7.5 (2.7.5.0) on .NET 4.0.30319.42000 (32-bit)
>>> import clr
>>> clr.GetClrType
<built-in function GetClrType>
>>> clr.GetClrType(1)
TypeError: expected Type, got int

>>> clr.GetClrType(int)
<System.RuntimeType object at 0x000000000000002B [System.Int32]>
>>> clr.GetClrType(int(1))
TypeError: expected Type, got int

>>> type(1)
<type 'int'>

>>> import System
>>> clr.GetClrType(System.Int32(1))
TypeError: expected Type, got int

>>>
```

denfromufa:
@vivainio besides the exception type, here is something else I found that needs to be handled like IPY for compatibility - uninitialized reference types. Can you add this to tests? Here is comparison between C# and IPY. Note that C# requires overloaded method TypeName (found it on SO somewhere):

```python
IronPython 2.7.7 (2.7.7.0) on .NET 4.0.30319.42000 (32-bit)
>>> import clr
>>> import System
>>> dblarr=System.Array[System.Double]
>>> clr.GetClrType(dblarr)
<System.RuntimeType object at 0x000000000000006C [System.Double[]]>
>>>
```

```csharp
scriptcs (ctrl-c to exit or :help for help)

> double[] dblarr;

> Type TypeName<T>(T obj) { return typeof(T); }
> Type TypeName(object obj) { if (obj != null) { return obj.GetType(); } else {
return null; } }
> TypeName(dblarr)
System.Double[], mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a
5c561934e089
> (Type)dblarr
(1,1): error CS0030: Cannot convert type 'double[]' to 'System.Type'
> typeof(dblarr)
(1,8): error CS0246: The type or namespace name 'dblarr' could not be found (are
 you missing a using directive or an assembly reference?)
> dblarr.GetType()
Object reference not set to an instance of an object.
```

vivainio:
It seems the implementation nicked from IronPython handles all the cases you described. I'll push a new version with that & requisite tests later today

src/tests/test_module.py
```diff
@@ -352,6 +352,26 @@ def test_clr_add_reference():
...
+    assert comparable.FullName == "System.IComparable"
+    assert comparable.IsInterface
+    assert GetClrType(int).FullName == "System.Int32"
+    assert GetClrType(str).FullName == "System.String"
+    assert GetClrType(float).FullName == "System.Double"
+    dblarr = System.Array[System.Double]
+    assert GetClrType(dblarr).FullName == "System.Double[]"
+
+    with pytest.raises(TypeError):
+        GetClrType(1)
```
denfromufa: can you compare also against types, not instance of types?

vivainio: Can you elaborate, what kind of comparison I should add?

denfromufa: GetClrType(int), GetClrType(str)

vivainio: Check lines 365-367 in the diff :)

denfromufa: Oh, I see now let's wait for @filmor to approve

obulat:
I'm working on Toga project (pybee), and getting an error when trying to import `GetClrType`:
`ImportError: cannot import name 'GetClrType'`

denfromufa:
@obulat this is in master branch, not on PYPI yet.

obulat:
Oh, thanks for the reply! Do you have any information about when it will be on PYPI?

denfromufa:
Whenever we fix one release blocking issue with numpy usage.

Issue #96, opened 2015-09-06
.NET Core support and CoreCLR embedding - cross-platform API


denfromufa:
Looks like with CoreCLR the unmanaged exports "hack" for embedding CLR can be extended by cross-platform hosting API:

https://github.com/dotnet/coreclr/issues/1256

This would allow simple C/C++ code to embed CLR

Current state of .NET Core support:

https://github.com/pythonnet/pythonnet/issues/96#issuecomment-430465993

denfromufa:
Note that for .NET 4+ still have to use unmanaged exports.

denfromufa:
apparently @lstratman has extensive experience porting Mono to CoreCLR embedding.

lstratman:
Thanks @denfromufa :).

It's funny, I actually had a fork of this repo from a while back with the goal of looking into this.  But, since my company has only a single customer that might want to integrate our .NET server application into their Python application on Linux and, even then, only at some point in the future, it's down towards the bottom of my priority list for development unfortunately.  However, I'm happy to offer some technical guidance.

I implemented basically this same functionality for Node.js for the [Edge.js project](https://github.com/tjanczuk/edge) and a lot of the code that I wrote there can probably be re-used or easily re-purposed for pythonnet.  https://github.com/tjanczuk/edge/blob/master/src/CoreCLREmbedding/coreclrembedding.cpp#L192 contains the code that spins up the .NET Core CLR in any process.  Towards the end of the method (https://github.com/tjanczuk/edge/blob/master/src/CoreCLREmbedding/coreclrembedding.cpp#L673), we get pointers to a bunch of .NET methods that serve as the managed end of the Node.js->.NET interop process.  These function pointers are then invoked anytime that we need to call from Node.js to .NET.  Implementations of those .NET methods are in https://github.com/tjanczuk/edge/blob/master/src/double/Edge.js/dotnetcore/coreclrembedding.cs and are included in the .NET Core application that's run from within Edge.js through a NuGet package that needs to be referenced (see https://github.com/tjanczuk/edge#how-to-specify-additional-clr-assembly-references-in-c-code).

[...]

PR #444, opened 2017-03-28
Added side-by-side VS 2017 build with NetStandard 1.5 support.

dmitriyse:
Added new files that allows to build PythonNet targeting NetStarndard 1.5
Current build scripts are not affected.

All code modification should be loss-less for current build configurations.

New VS 2017 build files are capable to build net40 and NetStandard 1.5 assemblies.

This change is a first CoreCLR compilable version. Probably it is not functional yet.
The work is in progress.

Also CoreCLR assembly/package now have 3.0.0-alpha version. Probably it's ok, if PythonNet with CoreCLR support is scheduled to 3.0.0
...

Not closes yet but relates to https://github.com/pythonnet/pythonnet/issues/96
...

### Any other comments?

We can join this branch into the master even now, and continue to develop with the additional target (NetStandard 1.5).
But it is reasonable to do it after the point where both Windows and Linux configurations will be tested.
...

### Checklist


dmitriyse:
I will inject ReflectionBridge right into PythonNet, to save single-Dll deployments.

vmuriart:
Alot of your changes are also changing the encoding of the files. 
I'll review this in more detail, but I think there's still room to improve compatibility to avoid having multiple `#if NETSTANDARD 1.5` sections.

denfromufa:
@vmuriart did you have a chance to review this .net core integration?

vmuriart:
Very much against it in the current form.
+ Adds too many build directives, making the code complicated. We just removed a bunch of them for PY2/PY3, don't want to readd them now for this
+ No new CI added for `.net core`. If something is added and breaks we won't know.

A better path would be to either refactor some more or completely move to it.

denfromufa:
We can't move completely to .NET Core, it doesn't support DllExport and
demo apps will break. So .NET Framework should still be tested.

denfromufa:
@dmitriyse can we close this pull request, or is this still part of other pull requests for .NET Core support?

dmitriyse:
Yes, we can definitely close this PR. It's outdated. We will not support .NetStandard lower than 2.0.

PR #450, opened 2017-04-05
WIP: 449 Py_Initialize_SignalConfiguration

### LINKED ISSUE #449: Call to Py_Initialize changes signal handler configuration
### Environment

-   Pythonnet version:
2.3.0
-   Python version:
Python 2.7.13 (v2.7.13:a06454b1afa1, Dec 17 2016, 20:42:59) [MSC v.1500 32 bit (Intel)] on win32
-   Operating System:
Win 7, 64 bit
### Details

* What I'm trying to do:
        Console application with CTRL+C handling.

According to https://docs.python.org/2/c-api/init.html, Py_Initialize() changes the current signal configuration. In practice, signals like interrupt from CTRL+C in a console application seems to be ignored, which is not the desired behavior. This means that the event handler configured as this ```Console.CancelKeyPress += Console_CancelKeyPress;``` does nothing.

It seems Py_InitializeEx(0) should be used for general applications that does not include a python terminal. Alternatively, whether to use the Py_InitalizeEx(0) could be a argument to Python.Initialize()

To reproduce try running this:

```cs
class Program
    {
        {
            PythonEngine.Initialize(); // Comment this out to support CTRL+C again.
            Console.WriteLine("Waiting 5s.. Try CTRL+C.");
            System.Threading.Thread.Sleep(5000);
            Console.WriteLine("Done");
        }
    }
```

rmadsen-ks:
Note: This PR is not ready yet. I just added the PR to see if it builds.

Avoiding overriding signals from the application so that CTRL-C might be handled.

Using Py_InitializeEx(0) (instead Py_Initialize) to avoid overriding signals.

This fixes #449 

...

### Any other comments?

...

### Checklist


src/runtime/pythonengine.cs
```diff
@@ -153,8 +153,9 @@ public static void Initialize(bool setSysArgv = true)
         /// more than once, though initialization will only happen on the
         /// first call. It is *not* necessary to hold the Python global
         /// interpreter lock (GIL) to call this method.
+        /// initSigs can be set to 1 to do default python signal configuration. This will override the way signals are handled by the application.
         /// </remarks>
-        public static void Initialize(IEnumerable<string> args, bool setSysArgv = true)
+        public static void Initialize(IEnumerable<string> args, bool setSysArgv = true, int initSigs = 0)
```
filmor: `initSigs` should be a `bool`.

rmadsen-ks:
@denfromufa, it seems some of the builds takes an extraordinary amount of time (59 min). Is there some way of canceling the builds so we dont have to wait for timeout?

denfromufa:
@rmadsen-ks you may need to have access to CI build servers to cancel the builds, what do you think @vmuriart ?

denfromufa:
@rmadsen-ks @filmor @vmuriart do you guys agree on removing the test for this behavior and merge pull request without the tests in CI? We can leave the original issue open, until proper testing is added.

Here is a user in need of this pull request:

https://github.com/pythonnet/pythonnet/issues/511

rmadsen-ks:
@denfromufa, unfortunately I dont have the bandwidth to figure how to make a unit test that works reliably. It turned out to be harder than expected to signal CTRL+C to an application running in the console on windows. Also I had decided to use CSharpProvider for runtime assembly generation, which I think is not compatible with .NET core, so that will probably need to move to Roslyn.

Do you have another idea of how to test this in an automated way?

In any case, the non-test part I think is low risk, but I suggest you do a bit of manual testing first anyway.

denfromufa:
I did not understand the part about CSharpProvider :)

But the Ctrl+c keyboard interrupt can be emulated with this threading call:

`thread.interrupt_main() # raises KeyboardInterrupt`

More details in this new issue that I linked above.

rmadsen-ks:
@denfromufa, ```thread.interrupt_main()``` does not work in our case because we have a C# console application running embedded python code. 

This is why I wanted to use CSharpProvider. I wanted to build a short console application that delays for a few s, but is interrupted by a keyboard interrupt. This I could also do by adding an assembly to the git repo, but I figured that would not be a popular decision. In any case, i have not been able to emulate CTRL+C on windows.

In my testing running _thread.interrupt_main() (python 3) as the embedded python script does not have the desired effect.

filmor:
Let's continue work on this one in #724.

PR #456, opened 2017-04-25
Binary substitution of Python.Runtime.dll becomes safe.

dmitriyse:
Public constants replaced to static readonly fields. Relaxed platform binding.
C# Compiler copies constants of assembly. This may cause to errors when
code was compiled with reference to one build of Python.Runtime.dll and then
after compile other build of Python.Runtime.dll is used. 
...

No, but it work towards cross-platform nuget package https://github.com/pythonnet/pythonnet/issues/165
...

### Any other comments?

...

### Checklist


filmor:
Why do you rename all of those fields?

dmitriyse:
Python.Runtime.dll consumers should not use public constants that have different values in different builds (platform, python version, etc). Otherwise when Python.Runtime.dll replaced in binary distribution, will have mismatch between constants values used in replaced Python.Runtime.dll and constant values copied to assemblies that referenced non replaced version of Python.Runtime.dll at compile time.

So we needs to replace constants to static read only fields.

Also we needs to honor source code backward compatibility and produce no any breaking change in an existing consumer source code.

So the only way to satisfy both "source code backward compatibility" and const->fields change is to rename constants and make them internal.

denfromufa:
closed and re-opened to check on Travis CI again

src/runtime/runtime.cs
```diff
@@ -80,16 +80,22 @@ public static IntPtr GetProcAddress(IntPtr dllHandle, string name)
...
+        internal const int _UCS = 4;
 
         /// <summary>
         /// EntryPoint to be used in DllImport to map to correct Unicode
         /// methods prior to PEP393. Only used for PY27.
         /// </summary>
         private const string PyUnicodeEntryPoint = "PyUnicodeUCS4_";
 #elif UCS2
-        public const int UCS = 2;
+        public const int _UCS = 2;
```
denfromufa: @filmor do you mean that only one if-branch is changed to from public to internal for _UCS? @dmitriyse can you please address this?

src/runtime/runtime.cs
```diff
@@ -100,32 +106,39 @@ public class Runtime
...
 #elif PYTHON36
-        public const string pyversion = "3.6";
-        public const string pyver = "36";
+        public const string _pyversion = "3.6";
+        public const string _pyver = "36";
 #elif PYTHON37 // TODO: Add `interop37.cs` after PY37 is released
-        public const string pyversion = "3.7";
-        public const string pyver = "37";
+        public const string _pyversion = "3.7";
+        public const string _pyver = "37";
```
denfromufa: all these `_pyver*` above should be `internal` @dmitriyse

dmitriyse: Yes, I will fix this soon.

filmor:
@dmitriyse Are you going to continue on this branch? I fear we have to close it at some point as it will have deviated to far from master.

dmitriyse:
I will try to do update it tomorrow. Or we will close this branch.

src/runtime/runtime.cs
dmitriyse: I fixed this and some similar problems.

src/runtime/runtime.cs
dmitriyse: Fixed.

src/runtime/runtime.cs
```diff
@@ -80,16 +80,22 @@ public static IntPtr GetProcAddress(IntPtr dllHandle, string name)
     /// </summary>
     public class Runtime
     {
+        // C# compiler copies constants to the assemblies that references this library.
+        // We needs to replace all public constants to static readonly fields to allow 
+        // binary substitution of different Python.Runtime.dll builds in a target application.
+
+        public static readonly int UCS = _UCS;
```
filmor: Please change these to properties. Allows us to make this fully dynamic in the future without breaking user code.

src/runtime/runtime.cs
```diff
@@ -100,32 +106,39 @@ public class Runtime
 #error You must define either UCS2 or UCS4!
 #endif
 
+        // C# compiler copies constants to the assemblies that references this library.
+        // We needs to replace all public constants to static readonly fields to allow 
+        // binary substitution of different Python.Runtime.dll builds in a target application.
+
+        public readonly string pyversion = _pyversion;
+        public readonly string pyver = _pyver;
```
filmor: These two too.

src/runtime/runtime.cs
dmitriyse: Yes, I will replace them to properties.
But I did not understand this: 
> fully dynamic in the future without breaking user code.

Probably I need some example.

filmor: If we ever want to have a single DLL for all platforms and Python versions, the values of `pyversion` or `UCS` will have to be evaluated at runtime on startup, that's what I meant.

PR #460, opened 2017-05-08
Allow passing None for nullable args

### LINKED ISSUE #59: Creating a C# Nullable Int32 within Python (using Python.NET) to call a C# method with an optional int argument
http://stackoverflow.com/questions/26742837

dhirschfeld:
Fixes #59

### DIFF
--- a/src/runtime/converter.cs
+++ b/src/runtime/converter.cs
@@ -318,6 +318,17 @@ namespace Python.Runtime
                 return true;
             }
 
+            if (obType.IsGenericType && obType.GetGenericTypeDefinition() == typeof(Nullable<>))
+            {
+                if( value == Runtime.PyNone )
+                {
+                    result = null;
+                    return true;
+                }
+                // Set type to underlying type
+                obType = obType.GetGenericArguments()[0];
+            }
+
             if (obType.IsArray)
             {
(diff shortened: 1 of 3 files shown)


dhirschfeld:
Before this PR:
image
...and after:
image

src/runtime/converter.cs
```diff
@@ -318,6 +318,17 @@ internal static IntPtr ToPythonImplicit(object value)
                 return true;
             }
 
+            if( obType.IsGenericType && obType.GetGenericTypeDefinition() == typeof(Nullable<>) )
```
filmor: Please adjust the style here.

dhirschfeld:
@filmor - comments addressed and commit amended.

filmor:
Thank you very much for your contribution :)

filmor:
Ah, sorry, I forgot:

You didn't use the PR-template, so please open another PR in which you
- [ ] Add yourself to AUTHORS
- [ ] Update the CHANGELOG

dhirschfeld:
@filmor, thanks for merging - I think it's a very useful improvement. Changelog & Authors updated in #462

davesliu:
@dhirschfeld  I have a similar issue that for pass null parameter with string type, the code fix in #460 seems not work for it.

PR #461, opened 2017-05-10
Implement named arguments and With semantics in C# embedding side

dsuarezv:
## Named Arguments

I have implemented support for calling methods using named arguments on the C# side. Say we have this code in Python:

`yes = numpy.ones(9, dtype = numpy.int)`
`logger = k.callbacks.CSVLogger('log.csv', separator = ',', append = False)`

Here is how this translated to C# with the old syntax (using keyword arguments in Py): 

`dynamic yes = numpy.ones(9, Py.pw("dtype", numpy.@int));`
`dynamic logger = k.callbacks.CSVLogger("log.csv", Py.kw("separator", ",", "append", false));`

With the new named argument support: 

`dynamic yes = numpy.ones(9, dtype: numpy.@int);`
`dynamic logger = k.callbacks.CSVLogger("log.csv", separator:",", append: false);`

I think the new syntax has better readability. 

## With semantics

Also implemented the semantics of the Python _with_ statement in the embedded side. Take this pice of python code: 

``` 
with tf.device('/cpu:0') as dev:
    model = GetSequenceModel()
    model.fit(X, Y, epochs = 150, batch_size = 32)
```

This was tricky to replicate before in embedded C#. Now it can be called like this: 

```
Py.With((PyObject)tf.device("/cpu:0"), dev =>
{
    dynamic model = GetSequenceModel();
    model.fit(X, Y, epochs: 150, batch_size: 32);
});
```
I implemented the semantics of calling __enter__ and __exit__ in context managers as described in [the python docs](https://docs.python.org/2/reference/datamodel.html#with-statement-context-managers).

### Tests

Some new NUnit tests have been created and added to the embeddingTest project for both named arguments and the new With method. 

Hope you find it useful for the project.


src/embed_tests/Python.EmbeddingTest.csproj
```diff
@@ -1,120 +1,122 @@
...
-  <PropertyGroup>
-    <TargetAssembly>$(TargetPath)</TargetAssembly>
-    <TargetAssemblyPdb>$(TargetDir)$(TargetName).pdb</TargetAssemblyPdb>
-  </PropertyGroup>
-  <Target Name="AfterBuild">
-    <Copy SourceFiles="$(TargetAssembly)" DestinationFolder="$(PythonBuildDir)" />
-    <Copy SourceFiles="$(TargetAssemblyPdb)" Condition="Exists('$(TargetAssemblyPdb)')" DestinationFolder="$(PythonBuildDir)" />
-  </Target>
-</Project>
+﻿<?xml version="1.0" encoding="utf-8"?>
```
filmor: Please try to keep the line-endings as they were (is `autocrlf` set?).

src/embed_tests/TestPyWith.cs
```diff
@@ -0,0 +1,88 @@
...
+        /// <summary>
+        /// Test that exception is raised in context manager that ignores it.
+        /// </summary>
+        [Test]
+        public void TestWithPositive()
+        {
+            var locals = new PyDict();
+            
+            PythonEngine.Exec(@"
+class cmTest:
```
filmor: `CmTest`.

denfromufa:
@dsuarezv thanks for putting this together! Python 2.7 is failing in CI:

https://ci.appveyor.com/project/pythonnet/pythonnet/build/master-655/job/vshxmeaww7rngtjv#L188

```
1) Failed : Python.EmbeddingTest.TestPyWith.TestWithPositive
  Expected: True
  But was:  False
at Python.EmbeddingTest.TestPyWith.TestWithPositive() in C:\projects\pythonnet\src\embed_tests\TestPyWith.cs:line 55

```

src/runtime/pyobject.cs
```diff
@@ -915,6 +915,34 @@ public override bool TrySetMember(SetMemberBinder binder, object value)
             return true;
         }
 
+        private void GetArgs(object[] inargs, CallInfo callInfo, out PyTuple args, out PyDict kwargs)
```
filmor: Very nice :)

src/runtime/pyobject.cs
```diff
@@ -915,6 +915,34 @@ public override bool TrySetMember(SetMemberBinder binder, object value)
...
+            var regularArgumentCount = callInfo.ArgumentCount - namedArgumentCount;
+
+            var argTuple = Runtime.PyTuple_New(regularArgumentCount);
+            for (int i = 0; i < regularArgumentCount; ++i)
+            {
+                AddArgument(argTuple, i, inargs[i]);
+            }
+            args = new PyTuple(argTuple);
+
+            var namedArgs = new object[namedArgumentCount * 2];
```
filmor: You could just use `PyDict` here, couldn't you?

src/runtime/pyobject.cs
```diff
@@ -959,6 +975,32 @@ private void GetArgs(object[] inargs, out PyTuple args, out PyDict kwargs)
...
+        {
+            IntPtr ptr = GetPythonObject(target);
+
+            if (Runtime.PyTuple_SetItem(argtuple, i, ptr) < 0)
+            {
+                throw new PythonException();
+            }
+        }
+
+        private static IntPtr GetPythonObject(object target)
```
filmor: I'm pretty sure that we already have a function like this, I'll check.

src/embed_tests/TestPyWith.cs
denfromufa: @filmor what do you mean here?

filmor: Python convention for classes is `CamelCase`.

dsuarezv:
Hi, thanks for your comments.  

@denfromufa I'll check the tests in 2.7. 

@filmor about implementing With as IDisposable, that was my first feeling as well, the _using_ statement in c# is close to _with_ in Python but, you know, devil is in the details. Here are a couple issues I see, maybe I'm missing a more obvious way of implementing it. 

1. the object in _with ... as obj:_ is the result of the __enter__ method, and it doesn't have to be _self_ (it isn't in many cases). To deal with this, maybe a wrapper class implemeting IDisposable and inheriting from PyObject could be created by this hypothetical With C# method. More on this below. 

2. then, in Python, if a exception happens inside the _with_ block, __exit__ is called and it can decide to swallow the exception or let it raise out. In C#, a _using_ statement just ensures that Dispose is called, but if a exception happens inside the block, it will be raised out of it (unless a try catches it inside).

So, about a wrapper class inheriting from PyObject that implements IDisposable. A method like Py.With would call __enter__ on the parameter (tf.device() on the code above) and get the result, create a PyDisposableObject from that PyObject result (somehow), and return this so _using_ will call its Dispose method that will in turn call _exit_ and raise or ignore the exception. But then again, we have _Py.With_ to start the whole thing, and this would only add a lot of complexity to get to the same result.

denfromufa:
@dsuarezv good finding on differences between C# using and Python with! So I agree with your implementation, especially due to differences in exception handling.

filmor:
@dsuarezv Yeah, this semantical difference and the different exception handling were also what kept me away from implementing this :)

I think you are right, for now this solution is a very very good starting point for further movements in this direction, so I'm fine with merging this in general. Could you address the other commented points while I try to tame our CI?

src/runtime/pyobject.cs
dsuarezv: yes, but this way I'm reusing the validation and conversion in the kw method: 
```
...
if (kv[i + 1] is PyObject)
{
    value = ((PyObject)kv[i + 1]).Handle;
}
else
{
    value = Converter.ToPython(kv[i + 1], kv[i + 1]?.GetType());
}
...
```
Maybe the conversion/validation code could be refactored and called by the 2 methods, I have to look into it.

denfromufa:
@dsuarezv @filmor @vmuriart Travis CI has upgraded to Mono 5.0 (issue #464), so xbuild is now failing and deprecated:

```
/home/travis/build/pythonnet/pythonnet/src/runtime/Python.Runtime.csproj: error : Error executing task Copy: Error converting Property named 'SourceFiles' with value '$(TargetAssemblyPdb)' to type Microsoft.Build.Framework.ITaskItem[]: Can't concatenate Item list with other strings where an item list is expected ('@(_OutDirItem->'%(FullPath)')', 'Python.Runtime.pdb'). Use semi colon to separate items.
```

denfromufa:
closing and re-opening for CI

PR #471, opened 2017-05-21
Fixing Travis CI for mono 5.0

### LINKED ISSUE #464: mono 5.0 - xbuild deprecated, C# 7.0,  concurrent garbage collector
http://www.mono-project.com/docs/about-mono/releases/5.0.0/

dsuarezv:
My attempt to fix Travis CI after the upgrade to mono 5.0. Should fix #464 and let #461 be merged.

I have checked setup.py on osx with mono 5.0, the problem is the escaping of constants to the /p:DefineConstants arguments to msbuild. After trying several things, I'm escaping with %3B (semicolon), should be fine in windows as well. 

I don't have a linux box to test now or a travis setup to verify further, so I'm pushing the PR and will see the results of pythonnet travis. We can take it from there.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -198,7 +198,7 @@ class BuildExtPythonnet(build_ext.build_ext):
             _config = "{0}Win".format(CONFIG)
 
         elif DEVTOOLS == "Mono":
-            _xbuild = "msbuild"
+            _xbuild = "xbuild"
             _config = "{0}Mono".format(CONFIG)
         else:
             raise NotImplementedError(
--- a/setup.py
+++ b/setup.py
@@ -198,7 +198,7 @@ class BuildExtPythonnet(build_ext.build_ext):
             _config = "{0}Win".format(CONFIG)
 
         elif DEVTOOLS == "Mono":
-            _xbuild = "xbuild"
+            _xbuild = "msbuild"
             _config = "{0}Mono".format(CONFIG)
         else:
             raise NotImplementedError(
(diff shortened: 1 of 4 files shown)


dsuarezv:
Fails with 'msbuild not found', any ideas?

fractus:
Your PR builds without problems on my OSX setup with Mono 4.8 though. I would suggest to change the CI config to display additional info, like `mono --version`, `which mono`, env variables.

dsuarezv:
It seems that the problem is this definition in project files (.csproj): 

```<TargetAssemblyPdb>$(TargetDir)$(TargetName).pdb</TargetAssemblyPdb>```

When replaced in this line, throws the error:

```<Copy SourceFiles="$(TargetAssemblyPdb)" Condition="Exists('$(TargetAssemblyPdb)')" DestinationFolder="$(PythonBuildDir)" />```

if ```$(TargetDir)$(TargetName).pdb``` is replaced with ```$(TargetPath)``` it works fine. 

I keep digging...

denfromufa:
@dsuarezv the Travis CI is passing now, are you still making other changes to PDB output? We can open a separate issue if the problem is not easy to resolve since this is secondary. The mono runtime is now 5.0:

https://travis-ci.org/pythonnet/pythonnet/jobs/234974607#L872

dsuarezv:
I thnk it can be merged, the only issue is that PDBs are not copied to the /bin dir in the root of the project after build, even in debug. I think it's not a big issue.  This seems to be some weird issue with xbuild.

vmuriart:
It wasn't an issue previously because mono used to make Mdb files instead. Is coverage on c# still working? I forget which pdb file I mapped it to

filmor:
From my side this looks fine, but as I'm not using C# on OS X or Linux, so my opinion is not super important here :)

denfromufa:
@vmuriart coverage is working!

vmuriart:
Just documenting that Nuget got upgraded from v3.5 to v4.1 with this PR. I couldn't find that info referenced anywhere

Issue #486, opened 2017-06-12
Invalid method binding with specified .NET parameters


Konstantin-Posudevskiy:
### Environment

-   Pythonnet version: 2.3.0.0
-   Python version: 3.5.3
-   Operating System: Windows 10.

### Details

-   Call to .NET method fails with passed arguments (Double.NegativeInfinity, Double.PositiveInfinity, Single.NegativeInfinity, Single.PositiveInfinity).
There is an example, how you can reproduce this bug.

C# code

```csharp
namespace Bugs
{
    public class InvalidMethodBinding
    {
        public static void WriteDouble(System.Double value)
        {
            System.Console.WriteLine(value);
        }
    }
}
```
Python code:

```python
import clr

clr.AddReference('PathToDll')

from System import Double
from Bugs import *

InvalidMethodBinding.WriteDouble(0.01)
InvalidMethodBinding.WriteDouble(Double.NegativeInfinity)
```

-   Traceback:

```python
Traceback (most recent call last):
0.01
  File "X", line Y, in <module>
    Program.WriteDouble(Double.NegativeInfinity)
TypeError: No method matches given arguments
```

denfromufa:
related: https://github.com/pythonnet/pythonnet/issues/100

PR #487, opened 2017-06-12
Fix conversion of 'float' and 'double' values

Konstantin-Posudevskiy:
### Fix problem of conversion 'float' and 'double' values in converter.cs.

As there was a range check both for `float` and `double` values, which are less or greater than its `MinValue` and `MaxValue` accordingly,
several values like `float.NegativeInfinity`, `float.PositiveInfinity` and the same 'double' values cannot be converted from Python to .NET values.

### Add error check after `PyFloat_AsDouble` call
Due to Python C API documentation, method `PyFloat_AsDouble` can return `-1.0` upon failure. So it requires error check. This rule forces to check for error and throw exception in case of error.

### Add tests, which cover problem of conversion `float` and `double` values.

Resolves: #486

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -27,6 +27,7 @@
 -   Joe Frayne ([@jfrayne](https://github.com/jfrayne))
 -   John Burnett ([@johnburnett](https://github.com/johnburnett))
 -   Luke Stratman ([@lstratman](https://github.com/lstratman))
+-   Konstantin Posudevskiy ([@konstantin-posudevskiy](https://github.com/konstantin-posudevskiy))
 -   Matthias Dittrich ([@matthid](https://github.com/matthid))
 -   Patrick Stewart ([@patstew](https://github.com/patstew))
 -   Raphael Nestler ([@rnestler](https://github.com/rnestler))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ This document follows the conventions laid out in [Keep a CHANGELOG][].
 
 -   Fixed Visual Studio 2017 compat (#434) for setup.py
 -   Fixed `FooBar` bug
+-   Fixed conversion of 'float' and 'double' values (#486)
 
 ## [2.3.0][] - 2017-03-11
 
--- a/src/embed_tests/Python.EmbeddingTest.csproj
+++ b/src/embed_tests/Python.EmbeddingTest.csproj
@@ -84,6 +84,7 @@
     <Compile Include="pyimport.cs" />
     <Compile Include="pyinitialize.cs" />
     <Compile Include="pyrunstring.cs" />
+    <Compile Include="TestConverter.cs" />
     <Compile Include="TestCustomMarshal.cs" />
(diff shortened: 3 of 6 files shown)


denfromufa:
@Konstantin-Posudevskiy in .NET I get overflow exception, but not in python. Also precision in .NET is much less for some reason:

```python
Python 2.7.13 |Anaconda custom (32-bit)| (default, Dec 19 2016, 13:36:02) [MSC v
.1500 32 bit (Intel)] on win32
Anaconda is brought to you by Continuum Analytics.
Please check out: http://continuum.io/thanks and https://anaconda.org
>>> dmax = 1.7976931348623159e308
>>> dmax
inf
>>> dmax = 1.7976931348623156e308
>>> dmax
1.7976931348623155e+308
>>> dmax = 1.7976931348623154e308
>>> dmax
1.7976931348623153e+308
>>> dmax = 1.7976931348624159e308
>>> dmax
inf
>>> dmax+1e308
inf
```

```csharp
scriptcs (ctrl-c to exit or :help for help)

> double dmax = 1.7976931348623159e308d
* ;
(2,15): error CS0594: Floating-point constant is outside the range of type 'doub
le'
> double dmax = 1.7976931348623156e308d;
> dmax
1.79769313486232E+308
> dmax = 1.7976931348623146e308d;
1.79769313486231E+308
```

Konstantin-Posudevskiy:
@denfromufa, I absolutely agree with you.

This pull request addresses another issues. It only 

1. fixes invalid usages of Python C API method '_PyFloat_AsDouble_';
2. supports conversion from Python _'inf'_ and _'-inf'_ to _PositiveInfinity_ and _NegativeInfinity_ for _System.Double_ (using method '_PyFloat_AsDouble_')
3. supports conversion from Python float value to _PositiveInfinity_ and _NegativeInfinity_ for _System.Single_. As I am concerned, Python float values, which are in range of four bytes can be represented with the similar byte representation of _System.Single_ in .NET, even if they are values of different precision.
**OverflowException** is thrown for values, that are represented with 5-8 bytes
4. adds tests for paragraphs (2) and (3).

I also checked failing Python tests. **Please, take a look** at method [test_double_conversion](https://github.com/pythonnet/pythonnet/blob/master/src/tests/test_conversion.py#L469).
Python tests expects for .NET overflow exception, in case when it passes _'inf'_ value. IMO, it is incorrect, as **infinity is valid** double value.

Python output for test values from test method _test_double_conversion_:

``` python
>>> dmax = 1.7976931348623159e308
>>> dmax
inf
>>> dmin = -1.7976931348623159e308
>>> dmin
-inf
```

denfromufa:
@Konstantin-Posudevskiy can you fix this failing test?

```
        with pytest.raises(OverflowError):
>           ConversionTest().DoubleField = 1.7976931348623159e308
E           Failed: DID NOT RAISE <type 'exceptions.OverflowError'>
```
OverflowError is not .NET exception, it is Python exception and it should not be expected in this case.

Konstantin-Posudevskiy:
@denfromufa sure, I will do.

Issue #492, opened 2017-06-14
Error when exposing a python method to .Net in python 2.


klasma:
### Environment

-   Pythonnet version: 2.3.0
-   Python version: 2.7.13
-   Operating System: Windows 10

### Details

We are trying to expose a python method in .Net, in a class that inherits from System.Object.

We are following the example code in clr.py in the pythonnet repository except that we are inheriting from System.Object and that we are using a namespace. When we run

```python
class X(System.Object):
    __namespace__ = "PyTest"
    @clrmethod(int, [str])
    def test(self, x):
        return len(x)
```
we get the error message

Traceback (most recent call last):
  File "test.py", line 80, in <module>
    class X(System.Object):
TypeError: Error when calling the metaclass bases
    Unable to find an entry point named 'PyIter_Check' in DLL 'python27'.

The error seems to be due to the fact that PyIter_Check is a #define statement and not a function in python. Therefore the import of the function fails in runtime.cs. The error does not occur in python 3 because PyIter_Check is not imported when python 3 is used.

klasma:
The definition of PyIterCheck in runtime.cs is

```
#if PYTHON2
        [DllImport(PythonDll, CallingConvention = CallingConvention.Cdecl)]
        internal static extern bool PyIter_Check(IntPtr pointer);
#elif PYTHON3
        internal static bool PyIter_Check(IntPtr pointer)
        {
            var ob_type = (IntPtr)Marshal.PtrToStructure(pointer + ObjectOffset.ob_type, typeof(IntPtr));
            IntPtr tp_iternext = ob_type + TypeOffset.tp_iternext;
            return tp_iternext != null && tp_iternext != _PyObject_NextNotImplemented;
        }
#endif
```

We have tried to implement PyIter_Check for python 2, similarly to the way it is done for python 3, but we have not gotten it to work. Here is what we tried

```
internal static bool PyIter_Check(IntPtr pointer)
{
    IntPtr pyType = Marshal.ReadIntPtr(pointer, ObjectOffset.ob_type);
    long flags = Marshal.ReadInt64(pyType, TypeOffset.tp_flags);
    IntPtr iternext = Marshal.ReadIntPtr(pyType, TypeOffset.tp_iternext);
    return (flags & Py_TPFLAGS_HAVE_ITER) != 0 && (Int64)iternext != 0;
}
```

Maybe you can tell us what we are doing wrong.

rickardraysearch:
[...]

PR #494, opened 2017-06-15
clrmethod working for python 2

rickardraysearch:
The issue #492 . It does this by replacing the call to PyIter_Check (which checks whether the object is an _iterator_) in PyObject.IsIterable by a call to the new method PyObject_IsIterable (which checks whether the object is an _iterable_). It also fixes the PyIter_Check for python 2, which has probably never worked.
...

#492 
...

### Any other comments?

...

### Checklist


rickardraysearch:
It's a bit strange to do the Runtime.PyObject_IsIterable and Runtime.PyIter_Check tests from python, but I didn't want to add a new nunit project for the new unit tests. Possibly, that's something to consider for the future. Or is there a good place for nunit tests not related to the embedding?

filmor:
I don't see a problem with adding those tests to what is currently called `EmbeddingTests`.

filmor:
Thank you very much for this :)

Issue #495, opened 2017-06-15
Constructor executed twice when overriding constructor in System.Object.


klasma:
### Environment

-   Pythonnet version: 2.3.0
-   Python version: 2.7.13
-   Operating System: Windows 10

### Details

We are trying to create a subclass of System.Object (using namespace) and override the constructor.

For some reason, the conctructor is executed twice when an instance of the class is created. Below is an example that reproduces the problem.

```python
class X(System.Object):
    __namespace__ = "PyTest"

    def __init__(self):
        print("Running X constructor")

x = X()
```
The output is

```
Running X constructor
Running X constructor
```

rickardraysearch:
Furthermore, if you add arguments to the constructor, the arguments are only passed the second time. 

```csharp
class X(System.Object):
    __namespace__ = "PyTest"

    def __init__(self, *args):
        print("Running X constructor, args = {0}".format(args))

x = X(1,2)
```
produces the output 
```
Running X constructor, args = ()
Running X constructor, args = (1, 2)
```

It seems that the first invocation is from the call in `Python.Runtime.MetaType.tp_call` to `tp_new`, whereas the second is from the call to `py__init__` using `Runtime.PyObject_Call`. 

Is the error that somehow the call to `tp_new` also calls `__init__`?

filmor:
Hmm, the comment says

        /// Metatype __call__ implementation. This is needed to ensure correct
        /// initialization (__init__ support), because the tp_call we inherit
        /// from PyType_Type won't call __init__ for metatypes it doesn't know.

Maybe this changed in Python? I'll have a look.

filmor:
Fixed by #503.

PR #496, opened 2017-06-16
[WIP] Drop tp_call implementation from metatype

filmor:
This seems to be a patch for broken behaviour in Python. Removing it
should solve #495 (`__init__` called twice on construction).

#495

#495

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ This document follows the conventions laid out in [Keep a CHANGELOG][].
     attribute (#481)
 -   Fixed conversion of 'float' and 'double' values (#486)
 -   Fixed 'clrmethod' for python 2 (#492)
+-   Fixed double calling of constructor when deriving from .NET class (#495)
 
 
 ## [2.3.0][] - 2017-03-11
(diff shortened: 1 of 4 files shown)


filmor:
Don't merge this yet, apparently the example given by @rickardraysearch is not tested. When running his test:

    class X(System.Object):
        __namespace__ = "PyTest"

        def __init__(self, *args):
            print("Running X constructor, args = {0}".format(args))

    x = X(1,2)

one can see that without the `tp_call` implementation given here, `__init__` is always called without parameters. I'll add the test and try to make it work.

rickardraysearch:
My very naive understanding of what is being done in pythonnet is as follows: Since there's no exact equivalence to a constructor in python indicating which base constructor to use (apart from calling one in `__new__`), pythonnet solves that by copying all base constructors to the subclass (`AddConstructor` in `ClassDerivedObject.CreateDerivedType`), but tries to run the python-defined `__init__` with any remaining arguments from the constructor call (the `eargs` thingy in `ConstructorBinder.InvokeRaw`). That fails in this case, since `InvokeRaw` is called by the created CLR constructor, which has already lost the extra arguments.

Wouldn't it make more sense to have a `ClassDerived.tp_init` method that calls `__init__` with whatever arguments are passed there? I have a commit at rickardraysearch/pythonnet@ a06fc321bcf40033997ae3fc0df7a69ea0d8367f that does more or less that with some bugs.

filmor:
This is exactly what I was looking into right now :)

PR #497, opened 2017-06-16
Handle C long size on Windows.

filmor:
On Windows 64bit systems, a C `long` is a 32 bit integer while the rest
of the world has agreed on making `sizeof(long) == sizeof(void*)`.

New attempt on #376

### Any other comments?

@vmuriart Do you have a test-case for PR #376?

### Checklist

### DIFF
--- a/src/runtime/Python.Runtime.csproj
+++ b/src/runtime/Python.Runtime.csproj
@@ -136,6 +136,7 @@
     <Compile Include="runtime.cs" />
     <Compile Include="typemanager.cs" />
     <Compile Include="typemethod.cs" />
+    <Compile Include="Util.cs" />
   </ItemGroup>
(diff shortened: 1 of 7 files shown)


denfromufa:
@filmor tp_flags is c_long in Python 2.7 and unsigned c_long in Python 3.6. This may not fit into CLR `int` (`System.Int32`) on *NIX and should be unsigned UInt32 or UInt64 for `PYTHON3`.

denfromufa:
https://docs.python.org/3/c-api/typeobj.html
https://docs.python.org/2/c-api/typeobj.html

filmor:
Does it matter, though? Bitwise it doesn't make a difference, and I'm reading the actual value of `tp_flags` using the `ReadInt` functions, so they will just interpret too large numbers as negative.

PR #503, opened 2017-06-21
Another shot at #495

rickardraysearch:
Remove the extraneous call to `__init__` from `ClassDerived.InvokeCtor`.

#495 

### Any other comments?
This is a different shot at fixing #495 . @filmor has one at #496 but it fails a couple of tests. This one might have some bugs wrt `XDecref` in `MetaType.tp_call` still.

### Checklist

### DIFF
--- a/src/runtime/metatype.cs
+++ b/src/runtime/metatype.cs
@@ -163,6 +163,7 @@ namespace Python.Runtime
             if (init != IntPtr.Zero)
             {
                 IntPtr result = Runtime.PyObject_Call(init, args, kw);
+                Runtime.XDecref(init);
 
                 if (result == IntPtr.Zero)
                 {
--- a/src/runtime/metatype.cs
+++ b/src/runtime/metatype.cs
@@ -1,4 +1,4 @@
-using System;
+﻿using System;
 using System.Runtime.InteropServices;
 
 namespace Python.Runtime
@@ -157,23 +157,12 @@ namespace Python.Runtime
                 return IntPtr.Zero;
             }
 
-            IntPtr py__init__ = Runtime.PyString_FromString("__init__");
-            IntPtr type = Runtime.PyObject_TYPE(obj);
-            IntPtr init = Runtime._PyType_Lookup(type, py__init__);
-            Runtime.XDecref(py__init__);
+            var init = Runtime.PyObject_GetAttrString(obj, "__init__");
             Runtime.PyErr_Clear();
 
             if (init != IntPtr.Zero)
             {
-                IntPtr bound = Runtime.GetBoundArgTuple(obj, args);
-                if (bound == IntPtr.Zero)
-                {
-                    Runtime.XDecref(obj);
-                    return IntPtr.Zero;
-                }
-
-                IntPtr result = Runtime.PyObject_Call(init, bound, kw);
-                Runtime.XDecref(bound);
+                IntPtr result = Runtime.PyObject_Call(init, args, kw);
 
                 if (result == IntPtr.Zero)
                 {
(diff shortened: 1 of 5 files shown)


src/tests/test_subclass.py
```diff
@@ -190,3 +190,59 @@ def test_isinstance_check():
     for x in b:
         assert isinstance(x, System.Object)
         assert isinstance(x, System.String)
+
+def test_clr_subclass_with_init_args():
```
denfromufa: please rename test_clr_subclass_init_with_args

src/tests/test_subclass.py
```diff
@@ -190,3 +190,59 @@ def test_isinstance_check():
...
+    class TestX(System.Object):
+        __namespace__ = "test_clr_subclass_with_init_args"
+        def __init__(self, *args, **kwargs):
+            calls.append((args, kwargs))
+    t = TestX(1,2,3,foo="bar")
+    assert len(calls) == 1
+    assert calls[0][0] == (1,2,3)
+    assert calls[0][1] == {"foo":"bar"}
+
+def test_clr_subclass_without_init_args():
```
denfromufa: please rename test_clr_subclass_init_without_args

src/runtime/metatype.cs
```diff
@@ -157,23 +157,12 @@ public static IntPtr tp_call(IntPtr tp, IntPtr args, IntPtr kw)
                 return IntPtr.Zero;
             }
 
-            IntPtr py__init__ = Runtime.PyString_FromString("__init__");
-            IntPtr type = Runtime.PyObject_TYPE(obj);
-            IntPtr init = Runtime._PyType_Lookup(type, py__init__);
-            Runtime.XDecref(py__init__);
+            var init = Runtime.PyObject_GetAttrString(obj, "__init__");
```
filmor: You are leaking this reference.

filmor:
Could you explain how and why this works?

rickardraysearch:
I believe that what's happening is that `MetaType.tp_call` is calling `__init__` on the class, just as PyType_Type does in [type_call](https://github.com/python/cpython/blob/master/Objects/typeobject.c#L903). It has access to all of the passed arguments, including those not available in the CLR constructor. 

Since the class' `__init__` is called from there, there is no need to call it from `ClassDerived.InvokeCtor`. 

When constructing python objects from e.g. C#, the `__init__` method would need to be called separately, since the CLR constructor only creates the object. If that should change, I believe we should have a look at actually calling `__new__` for python types with `__namespace__` and make those arguments match whatever is generated for the CLR constructors.

src/runtime/metatype.cs
rickardraysearch: Oh.

rickardraysearch:
From the commit history, it looks as if the call was added to invoke `__init__` when instantiating python classes from managed code. The tests that @tonyroberts added then still work though, so hopefully this PR isn't breaking any existing major use cases by removing the call to `__init__` when constructing python objects from managed code.

By the way, there are two tests in `test_subclass.py` that are currently skipped with the comment `FIXME: test randomly pass/fails`. I have run them a couple of hundred times, both on macos and windows, with python 2.7 and 3.6, and not seen a single fail. Can I try reenabling them? They are both relevant to this PR.

filmor:
Thank you very much for your contribution. We'll have to see, whether this breaks compatibility for anyone, we probably should put a note into the changelog, documenting this change of behaviour.

rickardraysearch:
Yes, it definitely makes sense to put a good explanation in the changelog. Also, could we have a discussion on how to bring back that functionality and exposing additional arguments to the managed constructors? Maybe a new issue is the best place for that.

filmor:
Feel free to open one :)

Issue #458, opened 2017-05-03
Propose allow gcc usage instead of clang if not available


testrunner123:
### Environment

-   Pythonnet version: 2.3.0
-   Python version: 3.4
-   Operating System: Ubuntu 14.04

### Details

-   Describe what you were trying to get done.

I was trying to build on my system. But I don't have clang. So I tried with my gcc 4.9.4.
It appears that it works beautifully and generates interop34m.cs with same content as the one made with clang. I suggest that geninterop.py should run with "gcc" if clang is not installed. At least if gcc has version >=4.9.4

-   What commands did you run to trigger this issue? If you can provide a
    [Minimal, Complete, and Verifiable example](http://stackoverflow.com/help/mcve)
    this will help us understand the issue.

1. I replaced string "clang" with "gcc" in geninterop.py. 
2. run setup.py.

denfromufa:
@testrunner123 pull requests are welcome! We should probably also update fake_libc_include:

https://github.com/eliben/pycparser/tree/master/utils/fake_libc_include

https://github.com/pythonnet/pythonnet/tree/master/tools/geninterop/fake_libc_include

testrunner123:
I do not think that we need update fake_libc_include. Python C-extension api do not use lot of things and current fake_libc fullfill it completely. Newer fake_libc has IMHO just more fake headers important for other projects.

I do not program in python :) I just replaced string "clang" with "gcc" and it appears to work. I also believe that on Windows VS2015 C compiler can be used, because it has /E and /EP options that yield similar results. The only difference is that VS2015 compiler uses / instead of - for options.

Besides that it is not clear for me why "clang" was used until now and what it should be in future. I would propose:

- always use "gcc" on linux, because it is by default already installed
- on Mac it may still be "clang" ( I do not have Mac)
- on Windows VS2015 or "gcc" (I know some people use mingw) can be probed and used ?

So what do you think?

denfromufa:
I think @tonyroberts chose clang because this is recommended by pycparser and is very light dependency on all platforms.

QuantApp:
Any ideas on the following?

I forked the python net project and changed the clang flag for gcc.
I wish to run this Azure Notebooks

!pip install git+https://github.com/quantapp/pythonnet



Python 3.6.0 |Anaconda custom (64-bit)| (default, Dec 23 2016, 12:22:00)
[GCC 4.4.7 20120313 (Red Hat 4.4.7-1)] on linux
[...]

PR #510, opened 2017-07-11
Use gcc on Linux, VS on Windows and CLang in other cases for geninterop

testrunner123:
Improve generating on geninterop data

#458

### Any other comments?

On Windows "cl.exe" must be in PATH, on Linux "gcc".

### Checklist


filmor:
We have some lookup code in `setup.py`, could you incorporate this somehow here?

denfromufa:
I agree with @filmor about locating cl.exe from default directories, if the build is not started from VS command-line prompt with pre-configured paths. Also we should probably still allow to fall back to clang, since it is much lighter dependency on Windows than MSVC. Even pycparser recommends Clang for Windows:

https://github.com/eliben/pycparser#31interaction-with-the-c-preprocessor

For Linux only gcc should be fine.

Finally right now this pull request is not testing this functionality with compilers due to `interopXX.cs` files already available. These files are only re-generated for new versions of Python or new combinations of Python flags.

denfromufa:
I just tried to search the paths for `mt.exe`, `msbuild.exe`, and `cl.exe` used during build on Windows and the paths are totally unrelated to each other:

image

denfromufa:
Here is proper way to locate cl.exe, or more precisely to setup environment variables to get cl.exe into path:

https://github.com/pypa/setuptools/blob/master/setuptools/msvc.py

This above module monkey-patches distutils (just import setuptools) to provide location of cl.exe for up-to-date compiler locations (distutils.py updates on much slower cycle):

https://github.com/python/cpython/blob/master/Lib/distutils/msvccompiler.py#L267

More details about this spaghetti code for msvc compilers in this article (see also comments):

https://blogs.msdn.microsoft.com/pythonengineering/2016/04/11/unable-to-find-vcvarsall-bat/

testrunner123:
@denfromufa
I personally will not use clang on Win/Lin for several reasons:

1. clang based geninterop is potentially error prone
Python for Windows (that I download from python.org) is compiled with MSVC. Python for my Ubuntu was compiled with Gcc. If I use another compiler with another runtime as the one used to build Python interpreter then I would risk to create bugs that should not be there. CPython Module Extension API is also platform dependent.
2. clang is unnecessary ballast for build system
Clang introduces unnecessary effort when integrating pythonnet into complex project. In fact you need any "C Preprocessor" for geninterop, so why not to use the one that I already have (native to the platform). The purpose of my pull request was to provide information how this can be done.

I use CMake for build and it finds compiler paths automatically. I do not think that each package must do it on its own, because it is pretty complex problem.

> Even pycparser recommends Clang for Windows:

There is no recommendation for Clang on linked page. Also there are no reports about problems with msvc/gcc on it. And I didn't meet any problems.
> Finally right now this pull request is not testing this functionality with compilers due to interopXX.cs files already available. These files are only re-generated for new versions of Python or new combinations of Python flags.

But there is no test in existing codebase for clang-based geinterop either. I believe it is a different topic to create automatic test for geninterop script. I have tested it manually with Python 2.7/3.4 on Ubuntu 64-bit, Python 3.5 on Windows 32-/64-bit.

testrunner123:
@filmor
> We have some lookup code for Visual Studio paths in setup.py, could you incorporate this somehow here?

It looks like now clang on Windows is used from PATH. Or where is its path set ?

denfromufa:
@testrunner123 please address merge conflicts

filmor:
Since geninterop is an internal script that only needs to be triggered when we want to support a new Python version, I don't thinkt it makes sense to change it right now.

PR #518, opened 2017-07-28
CoreCLR msbuild (xplat) support. Initial compilable version.

dmitriyse:
This is initial step in a long road to add "xplat" build tools support into python.net.
...

Not closes but will helps to solve those:
https://github.com/pythonnet/pythonnet/issues/96

...

### Any other comments?
Or even do everything in the master, because this xplat support should be implemented as side-by-side to the current build system.
Later in the branch coreclr2 we can track master and also allow to build NetStandart 2.0 target.
...

### Checklist

-   [] Make sure to include one or more tests for your change
-   [] If an enhancement PR, please create docs and at best an example
-   [] Add yourself to [`AUTHORS`](../blob/master/AUTHORS.md)
-   [] Updated the [`CHANGELOG`](../blob/master/CHANGELOG.md)


denfromufa:
@dmitriyse can you merge `*.15.csproj` and `*.15.sln` as options to Solution Configurations next to `*Win*` and `*Mono*` options in the regular `pythonnet.sln` and its projects?

dmitriyse:
> @dmitriyse can you merge *.15.csproj and *.15.sln as options to Solution Configurations next to *Win* and *Mono* options in the regular pythonnet.sln and its projects?
May be it is possible to use single .sln file (i will try). But it is definitely impossible to fit xplat and classic builds into one csproj file.
Take a look at the first lines of classic a *.csproj and a new *.csproj 2017.
2017:
```Xml
<Project Sdk="Microsoft.NET.Sdk">
```
2000+:
```Xml
<Project DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003" ToolsVersion="4.0">
```

Msbuild and VS uses totally different toolsets for those project types.

dmitriyse:
Also I think we need use only xplat in the future.
It can compile totally all targets, any runtime, any platform, any CLR version.

So current build system should become legacy. But some part of time (migration period) both build system should co-exits.

So current build system and xplat build system should have the same set of solution configurations.
Thus I also think usage of single .sln file is a bad idea (even if is possible).

denfromufa:
@dmitriyse can you point to any Microsoft references for understanding about the migration of build systems?

Also have you read this document, which suggests that it is possible to support both runtimes from the same solution?

https://docs.microsoft.com/en-us/dotnet/core/porting/project-structure

dmitriyse:
P.S. Sorry, but I will create PR only for parts that I use in my commercial projects. I have no free time to support xplat build system for all pythonnet targets. But my contributions contain well-tested in the production solutions, and even if some PR is not complete, it can be merged into the main repo to move work forward.

dmitriyse:
> https://docs.microsoft.com/en-us/dotnet/core/porting/project-structure

Thank you for pointing. I will review this article, probably there is some better solution than in this PR.

dmitriyse:
image
MS suggest the near the same solution that contained in this PR.
But, if we will do exactly like MS suggest, than current pythonnet build system will completely broken.
So this PR is still relevant.

dmitriyse:
> @dmitriyse can you point to any Microsoft references for understanding about the migration of build systems?

Unfortunately, no.
I am just sitting on the bleeding edge tools and know what they really can (or cannot) now.

denfromufa:
@dmitriyse note that you skipped the first option suggested in that article :)

https://docs.microsoft.com/en-us/dotnet/core/porting/project-structure#replace-existing-projects-with-a-multi-targeted-net-core-project

dmitriyse:
>@dmitriyse note that you skipped the first option suggested in that article :)
https://docs.microsoft.com/en-us/dotnet/core/porting/project-structure#replace-existing-projects-with-a-multi-targeted-net-core-project

This is the revolution way. xbuild and old msbuild (prior to 15.x) will fail to build anything.
This would be a huge breaking change for python.net users, and will stress contributors to solve big number of problems at once.
IMO this way is a worst one.

vmuriart:
I agree w @denfromufa  that I'd rather combine the solution files into a single solution. If this means that contributors will need to use VS2017 to build but otherwise it outputs the same, I'm ok with this. Most windows users likely install from `pypi` anyways.

If combining them creates issues for mono, then I'd hold off altogether. My main concern for it is based on the clean-up I did earlier in the year in which we had 3 or 4 different solution files because of multiple upgrades of VS and then those files lingered around for a **long** time.
Second problem is that these changes are not being tested at all as part of the CI. If a breaking change is introduced somewhere in the code, we will be oblivious to it for a while.

dmitriyse:
> If combining them creates issues for mono, then I'd hold off altogether. 

What do you mean ?

>If combining them creates issues for mono, then I'd hold off altogether. My main concern for it is based on the clean-up I did earlier in the year in which we had 3 or 4 different solution files because of multiple upgrades of VS and then those files lingered around for a long time.

We can contain one set of csproj files and on sln file only if we totally move to VS 2017/Xplat build system. And MonoDevelop users should migrate to JetBrains Project Rider. 
This work is too long, thus even in this revolutionary case python.net project should have
have two set of csproj files and two sln files, for new and old build system, but in different git branches.

So the choice is between 1) Avoid NetCoreApp 2.0 target and use workarounds(see later) 2) use two sets of csproj files in the same branch 3) some period of time maintain two branches with old and new build systems.

I prefer option 2) - use two sets of csproj files, because option 3) is a different type of hell with merging with even more required efforts that maintain two sln files.

dmitriyse:
NetCoreApp 2.0 offer ability to reference binaries that targets framework 4.6.1 and earlier. I did not try this ability. But in this case existing pythonnet binaries can be referenced to Linux CoreCLR 2.0 projects as is.

But even if this magic is really exists, we needs to duplicate all CI builds for CoreCLR 2.0 hosts.

dmitriyse:
Off topic. I have a bunch of stability&performance fixes for pythonnet/CoreCLR from the high-load production project (Currently they are contained here https://github.com/dmitriyse/pythonnet/tree/pythonnet-plus). 

Lacks of official support of CoreCLR stops me from creating pull requests.

denfromufa:
@dmitriyse even if we decide together to split into 2 projects/solutions, still this pull request does not include testing for new set of .NET Core DLLs. Can you please add this?

After some more thinking, I might actually incline towards two solutions. The reason is that the current setup.py assumes .NET Framework for Windows, and Mono for other platforms.

But we need a new setup.py that uses .NET Core on all platforms.
Of course we could also add a flag to setup.py to be able to use .NET Core as a non-default option like described here:

https://stackoverflow.com/questions/677577/distutils-how-to-pass-a-user-defined-parameter-to-setup-py

dmitriyse:
Yes, I will try to inject some option in the setup.py to allow build the same output but with xplat build tools and new *.15.csproj + *.15.sln files.
Finally we needs to have 3 times more CI build configurations:
1) net40 + Current Tools (msbuild < 15.0, mono)
2) net40 + xplat build tools 
3) NetStandard20 + xplat build tools
Only with this set of CI builds we can ensure that codebase have no regression.

vmuriart:
@dmitriyse what I meant earlier is due to the fact that Linux users will be building the project alot more often than windows users, so if there's any regressions introduced there because of having a single solution on then I'd rather keep the current solution file.

I'm ok with having the dual solutions as long as we add the CI for them and open an issue to *eventually* merge them back into one but leave a transition period in which we'll have both solution files in existence.

On the CI, I'm ok tripling the build time for now so that we test out all configurations. After the PR is merged in, I'll rewrite the CI config to test all PY version on one configuration, and then limit the other configuration to subset of them.

dmitriyse:
Eventually we can totally move from mono/msbuild < 15.0 to xplat/msbuild >15,0 and keep all users onboard. Eventually even MonoDevelop/XamarinStudio will obtain xplat/NetCoreApp 2.0 support (The work is in progress http://lastexitcode.com/blog/2016/06/05/AspNetCoreRC2SupportInXamarinStudio).

Thank you for acceptance. I will try to find a time soon to improve setup.py as a part of this PR.Eventually we can totally move from mono/msbuild < 15.0 to xplat/msbuild >15,0 and keep all users onboard. Eventually even MonoDevelop/XamarinStudio will obtain xplat/NetCoreApp 2.0 support (The work is in progress http://lastexitcode.com/blog/2016/06/05/AspNetCoreRC2SupportInXamarinStudio).

Thank you for acceptance. I will try to find a time soon to improve setup.py as a part of this PR.

dmitriyse:
Now it's not the same pull requests :).
As I already say, this PR covers only injection of the new build system (xplat).
So we can build just the same output (net40 assemblies) under linux and under windows but with CoreCLR tooling.

New tooling requires completely different .sln and csproj files.

NetStandard 2.0 target covered by anther PR. 

We needs to ensure that with the same sources and the same build targets new build system works good.
So "step by step".

dmitriyse:
> On the CI, I'm ok tripling the build time for now so that we test out all configurations. After the PR is merged in, I'll rewrite the CI config to test all PY version on one configuration, and then limit the other configuration to subset of them.
@vmuriart, you can create new CI configs with setup.py build_ext --inplace --xplat build command and CoreCLR 2.0 Preview 2.0+ installed.
(see https://www.microsoft.com/net/core/preview#linuxubuntu
and  https://www.microsoft.com/net/core/preview#windowscmd)

Currently --xplat option implementation is just in very beginning, but it passes package restore and build.
(setup.py, *15.csproj , pythonnet.15.sln should be improved, to create the same output as mono and msbuild)

denfromufa:
@dmitriyse i don't see `--xplat` being passed to `setup.py` in appveyor.yml and travis.yml.

denfromufa:
looks like hitting this issue with python 3.6:

https://github.com/pypa/setuptools/issues/1101

dmitriyse:
> @dmitriyse i don't see --xplat being passed to setup.py in appveyor.yml and travis.yml.
I did not figure out that I can change build options just in the PR to perform tests, until it's merged to the master.
After some improvements it will be reasonable to perform CI tests.

P.S. 
Just before merging to the master we needs to find a way to multiply build configurations (with --xplat and without --xplat).

denfromufa:
@dmitriyse expanding test matrix to include coreclr is possible, see here:

https://docs.travis-ci.com/user/customizing-the-build/#Build-Matrix
https://www.appveyor.com/docs/build-configuration/#build-matrix

setup.py
```diff
@@ -193,7 +203,10 @@ def build_extension(self, ext):
             geninterop = os.path.join("tools", "geninterop", "geninterop.py")
             subprocess.check_call([sys.executable, geninterop, interop_file])
 
-        if DEVTOOLS == "MsDev":
+        if self.xplat:
```
vmuriart: why are you defining a new variable instead of reusing the existing one?

dmitriyse:
I updated xplat csproj/sln files. Now it's fully implemented. setup.py with --xplat key should produce exactly the same output as xbuild/msbuild.14, but with msbuild.15 toolset (even under linux).

dmitriyse:
It's a real nightmare to debug all this stuff. 100500 small build issues. I spent near two working days to get linux builds works.

dmitriyse:
But we have a good news, it's already possible to completely migrate to the new build system and new sln/csproj files. Once new build system will be widely supported in IDEs (Project Rider, MonoDevelop), we can make MsDev15/dotnet build as a default.

dmitriyse:
AppVeyor builds I will debug in another day.

dmitriyse:
Travis and AppVeyor successfully passes all builds! 
But ... nunit-console randomly hangs under travis-ci. The only change is version of the mono and ubuntu 12.04 -> 14.04 upgrade.
I will try to solve this issue soon.

dmitriyse:
All issues fixed. Now we have two working build system:
1) Mono XBuild  + MSBuild < 15.0
2) dotnet core  + MSBuild >= 15.0
Those two build system are able to build exactly the same assemblies with the same sources.
Second build system is required for build NetCoreApp version of pythonnet.
Work toward NetCoreApp target is tracked in another PR, this PR contains only second build system.

After some time we can remove first (current) build system. We needs to ensure that linux and mac IDEs correctly supports this new build system.

denfromufa:
@dmitriyse do we still need this other PR? if not, can you close it:

https://github.com/pythonnet/pythonnet/pull/519

dmitriyse:
Yes, we needs both https://github.com/pythonnet/pythonnet/pull/518 and https://github.com/pythonnet/pythonnet/pull/519.

If we merge both works 1) New build system and 2) NetCoreApp 2.0 support, it will be huge super PR that is hard to review.

https://github.com/pythonnet/pythonnet/pull/519 is a next step after https://github.com/pythonnet/pythonnet/pull/518. 
So we can skip reviewing https://github.com/pythonnet/pythonnet/pull/519 until https://github.com/pythonnet/pythonnet/pull/518 will be merged into the master.

Now I am working on https://github.com/pythonnet/pythonnet/pull/519 and finally it will contain enough diffs to https://github.com/pythonnet/pythonnet/pull/518

setup.py
```diff
@@ -319,6 +364,20 @@ def _find_msbuild_tool(self, tool="msbuild.exe", use_windows_sdk=False):
 
         raise RuntimeError("{0} could not be found".format(tool))
 
+    def _find_msbuild_tool_15(self):
+        """Return full path to one of the Microsoft build tools"""
+        try:
+            basePathes = subprocess.check_output(
+                ["vswhere", "-latest",
```
denfromufa: @dmitriyse can you change this to `tools\vswhere\vswhere.exe` like I did in my pull request #540?

setup.py
```diff
@@ -196,20 +216,33 @@ def build_extension(self, ext):
...
+            _custom_define_constants = True
         elif DEVTOOLS == "Mono":
-            _xbuild = "xbuild"
+            _xbuild = 'xbuild'
             _config = "{0}Mono".format(CONFIG)
+            _solution_file = 'pythonnet.sln'
+            _custom_define_constants = False
+        elif DEVTOOLS == "dotnet":
+            _xbuild = 'dotnet msbuild'
+            _config = "{0}Mono".format(CONFIG)
```
denfromufa: why dotnet DEVTOOLS corresponds to Mono _config?

denfromufa:
@dmitriyse you mentioned that you are not using MonoDevelop, how are you testing these solution on Linux and OSX? Are you using Rider IDE on these platforms?

I tested and both solutions look good on Windows in VS 2017, just don't try to open both solutions at the same time :)

denfromufa:
@dmitriyse can you please avoid a lot of duplication in travis.yml in the matrix by using excludes?

https://docs.travis-ci.com/user/customizing-the-build/#Excluding-Jobs

denfromufa:
@filmor @dmitriyse note that some comments are still not addressed:

https://github.com/pythonnet/pythonnet/pull/518#discussion_r138716245
https://github.com/pythonnet/pythonnet/pull/518#discussion_r138712339
https://github.com/pythonnet/pythonnet/pull/518#issuecomment-329285804

dmitriyse:
I will "address" those today. Please sorry for delay.

setup.py
dmitriyse: I added tools\vswhere\vswhere.exe file into the repository instead of chocolatey installation.
See PR #546

setup.py
dmitriyse: The first problem:
DebugMonoPY3, ReleaseMono, etc. build configuration is a historically names from times where nothing than mono had exists for the linux.
Now it is more correctly to name DebugLinuxPY3, ReleaseLinux, etc.

http://cs5.pikabu.ru/images/big_size_comm/2015-03_2/14257150104254.jpg

But we "cannot just rename" build configurations without affecting users.
We should introduce new build configurations (DebugLinuxPY3, ReleaseLinux), give a time to users to migrate to new names, and than remove old names DebugMonoPY3, ReleaseMono, etc.

The second problem:
We cannot use dotnet build command under windows due to DllExport project msbuild targets. We needs to use msbuild >=15 under windows 

The result:

DEVTOOLS=dotnet ==> linux (--xplat) === ReleaseLinuxPY3 <== ReleaseMonoPY3

So currently we have DOVTOOLS=dotnet <===> ReleaseMonoPY3

If I am wrong in somewhere in my conclusions, lets improve something.

dmitriyse:
Answering to the https://github.com/pythonnet/pythonnet/pull/518#issuecomment-329285804 
Unfortunately travis-ci currently does not supports conditional  "addons" section. 
See https://github.com/travis-ci/travis-ci/issues/3618
So exclude solution is totally impossible now and in near future.

But fortunately we have a little yaml reference sugar, I will try to apply it.

dmitriyse:
I was managed to solve copy-past .travis.yaml problem with help of yaml references Solution is included into https://github.com/pythonnet/pythonnet/pull/546

dmitriyse:
> @dmitriyse you mentioned that you are not using MonoDevelop, how are you testing these 2 solutions on Linux and OSX? Are you using Rider IDE on these platforms?
I tested and both solutions look good on Windows in VS 2017, just don't try to open both solutions at the same time :)

I tried Project rider 2017.1.2 for the pythonnet.15.sln and...
It's unbelievable:
1) I am successfully build, run, debug, and run/debug unit tests for .Net 4.0
2) I am successfully build and run CoreApp 2.0 target, but unable to debug it and unable to run/debug NUnit tests

Probably full support will be added soon.
image

But in general Project rider works great and looks like a very mature tool.
RIP MonoDevelop. (Nothing personal, this product played big role)

P.S. some fix for better usability inside IDEs are added to #546

dmitriyse:
We can advertise full migration to --xplat in the 2.5 release. IDEs already can handle development in any  OS.

Issue #475, opened 2017-05-29
Exceptions thrown from c# iterators (yield) cannot be caught


drlagos:
### Environment

-   Pythonnet version: 2.3.0
-   Python version: 2.7.10
-   Operating System: Windows 7

### Details

-   Describe what you were trying to get done.

Catch an exception thrown from a C# iterator (done with yield return) in a python script.    

-   What commands did you run to trigger this issue? If you can provide a
    [Minimal, Complete, and Verifiable example](http://stackoverflow.com/help/mcve)
    this will help us understand the issue.

```csharp
using System;
using System.Collections;
using System.Collections.Generic;
namespace Example
{
	public class Script 
	{
	    public static IEnumerable<int> Test()
	    {
	        for (var i = 0; i < 10; i++)
	        {
	            if (i == 5)
	                throw new Exception("Ooops!");
	            yield return i;
	        }
	    }
	}
}
```
compile:
```batch
"C:\Windows\Microsoft.NET\Framework\v4.0.30319\csc.exe" /t:library bad_yield.cs
```

```python
import clr
clr.AddReference(...)
import System
import Example


try:
    for r in Example.Script.Test():
        print r
except System.Exception as e:
    print e.Message
except Exception as e:
    print e.message
except:
    print "Unfortunately this one doesn't work either"
```

-   If there was a crash, please include the traceback here.

```csharp
Unhandled Exception: System.Exception: Ooops!
    at Example.Script.<Test>d__13.MoveNext() in ...
    at Python.Runtime.Iterator.tp_iternext(IntPtr ob)
```

denfromufa:
@drlagos thanks for detailed report. I edited your response for full example.

iantDev:
- Python 3.6.1
- Pythonnet 2.3
- Windows 7

Similar issue with SqlDataReader:
`   dbComm = SqlCommand(query, Conn)
    Conn.Open()        
    sqlReader = dbComm.ExecuteReader()
    sqlReader.Read() 
    sqlReader.Close()
    `
I have tried the following as subsequent code before closing the reader:
`
while sqlReader.Read():
   pass
`
but same error.

Error message:
`
Unhandled Exception: System.InvalidOperationException: Inlvaid attempt to call FieldCount when reader is closed.
     at System.Data.SqlClient.SqlDataReader.get_FieldCount()
     at System.Data.Common.DbEnumerator.BuildSchemaInfo()
     at System.Data.Common.DbEnumerator.MoveNext()
[...]

PR #523, opened 2017-08-09
Catching exceptions from iterator MoveNext() (related to #475) 

rickardraysearch:

Catches exceptions in iterator MoveNext() and throws them in python instead of crashing.

Issue #475 
### Any other comments?

...

### Checklist


rickardraysearch:
For future reference, the only place where a general "catch (Exception E)" is done without throwing away the outermost exception is in TypeMethod.Invoke.

denfromufa:
@rickardraysearch can you clarify about that TypeMethod.Invoke? Is this an issue or just a note?

rickardraysearch:
@denfromufa , I only wanted to note that the usage is not completely consistent across all invocations of Exceptions.SetError(). When we revisit how to do the SetErrors, I want to remember this without having to find it out again.

denfromufa:
@filmor @vmuriart please review!

vmuriart:
What happens if you call `next()` after the exception is thrown? Might be worth adding a test for it to make sure it has expected behavior.

Other than adding that test, ok to merge

denfromufa:
> What happens if you call next() after the exception is thrown? Might be worth adding a test for it to make sure it has expected behavior.
> 
> Other than adding that test, ok to merge

@vmuriart did you mean calling this part twice?

```
with pytest.raises(OverflowException) as cm:
    next(val)
```

denfromufa:
@vmuriart or did you mean to verify this behavior?

https://stackoverflow.com/a/11366139/2230844

vmuriart:
both :).

denfromufa:
@rickardraysearch can you please address the suggestions from @vmuriart and then we can merge this!

ArvidJB:
We just ran into the same issue that this pull request is going to fix. Is there any particular reason this was never merged? If there is further work to be done we would be happy to help.

denfromufa:
@ArvidJB i just updated this branch with latest master and CI tests look good. You can add 2 additional unit tests suggested by @vmuriart but personally I approved this PR. If one more core developers approves, then we can merge this. @filmor had a question and that probably can be addressed in another PR.

ArvidJB:
@denfromufa / @vmuriart I think this boils down to adding the following lines
```
    # after exception is thrown iterator is no longer valid
    with pytest.raises(StopIteration):
        next(val)
```
to `test_iteration_innerexception`. Should I create a new pull request for that or would you be willing to add those lines to @rickardraysearch 's request?
I tested @rickardraysearch 's changes and they work as intended.

denfromufa:
@ArvidJB please fork @rickardraysearch branch and make a new pull request with your changes. If you PR looks good, then we can close this PR.

ArvidJB:
@denfromufa I just submitted #693 with the enhanced testing

filmor:
Merged as #693.

PR #528, opened 2017-08-23
Optimizes implicit assembly loading. Helps to reduce amount of faulted LoadAssembly calls.

dmitriyse:

1) Assembly.GetTypes replaced to Assembly.GetExportedTypes. So we are really skipping non-public types now.
2) Type.IsNested types are excluded from the names collection that is used for implicit assembly loading.

no
...

### Any other comments?
VS 2017 + Docker is insensitive to DebuggerNonUserCode, and LoadAssembly exceptions produces long trash in the debug log.
+ This fix speedups unit tests (up to 3x). And probably speedups initialization of production projects.
...

### Checklist


src/runtime/assemblymanager.cs
```diff
@@ -343,8 +353,17 @@ internal static void ScanAssembly(Assembly assembly)
             // A couple of things we want to do here: first, we want to
             // gather a list of all of the namespaces contributed to by
             // the assembly.
+            Type[] types = new Type[0];
+            try
+            {
+                types = assembly.IsDynamic ? assembly.GetTypes():assembly.GetExportedTypes();
```
denfromufa: @dmitriyse  can you please add explanation for why you added check for `isDynamic`?

src/runtime/assemblymanager.cs
```diff
@@ -419,12 +438,15 @@ public static List<string> GetNames(string nsname)
...
                 foreach (Assembly a in namespaces[nsname].Keys)
                 {
-                    Type[] types = a.GetTypes();
+                    Type[] types = a.IsDynamic ? a.GetTypes(): a.GetExportedTypes();
                     foreach (Type t in types)
                     {
                         if ((t.Namespace ?? "") == nsname)
                         {
-                            names.Add(t.Name);
+                            if (!t.IsNested)
```
denfromufa: @why is this nested check needed? should there be a test for this?

src/runtime/assemblymanager.cs
denfromufa: should there be a test for this?

src/runtime/assemblymanager.cs
```diff
@@ -343,8 +353,17 @@ internal static void ScanAssembly(Assembly assembly)
...
             // the assembly.
+            Type[] types = new Type[0];
+            try
+            {
+                types = assembly.IsDynamic ? assembly.GetTypes():assembly.GetExportedTypes();
+            }
+            catch(TypeLoadException)
+            {
+                // Do nothing.
+                // This problem usually occurs when transitive dependencies have references to older packages than main application.
```
denfromufa: @dmitriyse what do you mean by transitive dependencies?

src/runtime/assemblymanager.cs
Cronan: According to the [documentation](https://msdn.microsoft.com/en-us/library/system.reflection.assembly.getexportedtypes%28v=vs.110%29.aspx?f=255&MSPPError=-2147217396), GetExportedTypes() is not supported on dynamic types and will throw *NotSupportedException*

src/runtime/assemblymanager.cs
Cronan: I may be wrong, but **ReflectionTypeLoadException**

denfromufa: @Cronan i think you are right about  TypeLoadException -> ReflectionTypeLoadException, these are 2 different exceptions and I believe this was a typo from @dmitriyse:

https://msdn.microsoft.com/en-us/library/system.typeloadexception(v=vs.110).aspx
https://msdn.microsoft.com/en-us/library/system.reflection.reflectiontypeloadexception(v=vs.110).aspx

src/runtime/assemblymanager.cs
denfromufa: @Cronan good explanation! still maybe worth adding a comment about this. especially explaining the switch from GetTypes to GetExportedTypes for normal types.

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ This document follows the conventions laid out in [Keep a CHANGELOG][].
 ## [unreleased][]
 
 ### Added
+-   Optimized implicit assembly loading on module import, PythonEngine.ImplicitAssemblyLoading event added.
```
denfromufa: @dmitriyse  please add reference to this pull request and issues from which you referred to this pull request.

src/runtime/assemblymanager.cs
denfromufa: @dmitriyse @Cronan ok, I see this comment in the header of the PR:

Assembly.GetTypes replaced to Assembly.GetExportedTypes. So we are really skipping non-public types now.
Type.IsNested types are excluded from the names collection that is used for implicit assembly loading.

src/runtime/assemblymanager.cs
denfromufa: @dmitriyse ok, I see this comment in the header of the PR:

Type.IsNested types are excluded from the names collection that is used for implicit assembly loading.

src/runtime/assemblymanager.cs
```diff
@@ -196,7 +196,17 @@ public static Assembly LoadAssembly(string name)
             Assembly assembly = null;
             try
             {
-                assembly = Assembly.Load(name);
+                var importEvent = new ImplicitAssemblyLoadingEventArgs(name);
```
denfromufa: @dmitriyse why did you augment assembly loading with this event, I cannot see the purpose of this.

denfromufa:
@dmitriyse I left some comments in this pull request. Once you explain the intent of this code, feel free to merge this one with all CI's passing tests. And please address @Cronan concern with likely incorrect exception names.

denfromufa:
@Cronan thanks for reviewing this!

@dmitriyse if you can come up with a test for nested types, that would be great.

filmor:
To be honest, I don't see the point of the whole event/event-args part of this pull request. `SkipAssemblyLoad` is never set explicitly, so I fail to see where we actually skip anything. I'll prepare a PR that only includes the `IsNested` check and `GetExportedTypes`, that should do the trick.

filmor:
Superceded by #723.

Issue #499, opened 2017-06-19
Memory leak when PythonEngine.Initialize/Shutdown is called in a loop


RomzMcGriess:
### Environment

-   Pythonnet version: 2.3.0
-   Python version: 3.6.0
-   Operating System: Win7 x64

### Details

Hello all together,
i started to use pythonnet for my application. I discovered a memory leak. I reduced my code to this to identify the leak. It appears that Python.Shutdown() does not clear all references and memory. Here is a picture of my memory diagnostics. Do i miss something? Maybe someone of you has some info.

[Memory diagnostics picture](https://picload.org/image/riwidglr/memory.png)

Here is my c# code:

        static void Main(string[] args)
        {
            for (int i = 0; i < 10000; i++)
            {
                Console.WriteLine($"Cycle: {i}");

                Console.WriteLine($"Status: {PythonEngine.IsInitialized}");

                if (!PythonEngine.IsInitialized)
                {
                    Console.WriteLine("Initialize");
                    PythonEngine.Initialize();
                }

                Console.WriteLine($"Status: {PythonEngine.IsInitialized}");

                if (PythonEngine.IsInitialized)
                {
                    Console.WriteLine("Shutdown");
                    PythonEngine.Shutdown();
                }

                Console.WriteLine($"Status: {PythonEngine.IsInitialized}");
            }

            Console.ReadLine();
        }

filmor:
Can you check this against `master`?

dmitriyse:
Flat Py_Initialize/ Py_Finalize call cylce will produce memory leak, it's directly described in the Python C Api documentation.
For example each run of all Python.EmbeddingTest, eats 5mb of memory. It's more python itself. Nothing to do with this.

If you have larger memory leak, please try this branch https://github.com/pythonnet/pythonnet/pull/532
This PR reduces memory leaks of a working Python.Engine.

PR #534, opened 2017-08-31
Fixed Second PythonEngine.Initialize call, all sensitive static variables now reseted.

dmitriyse:
This change fixes hidden bug. Once python cleaning up enough memory, objects from previous engine run becomes corrupted and produces segfault.
...

It's required but not enough to fix this issue
https://github.com/pythonnet/pythonnet/issues/499
...

### Any other comments?
Described problem never pop up, because ~PyObject currently does not called by the GC (due to some reasons). So many many time pythonnet lives without garbage collection with constant memory leak.
...

### Checklist


denfromufa:
@dmitriyse can you please resolve a conflict in this PR?

dmitriyse:
Yes, I will try to do it on this week.

filmor:
@denfromufa Could you review this again?

denfromufa:
@filmor i don't see any obvious issues.

PR #519, opened 2017-07-29
Added CoreCLR 2.0 build target. Compile issues fixed.

dmitriyse:
### What does this implement/fix? Explain your changes.
Added dotnet core 2.0 build target. All tests are passed under win, py35, x64.
...

### Does this close any currently open issues?
Work towards https://github.com/pythonnet/pythonnet/issues/96
...

### Any other comments?
After a half of a year CoreCLR 1.0 will go away, so we can only offer CoreCLR 2.0 support. 

Porting to CoreCLR 2.0 is much simpler than to CoreCLR 1.0. So we can do all the work right in the master. Xplat build fortunately can coexist side-by-side with the current build.
(See https://github.com/pythonnet/pythonnet/pull/518)

...

### Checklist

Check all those that are applicable and complete.

-   [ ] Make sure to include one or more tests for your change
-   [ ] If an enhancement PR, please create docs and at best an example
-   [ ] Add yourself to [`AUTHORS`](../blob/master/AUTHORS.md)
-   [ ] Updated the [`CHANGELOG`](../blob/master/CHANGELOG.md)

PR #538, opened 2017-09-06
Python to CLR string marshaling LRU cache.

dmitriyse:

Adds string marshaling LRU cache (10k entries per marshaling type).
This greatly speedups mixed clr/py code. + Saves big amount of memory on large structures translation.

...

### Any other comments?

This code is well tested in the high-load project. Memory usage was dramatically decreased after this improvement.

### Checklist


vmuriart:
Can you rebase this PR? looks like it was started out of a different PR that was already merged in.

dmitriyse:
This PR depends on one non merged PR https://github.com/pythonnet/pythonnet/pull/534

Also I rebased all PR that depends on merged PRs #518 and #519

denfromufa:
@jakrivan can you add your comments as inline according to the code changes in this PR:

https://github.com/pythonnet/pythonnet/pull/538/files

jakrivan:
@denfromufa - inline comment added: https://github.com/pythonnet/pythonnet/pull/625

However it's just one from many things that would need to be adjusted to support multi-domain usages. So feel free to reject - not to spoil the code with irrelevant comment.

filmor:
Note to self: This PR needs to be adjusted to the changes I made to PR #534, I renamed the `TestsSuite` class.

lostmsu:
As a performance change this needs to be accompanied by a performance test, that sets up a baseline.

E.g. first, a test must be merged, that sets the baseline performance goal without this change, then this PR should update the goal with the improved value.

filmor:
This is also probably unmergable in its current form, I have just kept it open for reference.

Issue #542, opened 2017-09-15
Incorrect error place reported in python traceback


testrunner123:
### Environment

-   Pythonnet version: 2.3.0
-   Python version: 3.4 64-bit
-   Operating System: Ubuntu 14.04

### Details

-   I tried to run test program and got following:

> GetPreferredSize
> Traceback (most recent call last):
>   File "t.py", line 29, in <module>
>     main()
>   File "t.py", line 24, in main
>     form = StatusWindow()
>   File "t.py", line 19, in __init__
>     self.Controls.Add(_StagesGroup())
> NameError: name 'xyz' is not defined

As we see, function GetPrefferedSize() got called. But stack traceback shows error in line 19 that is wrong. Bacause error is in line 14.

-   What commands did you run to trigger this issue? If you can provide a
    [Minimal, Complete, and Verifiable example](http://stackoverflow.com/help/mcve)
    this will help us understand the issue.

> python3 t.py

```python
import clr, sys
import System.Windows.Forms as WinForms
import System.Drawing as Drawing

class _StagesGroup(WinForms.GroupBox):
    __namespace__ = "System.Windows.Forms"
    def __init__(self):
        self.Text = "Stages"
        self.AutoSize = True

    def GetPreferredSize(self, proposedSize):
        print ("GetPreferredSize")
        retsz = WinForms.GroupBox.GetPreferredSize(self, proposedSize)
        print("GetPreferredSize",xyz)
        return retsz

class StatusWindow(WinForms.Form):
    def __init__(self):
        self.Controls.Add(_StagesGroup())
    def run(self):
        WinForms.Application.Run(self)

def main():
    form = StatusWindow()
    app = WinForms.Application
    app.Run(form)

if __name__ == '__main__':
    main()
```

So why the error place is incorrect in traceback?

denfromufa:
Oh, I see that this is caused by `__namespace__ = "System.Windows.Forms"`. Can you explain how `GetPreferredSize` gets called? The traceback is likely broken due to callback to CLR?

testrunner123:
>  Can you explain how GetPreferredSize gets called? 

GetPreferredSize() here is an overload for virtual method of System.Windows.Forms.GroupBox class. It is called from .NET code inside self.Controls.Add() method of System.Windows.Forms.Form class. So callstack is: Python -> .NET ->Python GetPreferredSize().

[...]

PR #545, opened 2017-09-18
Add traceback information to exception

testrunner123:

Attach traceback information to newly created exception.

#542

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,8 @@ This document follows the conventions laid out in [Keep a CHANGELOG][].
 
 ### Changed
 
+-   Reattach python exception traceback information (#545)
+
 ### Fixed
 
 -   Fixed secondary PythonEngine.Initialize call, all sensitive static variables now reseted.
--- a/src/runtime/exceptions.cs
+++ b/src/runtime/exceptions.cs
@@ -256,7 +256,7 @@ namespace Python.Runtime
             var pe = e as PythonException;
             if (pe != null)
             {
-                Runtime.PyErr_SetObject(pe.PyType, pe.PyValue);
+                Runtime.PyErr_Restore(pe.PyType, pe.PyValue, pe.PyTB);
                 return;
             }
 
--- a/src/runtime/exceptions.cs
+++ b/src/runtime/exceptions.cs
@@ -256,6 +256,9 @@ namespace Python.Runtime
             var pe = e as PythonException;
             if (pe != null)
             {
+                Runtime.XIncref(pe.PyType);
+                Runtime.XIncref(pe.PyValue);
+                Runtime.XIncref(pe.PyTB);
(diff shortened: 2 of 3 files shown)


testrunner123:
Tested with Python 3.5 64-bit on Windows 7 Pro.

denfromufa:
@testrunner123 there is one conflict in changelog.md and also no testing included.

testrunner123:
@ denfromufa

I added line to CHANGELOG.md, but I can|t understand where conflict is. It was shown that merge automatically is possible.

testrunner123:
I have no write-access to pythonnet:master.

testrunner123:
Travis build is broken because it failed to install mono 4.x packages.

denfromufa:
@testrunner123 those Mono + Travis CI issues have been fixed in @dmitriyse pull requests, which you can cherry pick into your commit.

testrunner123:
> those Mono + Travis CI issues have been fixed in @dmitriyse pull requests, which you can cherry pick into your commit.

The problem is if I try to make a pull request to my branch or master then I am unable to do it. It says: everything is update. I am using web interface.

denfromufa:
@testrunner123 you can only reliably cherry pick from git user interface, e.g. I use sourcetree and @vmuriart uses gitextensions:

https://github.com/gitextensions/gitextensions

Don't try to do it from command-line :)

Only web interface in gitlab supports cherry picking, GitHub and BitBucket do not support this.

denfromufa:
@testrunner123 this should now pass the CI

testrunner123:
THX

filmor:
This isn't quite right (it wasn't before, so you didn't introduce it but you can fix it :)). The documentation of [PyErr_Restore](https://docs.python.org/3.5/c-api/exceptions.html#c.PyErr_Restore) states that you are giving up the references that you pass on, however, the `PythonException` object that the references stem from are `Decref`d on `Dispose`. Best is probably to mark the `PythonException` instance as already disposed in this case.

testrunner123:
I guess you propose to call pe.PythonException.Restore() instead of Runtime.PyErr_Restore(). But this will also take GIL lock, that is already taken at this time ?

I just do not have enough knowledge about python interpreter to make a change and test it...
So may be you can do it right ? :)

denfromufa:
@filmor can you clarify your suggestion a little bit more for @testrunner123 and for me?

denfromufa:
@filmor i suggest to merge this PR and open a separate issue for your suggestion?

filmor:
This PR decrefs an object that we hand over to another function, I don't think it's a good idea to merge it like this. I'll see whether I can do the needed changes.

filmor:
So, from a review point of view, this is fine by me, we just need to fix the whole refcount situation, introducing another Incref before calling restore should be enough.

denfromufa:
@filmor  what is going on today :)

Issue #591, opened 2017-12-30
PythonEngine.Exec scope between classes


evandavey:
### Environment

-   Pythonnet version: 2.3.0
-   Python version: 3.6
-   Operating System:  Windows 10        

### Details

I am embedding Python using PythonEngine.Exec to run code read from a file which seems to be losing global scope.

```csharp
string code;
using (var streamReader = new StreamReader(scriptFile, Encoding.UTF8)) {
   code = streamReader.ReadToEnd();
}
 
using (Py.GIL()) {
    PythonEngine.Exec(code)
}
```
The python code contains two classes:

```python

class Test1():
   pass

class Test2():
   def __init__(self):
       Test1()

Test2()
```

Error:

```
NameError: name 'Test1' is not defined
```

yagweb:
It's a bug of the Exec method. I just submit a [PR](https://github.com/pythonnet/pythonnet/pull/594) to fix it.

Another method is using the PyScope class,
```
string code;
using (var streamReader = new StreamReader(scriptFile, Encoding.UTF8)) {
   code = streamReader.ReadToEnd();
}
 
using (Py.GIL()) {
    var scope = Py.CreateScope();
    scope.Exec(code);
    var t1 = (scope as dynamic).Test1();
    var t2 = scope.Get<dynamic>("Test2")();
}
```

evandavey:
Thanks.  I also had some success using `PythonEngine.ModuleFromString(...)`.

denfromufa:
@evandavey @yagweb thanks for this bug report and pull request!

PR #594, opened 2018-01-10
fixed bug of Exec method #591

yagweb:
This PR makes the RunString method acts just like the exec/eval method in CPython.  

In CPython, the default value of parameter 'locals' of  exec/eval method is obtained by calling the method locals() when its given value is None.  So only when the exec/eval method is called in a function, the default value of parameter 'locals' is different from the value of  'globals'. 

Since in pythonnet the Eval/Exce method can only be called from the global/module scope, it meaningful to directly set the value of locals to the value of globals when it is null.

...

[#591](https://github.com/pythonnet/pythonnet/issues/591)

...

### Any other comments?
No.

...

### Checklist

### DIFF
--- a/src/embed_tests/pyrunstring.cs
+++ b/src/embed_tests/pyrunstring.cs
@@ -57,5 +57,20 @@ namespace Python.EmbeddingTest
             object c = locals.GetItem("c").AsManagedObject(typeof(int));
             Assert.AreEqual(111, c);
         }
+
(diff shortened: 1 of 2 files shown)


denfromufa:
@yagweb looks great! let's wait for one more approval. the tests on appveyor fail for 2 environments due to github connection failure with OpenSSL. so nothing to worry about!

Issue #606, opened 2018-01-28
handle is not inialized crash on `clr.GetClrType(System.__class__)`


denfromufa:
Python 3.5, 64-bit, Windows 10, pythonnet master

```python
clr.GetClrType(System.__class__)
```

```
Unhandled Exception: System.InvalidOperationException: Handle is not initialized.
   at System.Runtime.InteropServices.GCHandle.FromIntPtr(IntPtr value)
   at Python.Runtime.ManagedType.GetManagedObject(IntPtr ob)
   at Python.Runtime.Converter.ToManagedValue(IntPtr value, Type obType, Object& result, Boolean setError)
   at Python.Runtime.Converter.ToManaged(IntPtr value, Type type, Object& result, Boolean setError)
   at Python.Runtime.MethodBinder.Bind(IntPtr inst, IntPtr args, IntPtr kw, MethodBase info, MethodInfo[] methodinfo)
   at Python.Runtime.MethodBinder.Invoke(IntPtr inst, IntPtr args, IntPtr kw, MethodBase info, MethodInfo[] methodinfo)
   at Python.Runtime.MethodObject.Invoke(IntPtr target, IntPtr args, IntPtr kw, MethodBase info)
   at Python.Runtime.MethodObject.Invoke(IntPtr inst, IntPtr args, IntPtr kw)
   at Python.Runtime.ModuleFunctionObject.tp_call(IntPtr ob, IntPtr args, IntPtr kw)
```

denfromufa:
The reason why I encountered this error is when iterating over `System` members to find all built-in structs:

```python
structs=list()
for i,_type in enumerate(dir(System)):
    print((i, _type), end=", ")
    try:
        t=getattr(System, _type)
    except Exception as e:
        print('cannot get attribute {0}'.format(_type))
    else:
        try:
            tt=clr.GetClrType(t)
        except:
            pass
        else:
            if tt.IsValueType and (not tt.IsEnum):
                print()
                print()
                print(tt)
                print()
                structs.append((i, _type, tt))
```

denfromufa:
The fixed behavior is exception instead of crashing python interpreter:

```python
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-6-ef0439a47067> in <module>()
----> 1 clr.GetClrType(System.__class__)

TypeError: No method matches given arguments
```

PR #607, opened 2018-01-28
fix for "handle is not inialized" crash on `clr.GetClrType(System.__class__)`

denfromufa:

fix for "handle is not initalized" crash on `clr.GetClrType(System.__class__)`

https://github.com/pythonnet/pythonnet/issues/606

### Checklist


denfromufa:
@dmitriyse merging according to new rules, since no objection is received

Issue #614, opened 2018-02-02
clr.AddReference doesn't paly well with path with extensions or with forward slashes


jakrivan:
Cases that will fail:

Case 1:

clr.AddReference("C:\somepath\MyAssm.dll")

Case 2:

sys.path.insert(0,"C:/some/path")
clr.AddReference("MyAssm")

I'm creating PR with fix

denfromufa:
@jakrivan are you sure about case 1? It is not escaped as a raw string literal `r""`. Also it works for me without your PR.

The second case works for me without your pull request.

So please clarify on your breaking cases.

jakrivan:
My appology - not 100% sure - I should have tried more thoroughly.

I had a failing case that I resolved by my changes. However - I needed to transfer repo to Bitbucket, so I just ported my changes here, and likely different part of my changes was resolving the issue.

The problematic case is if you have A.dll and B.dll, A depending on B. You load A from path F in syspath; then B is attempted to be loaded from GAC and current dir - there is no attempt to load from F (since there is no code stripping just filename of given resolution path and combining it with sys.path).

I'll try to find time to create proving unit test with proper fix.

Until then you can disregard this - sorry about that.

jakrivan:
Closing as I don't have simple repro proov yet

PR #615, opened 2018-02-02
Assembly load improve: Support forward slashes in sys.path, path with…

jakrivan:
… extensions

fixing #614

forward and backward slashes can be mixed in path combined from sys.path and assemblymanager code - Path.GetFullPath is needed to normalize this.
pre-existing extension doesn't need to be reappended

#614

### Any other comments?

...

### Checklist


filmor:
This doesn't seem to be reproducible, the corresponding issue has been closed, so I'll close this PR as well.

PR #628, opened 2018-02-14
Fixes lock recursion bug in assembly loading

### LINKED ISSUE #627: AssemblyList throws LockRecursionException
### Environment

-   Pythonnet version: djoyce82's [core_clr candidate](https://github.com/djoyce82/pythonnet)
-   Python version: 2.7
-   Operating System: Red Hat Linux

### Details

Use Nuget to obtain the System.Configuration.ConfigurationManager package:

`<PackageReference Include="System.Configuration.ConfigurationManager" Version="4.4.1" />`

then:

```python
    from System.Configuration import ConfigurationManager
```

The code then throws a LockRecursionException in AssemblyList.GetEnumerator(), caused by [EnterReadLock()](https://msdn.microsoft.com/en-us/library/system.threading.readerwriterlockslim.enterreadlock(v=vs.110).aspx) when another lock is already held.

This is probably caused by the fact that System.Configuration and System.Configuration.ConfigurationManager are both assemblies that need to be loaded, and there is a race condition between the **AssemblyLoadHandler**, the **ResolveHandler**, **LookupType** and **FindLoadedAssembly**.

Cronan:
### What does this implement/fix?

The AssemblyList class has a bug where assembly loading race conditions can cause a LockRecursionException under certain circumstances. Replacing the custom implementation with a `ConcurrentBag<Assembly>` solves the problem.

This closes #627

### Any other comments?

The downside of ConcurrentBag<T> is that it doesn't support accessing items via the index, and it is an unordered collection, neither of which matters in the way we are using it.

### Checklist


denfromufa:
@Cronan related PR from @ArvidJB, where he introduced AssemblyList: 

https://github.com/pythonnet/pythonnet/pull/277

denfromufa:
@Cronan I'm all for not using our own implementation of concurrent collection and use .NET types :)

Here is one case to consider:

```
             var names = new List<AssemblyName>(assemblies.Count);
```

Is it possible that new elements are added to `assemblies` concurrently before the next line with iteration starts and hence `.Count` is not correct for the size of `names` anymore?
This may sound crazy scenario, but I would hate to debug concurrency bugs.
So maybe add a lock around this?

ArvidJB:
@denfromufa that line is just a hint to pre-allocate `assemblies.Count` slots in the List and if you get that wrong the only implication is that it needs to reallocate the internal array in the List.

I think the only real difference here is that ConcurrentBag does not try to maintain any ordering of the elements you add. The ordering of the assemblies would dictate the resolution for classes which are present in multiple assemblies (not sure how frequent this actually is). Switching the order during the evaluation of the program could lead to inconsistent class definitions.
I would think that's why I wrote that custom class, but I don't really remember, to be honest.

denfromufa:
@ArvidJB so what do you recommend?

Cronan:
@ArvidJB @denfromufa [ConcurrentQueue<T>](url) would preserve the ordering, and also allow threadsafe snapshot enumeration.
The only difference with the current implementation then would be that

Issue #276, opened 2016-10-24
.NET Remoting support broken


jcarole:
Opening this issue per request from Denis Akhiyarov in PythonDotNet mailing list discussion.

Tried uploading a .zip file but got error "We don't support that file type." So here is a link to that file: 

https://www.dropbox.com/s/66pkgd1m2l3kbmh/DotNetRemotingHelloWorld.zip?dl=0

jcarole:
Forgot to copy original text from PythonDotNet mailing list discussion:

Does pythonnet-2.1.0 support .NET Remoting? When I try to get an interface, python 2.7 crashes:

Unhandled Exception: System.NullReferenceException: Object reference not set to an instance of an object.
at Python.Runtime.ClassDerivedObject.ToPython(IPythonDerivedType obj)
at Python.Runtime.Converter.ToPython(Object value, Type type)
at Python.Runtime.MethodBinder.Invoke(IntPtr inst, IntPtr args, IntPtr kw, MethodBase info, MethodInfo[] methodinfo)
at Python.Runtime.MethodObject.Invoke(IntPtr target, IntPtr args, IntPtr kw,MethodBase info)
at Python.Runtime.MethodBinding.tp_call(IntPtr ob, IntPtr args, IntPtr kw)

It's getting past these lines in my script:
import clr
clr.Addreference(my remoting DLL)
from MyNamespace import *

and choking on a static method in my app's remoting DLL call which returns the interface. Through logging, I can tell that the code in that static method call is successfully executing:
(IMyInterface)Activator.GetObject(typeof(IMyInterface), url);
and returning the object back to Python.

This worked fine before when my app targeted .NET 3.5 and I used the CLR2 version of the Python for .NET package. Now my app targets .NET 4.5.2 so I:
- Updated to Python 2.7.12
- Updated the copy of my app's remoting DLL in c:\python27\dll
- Removed the old Python for .NET package
- Executed "pip install pythonnet", which gave me version 2.1.0.

In C:\Python27\Lib\site-packages\ I see:
clr.pyd
Python.Runtime.dll v 2.0.0.2
Python.Runtime.dll.config

jcarole:
Denis's reply on PythonDotNet mailing list discussion:

Can you provide minimal reproducible code sample for both C# and Python?
What is the latest pythonnet version in which this works?

Are you testing this from interactive python or ipython? There are some
differences between the 2 options.

Please also open an issue on github.

BTW, why have you not switched to WCF?

[...]

PR #632, opened 2018-02-24
Fixed errors breaking .NET Remoting on method invoke (#276)

denfromufa:

https://github.com/pythonnet/pythonnet/issues/276

### Checklist


denfromufa:
@callumnoble @jcarole this PR for .NET Remoting cannot be merged, because more general solution that would work on .NET Core is necessary. .NET Remoting is deprecated and its API is not available in .NET Core and .NET Standard.

jcarole:
@denfromufa Just to confirm, does this mean that Pythonnet 2.0.0 will be the last version that supports .NET Remoting?

callumnoble:
For my application which needs this I'm deploying a wheel I built with 2.3 + this patch. Is working well.

I see the fails from the CI build. Would you support an `#if / #endif` conditional for building .NET with version that supports remoting. By definition this check does not need to run on .NET Core since the remoting the library uses wont have been able to run on it!

The same define as used here could work?
https://github.com/pythonnet/pythonnet/blob/df2aa64f192d986bf3cd086e53ae7e8b72fa534e/src/runtime/nativecall.cs#L26

jcarole:
@denfromufa @callumnoble Exactly. Since .NET Remoting isn't supported by .NET Core (yet, and probably ever), a conditional compilation solution seems ideal.

callumnoble:
@denfromufa opened new pr for this #637

PR #653, opened 2018-03-30
Adds method name to "no method matches" error

AlexCatarino:
When a method does not find a match for given arguments, the thrown runtime error message does not informs us which method call failed.

No.

### Any other comments?
No.

### Checklist

### DIFF
--- a/src/runtime/methodbinder.cs
+++ b/src/runtime/methodbinder.cs
@@ -507,7 +507,12 @@ namespace Python.Runtime
 
             if (binding == null)
             {
-                Exceptions.SetError(Exceptions.TypeError, "No method matches given arguments");
+                var value = "No method matches given arguments";
+                if (methodinfo != null && methodinfo.Length > 0)
+                {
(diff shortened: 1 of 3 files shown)


src/runtime/methodbinder.cs
```diff
@@ -507,7 +507,7 @@ internal virtual IntPtr Invoke(IntPtr inst, IntPtr args, IntPtr kw, MethodBase i
 
             if (binding == null)
             {
-                Exceptions.SetError(Exceptions.TypeError, "No method matches given arguments");
+                Exceptions.SetError(Exceptions.TypeError, $"No method matches given arguments for {methodinfo[0].Name}");
```
filmor: How do you know that `methodinfo` is not `null` and that it contains at least one element?

AlexCatarino: Right, I will add a check for null and at least one element.
Thank you. I didn't came across those conditions in my application.

Issue #666, opened 2018-05-05
str() failure when C# ToString() return unicode character


EcmaXp:
### Environment

-   Pythonnet version: 2.4.0.dev0 (master; #19d854c)
-   Python version: 3.6.3 (v3.6.3:2c5fed8, Oct  3 2017, 18:11:49) [MSC v.1900 64 bit (AMD64)]
-   Operating System: Windows 10 Pro 64 Bit (1803; 17134.5)

### Details

-   Describe what you were trying to get done.

1. Just call `str` python built-in function with C# class
2. Raise UnicodeDecodeError if value is `안녕`
3. Return `안녕PADD` if value is `안녕PADDINGX"

It seems like failure str convert with ToString() function only.
(Another method, nor read value does not problem)

-   What commands did you run to trigger this issue? If you can provide a
    [Minimal, Complete, and Verifiable example](http://stackoverflow.com/help/mcve)
    this will help us understand the issue.

```c#
namespace Hello
{
    public class World
    {
        public string value = "안녕";

        public string GetString()
        {
            return value;
        }

        public override string ToString()
        {
            return value;
        }
    }
}
```

```python
import clr
clr.AddReference("hello")
from Hello import World
world = World()
print(1, str(world.value))
print(2, str(world.GetString()))
print(3, str(world))
```

-   If there was a crash, please include the traceback here.

```python
1 안녕
2 안녕
Traceback (most recent call last):
  File "main.py", line 7, in <module>
    str(hello)
UnicodeDecodeError: 'utf-8' codec can't decode bytes in position 0-1: unexpected end of data
```

yagweb:
It's a bug of the PyString_FromString method in runtime.cs, refer to #662

EcmaXp:
https://github.com/EcmaXp/execlib/blob/master/gen_lib/PyExcelLibrary/Excel.cs#L14-L33

Because, `PyString_FromStringAndSize` python c api accept the bytes length, not string length.

denfromufa:
@EcmaXp pull requests are welcome!

EcmaXp:
@denfromufa I can't build this repo, how can i build?

filmor:
The fix for this has been merged to master, so I'll close this. Feel free to reopen if the version in the master branch doesn't solve the issue.

Issue #662, opened 2018-04-18
When I try to use the Chinese namespace in C#


charSLee013:
### Environment

-   Pythonnet version:2.3.0
-   Python version:3.6.3
-   Operating System:Win 10

### Details

Hi,

- When I try to use the Chinese namespace in C#,Pythonnet will prompt memory corruption.
```c#
using System;
using Python.Runtime;

namespace 中文测试
{
    public class 中文
    {
        static void Main(string[] args)
        {
            using (Py.GIL())
            {
                dynamic sys = Py.Import("sys");
                Console.WriteLine(sys.verison);
                Console.ReadKey();
            }
        }
    }
}

```
- The visual studio then prompts the following error.

![error](https://user-images.githubusercontent.com/20462328/38931366-8396b8ba-4345-11e8-9de0-ffb64e25b684.png)

- How should I solve this problem in the Chinese namespace?

THANK YOU SO MUCH!

denfromufa:
@charSLee013 this is likely not going to be supported, but let's check with @yagweb if this makes any sense.

charSLee013:
This bug is very big, which is also a hindrance to the globalization of Pythonnet. We need to solve this problem urgently and hope to get your help :)

denfromufa:
Contributions are welcome!

charSLee013:
I tried to troubleshoot the problem, when running to ` moduleobject.cs` 1194 lines , ` pyname ` returns null pointer
```c#
moduleName = "中文";
IntPtr pyname = Runtime.PyString_FromString(moduleName);
out:>>>pyname  == 0x0000000000000

IntPtr pyfilename = Runtime.PyString_FromString(filename);    //error
```
I tried to find the problem, and found it in`runtime.cs` line 1194 , and changed it to the following code.
```c#
        internal static IntPtr PyString_FromString(string value)
        {
            byte[] len = Encoding.UTF8.GetBytes(value);
            IntPtr a = PyString_FromStringAndSize(value, value.Length);
            foreach (char t in value)
            {
                if (t >= 0x4e00 && t <= 0x9fbb)
                {
                    a = PyString_FromStringAndSize(value, len.Length);
                    return a;
                }

            }
            return a;
        }

```
The exciting thing is that it does.

But in `pythonengine.cs` in line 191 appeared error, unable to load ` clr._extras ` module
```c#
try
{
IntPtr module = Runtime.PyImport_AddModule("clr._extras");
out:>>>module  == 0x00000000000000

[...]

PR #670, opened 2018-05-10
fixed bug of method PyString_FromString

yagweb:
 The second parameter of PyString_FromStringAndSize should be filled with the length of the bytes array, not the the length of the raw string.

This method is a very basic method，sufficient tests should be applied.

#666 
A unit test method need be added.

### Any other comments?
No.

### Checklist


denfromufa:
@yagweb failing in python 2.7:

https://travis-ci.org/pythonnet/pythonnet/jobs/377211034

`runtime.cs(1213,56): error CS1503: Argument 1: cannot convert from 'System.IntPtr' to 'string'`

denfromufa:
@yagweb need a fix for appveyor pip update:

https://github.com/Microsoft/vscode-python/issues/1107

denfromufa:
@yagweb will this PR fix these 3 disabled tests?

https://github.com/pythonnet/pythonnet/blob/6e4cf9d0fc4443db6f63c2f3c55e838fbda35ee0/src/embed_tests/TestPyString.cs#L38

https://github.com/pythonnet/pythonnet/blob/6e4cf9d0fc4443db6f63c2f3c55e838fbda35ee0/src/embed_tests/TestPyString.cs#L73

https://github.com/pythonnet/pythonnet/blob/6e4cf9d0fc4443db6f63c2f3c55e838fbda35ee0/src/embed_tests/TestPyAnsiString.cs#L89

yagweb:
@denfromufa No.

I also found another bug of PyString/PyAnsiString, their ToString method don't work with non-ASCII strings.

I will add all these into the checklist and dig deeper when I have time. 

Maybe they are the cause of #662.

denfromufa:
@yagweb i added the unit test based on reported bug by @EcmaXp in #666 but the issue does not look resolved in Python 2.7, works on all Python 3+ versions:

Travis CI:
`UnicodeDecodeError: 'ascii' codec can't decode byte 0xec in position 0: ordinal not in range(128)`
https://travis-ci.org/pythonnet/pythonnet/jobs/387536002#L2111

Appveyor CI:

```
assert test_unicode_str == text_type(world)
\x1b[1m\x1b[31mE       AssertionError: assert '\uc548\ub155' == '??'\x1b[0m
\x1b[1m\x1b[31mE         - \uc548\ub155\x1b[0m
E         + ??
src\tests\test_conversion.py:544: AssertionError
```
https://ci.appveyor.com/project/pythonnet/pythonnet/build/master-1182/job/nvwn4ek3eskteg3m#L220

What do you think? Maybe the test should be modified for Python 2.7?

filmor:
I don't think this can work on Py2 without breaking backwards compatibility, as we identify `System.String` with `str` while it would need to be `unicode`.

denfromufa:
@filmor agreed, so then skip the last test on PY2?

src/runtime/runtime.cs
```diff
@@ -1193,7 +1193,11 @@ internal static bool PyString_Check(IntPtr ob)
 
         internal static IntPtr PyString_FromString(string value)
         {
+#if PYTHON3
+            return PyUnicode_FromKindAndData(_UCS, value, value.Length);
+#elif PYTHON2
             return PyString_FromStringAndSize(value, value.Length);
```
filmor: Maybe check here whether the string is ASCII? I'm not sure where this one is used, but we should probably distinguish the case where we actually want Unicode (and should also return that on Python 2) from the cases that should be ASCII-only on Python 2 (like identifiers).

denfromufa: @filmor can you file a separate issue for this with a suggested fix? this PR is now ready to merge!

PR #677, opened 2018-05-27
Fix PyObject.GetHashCode

### LINKED ISSUE #676: PyObject.GetHashCode overflow bug
`PyObject.GetHashCode()` can throw an OverflowException because `IntPtr.ToInt32()` does throw. See https://github.com/pythonnet/pythonnet/blob/b8cb6c498ff48bc45643751e05eeddf1b97f8b06/src/runtime/pyobject.cs#L912. I suggest fixing this by xoring the high and low parts of that IntPtr together. Like this:

    var ulongVal = (ulong)ptr;
    return (int)((uint)ulongVal ^ (uint)(ulongVal >> 32));

This happened to me when inserting into a dictionary using tensorflow objects as keys:

    new Dictionary<dynamic, dynamic>() { { x, 1.0 }, { y, 1.0 }, }

### Environment

-   Pythonnet version: 2.3
-   Python version: 3.5
-   Operating System: Windows 7

GSPP:
Fixes #676

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -48,4 +48,5 @@
 -   ([@rmadsen-ks](https://github.com/rmadsen-ks))
 -   ([@stonebig](https://github.com/stonebig))
 -   ([@testrunner123](https://github.com/testrunner123))
+-   ([@GSPP](https://github.com/GSPP))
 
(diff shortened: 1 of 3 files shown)


filmor:
Did you check that `ulong`'s `GetHashCode` takes all bits into account? I can't find documentation on this, and apparently `IntPtr.GetHashCode` used to be broken in this regard.

GSPP:
I checked this with Reflector on .NET 4.7.2. Even if this is not working on lower versions, which I doubt for ulong, then the fix is still better than nothing. But it should be totally fine.

filmor:
Cool, thank you very much :)

PR #687, opened 2018-06-16
Some details optimization

amos402:
* Supported debug version of Python
* Fixed some reference count leak on sources and tests. I tried to balance the reference count of builtin types such as `PyTypeType`, but I failed, I still can't find where I missed.  
The code is in `TypeManager.Shutdown`. If cancel the comments, Python would crash on test, cause `AssignObject` test in `DynamicTest`. Because it set a attribute into `sys` and it wouldn't be release.
* Use the return value to checking failure for save a C call. Also make sure throw a exception even C level didn't set one.

...

### Any other comments?

* I'm trying to use pythonnet on Unity3D. On editor mode, the .net vm will restart for play. Because of Python bugs (maybe should not called bugs), Python will cause memory leak after finalize every times, and the GC will hold the old object after finalize, and visit them on next `gc.collect`. But `pythonnet` objects are already invalid after last finalize, the pointers are invliad. So it would cause a crash as follow.

To avoid the problems above, I modified the python sources temporary for my own now (reset gc linkedlist after finalize, but it will cause more memory leaks).

* I have some confuse about the implement of `Python.Runtime.Interop.GetThunk`, I don't understand why not use `Marshal.GetFunctionPointerForDelegate`.
### Checklist


denfromufa:
Unity3d support would be really interesting. I noticed some activity recently.

@amos402 tests failing on Python 2.7 in Travis CI and time-outs in appveyor CI. Python 3.6 failed only in one instance.

Note that I have not reviewed this PR yet.

amos402:
I fixed the problem that cause crash on Python 2. The reason is my modification make `clr._extras` can be release after finalize and Python 2 got a different branch in `GetCLRModule`, but the `hacked` flag is still `true`. The old code always use the first created module.
And it also exposed a hidden problem on Python 3, the `m_free` of `PyModuleDef` pointer to a undefined memory because it didn't set to 0, when module module `clr` try to release, it would crash.
Now it seems that travis-ci just failed on install package on Python 3.5 test environment. @denfromufa

denfromufa:
@amos402 GetFunctionPointerForDelegate is most likely not used due to calling convention limitation:

https://stackoverflow.com/a/19740283/2230844

src/embed_tests/pyinitialize.cs
```diff
@@ -24,8 +24,9 @@ public static void StartAndStopTwice()
         public static void LoadDefaultArgs()
         {
             using (new PythonEngine())
-            using (var argv = new PyList(Runtime.Runtime.PySys_GetObject("argv")))
             {
+                // PySys_GetObject get a borrowed reference, it just no need for dispose
+                var argv = new PyList(Runtime.Runtime.PySys_GetObject("argv"));
```
filmor: Good catch 👍

src/runtime/classmanager.cs
```diff
@@ -34,6 +34,15 @@ static ClassManager()
             dtype = typeof(MulticastDelegate);
         }
 
+        internal static void Shutdown()
+        {
+            foreach (var item in cache.Values)
+            {
+                Runtime.Py_DecRef(item.pyHandle);
```
filmor: Again, use `XDecref`

src/runtime/importhook.cs
```diff
@@ -39,11 +39,25 @@ internal static void Initialize()
             IntPtr mod = Runtime.IsPython3
                 ? Runtime.PyImport_ImportModule("builtins")
                 : Runtime.PyDict_GetItemString(dict, "__builtin__");
+            if (mod == IntPtr.Zero)
+            {
+                throw new PythonException();
+            }
```
filmor: Has this ever happened for you? In any case, the call should be changed to `PyEval_GetBuiltins()`, while you're at it.

src/runtime/importhook.cs
```diff
@@ -39,11 +39,25 @@ internal static void Initialize()
...
+
             hook = new MethodWrapper(typeof(ImportHook), "__import__", "TernaryFunc");
             Runtime.PyObject_SetAttrString(mod, "__import__", hook.ptr);
-            Runtime.XDecref(hook.ptr);
+
+            if (Runtime.IsPython3)
+            {
+                // On Python3, PyImport_ImportModule got a new reference, so we need to decrease it.
+                Runtime.Py_DecRef(mod);
```
filmor: Not needed with `PyEval_GetBuiltins`, always returns a borrowed reference.

src/runtime/importhook.cs
```diff
@@ -142,7 +166,7 @@ public static IntPtr GetCLRModule(IntPtr? fromList = null)
                     }
                 }
             }
-            Runtime.XIncref(py_clr_module);
+            Runtime.Py_IncRef(py_clr_module);
```
filmor: Why did you change this?

src/runtime/interop.cs
```diff
@@ -227,7 +227,7 @@ public static IntPtr AllocModuleDef(string modulename)
             byte[] ascii = Encoding.ASCII.GetBytes(modulename);
             int size = name + ascii.Length + 1;
             IntPtr ptr = Marshal.AllocHGlobal(size);
-            for (int i = 0; i < m_free; i += IntPtr.Size)
+            for (int i = 0; i <= m_free; i += IntPtr.Size)
```
filmor: Very good, yes :)

src/runtime/methodwrapper.cs
```diff
@@ -27,9 +28,29 @@ public MethodWrapper(Type type, string name, string funcType = null)
...
             return Runtime.PyCFunction_Call(ptr, args, kw);
         }
+
+        public void Dispose()
+        {
+            if (_disposed)
+            {
+                return;
+            }
+            _disposed = true;
```
filmor: You are calling `Dispose` from the finalizer which may execute without the GIL, so you probably have to wrap this in a `using (Py.GIL())`

src/runtime/pyint.cs
```diff
@@ -51,7 +51,10 @@ public PyInt(PyObject o)
         public PyInt(int value)
         {
             obj = Runtime.PyInt_FromInt32(value);
-            Runtime.CheckExceptionOccurred();
+            if (obj == IntPtr.Zero)
+            {
+                throw new PythonException();
+            }
```
filmor: Why? `Runtime.CheckExceptionOccurred()` is shorter and equivalent, according to the Python documentation.

src/runtime/pythonengine.cs
```diff
@@ -7,11 +7,18 @@
 
 namespace Python.Runtime
 {
+    public class ResolveStreamResouceArgs : EventArgs
```
filmor: Typo, "Resou*r*ce".

src/runtime/pythonengine.cs
```diff
@@ -520,6 +540,29 @@ internal static PyObject RunString(string code, IntPtr? globals, IntPtr? locals,
...
+                    if (stream != null)
+                    {
+                        return stream;
+                    }
+                }
+            }
+            Assembly assembly = Assembly.GetExecutingAssembly();
+            stream = assembly.GetManifestResourceStream("clr.py");
+            return stream;
+        }
```
filmor: I don't quite understand this, what issue does this resolve? Do you need to override `clr.py`? Why? This should probably be in a separate PR for more discussion.

src/runtime/pytuple.cs
```diff
@@ -68,12 +71,22 @@ public PyTuple(PyObject[] items)
...
                 IntPtr ptr = items[i].obj;
                 Runtime.XIncref(ptr);
-                Runtime.PyTuple_SetItem(obj, i, ptr);
-                Runtime.CheckExceptionOccurred();
+                if (Runtime.PyTuple_SetItem(obj, i, ptr) != 0)
+                {
+                    for (int j = i; j >= 0; j--)
+                    {
+                        Runtime.XDecref(items[j].obj);
+                    }
```
filmor: We should rather dec-ref `obj` I think, which should drag the already set elements along with it, shouldn't we?

src/runtime/runtime.cs
```diff
@@ -582,6 +611,7 @@ internal static unsafe long Refcount(IntPtr op)
             {
                 return 0;
             }
+            p = (byte*)p + TypeOffset.ob_refcnt;
```
filmor: This is safer, yes :)

src/runtime/typemanager.cs
```diff
@@ -14,13 +14,29 @@ internal class TypeManager
...
+            //foreach (var item in cache.Values)
+            //{
+            //    Runtime.Py_DecRef(item);
+            //}
+            cache.Clear();
+            //foreach (IntPtr tp in rawtypes)
+            //{
+            //    Runtime.Py_DecRef(tp);
+            //}
+            rawtypes.Clear();
```
filmor: Why did you comment this out?

src/runtime/typemanager.cs
```diff
@@ -537,5 +556,10 @@ internal static void CopySlot(IntPtr from, IntPtr to, int offset)
             IntPtr fp = Marshal.ReadIntPtr(from, offset);
             Marshal.WriteIntPtr(to, offset, fp);
         }
+
+        internal static void AddRawType(IntPtr tp)
+        {
+            rawtypes.Add(tp);
+        }
```
filmor: What's the plan here?

amos402:
Well... I just try getting a overview for this project at that moment and make some fixes passingly.
And I saw `CheckExceptionOccurred` just take more costs than a simple condition with return value, so I change them passingly. emm.... It doesn't matter a lot for now actually.
`pythonnet` is too slow because of its reflection mechanism of  binding, calling method..., converting and etc. Also it needs JIT for Python to inherit a .Net class(I can't use JIT because I need to use it on IL2CPP on Unity).
I already made a new branch rewrite these, emphasis on performance improvement and AOT mode support. I'm not certainly it can be merge into the official branch in future or not. It will change a lot but I will try to to make compatible with the official branch. It just can make some simple interactive calls now, I still need lots of work to do with it. 

https://github.com/amos402/pythonnet/tree/fast-call
You can preview it on above branch. I didn't clean up the code yet, so there are many useless  codes and garbage on it...
The code generator for static binding is in another repo that I didn't push yet, so you can't use the `NativeBinderManager` now.

src/embed_tests/pyimport.cs
```diff
@@ -40,6 +40,7 @@ public void SetUp()
             IntPtr str = Runtime.Runtime.PyString_FromString(testPath);
             IntPtr path = Runtime.Runtime.PySys_GetObject("path");
             Runtime.Runtime.PyList_Append(path, str);
+            Runtime.Runtime.Py_DecRef(str);
```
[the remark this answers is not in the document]
amos402: if `str` is NULL, it should be a error, we shouldn't ignore it.

src/runtime/importhook.cs
amos402: It should 100% be not null

src/runtime/importhook.cs
amos402: `PyEval_GetBuiltins` don't have __import__ certainly
default

src/runtime/methodwrapper.cs
amos402: `Dispose` will only call from `Shutdown` now, so I ignore it.
It will be modify after finalizer branch.
default

src/embed_tests/pyimport.cs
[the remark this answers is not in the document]
filmor: If you really want to check for that, do so after `PyString_FromString` and before it is appended to `sys.path`.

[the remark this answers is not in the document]
amos402: I don't need to check that, 'append' will

src/runtime/importhook.cs
filmor: That's odd, then `PyEval_GetBuiltins()` doesn't return `__builtin__` in Python 2. Sorry for the noise.

amos402:
Oh, I was wrong. I just found that `Py_IncrRef` is same as `Py_XINCREF`, I was always use `Py_INCREF` macro in C/C++ that make me didn't realize that,
Maybe we should add a new function for validating the `op`. @filmor

zhangyl08:
hi, has this compatibility merged to git master?

amos402:
@zhangyl08 I think it's not for now, since master has mush changes. Which part are you concerning about?

lostmsu:
I am not sure of the performance effect of that, but all pointers received from unmanaged code should be wrapped in [`SafeHandle`](https://stackoverflow.com/questions/155780/safehandle-in-c-sharp) to ensure proper disposal on domain unload and other exceptional cases. Also, using it instead of `IntPtr` ensures the lack of memory leaks.

zhangyl08:
Ive a mature project written in C# with Unity, but the server side is in Python, So I am trying to look for a pythonnet performance test see if it can replace parts of my C# code.

amos402:
> 
> 
> I am not sure of the performance effect of that, but all pointers received from unmanaged code should be wrapped in [`SafeHandle`](https://stackoverflow.com/questions/155780/safehandle-in-c-sharp) to ensure proper disposal on domain unload and other exceptional cases. Also, using it instead of `IntPtr` ensures the lack of memory leaks.

The main performance impact is calling `Runtime.CheckErrorOccurred()` due to it call a c function wrapper, the `Runtime.Py_Dec/IncRef` did not improve the either. As I said, maybe we just need a new interface for validating the refcount and decrease it.
Well, let's skip the reference count API, I should separate this into some several PRs cause some changes is about the bugs.

amos402:
@zhangyl08 I think you won't pass performance test for master now since reflection has too much costs.

Skip this unrelated message as follow,
详情也许我们可以走一波私信😂

denfromufa:
@amos402 this discussion maybe of interest to you - future of pythonnet development:

https://mail.python.org/pipermail/pythondotnet/2019-June/thread.html

Thanks,
Denis

filmor:
Closed as outdated.

Issue #691, opened 2018-06-18
Memory Leak on MethodBinding for generic method


williamsardar:
### Environment

-   Pythonnet version: 2.3.0
-   Python version: 3.6-32
-   Operating System: Windows 10

### Details

-   Getting a `MethodBinding` for a generic method results in the reference count being too high, so the `tp_dealloc` method on the `MethodBinding` never gets called, even when the `MethodBinding` goes out of scope.

```csharp
namespace PlainOldNamespace
{
    public class PlainOldClass
    {
        public void NonGenericMethod() { }

        public void GenericMethod<T>() { }
    }
}
```

```python
import clr, sys
clr.AddReference("PlainOldAssembly")
from PlainOldNamespace import PlainOldClass
poc = PlainOldClass()
sys.getrefcount(poc.NonGenericMethod)
# 1
sys.getrefcount(poc.GenericMethod[clr.System.Double])
# 2 - should be 1
```

I think this could be down to a bug in MethodBinding.cs. This line looks suspicious. Why should the reference count get incremented here?
https://github.com/pythonnet/pythonnet/blob/master/src/runtime/methodbinding.cs#L59

williamsardar:
I am happy to do a fix for this - would that be the fastest way to get this fixed?

denfromufa:
@williamsardar I'm not sure yet about this problem, can you provide a unit test where something fails because of this? There are 2 places like this in the codebase:

https://github.com/pythonnet/pythonnet/search?q=Runtime.XIncref%28mb.pyHandle%29&unscoped_q=Runtime.XIncref%28mb.pyHandle%29

williamsardar:
There don't appear to be any unit tests that will show this problem. I've seen there is the leaktest.py file. Not clear that this code gets run on the CI server as it contains syntax errors. Using the approach in leaktest.py (create and del an object 50,000 times), and adding a 1MB allocation to the PlainOldClass type shows that I get an OutOfMemory exception before making the fix, and don't get one after making the fix. Please could you let me know how likely it is to get this fix incorporated and a new version released? Happy to help where I can..



```csharp
namespace PlainOldNamespace
{
    public class PlainOldClass
    {
        private readonly byte[] payload = new byte[(int)Math.Pow(2, 20)]; //1 MB
        public void GenericMethod<T>() { }
    }
}
```



```python
import clr, sys
clr.AddReference("PlainOldAssembly")
from PlainOldNamespace import PlainOldClass

for i in range(50000):
[...]

PR #704, opened 2018-07-20
Fix memory leaks due to spurious handle increment and add unit tests

williamsardar:
Fix memory leaks due to spurious handle increment and add unit tests
...

Issue #691
...

### Any other comments?

...

### Checklist

-   [ x ] Make sure to include one or more tests for your change
-   [ x ] If an enhancement PR, please create docs and at best an example
-   [ x ] Add yourself to [`AUTHORS`](../blob/master/AUTHORS.md)
-   [ x ] Updated the [`CHANGELOG`](../blob/master/CHANGELOG.md)

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -42,6 +42,7 @@
 -   Ville M. Vainio ([@vivainio](https://github.com/vivainio))
 -   Virgil Dupras ([@hsoft](https://github.com/hsoft))
 -   Wenguang Yang ([@yagweb](https://github.com/yagweb))
+-   William Sardar ([@williamsardar])(https://github.com/williamsardar)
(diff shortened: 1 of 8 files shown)


williamsardar:
Hi @denfromufa. How can I add the psutil dependency to the CI system?

filmor:
There are probably more issues like this in the code, essentially every class derived from `ExtensionType` does not need an additional `Incref` as far as I can see, since the `PyType_GenericAlloc` call in the constructor will already lead to a refcount of 1.

filmor:
Thank you very much for your contribution :)

williamsardar:
@filmor thanks for the merge. Any indication as to when the next release might happen?

Issue #703, opened 2018-07-12
Very Slow Python Initialize


williamrubens:
### Environment

Pythonnet version: 2.3.0
Python version: Python 3.5 (conda environment)
Operating System: Windows 10 64bit
VS 2017

### Details

Hi guys, thanks for the great work building a super useful library. One question, and this could be expected behavior, I'm not sure. When I initialize the python interpreter on the .Net side, with this call:

`using (Py.GIL())
{`

I've timed it and it takes about 2 mins to complete the call. Further calls to grab python lock return immediately. I've also noticed that in visual studio my output window I see thousands of messages of this form:

> Exception thrown at 0x00007FFE37F73C58 in RSUser.exe: Microsoft C++ exception: EEFileLoadException at memory location 0x000000314CBFB8F0.
> Exception thrown at 0x00007FFE37F73C58 in RSUser.exe: Microsoft C++ exception: [rethrow] at memory location 0x0000000000000000.
> Exception thrown at 0x00007FFE37F73C58 in RSUser.exe: Microsoft C++ exception: EEFileLoadException at memory location 0x000000314CBFB8F0.
> Exception thrown: 'System.IO.FileNotFoundException' in mscorlib.dll
> Exception thrown at 0x00007FFE37F73C58 in RSUser.exe: Microsoft C++ exception: EEFileLoadException at memory location 0x000000314CBFB8F0.
> Exception thrown at 0x00007FFE37F73C58 in RSUser.exe: Microsoft C++ exception: [rethrow] at memory location 0x0000000000000000.

filmor:
I have never observed this, can you disclose your test sample?

denfromufa:
Did you mean that initialize takes 2 minutes under VS debugger? If yes,
then this could be due to VS or PTVS settings.

williamrubens:
I switched to running under release and it's much faster, takes around 40 seconds now, though I can still thousands of exceptions in the output window. Sounds like it could be some VS settings... any ideas which ones to play with?

inna-w:
This issue was discussed before: [#598](https://github.com/pythonnet/pythonnet/issues/598).
The solution proposed by @dmitriyse there worked for me - we are using build from [featured](https://github.com/pythonnet/pythonnet/tree/featured) branch. This also solved crashes on obfuscated types names in 3rd party assemblies.

williamrubens:
[...]

Issue #643, opened 2018-03-09
macOS: AddReference find the reference, but the module is not found


AlexCatarino:
### Environment

-   Pythonnet version: Latest
-   Python version: 3.6
-   Operating System: macOS

### Details
We have compiled `Python.Runtime.dll` for python 3.6 in Windows for each OS: Windows, macOS and Linux. Then we create a nuget package to distribute the library (along with `nPython.exe`) to be used with [Lean](https://github.com/QuantConnect/Lean). The following commands were used:

```
msbuild pythonnet.sln /nologo /v:quiet /t:Clean;Rebuild /p:Platform=x64 /p:PythonInteropFile="interop36.cs" /p:Configuration=ReleaseWin /p:DefineConstants="PYTHON36,PYTHON3,UCS2"
msbuild pythonnet.sln /nologo /v:quiet /t:Clean;Rebuild /p:Platform=x64 /p:PythonInteropFile="interop36.cs" /p:Configuration=ReleaseMono /p:DefineConstants="PYTHON36,PYTHON3,UCS2,MONO_OSX"
msbuild pythonnet.sln /nologo /v:quiet /t:Clean;Rebuild /p:Platform=x64 /p:PythonInteropFile="interop36.cs" /p:Configuration=ReleaseMono /p:DefineConstants="PYTHON36,PYTHON3,UCS4,MONO_LINUX"
```
When execute nPython.exe in Windows and try to import System, I can do it successfully:
```terminal
C:\Users\Alex\Lean\Launcher\bin\Debug>nPython.exe
Python 3.6.4 (v3.6.4:d48eceb, Dec 19 2017, 06:54:40) [MSC v.1900 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> from clr import AddReference
>>> ref = AddReference("System")
>>> ref.FullName
'System, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089'
>>> import System
>>>
```
In Linux:
```terminal
root@52578947ccd3:~/Lean/Launcher/bin/Debug# mono nPython.exe
Python 3.6.4 |Anaconda, Inc.| (default, Jan 16 2018, 18:12:19)
[GCC 7.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from clr import AddReference
>>> ref = AddReference("System")
>>> ref.FullName
[...]

PR #622, opened 2018-02-07
LookupType returns the first type found

Cronan:
### What does this implement/fix? Explain your changes.
In .Net Core 2.0, there are public and private versions of the same types in different assemblies. For example, there is a private version of System.Environment in mscorlib.dll, and a public version in System.Runtime.Extensions.dll. The existing code returns the first matching type found, even if it's private, and even if there are other matching types in other assemblies. 

The new version of this function returns the first public type, or the first type found.
### Does this close any currently open issues?

This closes some parts of #613 

### Checklist

Check all those that are applicable and complete.

-   [x] Make sure to include one or more tests for your change
-   [ ] If an enhancement PR, please create docs and at best an example
-   [x] Add yourself to [`AUTHORS`](../blob/master/AUTHORS.md)
-   [x] Updated the [`CHANGELOG`](../blob/master/CHANGELOG.md)

Issue #603, opened 2018-01-23
Infinite loop/crash in root.LoadNames()


barrybriggs:
### Environment

-   Pythonnet version: 2.3
-   Python version: 3.6
-   Operating System: Windows

### Details

(Pythonnet newbie). Trying to embed Pythonnet in a simple WPF app. Built Python.Runtime from source setting PYTHON3 and PYTHON36 on x64. Gets to GetCLRModule in importhook.cs and gets in infinite loop in root.LoadNames() line 98, ultimately crashes.

denfromufa:
@barrybriggs how can someone reproduce this issue? Did you look at troubleshooting and installation wikis? Does this happen without WPF (in console)?

barrybriggs:
Thanks for the quick response. The code is very, very simple -- click on a WPF menu and the click handler calls `PythonEngine.Initialize()`. When I trace into `root.LoadNames()`  there is the following code:

```      public void LoadNames()
        {
            ManagedType m = null;
            foreach (string name in AssemblyManager.GetNames(_namespace))
            {
                cache.TryGetValue(name, out m);
                if (m == null)
                {
                    ManagedType attr = GetAttribute(name, true);
                }
            }
        }
```

The 'names' that it loads are: "FXAssembly", "ThisAssembly", "AssemblyRef", "<PrivateImplementationDetails>", and then it starts trying to load what seems like types "_StaticArrayInitTypeSize=6" and then 8, and so on. It does this forever. 

I have a feeling this may be because there is some required assembly or .py file or something like that it's looking for, but I can't find it.

denfromufa:
Do the demo apps work for you? Your example is still not reproducible.

barrybriggs:
`npython` loads. However, the demo apps do not run (assuming I typed the command correctly, which, as a Python newbie, I may not have): 
```
>>> exec(open("Z:\demo\helloform.py").read())
C:\Windows\Microsoft.NET\Framework64\v4.0.30319\mscorlib.dll
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "<string>", line 8, in <module>
ModuleNotFoundError: No module named 'System'
```
`PATH` and `PYTHONHOME` are set correctly.

dmitriyse:
Please try "featured" branch. At least problems with "_StaticArrayInitTypeSize=6" types are eliminated.

barrybriggs:
@dmitriyse That seems to have cured the problem. Thanks!!

[...]

PR #723, opened 2018-08-21
Use GetExportedTypes where possible and filter nested types

filmor:
This is basically what @dmitriyse prepared in PR 528 without the event code that I don't understand.

Useless calls to load dependencies based on private and nested types.

#703 

### Any other comments?

...

### Checklist


filmor:
@denfromufa Any idea for a test?

denfromufa:
@filmor i don't have a good option for a test (requires multiple assemblies), unless we ask to test out this PR from all users @goerch @pierce-martin @williamrubens @inna-w @Cronan @barrybriggs @ZhukovGreen @AlexCatarino who reported this in the bug tracker:

#703 
#643
#622 
#603 
#598 
#528 
https://github.com/MSLNZ/msl-loadlib/issues/6

denfromufa:
@filmor i think this PR still misses this essential part about handling exceptions from "transitive" dependencies, which I experienced before (I'm going to find a link), see this comment from @Cronan:

https://github.com/pythonnet/pythonnet/pull/528/files#r167013554

src/runtime/assemblymanager.cs
```diff
@@ -461,5 +458,13 @@ public static Type LookupType(string qname)
             }
             return null;
         }
+
+        internal static Type[] GetTypes(Assembly a)
```
denfromufa: need a try-catch block here for transitive depedencies

denfromufa:
@filmor i agree with your comment about skiploadassembly: https://github.com/pythonnet/pythonnet/pull/528#issuecomment-414668945

src/runtime/assemblymanager.cs
filmor: Isn't it an error, though, if we can't load transitive dependencies?

filmor:
@denfromufa Can you please provide me with a test-case for the transitive dependency issue you mention? I have really no idea what this is about.

src/runtime/assemblymanager.cs
```diff
@@ -470,9 +471,28 @@ public static Type LookupType(string qname)
...
+                try
+                {
+                    return a.GetTypes();
+                }
+                catch (ReflectionTypeLoadException exc)
+                {
+                    // Return all types that were successfully loaded
+                    return exc.Types.Where(x => x != null).ToArray();
+                }
+            }
```
filmor: @denfromufa Is this what you had in mind?

denfromufa: @filmor this is even more than what i asked for :)

PR #724, opened 2018-08-21
Rebased PR 450 (Py_Initialize without overriding signals)

filmor:
@rmadsen-ks 

Avoiding overriding signals from the application so that CTRL-C might be handled.

Using Py_InitializeEx(0) (instead Py_Initialize) to avoid overriding signals.

#449 #450 

### Any other comments?

Just a rebased version of #450.

### Checklist


denfromufa:
@filmor why is this failing CI?

filmor:
Something about `CSharpCodeProvider`, I'll look into it. It's the same error that @rmadsen-ks had already when he pushed the PR, I think.

rmadsen-ks:
@filmor, the only change in this code is optionally calling InitializeEx instead of Initialize. This code is often being touched by tests. The actual change here is really simple, the unit test is excessively complex. Could you consider merging this without a unittest testing CTRL+C explicitly?

filmor:
I think you are right. I'll merge the PR without the unit test using a merge commit, that way the test is at least present in the history.

PR #725, opened 2018-08-24
Uni 56832 another crash fix attempt

benoithudson:
On a domain reload, some objects are leaked. When we reinitialize python, those leaked objects are still being tracked by gc. This means their types' tp_traverse, tp_clear, and tp_is_gc slots can get invoked. But since those are implemented in C#, and we unloaded the domain, those slots are pointing to garbage.

Our previous attempt was to unload the cpython library itself and reload it. This works on Windows, but does not work on linux and osx: any native modules that have been loaded depend on cpython, so those operating systems refuse to unload cpython.

This fix instead makes the slots point to functions that are still valid after reloading, but that have no side effects and return the same values (0 for tp_traverse and tp_clear; non-zero for tp_is_gc).

The function that returns zero is PyAST_Check. That means there's a latent bug if someone were to use C# to define a subtype of ast.AST (the python abstract syntax tree). We'd have to document that.

Very minor change embedded in this change: switch from Hashtable to HashSet<string> for cleaner code (and perhaps some speedup).

TODO:
- update authors and changelog
- update hotReloadCrash and put it in the EmbeddingTests


benoithudson:
Blah, sorry -- still trying to work out all the kinks on our fork.

Issue #714, opened 2018-08-09
Exception thrown when initializing Python .NET after a domain unload


lassond:
### Environment

-   Pythonnet version: latest master commit as of 2018-08-09 (fb84cd28254f1ed6bde82766dadd6ba18c3e7be1 )

-   Python version: 
Python 2.7.15 (v2.7.15:ca079a3ea3, Apr 30 2018, 16:30:26) [MSC v.1500 64 bit (AMD64)] on win32

-   Operating System:
Windows 10

- In the context of running in Unity, but also reproducible in a simple standalone app made from scratch

### Details

Unity, whenever a C# script is changed, the AppDomain in which the Assemblies are loaded gets unloaded. Then, Unity creates a fresh AppDomain, recompiles the C# assemblies and loads them again.

The Python.Runtime assembly gets unloaded, but the Python interpreter survives (it is a native library). This results in an inconsistent state between the Python .NET Engine/Runtime classes (which are not initialized) and the Python interpreter (which is initialized). 

An exception is raised when Python .NET accesses the interpreter (System.AppDomainUnloadedException: Attempted to access an unloaded AppDomain.)

-   Describe what you were trying to get done.

See above

-   What commands did you run to trigger this issue? If you can provide a
    [Minimal, Complete, and Verifiable example](http://stackoverflow.com/help/mcve)
    this will help us understand the issue.

I will attach my standalone application which reproduces this crash along with the repro steps. (either ere or on the pull request if I cannot attach here)

lassond:
[hotReloadCrashRepro.zip](https://github.com/pythonnet/pythonnet/files/2275521/hotReloadCrashRepro.zip)


(Note: with the upcoming fix, the application will still raise an exception, but on the second domain unload - a bit further. A fix for this second crash is under review)

benoithudson:
This is fixed and the test case integrated in #752. But I don't have permissions to close the bug.

filmor:
But I do, thanks again :)

dyeo:
> By the way, I'm working on a new binding mode for Python call that supported IL2CPP and with more higher performance, this pythonnet will a part of my unity package.
> https://github.com/amos402/pythonnet/tree/fast-call

@amos402 did you ever make any headway with the new binding mode that enables IL2CPP support?

PR #745, opened 2018-10-02
Fixes to integrate pythonnet into Unity

benoithudson:

This patch implements a fix to two successive crashes when integrating pythonnet into Unity3d, see #714

(1) On domain reload, shut down pythonnet. Otherwise there'll be a bunch of objects in python that point to functions implemented in C# code that isn't there anymore. In particular, the trick of replacing the import function means that on an import statement in the second domain, python is calling into deallocated memory.

Having fixed that crash we see a second crash, fixed by:
(2) Implement the garbage collection support in native code in memory that is not deallocated. Python's garbage collector keeps pointers to python objects that were allocated in the first domain and leaked on Py_Finalize. Python doesn't clear that list in Py_Initialize. This means it will try to collect these objects in the second domain. This led to a crash because garbage collection support for the python objects was implemented in C#, so on the first gc in the second domain, python was calling into deallocated memory.

Support for number 2:
(3) Use python's platform module to determine the platform at runtime so we know what native code to build and how to allocate an executable page. I'm preferring to check at runtime rather than at compile time so we can get away from shipping a .net DLL per configuration.

(4) Unit tests for domain reload, the platform module, and for allocating a page of memory.

Support is currently windows, linux, darwin; i386 and x86_64. Unit tests will fail with a descriptive message on other platforms.

#714 .

### Any other comments?

It will be nice to fix a lot of the leaks, which I see some action on. But we can't rely on there being no leaks: client C# code can always XIncref some of their own stuff one too many times (or client C++ code). Best not to have everything come crashing down with inexplicable errors if client code made that mistake.

It seems like the only reason we really need the native code page hack is that we use the size field for special magic stuff in metatype.cs:
```
            // Hmm - the standard subtype_traverse, clear look at ob_size to
            // do things, so to allow gc to work correctly we need to move
            // our hidden handle out of ob_size. Then, in theory we can
            // comment this out and still not crash.
```
If we can fix that, we can avoid leaking a page per domain reload, and it'll be easier to port to other platforms.

### Checklist

-   [ x ] Make sure to include one or more tests for your change
-   [ n/a ] If an enhancement PR, please create docs and at best an example
-   [ x ] Add yourself to [`AUTHORS`](../blob/master/AUTHORS.md)
-   [ x ] Updated the [`CHANGELOG`](../blob/master/CHANGELOG.md)

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -14,6 +14,7 @@
 
 -   Alexandre Catarino([@AlexCatarino](https://github.com/AlexCatarino))
 -   Arvid JB ([@ArvidJB](https://github.com/ArvidJB))
+-   Benoît Hudson ([@benoithudson](https://github.com/benoithudson))
 -   Bradley Friedman ([@leith-bartrich](https://github.com/leith-bartrich))
 -   Callum Noble ([@callumnoble](https://github.com/callumnoble))
 -   Christian Heimes ([@tiran](https://github.com/tiran))
@@ -22,6 +23,7 @@
 -   Daniel Fernandez ([@fdanny](https://github.com/fdanny))
 -   Daniel Santana ([@dgsantana](https://github.com/dgsantana))
 -   Dave Hirschfeld ([@dhirschfeld](https://github.com/dhirschfeld))
+-   David Lassonde ([@lassond](https://github.com/lassond))
 -   David Lechner ([@dlech](https://github.com/dlech))
 -   Dmitriy Se ([@dmitriyse](https://github.com/dmitriyse))
 -   He-chien Tsai ([@t3476](https://github.com/t3476))
@@ -39,6 +41,7 @@
 -   Sam Winstanley ([@swinstanley](https://github.com/swinstanley))
 -   Sean Freitag ([@cowboygneox](https://github.com/cowboygneox))
 -   Serge Weinstock ([@sweinst](https://github.com/sweinst))
+-   Viktoria Kovescses ([@vkovec](https://github.com/vkovec))
 -   Ville M. Vainio ([@vivainio](https://github.com/vivainio))
 -   Virgil Dupras ([@hsoft](https://github.com/hsoft))
 -   Wenguang Yang ([@yagweb](https://github.com/yagweb))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,8 @@ This document follows the conventions laid out in [Keep a CHANGELOG][].
 ### Fixed
 
 -   Fixed Visual Studio 2017 compat ([#434][i434]) for setup.py
+-   Fixed crashes when integrating pythonnet in Unity3d ([#714][i714]),
+    related to unloading the Application Domain
 -   Fixed crash on exit of the Python interpreter if a python class
     derived from a .NET class has a `__namespace__` or `__assembly__`
     attribute ([#481][i481])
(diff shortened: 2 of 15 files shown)


benoithudson:
Travis uses netstandard rather than 4.0 so it won't compile the unit test for domain reloading (everything else should go fine).

Suggestion on how best to proceed? Disabling the unit test on travis seems sad, but netstandard doesn't seem to fully support application domains... which is what I'm testing.

denfromufa:
@benoithudson does AppDomain parts that you use work in Mono? By "netstandard" I presume you tested it in .NET Core? I'm ok with skipping otherwise.

What is relation of this code to Unity Technologies?

Note that at least 2 core developers should review and approve before this can be merged.

benoithudson:
This works in Mono, indeed -- initial dev was on Darwin. I then ported the fix to windows using the microsoft toolchain, and to linux using mono.

I didn't test on .NET Core myself, but that's what continuous integration is using. CI is failing to compile the test case (and the test case would fail trying to create domains even if it did compile). I've just pushed up disabling the test case. It's just unfortunate that this test case will have to rely on humans.

Unity funded my team at Imaginary Spaces to integrate pythonnet for an internal project and to ship the results to the public by posting a PR to pythonnet, so anyone handy enough can use python in Unity. If you need anything from their legal I can ask them to provide it.

filmor:
Seeing in the CPython code that this apparently a known problem (essentially https://github.com/python/cpython/blob/5903296045b586b9cd1fce0b1e02caf896028d1d/Python/pylifecycle.c#L84-L89), I don't think we have another choice but doing something hacky.

Did you check if it was feasible to iterate over all still unreleased .NET objects and just overwriting their `tp_traverse` etc. slots with 0?

benoithudson:
That crashes. I didn't investigate whether it was calling 0 or calling a default function, which then failed on metatype only.

python3 is crashing in a new and interesting way to me; I initially developed on python3/net40/mac but apparently the netcore path has some important differences.

benoithudson:
I'll need help debugging those remaining CI errors.

Appveyor fails the python34 tests on `pip install --upgrade -r requirements.txt --quiet` which is before pythonnet even gets its boots on. It passes all other test cases.

Travis is having a parse error on the pythonnet.sln file for the non-xplat tests. The solution file hasn't changed since January. The xplat tests pass.

Is there a way to check if the master get a green light right now, or whether some new version of something killed it? The most recent green-light test run seems to be July.

benoithudson:
I get the same error running the commands from he travis CI manually on the master branch.

Has something changed in the version of packages we download since the last PR merge in July?

benoithudson:
Hallelujah!

I can make a separate PR for the CI fixes if that's helpful.

Travis is fixed by 14bc2e2. Use msbuild 14 rather than 15, now that travis has 15 available.

Appveyor is "fixed" by making python 3.4 an allowed-fail in ec6f6c5 and 386a034. The issue is that psutil is a source distribution on py34 and there's some error compiling it.

filmor:
Awesome, great work! If you can split this out into a separate, squashed PR, I'll merge both. @denfromufa, fine with you as well?

denfromufa:
Let me review tomorrow

benoithudson:
OK, I've created a PR #751 just for the travis/appveyor.

Once that's merged, I'll squash the domain reload work into a nice form.

benoithudson:
I'll close this PR, and create a new one with the squashed code.

PR #751, opened 2018-10-12
Fix travis, work around appveyor build breaks

benoithudson:
Continuous integration on pythonnet/master is broken at the moment. This patch gets to green-light on master.

(1) Travis now has msbuild 15 available, which breaks the no-arg build since it's using the pythonnet.sln which is msbuild 14. Fixed by forcing the use of msbuild 14.

(2) Appveyor on py34 is failing to build psutil. Newer python and py27 are using binary releases so the problem doesn't show up. I couldn't figure out how to fix it, so I made it an allowed failure.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -28,6 +28,12 @@ environment:
     - PYTHON_VERSION: 3.5
     - PYTHON_VERSION: 3.6
 
+matrix:
+    allow_failures:
+      - PYTHON_VERSION: 3.4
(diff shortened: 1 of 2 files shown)


benoithudson:
@filmor, @denfromufa -- this patch is ready for merge, and then I can squash #745.

filmor:
There you go, thank you very much :)

denfromufa:
i approve this!

PR #752, opened 2018-10-12
Enable pythonnet to survive in the Unity3d editor environment.

benoithudson:
The Unity editor unloads the C# domain and creates a new domain every time it recompiles C# sources.

This caused two crashes.
1. After a domain unload, python is now pointing to a bunch of objects whose C# side is now garbage. **Solution**: Shutdown() the engine on domain reload, which calls Py_Finalize.

2. After a domain unload, Py_Finalize, and a new Py_Intialize, python still keeps pointers in gc for any python objects that were leaked. And pythonnet leaks. This means that python's gc will be calling tp_traverse, tp_clear, and tp_is_gc on types that are implemented in C#. This crashes because the implementation was freed. **Solution**: implement those calls in code that is *not* released on domain unload. Side effect: we leak a page on every domain unload. I suspect but didn't test that python gc performance will be slightly faster.

Changes required to implement and test:
3. Use python's platform package to determine what we're running on, so we can use the right mmap/mprotect or VirtualAlloc/VirtualProtect routines, the right flags for mmap, and (theoretically) the right native code. Side effect: to port to a new platform requires updating some additional code now.

4. New unit tests. (a) A new test for domain reload. It doesn't run under .NET Core because you can't create and unload a domain in .NET Core. (b) New unit tests to help us find where to add support for new platforms.

#714 ; see also #745 and #751.

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -14,6 +14,7 @@
 
 -   Alexandre Catarino([@AlexCatarino](https://github.com/AlexCatarino))
 -   Arvid JB ([@ArvidJB](https://github.com/ArvidJB))
+-   Benoît Hudson ([@benoithudson](https://github.com/benoithudson))
 -   Bradley Friedman ([@leith-bartrich](https://github.com/leith-bartrich))
 -   Callum Noble ([@callumnoble](https://github.com/callumnoble))
 -   Christian Heimes ([@tiran](https://github.com/tiran))
@@ -22,6 +23,7 @@
 -   Daniel Fernandez ([@fdanny](https://github.com/fdanny))
 -   Daniel Santana ([@dgsantana](https://github.com/dgsantana))
 -   Dave Hirschfeld ([@dhirschfeld](https://github.com/dhirschfeld))
+-   David Lassonde ([@lassond](https://github.com/lassond))
 -   David Lechner ([@dlech](https://github.com/dlech))
 -   Dmitriy Se ([@dmitriyse](https://github.com/dmitriyse))
 -   He-chien Tsai ([@t3476](https://github.com/t3476))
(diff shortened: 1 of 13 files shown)


benoithudson:
Now we see if I flubbed the merge...

benoithudson:
> @benoithudson just to be clear this enables pythonnet survival in Unity3D editor even when AppDomains are not available on that platform? Is AppDomain only used for testing?

Exactly the other way: Unity uses app domains, loads and unloads them frequently. And that caused a crash.

CI runs tests in a setting without app domains (and also with app domains). So the patch works either way.

filmor:
@denfromufa Good for merge?

denfromufa:
I have not finished reviewing yet

denfromufa:
@benoithudson I'm mostly done reviewing the code, this is monumental achievement!

One thing that bothers me is if the bytes/assembly code can be replaced with a call to a compiled C code as part of pythonnet? @filmor @amos402  what do you think?

the native dll can even be embedded as a resource:

https://stackoverflow.com/a/768429/2230844

filmor:
The beauty of this solution is that it does not need all of the boilerplate that you have when using a proper DLL or .so. The code for the two functions is OS independent, it only depends on the CPU architecture. I'm okay with what is here, I would just like us to look into alternative solutions that don't leak the memory page (or at the very least only leak it once).

amos402:
It seems it's complex to do that.
To prevent Unity crash, only just need to reset the gc.

https://github.com/amos402/cpython/commit/3fb76b5fafc1fb87b408b751d25e80ad1b40b2b7
I made a modify in cpython before like above, it works well. I already use this for mouths.
I just made up a pure C# version by trick for reseting th GC. #754 
I still havn't test it fully for this new version.

filmor:
@benoithudson Can you check whether @amos402 solution works for you as well before we merge?

filmor:
I've triggered the build for #754 again, it seems to not be quite sound with regard to memory handling (though that doesn't necessarily have to be due to this PR). I'm currently leaning towards merging this one (#752) first such that we have a chance of preparing a 2.4.0 release before end of the year and afterwards look into #754 again.

@denfromufa @benoithudson @amos402 do you agree?

benoithudson:
@filmor : I'm in favour of your plan. I won't be sad if my code gets buried as soon as we have a better alternative.

amos402:
Yes. I still need more tests for fixing that PR.

denfromufa:
👍

filmor:
Thanks a lot for your and your colleagues efforts :)

Issue #552, opened 2017-10-04
Method precedence


AlexCatarino:
### Environment

-   Pythonnet version: 2.4.0
-   Python version: 2.7
-   Operating System: Windows 10

### Details

We have a series of overloads for a given method in existing base class and we want to add new overloads that should take precedence. To be more precise, we have the following case:
`public IEnumerable<Slice> History(int periods)`
that is consumed by C# classes and we have 
`public PyObject History(int periods)`
that is consumed by Python classes
To achieve it, we have created a subclass where the second overload is defined:
```
public class QCAlgorithm
{
    public IEnumerable<Slice> History(int periods) 
    {
    }
}
```
```
public class QCPyAlgorithm : QCAlgorithm
{
    public new PyObject History(int periods) 
    {
    }
}
```
When we use reflection to get the methods from QCPyAlgorithm, the new method comes first. However, after the sorting performed by pythonnet [here](https://github.com/pythonnet/pythonnet/blob/master/src/runtime/methodbinder.cs#L162), the method from QCAlgorithm comes first. Consequently it is called in my python script and returns an enumerable of Slice.

**The issue/question is:** 
How do I keep the methods from the declaring type on the top of the list of methods are try to bind?

My first guess is to send the first overload behind by adding the following line in the [GetPrecedence](https://github.com/pythonnet/pythonnet/blob/master/src/runtime/methodbinder.cs#L183) method: 
```
val += mi.DeclaringType == mi.ReflectedType ? 0 : 3000;
```

denfromufa:
Can you explicitly use .Overloads on your methods to enforce the correct
overload?

AlexCatarino:
In our particular case ([QuantConnect/Lean](https://github.com/QuantConnect/Lean/)), we want users coding in C# and Python to use the exact same method call. We don't want them to learn all the overloads or use a different method signature.

filmor:
You can fix the method sorting by not trying to do it via a mere `.Sort` call but instead by building that list yourself, starting with the reflected type and going upwards in the inheritance tree, appending the methods (sorted using the existing `MethodSorter`) as you go.

I usually don't expose the .NET objects directly to the user and instead wrap it in a more pythonic library, essentially using Python.NET as an FFI, that's another option.

denfromufa:
https://github.com/pythonnet/pythonnet/issues/552

christabella:
[...]

PR #756, opened 2018-10-22
Inheritance not working with non-abstract base methods

### LINKED ISSUE #755: Inheritance not working with non-abstract base methods
### Environment

-   Pythonnet version: 2.3.0
-   Python version: 3.6.3

### Details

    Trying to use a C# DLL which has a class that overrides a non-abstract method of its parent class. Although the new method should be used, pythonnet still calls the parent method. The code for a minimal DLL to reproduce this problem looks like this:

```c#
using System;

namespace InheritanceTest
{
    public class BaseClass
    {
        public bool Transmit()
        {
            throw new NotImplementedException();
        }
    }

    public class InheritedClass: BaseClass
    {
        public new bool Transmit()
        {
            Console.WriteLine("Success!");
            return true;
        }
    }
}
```

In the following python code, `BaseClass.Transmit` should be called when `BaseClass` is directly instantiated and `InheritedClass.Transmit` when `InheritedClass` is instantiated. However, `BaseClass.Transmit` is called both times.

```python
## setup
import clr
import os

clr.AddReference(os.getcwd() + '\\InheritanceTest.dll')
import InheritanceTest

## method test

base_class = InheritanceTest.BaseClass()
base_class.Transmit() # throws a NotImplementedException as expected

inherited_class = InheritanceTest.InheritedClass()
inherited_class.Transmit() # still throws a NotImplementedException, although it should call InheritedClass.Transmit
```

smourier:
Fixes #755. Also check https://stackoverflow.com/questions/52868552/why-does-python-net-use-the-base-method-instead-of-the-method-from-a-derived-cla

### DIFF
--- a/src/runtime/methodbinder.cs
+++ b/src/runtime/methodbinder.cs
@@ -601,6 +601,19 @@ namespace Python.Runtime
     {
         int IComparer.Compare(object m1, object m2)
         {
+            var me1 = (MethodBase)m1;
+            var me2 = (MethodBase)m2;
+            if (me1.DeclaringType != me2.DeclaringType)
+            {
+                // m2's type derives from m1's type, favor m2
+                if (me1.DeclaringType.IsAssignableFrom(me2.DeclaringType))
+                    return 1;
+
+                // m1's type derives from m2's type, favor m1
+                if (me2.DeclaringType.IsAssignableFrom(me1.DeclaringType))
+                    return -1;
+            }
+
             int p1 = MethodBinder.GetPrecedence((MethodBase)m1);
             int p2 = MethodBinder.GetPrecedence((MethodBase)m2);
             if (p1 < p2)


denfromufa:
@smourier what should be the logic if the types are not equal and they don't derive from one another?

Not sure what is going on with both CI, first time this happens! I just tried to restart some builds, but looks like the problem is on the github side.

smourier:
I choose to modify the existing Comparer to minimize regression risks since I don't know this project much, and I don't know python much either :-).

Like I said, the whole Bind code that chooses what method should be taken is quite fragile (also I don't really understand the funky GetPrecedence algorithm?) and cannot support all different possibilities, it's not only a matter of sorting the methods.

If you want to support all the different types of methods that the .NET framework supports (including generics), you probably need to add some options that the caller could use on the python side, IMHO

PS: yes, github seems in trouble today!

denfromufa:
github is still recovering from network/database failure:

https://blog.github.com/2018-10-21-october21-incident-report/

denfromufa:
All existing tests are passing, but this new code needs a unit test.

denfromufa:
@AlexCatarino this looks like a fix for your issue https://github.com/pythonnet/pythonnet/issues/552?

AlexCatarino:
Yes, @denfromufa , thank you @smourier!

filmor:
@smourier Thanks a lot for this patch, from my point of view this is good to go, but please, if possible

1. Add yourself to the AUTHORS file
2. Add a test
3. Add the fix to the Changelog

If you don't have the time for these, say the word and I'll complete the PR s.t. it can go into the next release :)

smourier:
Well, originally, I was just answering a question on stackoverflow, so I let you guys the details complete what needs to be :)

PR #757, opened 2018-10-30
Allow user code to register for a callback on shutdown.

benoithudson:

New feature: allow user code to clean up before python.net shuts down. We use it in a client-server architecture, with python.net on the server and a pyside UI on the client. This callback lets us clean up gracefully.

I decided to also use the feature in order to help PythonEngine itself set up / shut down correctly, because dog food is tasty. And there's a unit test.

Not on pythonnet.

### Any other comments?

I'm not clear on the division of labour between PythonEngine and Runtime. User code tells the PythonEngine to initialize, not the runtime, so that seemed the place to put the shutdown handler.

But it could just as well be in the runtime, and the runtime also needs the same pattern of building up / tearing down in the opposite order. User code could still think it's PythonEngine handling it (just make PythonEngine.AddShutdownHandler call Runtime.AddShutdownHandler).

Thoughts?

### Checklist

===> where? On the wiki?

### DIFF
--- a/src/runtime/pythonengine.cs
+++ b/src/runtime/pythonengine.cs
@@ -168,6 +168,16 @@ namespace Python.Runtime
                 initialized = true;
                 Exceptions.Clear();
 
+                // Make sure we clean up properly on app domain unload.
+                AppDomain.CurrentDomain.DomainUnload += OnDomainUnload;
+
+                // Remember to shut down the runtime.
+                AddShutdownHandler(Runtime.Shutdown);
+
+                // The global scope gets used implicitly quite early on, remember
+                // to clear it out when we shut down.
+                AddShutdownHandler(PyScopeManager.Global.Clear);
+
                 if (setSysArgv)
                 {
                     Py.SetArgv(args);
@@ -220,9 +230,6 @@ namespace Python.Runtime
                 {
                     locals.Dispose();
                 }
-
-                // Make sure we clean up properly on app domain unload.
-                AppDomain.CurrentDomain.DomainUnload += OnDomainUnload;
             }
         }
 
@@ -302,7 +309,12 @@ namespace Python.Runtime
         {
             if (initialized)
             {
-                PyScopeManager.Global.Clear();
+                // If the shutdown handlers trigger a domain unload,
+                // don't call shutdown again.
+                AppDomain.CurrentDomain.DomainUnload -= OnDomainUnload;
+
+                ExecuteShutdownHandlers();
+
                 Marshal.FreeHGlobal(_pythonHome);
                 _pythonHome = IntPtr.Zero;
                 Marshal.FreeHGlobal(_programName);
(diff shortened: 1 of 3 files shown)


src/runtime/pythonengine.cs
```diff
@@ -302,21 +309,86 @@ public static void Shutdown()
...
+                AppDomain.CurrentDomain.DomainUnload -= OnDomainUnload;
+
+                ExecuteShutdownHandlers();
+
                 Marshal.FreeHGlobal(_pythonHome);
                 _pythonHome = IntPtr.Zero;
                 Marshal.FreeHGlobal(_programName);
                 _programName = IntPtr.Zero;
                 Marshal.FreeHGlobal(_pythonPath);
                 _pythonPath = IntPtr.Zero;
```
benoithudson: These strings are a bit weird. We're supposed not to free these ever, because they can be used even when python is not initialized. I decided that at the very least, we shouldn't free them until *after* shutting down the engine.

In python 3.7, these strings are no longer necessary: SetPythonHome and friends now copy their argument (https://github.com/python/cpython/commit/af5a895073c24637c094772b27526b94a12ec897#diff-87aed37b4704d4e1513be6378c9c7fe6) -- though that's not actually documented.

benoithudson:
Making travis run again...

benoithudson:
Deadlock race in the travis-ci? It timed out on 3.7 the first time, 3.6 the second time.

src/runtime/pythonengine.cs
denfromufa: @benoithudson  good finding, so you are cleaning up these special strings to avoid possible runtime leftovers between python re-initializations?

denfromufa:
@benoithudson i see the same flaky issue with embedding tests getting deadlocked (restarted all 3.6-3.7 builds)

src/runtime/pythonengine.cs
benoithudson: In this PR I changed the order to be that we shut down python before freeing the strings, but I don't plan to change anything else.

Probably nobody will ever really notice the problem: people who set the strings will set them again next time we initialize, people who don't, won't ever set them. And in python 3.7, the problem goes away entirely. I'll create an issue, in case someone else feels motivated.

benoithudson:
OK, finally the tests didn't flake out. This is ready from my perspective, just squash it all. Once in I can do the wiki thing.

denfromufa:
@benoithudson i don't get the point of the reverse order for shutdown handlers, can you give a more realistic example instead of arithmetic operations?

filmor:
Think Subsystem `A` depends on Subsystem `Base`. On initialisation of `A`, `Base` must be already initialised.

During the respective initialisations, a shutdown handler will be registered.

No, if you undo this, you want `Base` to shut down after `A` (s.t. `Base` "runs" at least as long as `A`), so you have to reverse the order of the handlers.

Issue #720, opened 2018-08-15
Support for Python 3.7


iarspider:
Installation with Python 3.7 fails because there is no `interop37.cs`

MisterY:
What is the solution for this? 
I have pythonnet-2.4.0.dev0-cp37 that I'm using at the moment but still getting

![image](https://user-images.githubusercontent.com/462445/47842332-ad8ba380-ddbc-11e8-859a-17234f41d4c0.png)

so I assume this is still valid?

denfromufa:
3 steps to install latest pythonnet with python3.7 support:

```
git clone https://github.com/pythonnet/pythonnet
cd pythonnet
python setup.py install
```

Note that I'm pushing a fix when only VS 2017 is installed on machine:

https://github.com/pythonnet/pythonnet/pull/758

Somehow Appveyor CI still works without this fix and you can also find the wheels from these builds in the Appveyor artifacts.

denfromufa:
@MisterY you probably installed pybee from pypi and it depends on python 3.7, but pythonnet does not have python3.7 wheel on pypi yet.

pallavi1292:
Pythonnet works for Python 3.7 now?

Issue #609, opened 2018-01-28
python 3.7 issues


denfromufa:
pycparser error:

https://travis-ci.org/pythonnet/pythonnet/jobs/334432678#L1991

`pycparser.plyparser.ParseError: /usr/include/crypt.h:33:6: before: __THROW`

jonemo:
I am seeing the same error when attempting to build Docker images for https://github.com/jonemo/pythonnet-docker for all combinations that involve Python 3.6.5:

* Python 3.6.5, Mono 4.8.0.524, Python.NET 2.3.0
* Python 3.6.5, Mono 4.8.0.524, Python.NET master branch
* Python 3.6.5, Mono 5.2.0.224, Python.NET master branch
* Python 3.6.5, Mono 5.4.1.6, Python.NET master branch

Python 3.5.5 and 3.4.8 are not affected.

Here is the output I get for the command `pip install pythonnet==2.3.0` (almost identical to what's behind the Travis link above):

```
Step 3/3 : RUN pip install pycparser   && pip install pythonnet==2.3.0
 ---> Running in 306362b2258a
Collecting pycparser
  Downloading pycparser-2.18.tar.gz (245kB)
Building wheels for collected packages: pycparser
  Running setup.py bdist_wheel for pycparser: started
  Running setup.py bdist_wheel for pycparser: finished with status 'done'
  Stored in directory: /root/.cache/pip/wheels/95/14/9a/5e7b9024459d2a6600aaa64e0ba485325aff7a9ac7489db1b6
Successfully built pycparser
Installing collected packages: pycparser
Successfully installed pycparser-2.18
Collecting pythonnet==2.3.0
  Downloading pythonnet-2.3.0.tar.gz (1.5MB)
Building wheels for collected packages: pythonnet
  Running setup.py bdist_wheel for pythonnet: started
  Running setup.py bdist_wheel for pythonnet: finished with status 'error'
  Complete output from command /usr/local/bin/python -u -c "import setuptools, tokenize;__file__='/tmp/pip-build-zrnh3win/pythonnet/setup.py';f=getattr(tokenize, 'open', open)(__file__);code=f.read().replace('\r\n', '\n');f.close();exec(compile(code, __file__, 'exec'))" bdist_wheel -d /tmp/tmp9rlc9nv2pip-wheel- --python-tag cp36:
  running bdist_wheel
  running build
  running build_ext
  Checking for updates from https://www.nuget.org/api/v2/.
[...]

Issue #366, opened 2017-02-08
consider updating from unmanaged exports to dllexport friendly fork


denfromufa:
https://github.com/3F/DllExport

Note that this may resolve one issue for us:

https://github.com/3F/DllExport/issues/7

vmuriart:
I tried it yesterday, albeit for all of 5 minutes, and it didn't work as plug and play. It required some configuration to setup correctly i guess (and i gave up). 

It probably needs some fiddling around to get right, but I agree, if we can move to it we should since its oss.

filmor:
This is not relevant anymore.

PR #759, opened 2018-11-01
adding iterop37.cs to Python.Runtime.csproj and removing old TODO com…

paulie4:
…ment, see #609, #720, and https://github.com/pythonnet/pythonnet/issues/609#issuecomment-433637591

`iterop37.cs` was missing from `Python.Runtime.csproj`

no, #609 and #720 were already closed, although, I don't know how #720 closed without this change...

### Any other comments?
NOTE: to be able to build a usable `clr.pyd` using Visual Studio 2017, we had to run the following:
```
mklink /j "C:\Program Files (x86)\Microsoft Visual Studio\2017\Enterprise\VC\bin" "C:\Program Files (x86)\Microsoft Visual Studio\2017\Enterprise\VC\Tools\MSVC\14.15.26726\bin\Hostx86\x86"
```
This is because we got the following error message in the build's output (even though the project looks like it built correctly), i.e. since the [UnmanagedExports package](https://www.nuget.org/packages/UnmanagedExports) seems to be looking for `lib.exe` in the wrong place for VS2017, we put the files where it was expecting to find them:
```
2>RGieseckeDllExport:
2>  Cannot find lib.exe in 'C:\Program Files (x86)\Microsoft Visual Studio\2017\Enterprise\Common7\IDE\\..\..\VC\bin'.
2>  Found method: clrModule..method public hidebysig static void  'initclr'() cil managed
2>  	exporting as initclr and index 0
```

### DIFF
--- a/src/runtime/Python.Runtime.csproj
+++ b/src/runtime/Python.Runtime.csproj
@@ -148,6 +148,7 @@
     <Compile Include="interop34.cs" />
     <Compile Include="interop35.cs" />
     <Compile Include="interop36.cs" />
+    <Compile Include="interop37.cs" />
   </ItemGroup>
   <ItemGroup>
     <None Include="..\pythonnet.snk" />
--- a/src/runtime/runtime.cs
+++ b/src/runtime/runtime.cs
@@ -149,7 +149,7 @@ namespace Python.Runtime
 #elif PYTHON36
         internal const string _pyversion = "3.6";
         internal const string _pyver = "36";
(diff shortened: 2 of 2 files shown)


denfromufa:
@paulie4 I don't see the same issue in latest VS 2017 Enterprise (15.8.5) . 

But still we need to transition from unmanaged exports to a friendly fork dllexport:

https://github.com/pythonnet/pythonnet/issues/366

It builds fine using setup.py too.

Issue #588, opened 2017-12-12
Tests fail in Linux: AttributeError: Xml


Cronan:
### Environment

-   Pythonnet version: 2.4.0
-   Python version: 2.7.5
-   Operating System: Red Hat Enterprise Linux Server release 7.3 (Maipo)
-   DotNet version: 2.0.0
### Details

-   Tests fail in Linux: 
-- src/tests/test_compat.py::test_module_get_attr

`dotnet ./build/lib.linux-x86_64-2.7/netcoreapp2.0/nPython.dll`

```python
>>> import sys
>>> sys.path.append('/users/isys/icronyn/pyenvs/pythonnet/lib/python2.7/site-packages/pytest-2.9.1-py2.7.egg')
>>> sys.path.append('/users/isys/icronyn/pyenvs/pythonnet/lib/python2.7/site-packages/py-1.4.31-py2.7.egg')
>>> import pytest
>>> pytest.main(['-v', 'src/tests/test_compat.py::test_module_get_attr'])
============================================================================== test session starts ==============================================================================
platform linux2 -- Python 2.7.5, pytest-2.9.1, py-1.4.31, pluggy-0.3.1 -- /users/isys/icronyn/pyenvs/pythonnet/bin/python
cachedir: .cache
Arch: x64, UCS: 4, LIBDIR: /usr/lib64, Py_ENABLE_SHARED: True
rootdir: /users/isys/icronyn/git/pythonnet, inifile: 
collected 17 items 

src/tests/test_compat.py::test_module_get_attr FAILED

=================================================================================== FAILURES ====================================================================================
_____________________________________________________________________________ test_module_get_attr ______________________________________________________________________________

    def test_module_get_attr():
        """Test module getattr behavior."""
        import CLR.System as System
    
        int_type = System.Int32
        assert is_clr_class(int_type)
    
>       module = System.Xml
E       AttributeError: Xml

src/tests/test_compat.py:228: AttributeError
=========================================================================== 1 failed in 0.03 seconds ============================================================================
```

dmitriyse:
More general problem is now tracked here #590
This hell is going due to using Mono compiled Python.Test.dll under .net core 2.0

PR #769, opened 2018-11-09
Fix: do not cp System.XML on case-insensitive fs

### LINKED ISSUE #768: cp ...System.XML.dll fails on mac
### Environment

-   Pythonnet version: master https://github.com/pythonnet/pythonnet/commit/f8e147b97112c526c6548e900dee0fc41cdb39b1
-   Python version: 3.7.0
-   Operating System: macOS 10.14.1

### Details

When compiling pythonnet from source on macOS, I noticed that this line:
https://github.com/pythonnet/pythonnet/blob/f8e147b97112c526c6548e900dee0fc41cdb39b1/src/runtime/Python.Runtime.15.csproj#L134

causes the build to fail:
```
       "/.../pythonnet/pythonnet.15.sln" (Build target) (1) ->
       "/.../pythonnet/src/runtime/Python.Runtime.15.csproj" (default target) (2) ->
       "/.../pythonnet/src/runtime/Python.Runtime.15.csproj" (Build target) (2:3) ->
       (BeforeBuild target) ->
         /.../pythonnet/src/runtime/Python.Runtime.15.csproj(134,5): error MSB3073: The command "cp /Users/ztl8702/.nuget/packages//microsoft.targetingpack.netframework.v4.5/1.0.1/lib/net45/System.XML.dll /Users/ztl8702/.nuget/packages//microsoft.targetingpack.netframework.v4.5/1.0.1/lib/net45/System.Xml.dll" exited with code 1.

    5 Warning(s)
    1 Error(s)
```

And when I ran:
```bash
cp /Users/ztl8702/.nuget/packages//microsoft.targetingpack.netframework.v4.5/1.0.1/lib/net45/System.XML.dll /Users/ztl8702/.nuget/packages//microsoft.targetingpack.netframework.v4.5/1.0.1/lib/net45/System.Xml.dll
```

in the Terminal, I got:
```
cp: .../net45/System.Xml.dll and .../net45/System.XML.dll are identical (not copied).
```

So it appears to me that macOS by default has [a case-insensitive filesystem](https://developer.apple.com/library/archive/documentation/FileManagement/Conceptual/APFS_Guide/FAQ/FAQ.html)

> APFS, like HFS+, is case-sensitive on iOS and is available in case-sensitive and case-insensitive variants on macOS, with **case-insensitive being the default**.
 
My hack was to check for existence of `System.Xml.dll` before `cp`. Maybe there is a more elegant way. (I actually wonder why the `cp` command is there in the first place)

ztl8702:

On case-insensitive UNIX-like systems (e.g. macOS), copying `System.XML.dll` to `System.Xml.dll` will cause the build to fail. 

The fix (or hack) is to check for existence of `System.Xml.dll` before running `cp`.

Fixes #768 

### Checklist

### DIFF
--- a/src/runtime/Python.Runtime.15.csproj
+++ b/src/runtime/Python.Runtime.15.csproj
@@ -131,7 +131,7 @@
 
   <Target Name="BeforeBuild" Condition="'$(TargetFramework)'=='net40' AND $(Configuration.Contains('Mono')) AND '$(OS)' != 'Windows_NT'">
     <!--Endless war!-->
-    <Exec Command="cp $(NuGetPackageRoot)/microsoft.targetingpack.netframework.v4.5/1.0.1/lib/net45/System.XML.dll $(NuGetPackageRoot)/microsoft.targetingpack.netframework.v4.5/1.0.1/lib/net45/System.Xml.dll" />
+    <Exec Command="[[ -e $(NuGetPackageRoot)/microsoft.targetingpack.netframework.v4.5/1.0.1/lib/net45/System.Xml.dll ]] || cp $(NuGetPackageRoot)/microsoft.targetingpack.netframework.v4.5/1.0.1/lib/net45/System.XML.dll $(NuGetPackageRoot)/microsoft.targetingpack.netframework.v4.5/1.0.1/lib/net45/System.Xml.dll" />
   </Target>
   <Target Name="AfterBuild">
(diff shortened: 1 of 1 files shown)


denfromufa:
@ztl8702 cp for System.Xml was added probably due to this issue? https://github.com/pythonnet/pythonnet/issues/588

Issue #770, opened 2018-11-09
pythonnet converting ObservableCollection into python list


fdv1:
### Environment

-   Pythonnet version: head
-   Python version: any
-   Operating System: any

### Details

-   https://github.com/pythonnet/pythonnet/commit/4df6105b98b302029e524c7ce36f7b3cb18f7040 slightly changed Pythonnet behaviour in that it automatically convert any list implementing Ilist into a python list. This conversion is certainly consistent with pythonnet but it does hide some functionalities from python in the the case of more advanced lists such as ObservableCollection<T>. This is preventing some WPF objects from being usable from python (eg: AvalonEdit). I will submit a proposal to fix this issue in a PR.

PR #771, opened 2018-11-09
Fix ObservableCollection<T> conversion to python list.

fdv1:
### Don't convert .Net objects implementing Ilist to python list if they also implement INotifyPropertyChanged.

...

### Fix proposal for https://github.com/pythonnet/pythonnet/issues/770

...

### Any other comments?

Not sure how to write a test for this...


filmor:
This doesn't look like the right branch to merge :)

Issue #514, opened 2017-07-20
Implicit List conversion breaks backwards compatibility + numpy support


BenjaminPelletier:
### Environment

Computer 1:
-   Pythonnet version: 2.3, installed via pip I think
-   Python version: 2.7.12
-   Operating System: Ubuntu 16.04.2

Computer 2:
-   Pythonnet version: manual build on master from commit ce14424 (currently 11 commits behind)
-   Python version: 2.7.12
-   Operating System: Ubuntu 16.04.2

### Details

Description: Python calls .NET function which returns a List<MyType>.  Python then passes the return value, without modification, to a second .NET function which accepts a List<MyType>.  Computer 1 executes this code just fine.  On Computer 2, there is no .NET function found that matches the arguments because the return value of the .NET function has been transformed into a Python list.

Python code:
```
import clr
from MyDotNetProject import PythonInterop

x = PythonInterop.GetDoubleList()
PythonInterop.PrintDoubleList(x)
```

.NET code:
```
    public class PythonInterop
    {
        public static List<Double> GetDoubleList() {
            var result = new List<Double>();
            result.Add(1);
            result.Add(2);
            result.Add(3);
            return result;
        }

        public static void PrintDoubleList(List<Double> list) {
            Console.WriteLine("You list has " + list.Count + " elements");
        }
    }
```

The Python code works fine on Computer 1.  On Computer 2, the PrintDoubleList call produces
TypeError: No method matches given arguments

If I print type(x) in Python, Computer 1 gives me a .NET List type while Computer 2 gives me a Python list.  I can print x.Count on Computer 1, but I get a missing attribute error on Computer 2.

If I build manually from the 2.3 tag, I get the same (good) behavior as on Computer 1.

It seems that some feature has been partially added to automatically convert .NET objects into Python objects when possible.  I suppose this is ok (though I would prefer that not happen because I don't want the mandatory performance hit of converting even when I don't want to convert), but if that's the intention, there *must* be automatic conversion of Python objects to .NET objects also.  One without the other is a critical bug.

denfromufa:
Is this the issue your are experiencing?

https://github.com/pythonnet/pythonnet/issues/451

BenjaminPelletier:
As far as I can tell, this isn't like #451 because the error in #451 occurs in .NET.  The error in this issue occurs in Python.  But, the same underlying problem may be causing both issues.  Given the .NET code above, here is a different demonstration of the poor behavior:

```
import clr
[...]

PR #773, opened 2018-11-11
=Avoid converting ObservableCollection<T> into python list

fdv1:
Avoid converting objects implementing INotifyPropertyChanged (eg:ObservableCollection<T>) into python list. This would fix https://github.com/pythonnet/pythonnet/issues/770

### DIFF
--- a/src/runtime/converter.cs
+++ b/src/runtime/converter.cs
@@ -5,6 +5,7 @@ using System.Globalization;
 using System.Reflection;
 using System.Runtime.InteropServices;
 using System.Security;
+using System.ComponentModel;
 
 namespace Python.Runtime
 {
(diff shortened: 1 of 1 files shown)


denfromufa:
i think as a temp fix this is ok, but we are planning to revert this list autoconversion completely once a viable alternative is decided.

denfromufa:
here is the original issue:

https://github.com/pythonnet/pythonnet/issues/514

fdv1:
Great! sorry for the mess with the featured branch...

filmor:
@fdv1 I guess you accidentally closed this one, is this correct?

jnsquire:
This is still broken for some use cases. See, for example:

https://github.com/icsharpcode/SharpDevelop/blob/master/src/AddIns/DisplayBindings/AvalonEdit.AddIn/Src/IconBarManager.cs

Where Bookmarks is returned as an IList and thus loses the ability to mutate the underlying ObservableCollection.

PR #777, opened 2018-11-15
Minimal Py_IntializeEx change

### LINKED ISSUE #449: Call to Py_Initialize changes signal handler configuration

filmor:

Minimal version of #724, allows the user to select whether an embedded Python should configure signal handlers or not, defaulting to "false" (which is different from what we had before as a straight call to `Py_Initialize`  will configure signal handlers).

Fixes #449.

### Any other comments?

We should follow up on #724 to get a unit test working for this, but this can wait until after 2.4.0.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@ This document follows the conventions laid out in [Keep a CHANGELOG][].
 ### Changed
 
 -   Reattach python exception traceback information (#545)
+-   PythonEngine.Intialize will now call `Py_InitializeEx` with a default value of 0, so signals will not be configured by default on embedding. This is different from the previous behaviour, where `Py_Initialize` was called instead, which sets initSigs to 1. ([#449][i449])
 
 ### Fixed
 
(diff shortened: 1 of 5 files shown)


denfromufa:
@filmor why did you switch both x86 & x64 to AnyCPU?

filmor:
Argh, sorry, I wanted to exclude this part from the branch :/ I'll revert that particular change.

denfromufa:
@filmor py27 got stuck in appveyor for master branch. I restarted the build.

PR #779, opened 2018-11-17
clr.FindAssembly() method,throw exception when file not found.

jiftle:

clr.FindAssembly() method,throw exception when file not found.

maybe

### Any other comments?

no

### Checklist

### DIFF
--- a/src/runtime/moduleobject.cs
+++ b/src/runtime/moduleobject.cs
@@ -451,7 +451,12 @@ namespace Python.Runtime
         public static string FindAssembly(string name)
         {
             AssemblyManager.UpdatePath();
-            return AssemblyManager.FindAssembly(name);
+            string assemblyName = AssemblyManager.FindAssembly(name);
+            if (assemblyName == null)
+            {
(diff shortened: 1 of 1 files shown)


denfromufa:
@jiftle this may break existing user code when developers search for assembly via FindAssembly and do not expect exception in the case of failure. You should just look at return value of the FindAssembly call.

jiftle:
> @jiftle this may break existing user code when developers search for assembly via FindAssembly call and do not expect exception in the case of not finding any assembly. You should just look at return value of the FindAssembly call, which returns the full path to assembly or `None`.
> 
> ```
> >>> import clr
> >>> clr.FindAssembly('testing')
> >>> clr.FindAssembly('Python.Runtime')
> u'C:\\Python\\Python27_32b\\lib\\site-packages\\Python.Runtime.dll'
> >>>
> ```

but not found about clr.FindAssembly() method's return value in help and wiki

jiftle:
@denfromufa 
ok.

PR #813, opened 2019-02-02
Switch to using TrueDivide instead of Divide everywhere

### LINKED ISSUE #802: Unable to find an entry point named 'PyNumber_Divide' when trying to call an overridden division operator
### Environment

-   Pythonnet version: 2.3
-   Python version: 3.6.6
-   Operating System: Windows 10 x64

### Details

I was trying to dynamically call overridden division operator:

```python
dynamic tf = Py.Import("tensorflow");
var c = tf.constant(42, tf.float32) / 2;
```

**Traceback**:

```
    System.EntryPointNotFoundException: 'Unable to find an entry point named 'PyNumber_Divide' in DLL 'python36'.'

at Python.Runtime.Runtime.PyNumber_Divide(IntPtr o1, IntPtr o2)
at Python.Runtime.PyObject.TryBinaryOperation(BinaryOperationBinder binder, Object arg, Object& result)
at CallSite.Target(Closure , CallSite , Object , Int32 )
at System.Dynamic.UpdateDelegates.UpdateAndExecute2[T0,T1,TRet](CallSite site, T0 arg0, T1 arg1)
```

It works for `+ 2` in place of `/ 2`, but might affect other operators too.

### Workaround

The workaround is to do `.__div__(2)` instead of `/ 2`

filmor:
Fixes #802.

TrueDivide is supported in both Python 2 and 3 while Divide exists only in Python 2. This does mean that a simple division of numbers will behave differently in C# vs Python for Python 2, but it will at least
behave the same way in C# independent of the Python version being used.

### DIFF
--- a/src/runtime/pyobject.cs
+++ b/src/runtime/pyobject.cs
@@ -101,7 +101,7 @@ namespace Python.Runtime
             }
             return result;
         }
-        
+
         /// <summary>
         /// As Method
         /// </summary>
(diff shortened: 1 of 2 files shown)


filmor:
The failing tests are completely unrelated, I don't have the time right now to fix the Travis build and it ran through on AppVeyor, so I'm going to merge this now.

PR #816, opened 2019-02-05
Limit the mono version to 5.16 for the time being

filmor:


filmor:
That's not how it works, the feed provided by the mono project only contains 5.18 :/

Issue #823, opened 2019-03-05
A way to override Python.NET .NET <-> Python wrapping for custom classes


lostmsu:
### Environment

-   Pythonnet version: 2.3.0
-   Python version: 3.6
-   Operating System: Windows 10 x64

### Details

I am trying to generate statically-typed wrappers for Python libraries, particularly TensorFlow (see [Gradient](https://github.com/losttech/Gradient)).

Since I am automatically generating .NET-side types for wrappers, I can make their methods to use my custom converter written in C# to convert function arguments to `PyObject` instances. And then for any result values, I can convert `PyObject` instances back into my wrapper types.

Problem I am facing is when a user of my wrapper wants to inherit from one of the Python-derived classes, and override and/or extend its behavior.

An artificial sample:
```csharp
// this part of my library is auto-generated
// represents tensorflow module
class tf {
  static dynamic tf = Py.Import("tensorflow");
  // set_default_session - made up method
  static void SetDefaultSession(Session session) => tf.set_default_session(session.underlyingSession);

  // made up method
  static void ComputeUsingDefaultSession() => tf.compute_using_default_session();
}

// represents tf.Tensor
class Tensor { PyObject underlyingTensor; }
// represents tf.Session
class Session {
  PyObject underlyingSession;
  virtual object Run(Tensor tensor) {
    var pyTensor = ConvertToPyObject(tensor); // simply does tensor.underlyingTensor here
    dynamic pySession = this.underlyingSession;
    var result = pySession.Run(pyTensor);
    return ConvertFromPyObject(result); // wraps any returned object into one of generated classes
  }
}

// this code is what user of my library wants to do:
class MyBetterSession : Session {
  override object Run(Tensor tensor) {
    ... here he writes custom code to run a Tensor ...
  }
}

// an attempts to use the above (functions made up):
var simpleSession = new Session();
tf.SetDefaultSession(simpleSession); // OK
tf.ComputeUsingDefaultSession(); // SUCCEEDS

var betterSession = new MyBetterSession();
tf.SetDefaultSession(betterSession); // OK
tf.ComputeUsingDefaultSession(); // FAILS
```

The last line will fail, because Python will attempt to call `MyBetterSession.Run` with Python's class `tf.Tensor`, but the method actually expects wrapped class `Tensor`. I need to somehow tell Python or rather Python.NET to invoke my `ConvertFromPyObject` on the argument, before trying to find a matching overload.

[...]

PR #829, opened 2019-03-12
Refactoring of MethodBinder.Bind

lostmsu:
This is a refactoring/cleaning up of MethodBinder.Bind method. No new features were introduced.

This is a preliminary work necessary for #823.

### Any other comments?
The refactoring was mostly done using ReSharper with little manual interventions. The semantics should be preserved.

### Checklist

### DIFF
--- a/src/runtime/methodbinder.cs
+++ b/src/runtime/methodbinder.cs
@@ -349,6 +349,18 @@ namespace Python.Runtime
             return null;
         }
 
+        /// <summary>
+        /// Attempts to convert Python argument tuple into an array of managed objects,
+        /// that can be passed to a method.
+        /// </summary>
+        /// <param name="pi">Information about expected parameters</param>
+        /// <param name="paramsArray"><c>true</c>, if the last parameter is a params array.</param>
+        /// <param name="args">A pointer to the Python argument tuple</param>
+        /// <param name="pyArgCount">Number of arguments, passed by Python</param>
+        /// <param name="defaultArgList">A list of default values for omitted parameters</param>
+        /// <param name="needsResolution"><c>true</c>, if overloading resolution is required</param>
+        /// <param name="outs">Returns number of output parameters</param>
+        /// <returns>An array of .NET arguments, that can be passed to a method.</returns>
         static object[] TryConvertArguments(ParameterInfo[] pi, bool paramsArray,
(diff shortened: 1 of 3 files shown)


lostmsu:
The build failure seems to be simple build timeout. Very unlikely caused by this change.

filmor:
I'll have a closer look at this at the weekend. No idea why travis is not getting triggered here, I've poked the one part that timed out though, so this one should go green.

Thanks already :)

filmor:
I'll just close and reopen this one to poke Travis...

lostmsu:
CI still having problems :/

src/runtime/methodbinder.cs
```diff
@@ -301,192 +299,242 @@ internal Binding Bind(IntPtr inst, IntPtr args, IntPtr kw, MethodBase info, Meth
...
+
+                object target = null;
+                if (!mi.IsStatic && inst != IntPtr.Zero)
                 {
-                    match = true;
-                    defaultArgList = new ArrayList();
-                    for (var v = pynargs; v < clrnargs; v++)
+                    //CLRObject co = (CLRObject)ManagedType.GetManagedObject(inst);
+                    // InvalidCastException: Unable to cast object of type
+                    // 'Python.Runtime.ClassObject' to type 'Python.Runtime.CLRObject'
```
filmor: This comment confuses me more than it helps ;). It's clear after reading the next one, but here not so much.

src/runtime/methodbinder.cs
```diff
@@ -301,192 +299,242 @@ internal Binding Bind(IntPtr inst, IntPtr args, IntPtr kw, MethodBase info, Meth
...
+                if (!TryConvertArgument(op, parameter.ParameterType, needsResolution, out margs[paramIndex], out isOut))
+                {
+                    return null;
+                }
+
+                if (arrayStart == paramIndex)
+                {
+                    // TODO: is this a bug? Should this happen even if the conversion fails?
+                    // GetSlice() creates a new reference but GetItem()
+                    // returns only a borrow reference.
```
filmor: This indeed looks like a bug to me, the code has to also run before `return null`.

filmor:
I think both comments are for things that you didn't change, just wanted to leave them around.

PR #838, opened 2019-04-02
List installed Win10 SDK's when looking for windows sdk tool

OneBlue:

When I try to run `setup.py install`, the installation failed with the following error:
```
RuntimeError: mt.exe could not be found
```

Because setup.py is searching for specific versions of the Windows 10 sdk.

This patch changes that behaviour to use the installed Windows 10 sdk instead of relying on specific versions.

No.

### Any other comments?
I kept the existing hard-coded sdk version to ensure that this is a non-breaking change.

The objective of this patch is make installation easier.

### Checklist

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -55,6 +55,7 @@
 -   ([@civilx64](https://github.com/civilx64))
 -   ([@GSPP](https://github.com/GSPP))
 -   ([@omnicognate](https://github.com/omnicognate))
+-   ([@OneBlue](https://github.com/OneBlue))
 -   ([@rico-chet](https://github.com/rico-chet))
 -   ([@rmadsen-ks](https://github.com/rmadsen-ks))
 -   ([@stonebig](https://github.com/stonebig))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ This document follows the conventions laid out in [Keep a CHANGELOG][].
 
 ## [unreleased][]
 
+-   Look for installed Windows 10 sdk's during installation instead of relying on specific versions.
+
 ### Added
 
 -   Added support for embedding python into dotnet core 2.0 (NetStandard 2.0)
--- a/setup.py
+++ b/setup.py
@@ -39,7 +39,7 @@ sdks_root = "SOFTWARE\\Microsoft\\Microsoft SDKs\\Windows\\v{0}Win32Tools"
 kits_root = "SOFTWARE\\Microsoft\\Windows Kits\\Installed Roots"
 kits_suffix = os.path.join("bin", ARCH)
 
-WIN_SDK_KEYS = (
+WIN_SDK_KEYS = [
     RegKey(sdk_name="Windows Kit 10.0", key=kits_root,
            value_name="KitsRoot10", suffix=os.path.join("bin", "10.0.16299.0", ARCH)),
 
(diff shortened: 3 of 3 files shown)


filmor:
Thank you very much :)

I'm going to merge this for now, I'm just wondering whether we'll need this in the future. In PR #830 I've experimented with `vswhere` which is available on newer VS installations, would that work for you as well?

OneBlue:
Looking at #830, using vswhere is a much better solution. 

I'm not sure when was vswhere introduced so I would recommend keeping the current logic as a fallback in case vswhere is not available on the machine.

denfromufa:
@oneblue a copy of vswhere is included with pythonnet in tools. Can you try updating that version and see if it resolves your issue?

OneBlue:
The tool that wasn't found was mt.exe, which is looked up like this:
``` 
mt = self._find_msbuild_tool("mt.exe", use_windows_sdk=True)
```
This code path doesn't use vswhere (because use_windows_sdk is True). So updating vswhere won't fix the problem.

I also tried to checkout #830, but when I tried to build, I got: 
```λ setup.py install
running install
running build
running build_ext
Microsoft (R) Build Engine version 15.9.21+g9802d43bc3 for .NET Framework
Copyright (C) Microsoft Corporation. All rights reserved.

Building the projects in this solution one at a time. To enable parallel build, please add the "/m" switch.
Build started 06/04/2019 12:35:47.
Project "C:\Users\Blue\Desktop\repos\pythonnet\pythonnet.sln" on node 1 (Clean target(s)).
ValidateSolutionConfiguration:
  Building solution configuration "ReleaseWin|x64".
Project "C:\Users\Blue\Desktop\repos\pythonnet\pythonnet.sln" (1) is building "C:\Users\Blue\Desktop\repos\pythonnet\src\testing\Python.Test.csproj" (2) on node 1 (Clean target(s)).
Project "C:\Users\Blue\Desktop\repos\pythonnet\src\testing\Python.Test.csproj" (2) is building "C:\Users\Blue\Desktop\repos\pythonnet\src\testing\Python.Test.csproj" (2:2) on node 1 (Clean target(s)).
CoreClean:
  Creating directory "obj\x64\ReleaseWin\".
Project "C:\Users\Blue\Desktop\repos\pythonnet\src\testing\Python.Test.csproj" (2:2) is building "C:\Users\Blue\Desktop\repos\pythonnet\src\runtime\Python.Runtime.csproj" (3:2) on node 1 (Clean target(s)).
CoreClean:
  Creating directory "obj\ReleaseWin\".
Done Building Project "C:\Users\Blue\Desktop\repos\pythonnet\src\runtime\Python.Runtime.csproj" (Clean target(s)).

Done Building Project "C:\Users\Blue\Desktop\repos\pythonnet\src\testing\Python.Test.csproj" (Clean target(s)).

Project "C:\Users\Blue\Desktop\repos\pythonnet\src\testing\Python.Test.csproj" (2) is building "C:\Users\Blue\Desktop\repos\pythonnet\src\testing\Python.Test.csproj" (2:3) on node 1 (Clean target(s)).
C:\Program Files (x86)\Microsoft Visual Studio\2017\Enterprise\MSBuild\Sdks\Microsoft.NET.Sdk\build\Microsoft.NET.TargetFrameworkInference.targets(126,5): error : The current .NET SDK does not support targeting .NET Standard 2.0.  Either  target .NET Standard 1.6 or lower, or use a version of the .NET SDK that supports .NET Standard 2.0. [C:\Users\Blue\Desktop\repos\pythonnet\src\testing\Python.Test.csproj]
``` 
While master builds correctly.

PR #852, opened 2019-04-23
Make finalizer use a task

Martin-Molinero:
- The callback set by `Runtime.Py_AddPendingCall()` was not being
triggered in some cases in a multithreaded environment. Replacing it
with a `Task`
Changes cherry-picked from https://github.com/QuantConnect/pythonnet/pull/29

After 
imagen

Before
imagen

### DIFF
--- a/src/runtime/finalizer.cs
+++ b/src/runtime/finalizer.cs
@@ -80,12 +80,9 @@ namespace Python.Runtime
             if (Instance._finalizerTask != null
                 && !Instance._finalizerTask.IsCompleted)
             {
-                using (Py.GIL())
-                {
-                    var ts = PythonEngine.BeginAllowThreads();
-                    Instance._finalizerTask.Wait();
-                    PythonEngine.EndAllowThreads(ts);
-                }
+                var ts = PythonEngine.BeginAllowThreads();
+                Instance._finalizerTask.Wait();
+                PythonEngine.EndAllowThreads(ts);
             }
             else if (forceDispose)
             {
@@ -147,8 +144,11 @@ namespace Python.Runtime
 
                         _finalizerTask = Task.Factory.StartNew(() =>
                         {
-                            Instance.DisposeAll();
-                            _pending = false;
+                            using (Py.GIL())
+                            {
+                                Instance.DisposeAll();
+                                _pending = false;
+                            }
                         });
                     }
                 }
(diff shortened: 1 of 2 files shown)


filmor:
Great, thanks for your continuous efforts to backport fixes, it's highly appreciated. I'll look into merging this before the 2.4 release :)

filmor:
The Linux build seems to be fine apart from an unrelated error (happens on master as well). On Windows, however, for some reason this seems to stall, the builds time out during the unit tests.

filmor:
It trips up the `DomainReloadAndGC` test, it looks like there is a deadlock in `DisposeAll`.

src/runtime/finalizer.cs
```diff
@@ -223,46 +169,32 @@ private void DisposeAll()
             lock (_queueLock)
 #endif
             {
+                using (Py.GIL())
```
filmor: Why do you acquire the GIL here?

filmor:
If I remove the GIL call from `DisposeAll` everything works fine as far as I can tell right now.

src/runtime/finalizer.cs
Martin-Molinero: I believe it's required for the task I'm starting at line 148, right? It can be moved inside the `Action` I'm passing to the task. It's safe to assume the thread running `Shutdown` already owns the GIL?

filmor: I'll have to look into this in more detail, I just noticed that you added this one and that it's where the deadlock occurs on Windows :)

Martin-Molinero: Was able to repro on my end, added a commit with the fix I mentioned, all unit tests seem to be correctly passing (executed them a bunch of times). Tested with Quantconnect python tests and everything seems green there (memory leak still fixed). Let me know what you think :)

filmor:
Great, this looks a lot better, thank you very much :)

amos402:
@Martin-Molinero Using `Py_AddPendingCall` is aiming for no Python thread, although we called `PyEval_InitThreads` at `Initialize`, but that is a optimize point that we can prevent create a `GIL` on single thread environment. Also using `Py.GIL` may create a temporary Python thread state which takes overhead.
I think we can make a better user experience for the finalizer, we can setup a timer for finalizing queue, if the time past threshold value then create a Python thread to dispose them. That may also help for the queue size don't reach count threshold for a long time but keep the object reference.

Issue #811, opened 2019-01-28
An overload with a parameter of type `object` is not used


lostmsu:
### Environment

-   Pythonnet version: 2.3.0
-   Python version: 3.6
-   Operating System: Win 10 x64 1809

### Details

I am trying to get Python call my C# method, which has a parameter of type `object` with an instance of Python's `KeyError`.

```csharp
var o = overloaded.ToPython();

dynamic callWithSelf = PythonEngine.Eval("lambda o: o.ObjOrClass(KeyError())");
callWithSelf(o);
```

Where `ObjOrClass` definition is: `public void ObjOrClass(object _) => this.Value = Object;`. The actual function body here is not important.

-  Error:

```
Python.Runtime.PythonException : TypeError : No method matches given arguments for ObjOrClass
```

### Possibly related:
https://github.com/pythonnet/pythonnet/issues/265
https://github.com/pythonnet/pythonnet/issues/782

### Thoughs
IMHO, Python.NET should create a `PyObject` and pass it here.

filmor:
Yes, it's a possibility to do that. The issue at hand is probably, that Python objects do not derive from `System.Object` and can thus not match the overload. Can you prepare a PR?

lostmsu:
@filmor I am not sure I understood your comment. You can see my fix in the commit e9712b4 above. This change is potentially backwards-incompatible in case there's a .NET method with two overloads, one having parameter type `object`, and another `PyObject` for some parameter.

filmor:
I'll need to look into this, could you make a PR from this commit?

PR #853, opened 2019-04-24
Enable C# parameters of type `object` accept any argument, passed from Python

lostmsu:

Previously when a C# method with a parameter of type `object` was called from Python with some argument of a native Python type (e.g. instance of `dict` rather than instance of `System.Random`), that would be considered a parameter type mismatch, so overload resolution would fail, and the C# method would not be called.

After this change, when Python tries to pass a argument of some Python type to a C# parameter of type `object`, that overload would be considered good for resolution, and the argument will be wrapped into `PyObject` and passed to the `object` parameter.

For example:

```csharp
void CallMe(object anything) => Console.WriteLine(anything);
````

```python
CallMe(dict())
```

Before change: overload not found
After change: `anything` in `CallMe` is an instance of `PyObject`, wrapping a `dict` instance, so `CallMe` outputs `{}`

#811

### Checklist


filmor:
Could you please reopen this?

lostmsu:
@filmor , am I understanding it correctly, that Python-based tests need to be fixed appropriately?

filmor:
I didn't check yet in detail what this PR broke, I just want to keep it on the radar :)

PR #854, opened 2019-04-27
Remove error-prone text comparison in exception test

filmor:


filmor:
I'll merge this even though one test regarding the Finalizer is still failing to see whether #852 fixes it.

PR #877, opened 2019-06-01
Mention conda-forge as installation option

m-rossi:
I've just added the shield to the README.

We should may also update
* https://github.com/pythonnet/pythonnet/wiki/Installation
* https://github.com/pythonnet/pythonnet.github.io


filmor:
Great, thank you very much :)

PR #880, opened 2019-06-13
Drop LoadLibrary

amos402:
There are some limitations to dynamic load a library on some systems(e. g. Android).
`LoadLibrary` just called for once for getting the `_PyObject_NextNotImplemented`,
we can use another ways to get this function.

No

### Checklist


filmor:
I'll have to look a bit more into this. I thought that this `LoadLibrary` call there is not only pulling in that one symbol but that it serves as a "proxy" to enforce loading the library into memory s.t. the P/Invoke bindings work. Since your patch passes all tests (apart from the known failing one) I must have misunderstood something and will test this in our use-case to verify. Thank you already for your contribution, if this works out this will help a lot in simplifying the build process :)

lostmsu:
@filmor @amos402 I made a single Python.Runtime.dll to work with any Python 3.x interpreter in my fork (could have included 2.x, but its lifetime is ending) via extensive use of `LoadLibrary` and related methods. Which I intended to upstream once 2.7 is dropped (or, for example, if we had a separate 2.x branch in maintenance mode).

filmor:
@lostmsu What do you mean by "extensive use"? Isn't it enough to load libpython*.so once at startup? Could you point me to that code?

lostmsu:
@filmor well, the `LoadLibrary` is called once, but all the functions are obtained using `GetProcAddress`/`dlsym`.
See https://github.com/losttech/pythonnet/commit/672258f590c6ad16db7f327f2690698cd3d4b38f

src/runtime/runtime.cs
```diff
@@ -297,16 +297,20 @@ internal static void Initialize(bool initSigs = false)
             IntPtr dllLocal = IntPtr.Zero;
             var loader = LibraryLoader.Get(OperatingSystem);
```
benoithudson: Seems like you should delete these lines too, and line 295 as well since you moved it below.

amos402:
I removed the dependency of `zipimport`. But still have no solution for `SetNoSiteFlag`, maybe we should add a docs for it(like please use `-S` when using static compile).

src/runtime/runtime.cs
```diff
@@ -1893,25 +1908,24 @@ internal static IntPtr PyMem_Realloc(IntPtr ptr, long size)
...
+            }
             var loader = LibraryLoader.Get(OperatingSystem);
-
-            IntPtr dllLocal;
-            if (_PythonDll != "__Internal")
+            IntPtr dllLocal = loader.Load(_PythonDll);
+            if (dllLocal == IntPtr.Zero)
             {
-                dllLocal = loader.Load(_PythonDll);
+                throw new Exception($"Cannot load {_PythonDll}");
```
lostmsu: Can you throw more informative exception? `Win32Exception` would automatically read last error on Windows, but I don't know what to do for *nix

src/runtime/runtime.cs
```diff
@@ -328,6 +320,29 @@ internal static void Initialize(bool initSigs = false)
             AssemblyManager.UpdatePath();
         }
 
+        private static IntPtr Get_PyObject_NextNotImplemented()
+        {
+            IntPtr globals = PyDict_New();
+            if (PyDict_SetItemString(globals, "__builtins__", PyEval_GetBuiltins()) != 0)
```
lostmsu: Can `PyEval_GetBuiltins()` return an error?

src/runtime/runtime.cs
```diff
@@ -328,6 +320,29 @@ internal static void Initialize(bool initSigs = false)
...
+            {
+                XDecref(globals);
+                throw new PythonException();
+            }
+            const string code = "class A(object): pass";
+            IntPtr res = PyRun_String(code, (IntPtr)RunFlagType.File, globals, globals);
+            if (res == IntPtr.Zero)
+            {
+                XDecref(globals);
+                throw new PythonException();
```
lostmsu: You might want to construct an instance of `PythonException` before doing `XDecref`, otherwise last error could potentially be overwritten by `XDecref` with something new.

src/runtime/runtime.cs
filmor: From the docs it doesn't look like it can (https://docs.python.org/3/c-api/reflection.html#c.PyEval_GetBuiltins).

benoithudson: The globals dict is needed, but we don't seem to be using the `__builtins__` in the string. Do we really need this line?

amos402: If you don't setup a `__builtins__ ` or set `object` explicit, `object` may not be found.

src/runtime/runtime.cs
amos402: Generally it should throw inside from `Load`, the nullptr checking just for in case(if the loader implementation didn't throw an exception). I'm not going to insert any platform specific code here.

src/runtime/runtime.cs
```diff
@@ -1925,25 +1946,24 @@ internal static IntPtr PyMem_Realloc(IntPtr ptr, long size)
 
         internal static void SetNoSiteFlag()
         {
+            if (_PythonDll == "__Internal")
+            {
+                throw new NotSupportedException("SetNoSiteFlag didn't support on static compile");
```
filmor: Why should this not be supported? The symbol is just in the running DLL, so `loader.Load(_PythonDll)` has to return essentially the result of `dlopen(NULL, ...)`.

amos402: Yes, it should be supported. Just a mistake according to my wrong tests. I will recover it back.

Jeff17Robbins:
I believe issues #891, #946, and #967 all point to a fundamental problem with this PR.

On certain Linux distributions (Debian for example), Python C extension libraries [are deliberately not linked](https://www.debian.org/doc/packaging-manuals/python-policy/ch-module_packages.html) to `libpython`.  

Yet .NET Core's `[DllImport]` does not load `libpython` with `RTLD_GLOBAL`.

So if this PR removes `dllLocal = loader.Load(_PythonDll);` from `runtime.cs`, where the salient line in `LibraryLoader.cs` is `var res = dlopen(filename, RTLD_NOW | RTLD_GLOBAL);`, won't this PR break Python.NET's "Embedding" mode on .NET Core?

As Victor Stinner put it in [Python issue issue 21536](https://bugs.python.org/issue21536#msg340817): 

> ...libpython must always be loaded with RTLD_GLOBAL.

Unless we can get `[DllImport]` to set the `RTLD_GLOBAL` flag, I'm afraid we need the explicit `loader.Load`, at least for Linux .NET Core embedding.

amos402:
@Jeff17Robbins This PR won't break "Embedding" mode on .NET Core, moreover it improve the compatibility since it didn't need call the `dlopen`.

amos402:
Contained in #958

PR #882, opened 2019-06-17
Revert "Enable C# parameters of type `object` accept any argument, passed from Python"

filmor:
Reverts pythonnet/pythonnet#853 since it breaks for Python 2.


lostmsu:
@filmor what is the problem with Python 2? CI seems to be passing.

filmor:
Well, it does pass after reverting, this is a run on current master: https://travis-ci.org/pythonnet/pythonnet/jobs/546918277

Issue #873, opened 2019-05-28
ImportError: System.NotImplementedException: No support for armv7l


bimajatiwijaya:
### Environment

-   Pythonnet version: 
     - 2.4.0 & 2.4.1.dev0
-   Python version: 2.7.9
-   Operating System: Raspbian GNU/Linux 8 (jessie)

### Details
- Installation 
    - sudo pip install git+https://github.com/pythonnet/pythonnet
    - sudo pip install pythonnet==2.4.0 
-   Error triggered when import clr

```python
    import clr
    clr.AddReference("myLib")
```

-   Traceback (most recent call last):
>>> import clr
  File "<stdin>", line 1, in <module>
ImportError: System.NotImplementedException: No support for armv7l
  at Python.Runtime.TypeManager+NativeCode.get_Active () [0x0002a] in <9666540cfc85422bb7aded792ea4f880>:0
  at Python.Runtime.TypeManager.InitializeNativeCodePage () [0x00017] in <9666540cfc85422bb7aded792ea4f880>:0
  at Python.Runtime.TypeManager.InitializeSlots (System.IntPtr type, System.Type impl) [0x000ad] in <9666540cfc85422bb7aded792ea4f880>:0
  at Python.Runtime.TypeManager.CreateMetaType (System.Type impl) [0x00077] in <9666540cfc85422bb7aded792ea4f880>:0
  at Python.Runtime.MetaType.Initialize () [0x00000] in <9666540cfc85422bb7aded792ea4f880>:0
  at Python.Runtime.Runtime.Initialize (System.Boolean initSigs) [0x002a2] in <9666540cfc85422bb7aded792ea4f880>:0
  at Python.Runtime.PythonEngine.Initialize (System.Collections.Generic.IEnumerable`1[T] args, System.Boolean setSysArgv, System.Boolean initSigs) [0x00014] in <9666540cfc85422bb7aded792ea4f880>:0
  at Python.Runtime.PythonEngine.Initialize (System.Boolean setSysArgv, System.Boolean initSigs) [0x00005] in <9666540cfc85422bb7aded792ea4f880>:0
  at Python.Runtime.PythonEngine.InitExt () [0x00000] in <9666540cfc85422bb7aded792ea4f880>:0

bimajatiwijaya:
For now i reset commit to https://github.com/pythonnet/pythonnet/commit/6d68d70db037ed27fc0d8f24302202a2be8224f6 and "No support for armv7l" solved.

[...]

PR #887, opened 2019-06-20
Support ARM architectures again

filmor:

Adds ARM versions to the native page hack used to survive app-domain reloading.

#873

### DIFF
--- a/src/runtime/runtime.cs
+++ b/src/runtime/runtime.cs
@@ -229,6 +229,8 @@ namespace Python.Runtime
         {
             i386,
             x86_64,
+            armv7l,
+            armv8,
             Other
         };
 
@@ -247,6 +249,8 @@ namespace Python.Runtime
             ["amd64"] = MachineType.x86_64,
             ["x64"] = MachineType.x86_64,
             ["em64t"] = MachineType.x86_64,
+            ["armv7l"] = MachineType.armv7l,
+            ["armv8"] = MachineType.armv8,
         };
 
         /// <summary>
--- a/src/runtime/typemanager.cs
+++ b/src/runtime/typemanager.cs
@@ -510,6 +510,10 @@ namespace Python.Runtime
                             return I386;
                         case Runtime.MachineType.x86_64:
                             return X86_64;
+                        case Runtime.MachineType.armv7l:
+                            return Armv7l;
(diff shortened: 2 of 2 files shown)


filmor:
@benoithudson, could you have a look?
@bimajatiwijaya, could you check whether this branch works for you?

filmor:
The only thing I have is an old Raspberry Pi and a few phones, I just used godbolt to get the binaries. That's why I'd like people that had problems to test this branch ;)

amos402:
Ummm, I think it seems that it would better if we create a dynamic library and use P/Invoke to get a function address rather than making some hard codes like these?

filmor:
That would complicate the build process quite a lot though. I'd really like to get to a single arch independent assembly in the future that can just be compiled using dotnet cli without any additional params.

amos402:
But at first, you always need to build a sample binary and disassembly to get codes depend the system.
If you worry about the build process, why not just upload that prebuilt binary to repository, so that users would not need any compile.
What if we need to compatible systems like these:

How complicated would `typemanager.cs` be......

filmor:
The trick is that it's straight binary, so you only need one of these per cpu arch, not per os (no ELF vs PE distinction, no libc linking etc). Be my guest in providing a reasonable alternative.

amos402:
Oh, it is. I almost neglect it.

benoithudson:
> That would complicate the build process quite a lot though. I'd really like to get to a single arch independent assembly in the future that can just be compiled using dotnet cli without any additional params.

Me too, which is why my patch that introduced this code was discovering the platform at runtime rather than with ifdefs.

If there's some rainy days I'll see about converting more of the code. Eliminating the MONO_LINUX and MONO_OSX defines would be nice; UCS as well.

amos402:
By the way, the implementation for `Runtime.InitializePlatformData` just have some limitations.
Not `platform.py` would import `subprocess`, but not all release binaries have it, I encounter it before and use ifdef skiped, any good suggestion?

benoithudson:
Probably inline the two functions. IIRC one of the calls has a CPython component, the other I didn't track down.

We're getting pretty far off the original topic, I wonder if comments on this PR are the best way to gather our thoughts?

PR #889, opened 2019-06-20
Retry: Enable C# parameters of type `object` accept any argument, passed from Python

lostmsu:

Previously when a C# method with a parameter of type `object` was called from Python with some argument of a native Python type (e.g. instance of `dict` rather than instance of `System.Random`), that would be considered a parameter type mismatch, so overload resolution would fail, and the C# method would not be called.

After this change, when Python tries to pass a argument of some Python type to a C# parameter of type `object`, that overload would be considered good for resolution, and the argument will be wrapped into `PyObject` and passed to the `object` parameter.

For example:

```csharp
void CallMe(object anything) => Console.WriteLine(anything);
````

```python
CallMe(dict())
```

Before change: overload not found
After change: `anything` in `CallMe` is an instance of `PyObject`, wrapping a `dict` instance, so `CallMe` outputs `{}`

#811

### Checklist

### P.S.
This is a retry of #853

### DIFF
--- a/src/runtime/converter.cs
+++ b/src/runtime/converter.cs
@@ -398,12 +398,9 @@ namespace Python.Runtime
                     return ToArray(value, typeof(object[]), out result, setError);
                 }
 
-                if (setError)
-                {
-                    Exceptions.SetError(Exceptions.TypeError, "value cannot be converted to Object");
-                }
-
-                return false;
+                Runtime.XIncref(value); // PyObject() assumes ownership
+                result = new PyObject(value);
+                return true;
             }
 
             // Conversion to 'Type' is done using the same mappings as above for objects.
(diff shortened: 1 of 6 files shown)


lostmsu:
@filmor is there an instruction to run Python unit tests?

filmor:
You can run the unit tests by using pytest (either just `pytest` or `python -m pytest`) after installing.

lostmsu:
@filmor AppVeyor failed for 3.6 x86 due to connectivity issue on the VM: https://ci.appveyor.com/project/pythonnet/pythonnet/builds/28271316 . Results are not being picked up.

src/runtime/converter.cs
```diff
@@ -859,7 +856,22 @@ private static bool ToArray(IntPtr value, Type obType, out object result, bool s
...
+                    if (setError)
+                    {
+                        Exceptions.SetError(Exceptions.NotImplementedError, ".NET does not support lists with over 2^31 elements");
+                    }
+
+                    return false;
+                }
+
+                len = longLen < 0 ? -1 : checked((int)longLen);
+            }
```
filmor: This still fails to finish on Windows 32bit right now, which hints a problem with the way you handle the result of `Runtime.PySequence_Size`. As far as I can tell, the changes you did there are irrelevant for the actual problem you wanted to fix, could you please revert these changes as well as the one below and see whether it passes then?

lostmsu: They are sort of relevant. `KeyError` from the test I gets `true` on `PySequence_Check`. I will rewrite the test, and move the fix into a separate PR for clarity.

PR #894, opened 2019-06-25
Get the correct library loading functions at runtime

filmor:

Moves the library loader to three separate classes that contain implementations for Darwin, Linux and Windows s.t. they can be initialized at runtime.

Nope, but it brings us further towards a single build.

### DIFF
--- a/src/runtime/platform/LibraryLoader.cs
+++ b/src/runtime/platform/LibraryLoader.cs
@@ -1,0 +1,153 @@
+using System;
+using System.Runtime.InteropServices;
+
+namespace Python.Runtime.Platform
+{
+    interface ILibraryLoader
+    {
+        IntPtr Load(string dllToLoad);
+
+        IntPtr GetFunction(IntPtr hModule, string procedureName);
+
+        bool Free(IntPtr hModule);
+    }
+
+    static class LibraryLoader
+    {
+        public static ILibraryLoader Get(OperatingSystemType os)
+        {
+            switch (os)
+            {
+                case OperatingSystemType.Windows:
+                    return new WindowsLoader();
+                case OperatingSystemType.Darwin:
+                    return new DarwinLoader();
+                case OperatingSystemType.Linux:
+                    return new LinuxLoader();
+                default:
+                    throw new Exception($"This operating system ({os}) is not supported");
+            }
+        }
+    }
+
+    class LinuxLoader : ILibraryLoader
+    {
+        private static int RTLD_NOW = 0x2;
+        private static int RTLD_GLOBAL = 0x100;
+        private static IntPtr RTLD_DEFAULT = IntPtr.Zero;
+        private const string NativeDll = "libdl.so";
+
+        public IntPtr Load(string fileName)
+        {
+            return dlopen($"lib{fileName}.so", RTLD_NOW | RTLD_GLOBAL);
+        }
+
+        public bool Free(IntPtr handle)
+        {
+            dlclose(handle);
+            return true;
+        }
+
+        public IntPtr GetFunction(IntPtr dllHandle, string name)
+        {
+            // look in the exe if dllHandle is NULL
+            if (dllHandle == IntPtr.Zero)
+            {
+                dllHandle = RTLD_DEFAULT;
+            }
+
+            // clear previous errors if any
+            dlerror();
+            IntPtr res = dlsym(dllHandle, name);
+            IntPtr errPtr = dlerror();
+            if (errPtr != IntPtr.Zero)
+            {
+                throw new Exception("dlsym: " + Marshal.PtrToStringAnsi(errPtr));
+            }
+            return res;
+        }
+
+        [DllImport(NativeDll, CallingConvention = CallingConvention.Cdecl, CharSet = CharSet.Ansi)]
+        public static extern IntPtr dlopen(String fileName, int flags);
+
+        [DllImport(NativeDll, CallingConvention = CallingConvention.Cdecl, CharSet = CharSet.Ansi)]
+        private static extern IntPtr dlsym(IntPtr handle, String symbol);
+
+        [DllImport(NativeDll, CallingConvention = CallingConvention.Cdecl)]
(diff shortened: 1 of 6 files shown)


src/runtime/platform/LibraryLoader.cs
```diff
@@ -0,0 +1,153 @@
...
+                    return new LinuxLoader();
+                default:
+                    throw new Exception($"This operating system ({os}) is not supported");
+            }
+        }
+    }
+
+    class LinuxLoader : ILibraryLoader
+    {
+        private static int RTLD_NOW = 0x2;
```
lostmsu: const?

src/runtime/platform/LibraryLoader.cs
```diff
@@ -0,0 +1,153 @@
...
+                dllHandle = RTLD_DEFAULT;
+            }
+
+            // clear previous errors if any
+            dlerror();
+            IntPtr res = dlsym(dllHandle, name);
+            IntPtr errPtr = dlerror();
+            if (errPtr != IntPtr.Zero)
+            {
+                throw new Exception("dlsym: " + Marshal.PtrToStringAnsi(errPtr));
```
lostmsu: Here and in similar places should throw more specific exception.

src/runtime/runtime.cs
```diff
@@ -7,96 +7,7 @@
...
-        public static extern IntPtr LoadLibrary(string dllToLoad);
-
-        [DllImport(NativeDll)]
-        public static extern IntPtr GetProcAddress(IntPtr hModule, string procedureName);
-
-        [DllImport(NativeDll)]
-        public static extern bool FreeLibrary(IntPtr hModule);
-#endif
-    }
+    using Python.Runtime.Platform;
```
lostmsu: Looks like should be outside. Or let's move them all inside in a separate commit.

src/runtime/platform/LibraryLoader.cs
```diff
@@ -0,0 +1,153 @@
...
+            switch (os)
+            {
+                case OperatingSystemType.Windows:
+                    return new WindowsLoader();
+                case OperatingSystemType.Darwin:
+                    return new DarwinLoader();
+                case OperatingSystemType.Linux:
+                    return new LinuxLoader();
+                default:
+                    throw new Exception($"This operating system ({os}) is not supported");
```
lostmsu: PlatformNotSupportedException

src/runtime/platform/LibraryLoader.cs
```diff
@@ -0,0 +1,153 @@
...
+    class LinuxLoader : ILibraryLoader
+    {
+        private static int RTLD_NOW = 0x2;
+        private static int RTLD_GLOBAL = 0x100;
+        private static IntPtr RTLD_DEFAULT = IntPtr.Zero;
+        private const string NativeDll = "libdl.so";
+
+        public IntPtr Load(string fileName)
+        {
+            return dlopen($"lib{fileName}.so", RTLD_NOW | RTLD_GLOBAL);
```
lostmsu: Add error handling

src/runtime/platform/LibraryLoader.cs
```diff
@@ -0,0 +1,153 @@
...
+    class DarwinLoader : ILibraryLoader
+    {
+        private static int RTLD_NOW = 0x2;
+        private static int RTLD_GLOBAL = 0x8;
+        private const string NativeDll = "/usr/lib/libSystem.dylib";
+        private static IntPtr RTLD_DEFAULT = new IntPtr(-2);
+
+        public IntPtr Load(string fileName)
+        {
+            return dlopen($"lib{fileName}.dylib", RTLD_NOW | RTLD_GLOBAL);
```
lostmsu: Add error handling

src/runtime/platform/LibraryLoader.cs
```diff
@@ -0,0 +1,153 @@
...
+    }
+
+    class WindowsLoader : ILibraryLoader
+    {
+        private const string NativeDll = "kernel32.dll";
+
+        [DllImport(NativeDll)]
+        static extern IntPtr LoadLibrary(string dllToLoad);
+
+        public IntPtr Load(string dllToLoad) => WindowsLoader.LoadLibrary(dllToLoad);
```
lostmsu: Add error handling

src/runtime/platform/LibraryLoader.cs
```diff
@@ -0,0 +1,153 @@
...
+
+        [DllImport(NativeDll)]
+        static extern IntPtr GetProcAddress(IntPtr hModule, string procedureName);
+
+        public IntPtr GetFunction(IntPtr hModule, string procedureName) => WindowsLoader.GetProcAddress(hModule, procedureName);
```
lostmsu: Add error handling

lostmsu:
Done with a review pass.

src/runtime/platform/LibraryLoader.cs
lostmsu: Also, Ansi? I'd double-check.

src/runtime/platform/LibraryLoader.cs
filmor: I think `static` is fine, `const` not being available at runtime has created problems in the past and makes debugging more annoying.

src/runtime/platform/LibraryLoader.cs
filmor: Done.

src/runtime/platform/LibraryLoader.cs
filmor: Done.

src/runtime/runtime.cs
filmor: Done.

src/runtime/platform/LibraryLoader.cs
filmor: Well, it's expected to be ASCII only, so this should be fine. Also, this whole PR is just a refactoring, it should be functionally equivalent to the previous state.

PR #895, opened 2019-06-25
[QC] QuantConnect: Rolling Pull Request Master

jaredbroad:
Long term open PR for PythonNet


AUTHORS.md
```diff
@@ -14,7 +14,6 @@
 
 -   Alexandre Catarino([@AlexCatarino](https://github.com/AlexCatarino))
 -   Arvid JB ([@ArvidJB](https://github.com/ArvidJB))
--   Benoît Hudson ([@benoithudson](https://github.com/benoithudson))
```
lostmsu: HM, I wonder how these guys were removed?

src/embed_tests/TestPyAnsiString.cs
```diff
@@ -63,6 +63,7 @@ public void TestCtorPtr()
             const string expected = "foo";
 
             var t = new PyAnsiString(expected);
+            Runtime.Runtime.XIncref(t.Handle);
```
lostmsu: That should not be here. PyAnsiString constructor should call XIncref

src/embed_tests/TestPyFloat.cs
```diff
@@ -25,6 +25,7 @@ public void Dispose()
         public void IntPtrCtor()
         {
             var i = new PyFloat(1);
+            Runtime.Runtime.XIncref(i.Handle);
```
lostmsu: Same.
Since that might affect existing code, an overload should be added, that specifies ownership

src/embed_tests/TestPythonEngineProperties.cs
```diff
@@ -190,7 +156,7 @@ public void SetPythonPath()
             string path = PythonEngine.PythonPath;
             PythonEngine.Shutdown();
```
lostmsu: Why has this been changed? Test method name implies it should test `PythonPath`

AUTHORS.md
filmor: This is probably just due to the QuantConnect fork missing a few changes (in this case the domain unload fixes).

mjmvisser: @jaredbroad can you cherry-pick the domain unload fixes from master onto your branch?

lostmsu:
I just did not expect GitHub to show missing commits from target as a part of the PR. Good to know.

Issue #898, opened 2019-06-26
Crash on Raspberry Pi


climblinne:
### Environment

-   Pythonnet version: actual dev version  fc7d8a4  
-   Python version: 3.7.3
-   Operating System: Raspian (2019-06-20-raspbian-buster-lite)

### Details

-   I installed the following:

```
 # install 2019-06-20-raspbian-buster-lite
 raspi-config : change password, name & extend sd card
 sudo apt update 
 sudo apt dist-upgrade
 sudo reboot
 
 # Mono install (5.20.1.19)
sudo apt install apt-transport-https dirmngr gnupg ca-certificates
sudo apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys 3FA7E0328081BFF6A14DA29AA6A19B38D3D831EF
echo "deb https://download.mono-project.com/repo/debian stable-raspbianstretch main" | sudo tee /etc/apt/sources.list.d/mono-official-stable.list
sudo apt update 
sudo apt install mono-complete
sudo apt install ca-certificates-mono
 
 
 sudo apt install git
 sudo apt install clang
 sudo apt install libglib2.0-dev
 sudo apt-get install python3-pip # Python 3.7.3 already installed
 pip3 install -U setuptools
 pip3 install -U wheel
 pip3 install -U pip
 pip3 install --user pycparser
 pip3 install --user git+https://github.com/pythonnet/pythonnet
 
 pip3 install --user pytest
 
 python3 -m pytest

```
When run test, the following output happen:

```
    :~/pythonnet $ python3 -m pytest
======================================= test session starts =======================================
platform linux -- Python 3.7.3, pytest-4.6.3, py-1.8.0, pluggy-0.12.0
rootdir: /home/pi/pythonnet, inifile: setup.cfg
collecting ...
=================================================================
        Native Crash Reporting
=================================================================
Got a SIGILL while executing native code. This usually indicates
a fatal error in the mono runtime or one of the native libraries
used by your application.
=================================================================
/proc/self/maps:
00010000-003a6000 r-xp 00000000 b3:02 9763       /usr/bin/python3.7
003b5000-003b6000 r--p 00395000 b3:02 9763       /usr/bin/python3.7
003b6000-00434000 rw-p 00396000 b3:02 9763       /usr/bin/python3.7
00434000-00447000 rw-p 00000000 00:00 0
01bfa000-02026000 rw-p 00000000 00:00 0          [heap]
[...]

PR #899, opened 2019-06-26
Try to align the native code to the next 8 byte boundary

filmor:

Aligns the ARM code to the next 8 byte boundary.

Maybe #898

### DIFF
--- a/src/runtime/typemanager.cs
+++ b/src/runtime/typemanager.cs
@@ -504,7 +504,7 @@ namespace Python.Runtime
             {
                 get
                 {
-                    switch(Runtime.Machine)
+                    switch (Runtime.Machine)
                     {
                         case Runtime.MachineType.i386:
                             return I386;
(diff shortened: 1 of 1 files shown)


src/runtime/typemanager.cs
```diff
@@ -696,8 +696,21 @@ internal static void InitializeNativeCodePage()
...
             IMemoryMapper mapper = CreateMemoryMapper();
             int codeLength = NativeCode.Active.Code.Length;
-            NativeCodePage = mapper.MapWriteable(codeLength);
-            Marshal.Copy(NativeCode.Active.Code, 0, NativeCodePage, codeLength);
+
+            // Request 7 bytes more than required such that we can align to an 8 byte boundary on ARM
+            NativeCodePage = mapper.MapWriteable(codeLength + 7);
+
+            // Round to 8 byte boundary
+            var start = new IntPtr(((NativeCodePage.ToInt64() + 7) >> 3) << 3);
```
lostmsu: `>> 3) << 3)` can be replaced with `& ~0x7L`
IMHO, even better to write a generic align function (and unit test it)

src/runtime/typemanager.cs
```diff
@@ -698,8 +698,21 @@ internal static void InitializeNativeCodePage()
             // to be executable.
             IMemoryMapper mapper = CreateMemoryMapper();
             int codeLength = NativeCode.Active.Code.Length;
-            NativeCodePage = mapper.MapWriteable(codeLength);
-            Marshal.Copy(NativeCode.Active.Code, 0, NativeCodePage, codeLength);
+
+            // Request 7 bytes more than required such that we can align to an 8 byte boundary on ARM
+            NativeCodePage = mapper.MapWriteable(codeLength + 7);
```
benoithudson: mmap will return minimum a full page, and pages are typically 4096-byte-aligned (certainly more than 8-byte aligned). Add some logging to make sure, but I don't expect your changes are actually doing anything.

After the bx lr you might need a nop? I don't know this ISA myself, but it's not uncommon in architectures with branch prediction (i.e. basically all of them) that it'll unconditionally run an opcode right after a branch.

benoithudson:
You mentioned compiling on the raspberry pi was slow. You can compile the C# dll on any platform -- as long as you set the `#defines` properly, the generated code is compatible. It's just the platform you run on that has to match your `#defines`. We generally build on windows for all three platforms (windows/mac/linux), I've done it on mac as well.

src/runtime/typemanager.cs
benoithudson: https://community.arm.com/developer/ip-products/processors/f/cortex-m-forum/7154/alignment-in-arm

Looks like you need 16-byte-aligned if it's 64-bit ARM. So your Return1 is not properly aligned.

filmor: The older RPis are 32bit, though, so this (8 byte alignment) can't be the current issue.

PR #901, opened 2019-06-26
allow .NET classes to override __getattr__ and __setattr__

lostmsu:

This enables .NET classes to override `__getattr__` and `__setattr__` by implementing `IGetAttr` and `ISetAttr` interfaces correspondingly.

### How does it work

When constructing a Python type, that will represent C# type, if the C# type implements one of the interfaces, fill corresponding slots in the Python type with custom handlers for `tp_getattro` or `tp_setattro` with implementations from `slots.cs`.

### Checklist

### P.S.
This is a single-commit change. Probably could be cherry-picked merged without a merge commit.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ This document follows the conventions laid out in [Keep a CHANGELOG][].
 ### Added
 
 -   Added automatic NuGet package generation in appveyor and local builds
+-   Added IGetAttr and ISetAttr, so that .NET classes could override __getattr__ and __setattr__
 
 ### Changed
 
--- a/src/embed_tests/Python.EmbeddingTest.csproj
+++ b/src/embed_tests/Python.EmbeddingTest.csproj
@@ -89,6 +89,7 @@
     <Compile Include="TestDomainReload.cs" />
     <Compile Include="TestExample.cs" />
     <Compile Include="TestFinalizer.cs" />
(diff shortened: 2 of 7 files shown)


lostmsu:
Looks like .csproj files have varying line endings. Do you guys know how to enforce them in the future?

lostmsu:
@filmor any idea why that change could have broken x86 tests? I just tried to run Python.EmbeddingTest on Windows with x86, and did not get any crashes or issue? Could it be, that a simple rerun needed (then how to do it?)?

filmor:
It looks good now. Regarding the file endings: IMO we should always have Unix style line-endings everywhere. We could set up a `.gitattributes` file for this (like outlined here: https://stackoverflow.com/questions/170961/whats-the-best-crlf-carriage-return-line-feed-handling-strategy-with-git), be my guest :)

lostmsu:
@filmor I don't think it does. x86 builds are still failing in appveyor (unlike, for example, trivial C# version change).

lostmsu:
Damn, Hyper-V on Windows 10 does not boot x86 linuxes (tried Ubuntu 16.04, 18.04 and the latest Debian)

lostmsu:
Closing this out, as we decided on ad-hoc approach instead (e.g. one should simply have a `__getattr__` method in .NET if he wants to override Python's `__getattr__`).

PR #902, opened 2019-06-28
Qc patches

filmor:
This is a rollup of 6 patches from the Quantconnect fork that can be easily applied on master. With this merged, the only things separating the fork from upstream will be

1. The type conversion logic
2. A few changes to the finalizer code
3. The different versioning scheme

@AlexCatarino @Martin-Molinero @AlexCatarino, could you verify that I didn't botch your changes?

### DIFF
--- a/src/runtime/methodbinder.cs
+++ b/src/runtime/methodbinder.cs
@@ -187,6 +187,13 @@ namespace Python.Runtime
                 val += ArgPrecedence(pi[i].ParameterType);
             }
 
+            var info = mi as MethodInfo;
+            if (info != null)
+            {
+                val += ArgPrecedence(info.ReturnType);
+                val += mi.DeclaringType == mi.ReflectedType ? 0 : 3000;
+            }
+
             return val;
         }
 
@@ -199,6 +206,11 @@ namespace Python.Runtime
             if (t == objectType)
             {
                 return 3000;
+            }
+
+            if (t.IsAssignableFrom(typeof(PyObject)))
+            {
+                return -1;
             }
 
             TypeCode tc = Type.GetTypeCode(t);
--- a/src/runtime/propertyobject.cs
+++ b/src/runtime/propertyobject.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Linq.Expressions;
 using System.Reflection;
 using System.Security.Permissions;
 
@@ -12,6 +13,10 @@ namespace Python.Runtime
         private PropertyInfo info;
         private MethodInfo getter;
         private MethodInfo setter;
+        private bool getterCacheFailed;
+        private bool setterCacheFailed;
+        private Func<object, object> getterCache;
+        private Action<object, object> setterCache;
 
         [StrongNameIdentityPermission(SecurityAction.Assert)]
         public PropertyObject(PropertyInfo md)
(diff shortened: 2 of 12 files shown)


src/runtime/methodbinder.cs
```diff
@@ -187,6 +187,13 @@ internal static int GetPrecedence(MethodBase mi)
                 val += ArgPrecedence(pi[i].ParameterType);
             }
 
+            var info = mi as MethodInfo;
+            if (info != null)
+            {
+                val += ArgPrecedence(info.ReturnType);
+                val += mi.DeclaringType == mi.ReflectedType ? 0 : 3000;
```
lostmsu: What does this 3000 mean? Can this be a named constant?

src/runtime/methodbinder.cs
```diff
@@ -201,6 +208,11 @@ internal static int ArgPrecedence(Type t)
                 return 3000;
             }
 
+            if (t.IsAssignableFrom(typeof(PyObject)))
+            {
+                return -1;
```
lostmsu: That makes `TypeCode.Object` entry in the `switch` below unreachable. Is it by design?
Also, negative precedence does not sound good. What is its purpose?

src/runtime/propertyobject.cs
```diff
@@ -160,5 +198,54 @@ public static IntPtr tp_repr(IntPtr ob)
             var self = (PropertyObject)GetManagedObject(ob);
             return Runtime.PyString_FromString($"<property '{self.info.Name}'>");
         }
+
+        private static Func<object, object> BuildGetter(PropertyInfo propertyInfo)
+        {
+            var methodInfo = propertyInfo.GetGetMethod();
+            if (methodInfo == null)
+            {
+                // if the getter is not public 'GetGetMethod' will not find it
```
lostmsu: According to [MSDN](https://docs.microsoft.com/en-us/dotnet/api/system.reflection.propertyinfo.getgetmethod?view=netframework-4.0#System_Reflection_PropertyInfo_GetGetMethod_System_Boolean_), `GetGetMethod` has an overload, that returns non-public getters, so this condition can be removed, along with `getterCacheFailed`. Same for the setters I assume

src/runtime/runtime.cs
```diff
@@ -1741,6 +1741,11 @@ internal static bool PyObject_TypeCheck(IntPtr ob, IntPtr tp)
             return (t == tp) || PyType_IsSubtype(t, tp);
         }
 
+        internal static bool PyObjectType_TypeCheck(IntPtr type, IntPtr tp)
```
lostmsu: For clarity, should rewrite the function above to use this function.

lostmsu:
Also, some tests are failing apparently due to property access changes.

One test crashes during getter/setter delegate construction.

In two other cases the generated lambda converts incoming object into the type, that defined the property. For setters on value types that causes the object to be copied (via unboxing), and the original object is therefore left unmodified. Because nothing enforces in .NET requires getters to be immutable, it might affect getters too (for example, if `Lazy<T>` would be `struct`, its getter would be unable to store the created instance).

filmor:
The tests are probably also failing due to the missing conversion changes that I'd like to discuss separately.

filmor:
I don't see this ever getting merged, so I'll close it.

PR #905, opened 2019-07-02
Safe wheel import

chrisjbremner:
I upgraded from `2.3.0` to `2.4.0` today, which seems to include [the addition of a `wheel` import in the `setup.py` file](https://github.com/pythonnet/pythonnet/blob/78d5d463dd41f54ce3c400e6288f418544c7ae81/setup.py#L18). I am trying to install my own package that includes `pythonnet` on a machine that does not typically have `wheel` installed, therefore, I have to include `wheel` in my `install_requires` file. The line looks like this, note that I have put `wheel` first to install it before `pythonnet`:

```
install_requires = ["wheel", "pythonnet==2.4.0"]  # pythonnet v2.4.0 requires wheel
```

Unfortunately, this still fails due to pip collecting all the packages first, which calls `python setup.py egg_info` for `pythonnet`, before `wheel` is installed (which occurs after all package collection is complete):

```
Collecting wheel (from mypackage)
Downloading pythonnet-2.4.0 (1.8MB)
100% \|████████████████████████████████\| 1.9MB 45.9MB/s
Complete output from command python setup.py egg_info:
  File "", line 1, in
    File "/tmp/pip-install-fvrk22zh/pythonnet/setup.py", line 18, in
      from wheel import bdist_wheel

ModuleNotFoundError: No module named 'wheel'
```

To resolve this, I put the `wheel` import inside of a try/except, so that it doesn't fail while collecting the package, and would only fail if I tried to call `bdist_wheel` if I didn't have `wheel` installed.

[Here's](https://github.com/numba/llvmlite/blob/2aab7f4ba0fd7f6323aa305752644da382933366/setup.py#L10-L13) an example of another `setup.py` file that also try/excepts around the import of `wheel` when setting up the `cmdclass` dictionary.

And a [blog post](https://lepture.com/en/2014/python-on-a-hard-wheel) that describes a similar try/except pattern near the bottom.
...

No
...

### Any other comments?

...

### Checklist

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -19,6 +19,7 @@
 -   Callum Noble ([@callumnoble](https://github.com/callumnoble))
 -   Christian Heimes ([@tiran](https://github.com/tiran))
 -   Christoph Gohlke ([@cgohlke](https://github.com/cgohlke))
+-   Christopher Bremner ([@chrisjbremner](https://github.com/chrisjbremner))
 -   Christopher Pow ([@christopherpow](https://github.com/christopherpow))
 -   Daniel Fernandez ([@fdanny](https://github.com/fdanny))
 -   Daniel Santana ([@dgsantana](https://github.com/dgsantana))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ This document follows the conventions laid out in [Keep a CHANGELOG][].
 ### Changed
 
 -   Added argument types information to "No method matches given arguments" message
+-   Moved wheel import in setup.py inside of a try/except to prevent pip collection failures
 
 ### Fixed
 
--- a/setup.py
+++ b/setup.py
@@ -637,7 +637,7 @@ except NameError:
 
 setup(
     name="pythonnet",
-    version="2.4.0",
+    version="2.4.1-dev",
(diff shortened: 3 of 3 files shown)


filmor:
Can you check whether it suffices to drop the filter in the `requirements.txt`?

chrisjbremner:
Isn't the `requirements.txt` just for development/deployment purposes? If I install it via pypi it doesn't make use of `requirements.txt`, so it doesn't seem like dropping the filter would resolve the issue?

filmor:
Yeah, I got a bit confused because I stumbled over this recently when testing something on a Linux machine where `wheel` wasn't installed by `pip` from the requirements file.

Anyways, thank you very much for your contribution :)

chrisjbremner:
Thanks for the speedy review! What's the process for requesting a minor version bump/release? Is there anything I could help with?

filmor:
There is no process as of now. I'll have a look whether I can do a quick 2.4.1 this month, there were a few improvements over 2.4.0 already.

chrisjbremner:
Great, thank you!

chrisjbremner:
@filmor any update on whether a 2.4.1 release is possible? Thanks!

Issue #906, opened 2019-07-03
CLR operator and Python operator are not bound


tarocco:
### Environment

-   Pythonnet version: 2.4.0
-   Python version: 3.6.5
-   Operating System: Windows 10

### Details

While trying to use the add operator with two CLR classes (structs, actually), I ran into trouble. I expect the + operator to work in Python like it would in C#, but instead it raises an exception.

```python
>>> import sys
>>> import clr
>>> sys.path.append('path/to/libs')
>>> clr.AddReference('LibreMetaverse')  # Contains OpenMetaverse namespace
<System.Reflection.RuntimeAssembly object at 0x0000023EB7403CC0>
>>> from OpenMetaverse import Vector3 
>>> a = Vector3(1.0, 2.0, 3.0)
>>> b = Vector3(2.0, 4.0, 6.0)
>>> a + b  # Results in an error...
```
```
Traceback (most recent call last):
  File "<input>", line 1, in <module>
TypeError: unsupported operand type(s) for +: 'Vector3' and 'Vector3'
```
```python
>>> str(Vector3.Add(a, b))  # But this works!
'<3, 6, 9>'
```

The `Vector3` struct contains
```csharp
...
public static Vector3 Add(Vector3 value1, Vector3 value2)
{
    value1.X += value2.X;
    value1.Y += value2.Y;
    value1.Z += value2.Z;
    return value1;
}
...
public static Vector3 operator +(Vector3 value1, Vector3 value2)
{
    value1.X += value2.X;
    value1.Y += value2.Y;
    value1.Z += value2.Z;
    return value1;
}
...

```
([Entire Vector3.cs here](https://github.com/cinderblocks/libremetaverse/blob/master/LibreMetaverseTypes/Vector3.cs))

Does Pythonnet not bind arithmetic operator methods in CLR to Python magic methods? Or am I missing something?

amos402:
For now, pythonnet just not support the operator, I made a PR with this #907 , but some code is conflict with master, thus this PR may not merge, but I think we will support the operator in the future.

PR #907, opened 2019-07-05
Basic operator overloads support

amos402:
Support the operator for objects.

I found this code from my branch before, not only support the operator but also refactored the `MethodBinder`. @lostmsu I realized we just did the same thing. But a little different was I intended to support the method like `foo(int x = 0, params object[] bar)` can be call with `foo(1,1,1)`, you can see the change from the pytest. I think we need to discuss how to merge them.
I develop it base on Python3, some test cases may fail on 2 now.
#906

### Any other comments?

...

### Checklist

### DIFF
--- a/src/runtime/methodbinder.cs
+++ b/src/runtime/methodbinder.cs
@@ -1,5 +1,7 @@
 using System;
-using System.Collections;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Linq;
 using System.Reflection;
 using System.Text;
 
@@ -13,19 +15,31 @@ namespace Python.Runtime
     /// </summary>
     internal class MethodBinder
     {
-        public ArrayList list;
+        public List<MethodBase> list;
         public MethodBase[] methods;
         public bool init = false;
         public bool allow_threads = true;
 
+        private readonly Dictionary<MethodBase, object[]> _defualtArgs = new Dictionary<MethodBase, object[]>();
+
+        private enum MethodMatchType
+        {
+            NotDefined = 0,
+            Normal,
+            Operator,
+            WithDefaultArgs,
+            WithParamArray,
+            WithDefaultAndParamArray,
+        }
+
         internal MethodBinder()
         {
-            list = new ArrayList();
+            list = new List<MethodBase>();
         }
 
         internal MethodBinder(MethodInfo mi)
         {
-            list = new ArrayList { mi };
+            list = new List<MethodBase>() { mi };
         }
 
         public int Count
@@ -98,12 +112,47 @@ namespace Python.Runtime
             return null;
         }
 
+        internal static IEnumerable<MethodInfo> MatchParamertersMethods(IEnumerable<MethodInfo> mi, Type[] tp)
+        {
+            if (tp == null)
+            {
+                yield break;
+            }
+            int count = tp.Length;
+            foreach (MethodInfo t in mi)
+            {
+                if (!t.IsGenericMethodDefinition)
+                {
+                    continue;
+                }
+                Type[] args = t.GetGenericArguments();
+                if (args.Length != count)
+                {
+                    continue;
+                }
+
+                MethodInfo method;
+                try
+                {
+                    method = t.MakeGenericMethod(tp);
+                }
+                catch (ArgumentException)
+                {
+                    method = null;
+                }
+                if (method == null)
+                {
+                    continue;
+                }
+                yield return method;
+            }
+        }
 
         /// <summary>
         /// Given a sequence of MethodInfo and two sequences of type parameters,
         /// return the MethodInfo that matches the signature and the closed generic.
         /// </summary>
-        internal static MethodInfo MatchSignatureAndParameters(MethodInfo[] mi, Type[] genericTp, Type[] sigTp)
+        internal static MethodInfo MatchSignatureAndParameters(IEnumerable<MethodInfo> mi, Type[] genericTp, Type[] sigTp)
         {
             if (genericTp == null || sigTp == null)
             {
@@ -127,9 +176,14 @@ namespace Python.Runtime
                 {
                     continue;
                 }
+
                 for (var n = 0; n < pi.Length; n++)
                 {
-                    if (sigTp[n] != pi[n].ParameterType)
+                    Type sig = sigTp[n];
+                    Type param = pi[n].ParameterType;
+
+                    if (!param.IsGenericParameter && !IsNullableOf(sig, param) &&
+                        !param.IsAssignableFrom(sig))
                     {
                         break;
                     }
@@ -138,15 +192,33 @@ namespace Python.Runtime
                         MethodInfo match = t;
                         if (match.IsGenericMethodDefinition)
                         {
-                            // FIXME: typeArgs not used
-                            Type[] typeArgs = match.GetGenericArguments();
-                            return match.MakeGenericMethod(genericTp);
+                            try
+                            {
+                                return match.MakeGenericMethod(genericTp);
+                            }
+                            catch (ArgumentException)
+                            {
+                                continue;
+                            }
                         }
                         return match;
                     }
                 }
             }
             return null;
+        }
+
+        private static bool IsNullableOf(Type sigType, Type target)
+        {
+            if (!sigType.IsValueType || !target.IsValueType)
+            {
+                return false;
+            }
+            if (target != typeof(Nullable<>))
+            {
+                return false;
+            }
+            return true;
         }
 
 
@@ -161,7 +233,7 @@ namespace Python.Runtime
             {
                 // I'm sure this could be made more efficient.
                 list.Sort(new MethodSorter());
-                methods = (MethodBase[])list.ToArray(typeof(MethodBase));
+                methods = list.ToArray();
                 init = true;
             }
             return methods;
@@ -201,6 +273,17 @@ namespace Python.Runtime
                 return 3000;
             }
 
+            // Due to array type must be a object, "IsArray" should check first.
+            if (t.IsArray)
+            {
+                Type e = t.GetElementType();
+                if (e == objectType)
+                {
+                    return 2500;
+                }
+                return 100 + ArgPrecedence(e);
+            }
+
             TypeCode tc = Type.GetTypeCode(t);
             // TODO: Clean up
             switch (tc)
@@ -211,23 +294,20 @@ namespace Python.Runtime
                 case TypeCode.UInt64:
                     return 10;
 
-                case TypeCode.UInt32:
+                case TypeCode.Int64:
                     return 11;
 
-                case TypeCode.UInt16:
+                case TypeCode.UInt32:
                     return 12;
 
-                case TypeCode.Int64:
+                case TypeCode.Int32:
                     return 13;
 
-                case TypeCode.Int32:
+                case TypeCode.UInt16:
                     return 14;
 
                 case TypeCode.Int16:
                     return 15;
-
-                case TypeCode.Char:
-                    return 16;
 
                 case TypeCode.SByte:
                     return 17;
@@ -235,27 +315,22 @@ namespace Python.Runtime
                 case TypeCode.Byte:
                     return 18;
 
-                case TypeCode.Single:
+                case TypeCode.Double:
                     return 20;
 
-                case TypeCode.Double:
+                case TypeCode.Single:
                     return 21;
 
                 case TypeCode.String:
                     return 30;
 
+                // A char can be extracted from a string,
+                // so 'char' should larger than string.
+                case TypeCode.Char:
+                    return 31;
+
                 case TypeCode.Boolean:
                     return 40;
-            }
-
-            if (t.IsArray)
-            {
-                Type e = t.GetElementType();
-                if (e == objectType)
-                {
-                    return 2500;
-                }
-                return 100 + ArgPrecedence(e);
             }
 
             return 2000;
@@ -293,29 +368,20 @@ namespace Python.Runtime
             }
 
             // TODO: Clean up
+            bool hasOverloads = _methods.Length > 1;
             foreach (MethodBase mi in _methods)
             {
                 if (mi.IsGenericMethod)
                 {
                     isGeneric = true;
                 }
-                ParameterInfo[] pi = mi.GetParameters();
-                ArrayList defaultArgList;
-                bool paramsArray;
 
-                if (!MatchesArgumentCount(pynargs, pi, out paramsArray, out defaultArgList)) {
-                    continue;
-                }
-                var outs = 0;
-                var margs = TryConvertArguments(pi, paramsArray, args, pynargs, defaultArgList,
-                    needsResolution: _methods.Length > 1,
-                    outs: out outs);
-
+                int outs;
+                var margs = GetInvokeArguments(inst, args, mi, pynargs, hasOverloads, out outs);
                 if (margs == null)
                 {
                     continue;
                 }
-
                 object target = null;
                 if (!mi.IsStatic && inst != IntPtr.Zero)
                 {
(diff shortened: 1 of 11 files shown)


src/runtime/methodbinder.cs
```diff
@@ -13,19 +15,31 @@ namespace Python.Runtime
     /// </summary>
     internal class MethodBinder
     {
-        public ArrayList list;
+        public List<MethodBase> list;
         public MethodBase[] methods;
```
lostmsu: Is there a reason to have both `list` and `methods`?

src/runtime/methodbinder.cs
```diff
@@ -98,12 +112,47 @@ internal static MethodInfo MatchParameters(MethodInfo[] mi, Type[] tp)
             return null;
         }
 
+        internal static IEnumerable<MethodInfo> MatchParamertersMethods(IEnumerable<MethodInfo> mi, Type[] tp)
+        {
+            if (tp == null)
+            {
+                yield break;
```
lostmsu: I think this should throw instead, and the caller should explicitly handle `null` case.

src/runtime/methodbinder.cs
```diff
@@ -98,12 +112,47 @@ internal static MethodInfo MatchParameters(MethodInfo[] mi, Type[] tp)
...
+                Type[] args = t.GetGenericArguments();
+                if (args.Length != count)
+                {
+                    continue;
+                }
+
+                MethodInfo method;
+                try
+                {
+                    method = t.MakeGenericMethod(tp);
```
lostmsu: You can `yield return` right here

src/runtime/methodbinder.cs
```diff
@@ -98,12 +112,47 @@ internal static MethodInfo MatchParameters(MethodInfo[] mi, Type[] tp)
...
+                }
+                catch (ArgumentException)
+                {
+                    method = null;
```
lostmsu: You can `continue` right here

src/runtime/methodbinder.cs
```diff
@@ -127,9 +176,14 @@ internal static MethodInfo MatchSignatureAndParameters(MethodInfo[] mi, Type[] g
...
                 }
+
                 for (var n = 0; n < pi.Length; n++)
                 {
-                    if (sigTp[n] != pi[n].ParameterType)
+                    Type sig = sigTp[n];
+                    Type param = pi[n].ParameterType;
+
+                    if (!param.IsGenericParameter && !IsNullableOf(sig, param) &&
+                        !param.IsAssignableFrom(sig))
```
lostmsu: Can you explain what this method was supposed to be doing, and what your change enables it to do now? It is actually not clear what genericTp and sigTp represent from their names :(

src/runtime/methodbinder.cs
```diff
@@ -149,6 +208,19 @@ internal static MethodInfo MatchSignatureAndParameters(MethodInfo[] mi, Type[] g
...
+        {
+            if (!sigType.IsValueType || !target.IsValueType)
+            {
+                return false;
+            }
+            if (target != typeof(Nullable<>))
+            {
+                return false;
+            }
+            return true;
```
lostmsu: Can you clarify what this method does?
It looks like for it to return `true` `target` must be `Nullable<>`, and `Nullable<int>` won't work.

src/runtime/methodbinder.cs
```diff
@@ -201,6 +273,17 @@ internal static int ArgPrecedence(Type t)
                 return 3000;
             }
 
+            // Due to array type must be a object, "IsArray" should check first.
+            if (t.IsArray)
```
lostmsu: Was this code never called before you moved it up? o-O

src/runtime/methodbinder.cs
```diff
@@ -211,53 +294,45 @@ internal static int ArgPrecedence(Type t)
                 case TypeCode.UInt64:
                     return 10;
 
-                case TypeCode.UInt32:
+                case TypeCode.Int64:
```
lostmsu: These changes to the argument type priority are breaking.
We can consider them for Python.NET 3.0+, but a rationale and corresponding documentation is required.

src/runtime/methodbinder.cs
```diff
@@ -350,192 +416,247 @@ internal Binding Bind(IntPtr inst, IntPtr args, IntPtr kw, MethodBase info, Meth
...
-            if (clrtype != null)
+        private bool GetMultiInvokeArguments(IntPtr inst, IntPtr args, int pynargs,
+            MethodBase mi, ParameterInfo[] pi,
+            bool hasOverloads, object[] margs, ref int outs)
+        {
+            int clrnargs = pi.Length;
+            Debug.Assert(clrnargs > 0);
+            bool isOperator = OperatorMethod.IsOperatorMethod(mi);
+            Type lastType = pi[clrnargs - 1].ParameterType;
+            bool hasArrayArgs = clrnargs > 0 &&
```
lostmsu: Name should be `hasParamsArray` or something similar.

src/runtime/methodbinder.cs
```diff
@@ -350,192 +416,247 @@ internal Binding Bind(IntPtr inst, IntPtr args, IntPtr kw, MethodBase info, Meth
...
-            } else if (argumentCount > parameters.Length && parameters.Length > 0 &&
-                       Attribute.IsDefined(parameters[parameters.Length - 1], typeof(ParamArrayAttribute)))
-            {
-                // This is a `foo(params object[] bar)` style method
-                match = true;
-                paramsArray = true;
             }
 
-            return match;
+            switch (matchType)
```
lostmsu: This method became very long. Can you cut it into smaller ones, so it would be easier to understand?

src/runtime/operator.cs
```diff
@@ -0,0 +1,125 @@
...
+using System.Diagnostics;
+using System.Reflection;
+using System.Runtime.InteropServices;
+using System.Text;
+
+namespace Python.Runtime
+{
+    internal static class OperatorMethod
+    {
+        public static Dictionary<string, Tuple<string, int>> OpMethodMap { get; private set; }
```
lostmsu: Add a comment, describing contents of this Dictionary, or name it so that it would become clear.

src/runtime/operator.cs
```diff
@@ -0,0 +1,125 @@
...
+
+        private static Dictionary<string, string> _pyOpNames;
```
lostmsu: Add a comment, describing contents of this Dictionary, or name it so that it would become clear.

src/runtime/operator.cs
```diff
@@ -0,0 +1,125 @@
...
+            foreach (var method in clrType.GetMethods(flags))
+            {
+                if (!IsOperatorMethod(method))
+                {
+                    continue;
+                }
+                var slotdef = OpMethodMap[method.Name];
+                int offset = slotdef.Item2;
+                IntPtr func = Marshal.ReadIntPtr(_opType.Handle, offset);
+                Marshal.WriteIntPtr(pyType, offset, func);
```
lostmsu: Just clarifying, that this takes the default implementations for operator slots from a class, that defines them explicitly, and copies them to the target type?

src/runtime/operator.cs
```diff
@@ -0,0 +1,125 @@
...
+    {
+        public static Dictionary<string, Tuple<string, int>> OpMethodMap { get; private set; }
+
+        private static Dictionary<string, string> _pyOpNames;
+        private static PyObject _opType;
+
+        static OperatorMethod()
+        {
+            // TODO: Rich compare, inplace operator support
+            OpMethodMap = new Dictionary<string, Tuple<string, int>>
```
lostmsu: This needs explicit key comparer for string keys, the Ordinal one. Otherwise it might fail to work under weird locales.

src/runtime/operator.cs
```diff
@@ -0,0 +1,125 @@
...
+                ["op_BitwiseAnd"] = Tuple.Create("__and__", TypeOffset.nb_and),
+                ["op_BitwiseOr"] = Tuple.Create("__or__", TypeOffset.nb_or),
+                ["op_ExclusiveOr"] = Tuple.Create("__xor__", TypeOffset.nb_xor),
+                ["op_LeftShift"] = Tuple.Create("__lshift__", TypeOffset.nb_lshift),
+                ["op_RightShift"] = Tuple.Create("__rshift__", TypeOffset.nb_rshift),
+                ["op_Modulus"] = Tuple.Create("__mod__", TypeOffset.nb_remainder),
+                ["op_OneComplement"] = Tuple.Create("__invert__", TypeOffset.nb_invert)
+            };
+
+            _pyOpNames = new Dictionary<string, string>();
```
lostmsu: his needs explicit key comparer for string keys, the Ordinal one. Otherwise it might fail to work under weird locales.

src/runtime/operator.cs
```diff
@@ -0,0 +1,125 @@
...
+            if (!method.IsSpecialName)
+            {
+                return false;
+            }
+            return OpMethodMap.ContainsKey(method.Name);
+        }
+
+        public static void FixupSlots(IntPtr pyType, Type clrType)
+        {
+            IntPtr tp_as_number = Marshal.ReadIntPtr(pyType, TypeOffset.tp_as_number);
```
lostmsu: This does not seem to be used anywhere

src/runtime/pythonengine.cs
```diff
@@ -586,7 +586,10 @@ internal static PyObject RunString(string code, IntPtr? globals, IntPtr? locals,
                     code, (IntPtr)flag, globals.Value, locals.Value
                 );
 
-                Runtime.CheckExceptionOccurred();
+                if (result == IntPtr.Zero)
+                {
+                    throw new PythonException();
```
lostmsu: What is the significance of this change?

src/runtime/typemanager.cs
```diff
@@ -172,7 +172,11 @@ internal static IntPtr CreateType(ManagedType impl, Type clrType)
             // that the type of the new type must PyType_Type at the time we
             // call this, else PyType_Ready will skip some slot initialization.
 
-            Runtime.PyType_Ready(type);
+            if (Runtime.PyType_Ready(type) < 0)
+            {
+                throw new PythonException();
+            }
+            OperatorMethod.FixupSlots(type, clrType);
```
lostmsu: Should not the fixup happen before PyType_Ready?

src/tests/test_array.py
```diff
@@ -1125,7 +1125,7 @@ def test_md_array_conversion():
 
     for i in range(5):
         for n in range(5):
-            items.SetValue(Spam(str((i, n))), (i, n))
+            items.SetValue(Spam(str((i, n))), i, n)
```
lostmsu: This looks like a breaking change to me. Why was this affected?

src/tests/test_method.py
```diff
@@ -214,11 +214,12 @@ def test_string_params_args():
     assert result[1] == 'two'
     assert result[2] == 'three'
 
-    result = MethodTest.TestStringParamsArg(['one', 'two', 'three'])
-    assert len(result) == 3
-    assert result[0] == 'one'
-    assert result[1] == 'two'
-    assert result[2] == 'three'
+    # Skip these temporally cause of the changes of array parameter calling
```
lostmsu: Is this change not ready?
.NET permits arrays to be passed in place of `params object[]` arguments. If it was allowed previously, but is not after your change, it removes a feature, which should not be done.

src/runtime/methodbinder.cs
amos402: I just change `ArrayList` to `List<MethodBase>`, the `methods` is the original member for cache in `GetMethods`.

src/runtime/pythonengine.cs
amos402: `CheckExceptionOccurred` take a P/Invoke calling which means take much overhead, I already test it before, that's a huge gap for some computing intensive code.
Besides, some C function may return `nullptr` without setting a error, we should use the return value as we can.
We can just some helpers methods like that
https://github.com/amos402/pythonnet/commit/3591d92ff0597846cdaa23685d56628e3079ceb0#diff-d5e22b611cf0ccb5d1176bd604442282

src/runtime/methodbinder.cs
lostmsu: Fair enough

src/runtime/typemanager.cs
amos402: It could be.

lostmsu: @amos402 From [Python doc](https://docs.python.org/3.5/c-api/type.html#c.PyType_Ready): Finalize a type object. This should be called on all type objects to finish their initialization. This function is responsible for adding inherited slots from a type’s base class. Return 0 on success, or return -1 and sets an exception on error.

Unfortunately, the description does not clarify if the behavior could change in the future :/ I'd try to move `FixupSlots` up, and see if it works that way.

src/tests/test_array.py
amos402: I made a support for calling `foo(int x = 0, params object[] bar)` by `foo(10, 1, 2, 3)` just as the C# code like, this feature is just conflict with master (it needs call `foo(10, [1, 2, 3])`) which I thought it's not pretty enough.

src/runtime/methodbinder.cs
amos402: Yep, it will return by case `case TypeCode.Object:`

src/runtime/operator.cs
amos402: 😂opps, some methods are just used in my code, so it maybe not used here.

src/runtime/operator.cs
amos402: It makes slot defines for Python type, so that we can override the operator in `GetClassInfo`, without this, even you set some operator methods, Python would accept them on operator operation, like `a + b`.

src/runtime/operator.cs
christabella: Maybe something like
> Maps the compiled method name in .NET CIL (e.g. `op_Addition`) to the Python operator (e.g. `__add__`).

src/runtime/operator.cs
tminka: Yes, from [How to compare strings in C#](https://docs.microsoft.com/en-us/dotnet/csharp/how-to/compare-strings#ordinal-sorting-and-searching-in-collections): Collection classes such as System.Collections.Hashtable, System.Collections.Generic.Dictionary<TKey,TValue>, and System.Collections.Generic.List<T> have constructors that take a System.StringComparer parameter when the type of the elements or keys is string. In general, you should use these constructors whenever possible, and specify either StringComparer.Ordinal or StringComparer.OrdinalIgnoreCase.

src/runtime/operator.cs
tminka: Yes, this copies the default implementations for operator slots from a class that defines them explicitly.  As explained in [Python behind the scenes #6: how Python object system works](https://tenthousandmeters.com/blog/python-behind-the-scenes-6-how-python-object-system-works/), the default implementation of the nb_add slot simply calls `__add__` on the type, so we can change `__add__` later and it will be called by the default implementation in the slot.  The code should be updated to have this information in a comment.

src/runtime/operator.cs
amos402: I don't get it, they're not for sorting, and it would only use `GetHasCode` and `Equals` inside, `GetHashCode` would be calculated by their chars, `Equals` would compare by their chars too.  Besides, we all know they're just ASCII strings.

lostmsu: Makes sense. Probably no need to pass comparator then.

src/runtime/operator.cs
lostmsu: @amos402  is correct here, it is not needed.

src/runtime/operator.cs
```diff
@@ -0,0 +1,125 @@
...
+                _opType = null;
+            }
+        }
+
+        public static bool IsOperatorMethod(string methodName)
+        {
+            return OpMethodMap.ContainsKey(methodName);
+        }
+
+        public static bool IsPyOperatorMethod(string pyMethodName)
```
tminka, line 65: This method seems to be unused.

src/runtime/operator.cs
tminka: I think this dictionary can be deleted since it is never used.

tminka:
Fixes #1312

christabella:
Could be closed in favour of the updated version at https://github.com/pythonnet/pythonnet/pull/1324/

Issue #874, opened 2019-05-30
some wrong in python.runtime


shmliao:
when using python.runtime,some questions puzzle me, many mistakes happens, like "System.IO.FileNotFoundException","System.IO.FileLoadException",System.IO.FileNotFoundException,maybe this is the reason leadd my python code running slow,very slowly,I cant tolerate.my code is just a simple demo.so l wanna know how you resolve this problem,thanks,my python version is 3.6

Zeraphil:
Can confirm this is happening to me as well. Memory usage skyrockets to ~500MB with Python 3.6 in Release mode x64 standard config. If the binary is run this doesn't happen

filmor:
@Zeraphil What do you mean by "if the binary is run this doesn't happen"?

Zeraphil:
Hi. I mean that if I run the executable without the attached debugger, memory usage is comparable to the Python interpreter. So the question is more appropriately, why does the visual studio debugger use up so much memory when attached to a program that references the PythonNet runtime.

rmadsen-ks:
I have also experienced slowdowns the Visual Studio debugger blocks execution for a tiny while when any exceptions happens. This means that exception-heavy code becomes very slow with the debugger attached.

filmor:
Could you please check whether the linked pull-request improves this?

rmadsen-ks:
I tried to build and add use the code. I still get exceptions, but note they are first-time exceptions. They don't get thrown out of the runtime binding methods.


![image](https://user-images.githubusercontent.com/24250062/60965882-4218af00-a317-11e9-8665-1fd199703c55.png)

Stack trace:
```
 	Microsoft.CSharp.dll!Microsoft.CSharp.RuntimeBinder.RuntimeBinderController.SubmitError(Microsoft.CSharp.RuntimeBinder.Errors.CError pError)	Unknown
 	Microsoft.CSharp.dll!Microsoft.CSharp.RuntimeBinder.Semantics.MemberLookup.ReportErrors()	Unknown
[...]

PR #911, opened 2019-07-10
Don't throw exceptions during dynamic access

filmor:
The return value is meant to carry the failure information. This should make code using `dynamic` quite a bit quicker.

Does not through exceptions anymore in dynamic member access and instead communicates failure via the return value.

#874


filmor:
I think this fixes a possible memory leak as well, in the case where we fail to set an attribute.

filmor:
Yes, I think we should detect `AttributeError`s and otherwise pass on the underlying exception. It does indeed break the API slightly, but it should have always worked this way.

lostmsu:
@filmor can you think of a scenario, where this change is necessary?
TBH, documentation on .NET docs is not clear about the expected behavior in case member is present, but attempting to access it throws some exception.

filmor:
The scenario is the linked issue where a lot of useless exceptions are being thrown, maybe during probing the dynamic objects.

lostmsu:
@filmor just came up with 2 ways to do it properly:
1. (obvious) Check using `HasAttr` first
2. Only in case of exception, check `HasAttr`. If it does not, return `false`, otherwise rethrow.

filmor:
Ah, sorry, I didn't see your message. Yes, option 2 sounds quite reasonable, although I still think it's a bit too defensive. Nevertheless, I'll change it like this and leave a comment.

lostmsu:
@filmor I'd argue if you can't set a member in Python, the exception you get must be meaningful in nearly 100% cases?

lostmsu:
I dunno. I liked the TryGetMember part, as it does reduce the debug output.

PR #913, opened 2019-07-10
Readd .NET implementations of GC slots again

filmor:
Since I failed to provide properly working implementations of Return0 and Return1 for ARM, I'll just add the old behaviour back for all platforms but x86 and amd64.

#898


Cronan:
Looks like this would work except for codecov

filmor:
Yes, I still think this is a good idea, but it sadly didn't fix #898 as a I hoped it would...

PR #925, opened 2019-07-30
Removes imports deprecated in Python3

Cronan:
### What does this implement/fix? Explain your changes.
A small number of existing imports are deprecated in Python3. 
This removes them or compiles them out.

### Does this close any currently open issues?
Fixes #924 
...

### Any other comments?

N/A

### Checklist

Check all those that are applicable and complete.

-   [N/A] Make sure to include one or more tests for your change
-   [N/A] If an enhancement PR, please create docs and at best an example
-   [X] Add yourself to [`AUTHORS`](../blob/master/AUTHORS.md)
-   [X] Updated the [`CHANGELOG`](../blob/master/CHANGELOG.md)

PR #927, opened 2019-08-01
Simplify travis config and pin mono to 5.20

filmor:

Pins Mono to 5.20 and simplifies the build.

Breaking build because of Mono 6.

### Any other comments?

Should help in running #925 .


Cronan:
Also needed for #928 (Joe is a colleague)

PR #928, opened 2019-08-01
Adds support to convert iterators to arrays

jmlidbetter:

Python 3 has much more liberal use of iterators, e.g. getting keys or values from a dictionary. This adds functionality to convert python object which support the iterator protocol to arrays in C#.

No

### Any other comments?

N/A

### Checklist

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -34,6 +34,7 @@
 -   Jan Krivanek ([@jakrivan](https://github.com/jakrivan))
 -   Jeff Reback ([@jreback](https://github.com/jreback))
 -   Joe Frayne ([@jfrayne](https://github.com/jfrayne))
+-   Joe Lidbetter ([@jmlidbetter](https://github.com/jmlidbetter))
 -   John Burnett ([@johnburnett](https://github.com/johnburnett))
 -   John Wilkes ([@jbw3](https://github.com/jbw3))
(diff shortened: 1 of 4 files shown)


filmor:
This looks like a good addition feature-wise, but there is quite a bit of duplicated code in it. I think it would be better if we had /one/ implementation based on the iterator protocol, that just uses an additional size hint if `len` is implemented.

jmlidbetter:
> This looks like a good addition feature-wise, but there is quite a bit of duplicated code in it. I think it would be better if we had /one/ implementation based on the iterator protocol, that just uses an additional size hint if `len` is implemented.

Sure. Presumably sequence protocol => iterator protocol?

filmor:
It says so in the docs: https://docs.python.org/2/library/stdtypes.html#iterator-types as well as https://docs.python.org/3/library/stdtypes.html#iterator-types:

>  Sequences [...] always support the iteration methods.

jmlidbetter:
> It says so in the docs: https://docs.python.org/2/library/stdtypes.html#iterator-types as well as https://docs.python.org/3/library/stdtypes.html#iterator-types:
> 
> > Sequences [...] always support the iteration methods.

Ok, have redone - always uses iterator protocol now.

Cronan:
@filmor Passed!

filmor:
This looks great, thanks :)

Issue #904, opened 2019-07-01
Crash when using pythonnet with nuitka


ftreurni:
### Environment

-   Pythonnet version: 2.4.0
-   Python version: Python 2.7.13
-   Operating System: Windows 10 

### Details

-   Describe what you were trying to get done.

    I want to use a .net library that I access through pythonnet in combination with a library that was compiled with Nuitka (the contents of the library do not seem to matter, nor the version of Nuitka). However, when I import the module in the pyd first, calling 'import clr' after that crashes pythonnet and then Python. 

I was wondering if this is a known issue/how easy it would be to fix it/if there would be some straightforward workaround so the order of imports would not matter?

-   What commands did you run to trigger this issue? If you can provide a
    [Minimal, Complete, and Verifiable example](http://stackoverflow.com/help/mcve)
    this will help us understand the issue.

```python
    import foo
    import clr
```

Will crash pythonnet and then Python. Changing the order of the imports will make it work, however it will make working with the .net library very annoying.

Relevant files are in the zip:
[minimal_example.zip](https://github.com/pythonnet/pythonnet/files/3339737/minimal_example.zip)

-   If there was a crash, please include the traceback here.

```python
    Unhandled Exception: System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation. ---> System.AccessViolationException: Attempted to read or write protected memory. This is often an indication that other memory is corrupt.
   at Python.Runtime.Runtime.PyModule_GetDict(IntPtr module)
   at Python.Runtime.PythonEngine.Initialize(IEnumerable`1 args, Boolean setSysArgv, Boolean initSigs)
   at Python.Runtime.PythonEngine.InitExt()
   --- End of inner exception stack trace ---
   at System.RuntimeMethodHandle.InvokeMethod(Object target, Object[] arguments, Signature sig, Boolean constructor)
   at System.Reflection.RuntimeMethodInfo.UnsafeInvokeInternal(Object obj, Object[] parameters, Object[] arguments)
   at System.Reflection.RuntimeMethodInfo.Invoke(Object obj, BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture)
   at System.RuntimeType.InvokeMember(String name, BindingFlags bindingFlags, Binder binder, Object target, Object[] providedArgs, ParameterModifier[] modifiers, CultureInfo culture, String[] namedParams)
   at clrModule.initclr()
```

amos402:
[...]

PR #929, opened 2019-08-01
Fix so that Runtime.PyModuleType is retrieved via Python.dll (#904)

ftreurni:

Currently the PyModuleType is deduced from the ``__builtin__`` module. However, this fails when some third party package messes with this module (Nuitka). In this PR the PyModuleType is deduced directly from the Python shared library instead, which should be a safer approach to achieve the same.

#904 

### Any other comments?

...

### Checklist


Cronan:
Seems like a good thing to me

filmor:
Thank you very much for this precise analysis and the fix, greatly appreciated! :)

PR #930, opened 2019-08-02
Improve performance of unwrapping .NET objects passed from Python

filmor:
Upstreams https://github.com/QuantConnect/pythonnet/pull/31 by @lostmsu.

Comment by the author:

This addresses the following scenario:

1. A C# object `a` is created and filled with some data.
2. `a` is passed via Python.NET to Python. To do that Python.NET creates a wrapper object `w`, and stores reference to `a` in one of its fields.
3. Python code later passes `w` back to C#, e.g. calls `SomeCSharpMethod(w)`.
4. Python.NET has to unwrap `w`, so it reads the reference to `a` from it.

Prior to this change in 4. Python.NET had to determine what kind of an object `a` is. If it is an exception, a different offset needed to be used. That check was very expensive (up to 4 calls into Python interpreter).

This change replaces that check with computing offset unconditionally by subtracting a constant from the object size (which is read from the wrapper), thus avoiding calls to Python interpreter.


lostmsu:
Now we need to update the perf goal accordingly. @filmor, can cherry-pick c2ddaa0e1aa366bf3b9cd2261e524fa33025bf18 into this branch?

filmor:
Sorry for the delay, cherry-picking didn't work (I assume the given commit is not in any of the branches on the main fork).

lostmsu:
Hm, it shows in this fork. This is a very small change though, you can copy it manually.

filmor:
I did (just before commenting ;)).

Issue #933, opened 2019-08-05
Calling a method that has a generic T[] argument causes interpreter crash


jmlidbetter:
### Environment

-   Pythonnet version: Commit hash c97a380db
-   Python version: 3.6.5
-   Operating System: Linux / mono

### Details

-   Describe what you were trying to get done.

    Call a method that takes a generically typed array as an argument without specifying the generic type.

-   What commands did you run to trigger this issue? If you can provide a
    [Minimal, Complete, and Verifiable example](http://stackoverflow.com/help/mcve)
    this will help us understand the issue.

```C#
namespace Test
{
    public class AClass
    {
        public AClass() { }

        public static int AMethod<T>(T[] genericArray)
        {
            return 1;
        }
    }
}    
```

```python
def method():
    from Test import AClass

    l = [1, 2, 3]
    return AClass.AMethod(l)
```

-   If there was a crash, please include the traceback here.

```
Unhandled Exception:
System.NotSupportedException: Array type can not be an open generic type
  at System.Array.CreateInstance (System.Type elementType, System.Int32[] lengths) [0x0007f] in <7660324592e4462eb91bfb91f5746699>:0 
  at System.Array.CreateInstance (System.Type elementType, System.Int64[] lengths) [0x0005f] in <7660324592e4462eb91bfb91f5746699>:0 
  at Python.Runtime.Converter.ToArray (System.IntPtr value, System.Type obType, System.Object& result, System.Boolean setError) [0x00022] in <0f2d060fb81447228b68596c3e5ddd48>:0 
  at Python.Runtime.Converter.ToManagedValue (System.IntPtr value, System.Type obType, System.Object& result, System.Boolean setError) [0x000de] in <0f2d060fb81447228b68596c3e5ddd48>:0 
  at Python.Runtime.Converter.ToManaged (System.IntPtr value, System.Type type, System.Object& result, System.Boolean setError) [0x00010] in <0f2d060fb81447228b68596c3e5ddd48>:0 
  at Python.Runtime.MethodBinder.TryConvertArgument (System.IntPtr op, System.Type parameterType, System.Boolean needsResolution, System.Object& arg, System.Boolean& isOut) [0x0001b] in <0f2d060fb81447228b68596c3e5ddd48>:0 
[...]

PR #934, opened 2019-08-05
Catch any unhandled exceptions in TryConvertArgument

jmlidbetter:

Fixes Issue #933

#933

### Any other comments?

TryConvertArgument ought not to throw any exceptions - exceptions => failure to convert so should do so gracefully.

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ This document follows the conventions laid out in [Keep a CHANGELOG][].
 
 - Fixed runtime that fails loading when using pythonnet in an environment
   together with Nuitka
+- Handle any exceptions thrown by TryConvertArgument (#933)
 
 ## [2.4.0][]
 
--- a/src/runtime/methodbinder.cs
+++ b/src/runtime/methodbinder.cs
@@ -417,22 +417,33 @@ namespace Python.Runtime
         static bool TryConvertArgument(IntPtr op, Type parameterType, bool needsResolution,
                                        out object arg, out bool isOut)
         {
-            arg = null;
-            isOut = false;
-            var clrtype = TryComputeClrArgumentType(parameterType, op, needsResolution: needsResolution);
-            if (clrtype == null)
+            try
             {
+                arg = null;
+                isOut = false;
+                var clrtype = TryComputeClrArgumentType(parameterType, op, needsResolution: needsResolution);
+                if (clrtype == null)
+                {
+                    return false;
+                }
+
+                if (!Converter.ToManaged(op, clrtype, out arg, false))
+                {
+                    Exceptions.Clear();
+                    return false;
+                }
+
+                isOut = clrtype.IsByRef;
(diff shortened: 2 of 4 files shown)


Cronan:
@filmor OK to merge?

filmor:
I'll have another look and merge pending PRs after I'm back from vacation next week

Cronan:
Thanks muchly. Hope you’re having a good time.

lostmsu:
This is a very poor solution to the problem, as it masks any real conversion problems, that might occur.
From the first look `ToArray` should explicitly check for this scenario.

jmlidbetter:
> 
> 
> This is a very poor solution to the problem, as it masks any real conversion problems, that might occur.
> From the first look `ToArray` should explicitly check its inputs to handle #933 , or `ToManagedValue` should to the same before calling `ToArray`.

jmlidbetter:
> 
> 
> This is a very poor solution to the problem, as it masks any real conversion problems, that might occur.
> From the first look `ToArray` should explicitly check its inputs to handle #933 , or `ToManagedValue` should to the same before calling `ToArray`.

I do partially agree with you that this should be handled at a lower level and you're more than welcome to propose such a fix yourself. Ultimately, however, a "TryXYZ" should not crash the python interpretter. If it returns False an error will get propogated through the correct channels and the user can report bugs as necessary. In my opinion, crashing the interpretter is a truly last resort and should be avoided at all costs. NB python is not like C# - often people will be running things in, for example,  stateful jupyter notebooks where a crash can destroy (rightfully or otherwise) hours of work.

lostmsu:
@jmlidbetter handling all exceptions can lead to something worse, than a crash: data corruption. You'd rather know there was a problem and redo 2 hours of work, than silently use flawed analysis results for years, until the consequences of corrupted data hit you in the face.

Cronan:
@jmlidbetter and I have discussed this in depth - the current state of pythonnet as far as uncaught exceptions go is not good - there are many places where non-fatal exceptions can bubble all the way to the top, which causes a segfault, and the Python interpreter to die.

This makes the library look flaky, and makes people waste time digging into hard to diagnose errors.
This kind of error should be writing warnings and errors to stdout (or better, to the python log), not causing the interpreter to crash. That should be reserved for genuinely fatal cases, as mentioned by @lostmsu above.

The reason this is in a PR because this specific case causes an interpreter crash in one of the tests in .NET Core.

lostmsu:
@Cronan at least me & @filmor agree about bad smell of the code, that swallows all exceptions.

I've already proposed one option: to handle this case in the downstream array converter. Another is to go up to `MethodBinder.Invoke`, and wrap the whole method in `try - catch` block, and in the `catch` use `Exceptions.SetError`

Cronan:
That all sounds good, and I’m sure someone will get to this at some point, we’ve got no further time to work on this right now - we’ve patched it in our fork.

PR #550, opened 2017-10-03
Marshalling fix for 3-IntPtr args cdecl calls.

dmitriyse:
### What does this implement/fix? Explain your changes.
Workaround for the DotNet Core 2.0 marshaling bug.
Problem described here https://github.com/pythonnet/pythonnet/issues/96#issuecomment-329542908
...

### Does this close any currently open issues?

...

### Any other comments?

...

### Checklist

Check all those that are applicable and complete.

-   [ ] Make sure to include one or more tests for your change
-   [ ] If an enhancement PR, please create docs and at best an example
-   [ ] Add yourself to [`AUTHORS`](../blob/master/AUTHORS.md)
-   [ ] Updated the [`CHANGELOG`](../blob/master/CHANGELOG.md)

PR #936, opened 2019-08-07
Remove delegate casting

jmlidbetter:

Fixes an issue where pythonnet attempts to cast one managed delegate to another. This is explicitly forbidden in the documentation: [see here](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.marshal.getdelegateforfunctionpointer?view=netstandard-2.0#System_Runtime_InteropServices_Marshal_GetDelegateForFunctionPointer__1_System_IntPtr_). In particular:

> You cannot use this method to create a delegate from a function pointer to another managed delegate

I have also removed the usage of the obselete function GetDelegateForFunctionPointer(IntPtr, Type) and replaced it with GetDelegateForFunctionPointer<TDelegate><IntPtr).

No

### Any other comments?

It seems like this issue got partially fixed in PR #550 and was mentioned in issue #96. See also this related [issue](https://github.com/dotnet/coreclr/issues/14362) in dotnet core.

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,11 +17,13 @@ This document follows the conventions laid out in [Keep a CHANGELOG][].
 -   Moved wheel import in setup.py inside of a try/except to prevent pip collection failures
 -   Removes PyLong_GetMax and PyClass_New when targetting Python3
 -   Added support for converting python iterators to C# arrays
+-   Changed usage of obselete function GetDelegateForFunctionPointer(IntPtr, Type) to GetDelegateForFunctionPointer<TDelegate>(IntPtr)
 
 ### Fixed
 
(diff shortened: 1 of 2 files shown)


Cronan:
@filmor I don't believe I can kick the build ...

lostmsu:
@jmlidbetter can you open a bug for the issue you are trying to fix with repro steps?

jmlidbetter:
@lostmsu You can repro the issue by compiling in dotnet core which I don't believe is officially supported so didnt think an issue is strictly necessary in this case. Check out pr 550 for more background / precedent on the changes as well as the other things mentioned above :)

lostmsu:
Looks good to me.

Issue #849, opened 2019-04-16
Support to call C# named parameters from Python


Lexmaul:
Hi, 
this is a change request:
### Environment

-   Pythonnet version: 2.3.0.0
-   Python version: 3.6.5 32-bit
-   Operating System: Win 10

### Details
In a python module, we want to call a C# Method, that has named optional parameters. But these are ignored. 
Reason: The MethodBinder.Bind(IntPtr inst, IntPtr args, IntPtr kw, MethodBase info, MethodInfo[] methodinfo) Method ignores the passed python keyword-arguments. But instead, it should assign them to the C# named parameters, if appropriate. Can you extend pythonnet to support named parameters?

### snip
-   C#
``` C#:
public class LibWithDefault
{
    public static string FuncWithDefault(int x, object b = null)
    {
        return $"x: {x} b: {b} ";
    }
}
```
- Python
```python
import clr
clr.AddReference("CS_DefaultParam")
from CS_DefaultParam import LibWithDefault
res = LibWithDefault.FuncWithDefault(x=3, b=4)
print(res)
```

filmor:
Yes, this would be a fine addition and we're very open to PRs on this subject. Related, there is #80 that could be tackled at the same time.

s4v4g3:
PR submitted for this feature.  By the way, it looks like #80 is already addressed since version 2.1.  I added some tests to make sure named parameters & default parameters can work together OK.

PR #953, opened 2019-09-10
Feature/named arg support

s4v4g3:
Adds support for kwarg/named arguments when calling .NET methods from python
...

Yes, #849 
...

### Any other comments?

...

### Checklist

-   [x ] Make sure to include one or more tests for your change

### DIFF
--- a/src/runtime/methodbinder.cs
+++ b/src/runtime/methodbinder.cs
@@ -2,6 +2,8 @@ using System;
 using System.Collections;
 using System.Reflection;
 using System.Text;
+using System.Collections.Generic;
+using System.Linq;
 
 namespace Python.Runtime
 {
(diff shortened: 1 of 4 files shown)


src/runtime/methodbinder.cs
```diff
@@ -505,29 +536,49 @@ static Type TryComputeClrArgumentType(Type parameterType, IntPtr argument, bool
...
+                    {
+                        // we have a keyword argument for this parameter,
+                        // no need to check for a default parameter, but put a null
+                        // placeholder in defaultArgList
+                        defaultArgList.Add(null);
+                    }
+                    else
+                    {
+                        // find default parameter
+                        if (parameters[v].DefaultValue == DBNull.Value)
```
filmor: This is not quite right, I think. According to https://stackoverflow.com/questions/9977530/default-parameters-and-reflection-if-parameterinfo-isoptional-then-is-defaultva?rq=1, the CLR allows for optional values that don't supply a specific default value, in which case `default(T)` should be used.

So essentially the logic needs to be:

    if (parameters[v].IsOptional) {
        object defaultValue;
        if (parameters[v].HasDefaultValue) { // Shorter than checking for DBNull.Value in 4.5
            defaultValue = parameters[v].DefaultValue;
        }
        else {
            var type = parameters[v].ParameterType;
            defaultValue = type.IsValueType ? Activator.CreateInstance(type) : null;
        }
        defaultArgList.Add(defaultValue);
    }
    else {
        match = false;
    }

s4v4g3: This makes sense, I had forgotten about the use of the [OptionalAttribute] tag to designate an optional parameter.  I will integrate your suggested change and add another test case.

s4v4g3: One complication I ran into is that .NET 4.0 doesn't support HasDefaultValue, so I used the Attributes property to test for ParameterAttributes.HasDefault.

Also, I found [here ](https://stackoverflow.com/questions/3416216/optionalattribute-parameters-default-value)that the logic for the default value of an [Optional] parameter was slightly more complicated, so I moved it to an extension method.

Issue #950, opened 2019-08-30
Dotnetcore System.UInt32 conversion issues


jmlidbetter:
### Environment

-   Pythonnet version: 2.4.0
-   Python version: 3.6.5
-   Operating System: Centos 7 64bit

### Details

-   Creating UInt32 in python with a value greater than 0x7FFFFFFF in dotnetcore will cause an overflow error. This is due (seemingly) to the fact that .NET core clr will cast the `uint` to a signed long (64 bits in size on my system) before calling the PyLong_FromUsignedLong function. Using GDB and breaking before the errornous overflow we can examine the created PyLongObject:

The `PyLongObject` given to C# by the python interpretter that can be converted to a `uint` using `PyLong_AsUnsignedLong`

![image](https://user-images.githubusercontent.com/53430310/64010679-2d37eb00-cb12-11e9-8323-3eb72be82f25.png)

Using `PyLong_FromUnsignedLong` on the resulting `uint` gives this:

![image](https://user-images.githubusercontent.com/53430310/64010712-3cb73400-cb12-11e9-96ec-3c7ae962433f.png)

I have raised an issue on the coreclr project: [Issue 26444](https://github.com/dotnet/coreclr/issues/26444)

FYI changing the function definition from
```C#
internal static extern IntPtr PyLong_FromUnsignedLong(uint value);
```
to
```C#
internal static extern IntPtr PyLong_FromUnsignedLong(ulong value);
```
fixes the issue.

filmor:
Well, this is problematic. We want to be platform-independent, but C's `unsigned long` that is used in Python is 64bit on most 64bit Unices and 32bit on Windows, so the `uint` parameter is indeed wrong.

Cronan:
I was wondering if there’s a Python C API call that might give us the unsigned king size at runtime, so we could decide which PInvoke call to make?

filmor:
As far as I can tell, Python is completely oblivious to this difference as it "lives" on the C platform, so all compiled modules will have the same idea of what `long` means. We can pretty reliably decide this by looking at the platform we are running on in .NET, though. We are already providing internally booleans for `IsWindows` and `Is32Bit` here https://github.com/pythonnet/pythonnet/blob/master/src/runtime/runtime.cs#L106-L109.

jmlidbetter:
> 
> 
[...]

PR #961, opened 2019-09-25
Detect py arch

jmlidbetter:

As described in Issue #950, the size of python's longs can vary depending on platform. This changes allows pythonnet to determine at runtime the architecture of the python interpreter and select the correct size of long/int for the affected methods.

#950 

### Any other comments?

No

### Checklist

### DIFF
--- a/src/runtime/runtime.cs
+++ b/src/runtime/runtime.cs
@@ -155,6 +155,11 @@ namespace Python.Runtime
         /// </summary>
         public static string MachineName { get; private set; }
 
+        /// <summary>
+        /// Gets the architecture the python interpreter is using as reported by python's struct.calcsize("P")
+        /// </summary>
+        public static MachineType PythonArchitecture { get; private set; }
+
         internal static bool IsPython2 = pyversionnumber < 30;
         internal static bool IsPython3 = pyversionnumber >= 30;
 
@@ -192,6 +197,9 @@ namespace Python.Runtime
 
             IntPtr op;
             IntPtr dict;
+
+            InitializePythonArch();
+
             if (IsPython3)
             {
                 op = PyImport_ImportModule("builtins");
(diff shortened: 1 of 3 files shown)


src/runtime/runtime.cs
```diff
@@ -371,6 +379,43 @@ private static void InitializePlatformData()
...
+                case 4:
+                    PythonArchitecture = MachineType.i386;
+                    break;
+                case 8:
+                    PythonArchitecture = MachineType.x86_64;
+                    break;
+                default:
+                    PythonArchitecture = MachineType.Other;
+                    break;
+            }
```
filmor: This is just not true.

1. On Windows (and possibly other systems) you will get 4 even when running in 64bit mode, that is what makes things complicated
2. ARMs exist

If we could always rely on `sizeof(void*) == sizeof(long)`, we could easily use `IntPtr` in places where `long` is used, but it's not that simple.

Also, I don't see why we would need to draw this information from Python, Python's architecture must always be the same as the arch .NET is running on.

We also have the information we need already: https://github.com/pythonnet/pythonnet/pull/497/files#diff-7e0cf65d55c3fae27decb3c07faf3e29

src/runtime/runtime.cs
```diff
@@ -1036,8 +1081,24 @@ internal static bool PyLong_Check(IntPtr ob)
...
+        internal static IntPtr PyLong_FromUnsignedLong(object value)
+        {
+            if(PythonArchitecture == MachineType.i386)
+                return PyLong_FromUnsignedLong32(Convert.ToUInt32(value));
+            else if(PythonArchitecture == MachineType.x86_64)
+                return PyLong_FromUnsignedLong64(Convert.ToUInt64(value));
+            else
+                // Couldn't determine interpreter arch, try 64 bit
+                return PyLong_FromUnsignedLong64(Convert.ToUInt64(value));
+        }
```
filmor: This is kind of fine, just use the already provided `Runtime.Is32Bit` and `Runtime.IsWindows` instead.

filmor:
Thanks for the contribution, the fix to the forwarded functions should do the trick, just use the already existing properties instead of trying to draw this from Python :)

src/runtime/runtime.cs
jmlidbetter: Yes you are correct. Should be calcsize("L") ... but if you're happy that checking Is32Bit and IsWindows is sufficent then will rework the other methods and remove this.

src/runtime/runtime.cs
```diff
@@ -371,6 +379,43 @@ private static void InitializePlatformData()
...
+        /// This method will allow pythonnet to determine at runtime how big
+        /// python's longs are.
+        /// </summary>
+        private static void InitializePythonArch()
+        {
+            IntPtr structModule = PyImport_ImportModule("struct");
+            IntPtr calcsizeMethod = PyObject_GetAttrString(structModule, "calcsize");
+            IntPtr methodArgs = PyTuple_New(1);
+            IntPtr pString = PyString_FromString("P");
+
```
amos402: All these Python objects didn't decrease the reference count.

PR #963, opened 2019-09-26
Drop List and IEnumerable conversions

filmor:

Drop implicit conversion of `List<>` and `IEnumerable`.

Heaps, I'll list them later

### Any other comments?

Breaks the interface, so we'll have to announce this loudly and raise the minor version number.

### Checklist

### DIFF
--- a/src/runtime/converter.cs
+++ b/src/runtime/converter.cs
@@ -155,7 +155,7 @@ namespace Python.Runtime
             // type is. we'd rather have the object bound to the actual
             // implementing class.
 
-            type = type ?? value.GetType();
+            type = type == null || type == typeof(object) ? value.GetType() : type;
 
             TypeCode tc = Type.GetTypeCode(type);
 
--- a/src/runtime/converter.cs
+++ b/src/runtime/converter.cs
@@ -155,7 +155,7 @@ namespace Python.Runtime
             // type is. we'd rather have the object bound to the actual
             // implementing class.
 
-            type = value.GetType();
+            type = type ?? value.GetType();
 
             TypeCode tc = Type.GetTypeCode(type);
 
--- a/src/runtime/converter.cs
+++ b/src/runtime/converter.cs
@@ -155,7 +155,7 @@ namespace Python.Runtime
             // type is. we'd rather have the object bound to the actual
             // implementing class.
 
-            type = type == null || type == typeof(object) ? value.GetType() : type;
+            type = type ?? value.GetType();
 
             TypeCode tc = Type.GetTypeCode(type);
 
--- a/src/runtime/converter.cs
+++ b/src/runtime/converter.cs
@@ -135,22 +135,6 @@ namespace Python.Runtime
                 return result;
             }
 
-			if (value is IList && !(value is INotifyPropertyChanged) && value.GetType().IsGenericType)
-			{
-                using (var resultlist = new PyList())
-                {
-                    foreach (object o in (IEnumerable)value)
-                    {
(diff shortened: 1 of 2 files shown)


lostmsu:
Just want to confirm if we are to bump up major version with it. Seems like a non-trivial change, that might break a lot of things, that are working now.

It is not immediately obvious from this change, but if I used to pass an array to a parameter, that expects Python `list`, would this no longer work anymore?

filmor:
I'm trying to keep the amount of changes in behaviour to a minimum. I'm not going to bump the major version (i.e. this will not be 3.0) but instead the minor version 2.5 as I've had enough of the "lists behave weirdly" kind of issues we are seeing since this was introduced.

.NET array objects are still `IEnumerable<>` and thus will keep being iterable, so for Python functions that don't (wrongly) check the type explicitly this will still work. We can as a further step also implement the Python Sequence Protocol for `ICollection<>`, but one step after another :)

lostmsu:
@filmor what about functions, that do `len(arg)`?

filmor:
That is the second paragraph of my comment ;)

`IEnumerable` doesn't allow you to do an efficient `len` either, that's what `ICollection<>` is for and we would have to implement at least `sq_len` then.

lostmsu:
@filmor I meant that previously when you pass a `List<T>` to a Python function, that uses `len(lst)` it would work, but if `List<T>` is no longer converted to `list`, that will stop working.

filmor:
`List<T>` implements `ICollection<T>` which has a `Count` property which should in turn be used to implement the `sq_len` slot of the sequence protocol.

lostmsu:
@filmor "should be used" but not "already used". Are you planning to implement `sq_length` too before the next release?

filmor:
This PR is a draft, I plan to work on it until all tests pass. If that means implementing `sq_length`, then that's what I am going to do.

Cronan:
What's the rationale for this please?

filmor:
The rationale is the huge number of issues related to the automatic conversion. Just being able to use a .NET `IEnumerable` as an iterable and `ICollection` as a sequence from Python catches most use cases of the conversion without breaking code like `l = obj.GetCSharpList(); obj.CallWithCSharpList(l)`.

Cronan:
> The rationale is the huge number of issues related to the automatic conversion. Just being able to use a .NET `IEnumerable` as an iterable and `ICollection` as a sequence from Python catches most use cases of the conversion without breaking code like `l = obj.GetCSharpList(); obj.CallWithCSharpList(l)`.

OK, thanks. It would be nice to close down a bunch of issues.

Cronan:
Any chance of getting a look at those issues?

koubaa:
@filmor @lostmsu Even if we drop it if we provide a codec (see 1022) that is not registered by default users can easily opt back into the current behavior by registering it themselves.

mmisol:
@filmor First of all, thanks for providing such a great library!

Do you think this can go in as part of the 2.5 release? I see it's not in the RC, but would love to see this fixed.

filmor:
@mmisol Sorry, 2.5 is a minor release, we can't break the API even though it would be an improvement.

lostmsu:
@mmisol injecting a no-op codec will allow you to override this behavior with 2.5: https://github.com/pythonnet/pythonnet/blob/master/src/runtime/Codecs/RawProxyEncoder.cs

lostmsu:
I think this has been superseded in 3.0 branch. These default conversions are no longer enabled but can be reimplemented by codecs.

PR #971, opened 2019-10-17
Add function to set Py_NoSiteFlag global variable to 1

alexhelms:
I'm using pythonnet to embed python 3.6 in a C# WPF application. I'm effectively booting up a venv with pythonnet and I noticed the user site folder kept getting included in `sys.path`.

There is a command line argument to disable this, (-S, more info here: https://docs.python.org/3/using/cmdline.html#id3) but as pythonnet loads python directly, this won't work.

There is a way to access this flag with the C API, info here: https://docs.python.org/3/c-api/init.html#c.Py_NoSiteFlag

This is a global variable that is exported, defined here: https://github.com/python/cpython/blob/3.6/Include/pydebug.h

My particular use case is for python 3.6, but it appears this flag is present in 2.7 and all currently supported 3.x versions.

Here is an example on how I am using this addition:

```
PythonEngine.PythonHome = pythonHome;
PythonEngine.PythonPath = pythonPath;
PythonEngine.SetNoSiteFlag();
PythonEngine.Initialize();
```

It seems to do the business and I no longer get the user site package path in `sys.path`.

Note: I'm not quite sure how to test this addition as really is dependent on the user's environment. Perhaps a test to verify the global var is set to 1? Please let me know your thoughts.

### DIFF
--- a/src/runtime/runtime.cs
+++ b/src/runtime/runtime.cs
@@ -1900,11 +1900,7 @@ namespace Python.Runtime
             try
             {
                 Py_NoSiteFlag = loader.GetFunction(dllLocal, "Py_NoSiteFlag");
-
-                if (Is32Bit)
-                    Marshal.WriteInt32(Py_NoSiteFlag, 1);
-                else
-                    Marshal.WriteInt64(Py_NoSiteFlag, 1);
+                Marshal.WriteInt32(Py_NoSiteFlag, 1);
             }
             finally
             {
--- a/src/runtime/runtime.cs
+++ b/src/runtime/runtime.cs
@@ -414,6 +414,8 @@ namespace Python.Runtime
         internal static IntPtr PyNoneType;
         internal static IntPtr PyTypeType;
 
+        internal static IntPtr Py_NoSiteFlag;
+
 #if PYTHON3
         internal static IntPtr PyBytesType;
 #endif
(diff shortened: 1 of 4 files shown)


src/runtime/runtime.cs
```diff
@@ -1884,5 +1886,33 @@ internal static IntPtr PyMem_Realloc(IntPtr ptr, long size)
...
+            }
+
+            try
+            {
+                Py_NoSiteFlag = loader.GetFunction(dllLocal, "Py_NoSiteFlag");
+
+                if (Is32Bit)
+                    Marshal.WriteInt32(Py_NoSiteFlag, 1);
+                else
+                    Marshal.WriteInt64(Py_NoSiteFlag, 1);
```
filmor: This is not correct, `int` on Windows is 32bit.

alexhelms: I don't understand, the code specifies Int32 or Int64 depending on arch. What would you suggest this be instead?

filmor: The arch defines the Pointer width, this flag is not a pointer but an `int`. It's actually 32bit pretty much everywhere, not only on Windows (the distinction has to be made for `long`). Drop the 64bit path and this is fine.

alexhelms: Thanks, changes pushed.

Issue #623, opened 2018-02-10
c# <-> python dictionary converter


denfromufa:
https://stackoverflow.com/questions/48689604/calling-into-python-code-using-pythonnet-from-c-sharp-and-custom-c-sharp-classes

yagweb:
This function is similar to the function I proposed here #484. These two functions can be  integrated into single one，“How to convert between python objects and CLR objects dynamically？”

Based on my code before, I just wrote a new [PyConverter](https://github.com/yagweb/pythonnetLab/blob/master/pynetLab/PyConverter.cs) here. I can make a PR if this design is accepted.

It can be used [in these ways](https://github.com/yagweb/pythonnetLab/blob/master/Test/TestPyConverter.cs),

1. Convert between [Python list and Clr List\<T\>][Test1] dynamically;
2. Convert between [Python dict and Clr Dictionary<K, V>][Test2] dynamically;
3. Convert between [Python  object and Clr object][Test3], looks like Serialize and Deserialize.
4. Convert any [complex objects][Test4], such as nested dictionaries, lists and objects etc.
4. It's extensible. Users can [define his own PyClrType][Func1].

[Test1]: https://github.com/yagweb/pythonnetLab/blob/aba0011a33812b304537f6f3b6a14e69bfab1787/Test/TestPyConverter.cs#L70-L92
[Test2]: https://github.com/yagweb/pythonnetLab/blob/aba0011a33812b304537f6f3b6a14e69bfab1787/Test/TestPyConverter.cs#L94-L119
[Test3]: https://github.com/yagweb/pythonnetLab/blob/aba0011a33812b304537f6f3b6a14e69bfab1787/Test/TestPyConverter.cs#L39-L68
[Test4]: https://github.com/yagweb/pythonnetLab/blob/aba0011a33812b304537f6f3b6a14e69bfab1787/Test/TestPyConverter.cs#L121-L169
[Func1]: https://github.com/yagweb/pythonnetLab/blob/aba0011a33812b304537f6f3b6a14e69bfab1787/Test/TestPyConverter.cs#L23-L37

For this issue from stackoverflow, we can use the PyConverter in this way,

First [define a PyConverter]() according to needs,
```
public PyConverter NewMyDictConverter()
{
    using (Py.GIL())
    {
        var converter = new PyConverter(); 
        converter.AddListType();
        converter.Add(new StringType());
        converter.Add(new Int64Type());
        converter.Add(new Int32Type());
[...]

PR #972, opened 2019-10-18
Add C# Dictionary to Python Dict conversion

SnGmng:

You can now convert a C# IDictionary to a Python Dict by calling ToPython()

    Dictionary<string, string> dict = new Dictionary<string, string>();
    dict.Add("hello", "world");

    dynamic d = dict.ToPython();
    string value = d.get("hello", "");

No, but it is progress for #623

### Any other comments?

Do i really have to include a test for this? I mean, the [TestConverter.cs](../blob/master/src/embed_tests/TestConverter.cs) file doesnt have any tests for other similar types like `IList` at all.

### Checklist

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -64,6 +64,7 @@
 -   ([@OneBlue](https://github.com/OneBlue))
 -   ([@rico-chet](https://github.com/rico-chet))
 -   ([@rmadsen-ks](https://github.com/rmadsen-ks))
+-   ([@SnGmng](https://github.com/SnGmng))
 -   ([@stonebig](https://github.com/stonebig))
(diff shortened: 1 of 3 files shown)


filmor:
Thanks a lot for this. However, as noted in #963, I'd really like to get away from the implicit conversions and instead allow the wrapped C# object to *act* like a Python `dict` without actually performing the conversion (unless explicitly requested).

SnGmng:
Oh thats nice :) Will it take long until the implicit conversions get replaced? Because maybe this PR could still be a temporary solution to allow the use of `IDictionary` as a python `dict`.

In #963 you only mention `IEnumerable` and `ICollection`. Will there be a solution for `IDictionary` too?

PR #974, opened 2019-10-21
Improve method binding

koubaa:
Two method binding changes:
Support passing lists/tuples to methods which take IEnumerable, ICollection, IList
Support passing functions to methods which take Action, Func, Task

Not that I am aware of.

### Checklist

### DIFF
--- a/src/runtime/ListWrapper.cs
+++ b/src/runtime/ListWrapper.cs
@@ -1,0 +1,201 @@
+using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+
+namespace Python.Runtime {
+    /// <summary>
+    /// Implements IEnumerable<typeparamref name="T"/> for any python iterable.
+    /// </summary>
+    internal class IterableWrapper<T> : IEnumerable<T> {
+        private IntPtr iterObject;
+
+        public IterableWrapper(IntPtr value) {
+            iterObject = Runtime.PyObject_GetIter(value);
+            if (iterObject == IntPtr.Zero)
+                Exceptions.RaiseTypeError("not an iterator");
+            Runtime.XIncref(iterObject);
+        }
+        ~IterableWrapper() {
+            Runtime.XDecref(iterObject);
+        }
+
+        public IEnumerator<T> GetEnumerator() {
+            return GetEnumerator();
+        }
+
+        IEnumerator IEnumerable.GetEnumerator() {
+            IntPtr item;
+            while ((item = Runtime.PyIter_Next(iterObject)) != IntPtr.Zero) {
+                object obj = null;
+                if (!Converter.ToManaged(item, typeof(T), out obj, true)) {
+                    Runtime.XDecref(item);
+                    Runtime.XDecref(iterObject);
+                    Exceptions.RaiseTypeError("wrong type in sequence");
+                }
+
+                Runtime.XDecref(item);
+                yield return obj;
+            }
+        }
+    }
+
+    /// <summary>
+    /// Implements IList<typeparamref name="T"/> for any python sequence.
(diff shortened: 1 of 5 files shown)


koubaa:
@filmor @lostmsu I am still working on the implementation here (the Action method binding doesn't work atm) but would appreciate some preliminary review to make sure I'm headed in the right direction.  Thanks!

koubaa:
allow_threads in methodbinder.cs Invoke is true which is causing the memory corruption.  If I set it to false there is no corruption so I need to look into why.  This looks like it may be new code

src/runtime/converter.cs
```diff
@@ -893,12 +894,89 @@ private static bool ToArray(IntPtr value, Type obType, out object result, bool s
...
+
+            Action action = () => {
+                PyObject py_action = new PyObject(value);
+                var py_args = new PyObject[0];
+                var py_result = py_action.Invoke(py_args);
+
+                //Discard the result since this is being converted to an Action
+                py_result.Dispose();
+                py_action.Dispose();
+            };
```
lostmsu: The object pointed to by `value` can be collected between the construction of this `action`, and the moment it is invoked. It would lead to a memory corruption.

src/runtime/converter.cs
```diff
@@ -847,44 +864,28 @@ private static void SetConversionError(IntPtr value, Type target)
...
 
-            while ((item = Runtime.PyIter_Next(IterObject)) != IntPtr.Zero)
-            {
+            while ((item = Runtime.PyIter_Next(IterObject)) != IntPtr.Zero) {
                 object obj = null;
 
-                if (!Converter.ToManaged(item, elementType, out obj, true))
-                {
+                if (!Converter.ToManaged(item, elementType, out obj, true)) {
                     Runtime.XDecref(item);
```
lostmsu: Please, revert whitespace-only changes.

src/runtime/converter.cs
koubaa: I had forgotten to push the latest change set, please check again

amos402:
Not suggesting this readonly implicit conversion, what if I want to use like this
```c#
void AddElem(IList<int> lst)
{
    lst.add(10);
}
```
```python
def func():
    lst = [1, 2, 3]
    AddElem(lst)
    # I want [1,2,3,10] actually
```
I suggest implement some subclasses from List<> so you can track the changes of the collection and apply them to the Python's list
Also the test didn't check any outputs, it only certify the runtime has no exceptions.

koubaa:
@amos402 see the latest changes.  I'm dealing with some crash/exception in native code but I think I'm going in the direction that you are proposing.

koubaa:
@lostmsu since we last spoke I looked into codec and it looks like string encoding, so I'm not sure how it could apply to method binding. What is your preferred way to support non-automatic conversions for me to consider/try?

src/runtime/ListWrapper.cs
```diff
@@ -0,0 +1,201 @@
...
+        private IntPtr iterObject;
+
+        public IterableWrapper(IntPtr value) {
+            iterObject = Runtime.PyObject_GetIter(value);
+            if (iterObject == IntPtr.Zero)
+                Exceptions.RaiseTypeError("not an iterator");
+            Runtime.XIncref(iterObject);
+        }
+        ~IterableWrapper() {
+            Runtime.XDecref(iterObject);
```
amos402: You can't use `XDecref` directly in destructor, see `Runtime.PyObject`'s destructor, you can just refer a `PyObject`, it can let you have no need concern about about the release.

lostmsu:
Opened https://github.com/pythonnet/pythonnet/pull/1022 for my take on generic approach.

koubaa:
This is superseded by #1084 and #1080

PR #975, opened 2019-10-22
Adds performance tests, that compare to published NuGet

lostmsu:
A new project `src\perf_tests\Python.PerformanceTests.csproj`, that uses BenchmarkDotNet to benchmark compare current version with the latest version published on NuGet (2.3 as of now).

During build the baseline and new version are copied into separate directories. Benchmark code then loads new or baseline based on an environment variable.

In the end, an "optimistic performance ratio" is computed, that has to be less than some target (usually close to 1.0) for the performance test to pass.


lostmsu:
Not sure I made the CI/build script changes correctly. Plus we might have to restrict testing to Python 3.5 for now, as it is the only version supported by a published NuGet.

lostmsu:
Excluded CI changes so that this can be merged independently.

lostmsu:
@filmor ready to review
AppVeyor result: https://ci.appveyor.com/project/pythonnet/pythonnet/builds/28304985

PR #980, opened 2019-10-27
Add python buffer api support

SnGmng:

Passing data like images from managed to python will be easyer than ever!
I implemented the python buffer api which makes copying big chunks of data directly into the python object buffer possible!

You can copy strings to python object:

    PyObject array = scope.Eval("bytearray(3)");
    byte[] managedArr = new UTF8Encoding().GetBytes(new char[] { 'a', 'b', 'c' });

    PyBuffer buf = array.GetBuffer(out bool success, 0);
    if (success)
    {
        buf.WriteToBuffer(managedArr, 0, managedArr.Length);
        buf.Release();
    }

Or copy images to python byte arrays:

    byte[] imageArr = LoadImage();
    PyObject array = scope.Eval("bytearray(" + managedArr.Length + ")");

    PyBuffer buf = array.GetBuffer(out bool success, 0);
    if (success)
    {
        buf.WriteToBuffer(managedArr, 0, managedArr.Length);
        buf.Release();
    }

No

### Any other comments?

...

### Checklist


src/runtime/pyobject.cs
```diff
@@ -942,6 +945,27 @@ public override int GetHashCode()
...
+        /// </remarks>
+        public PyBuffer GetBuffer(out bool success, int flags)
+        {
+            int size = Marshal.SizeOf(typeof(Runtime.Py_buffer));
+            byte[] rawData = new byte[size];
+            GCHandle handle = GCHandle.Alloc(rawData, GCHandleType.Pinned);
+            IntPtr view = handle.AddrOfPinnedObject();
+
+            success = Runtime.PyObject_GetBuffer(obj, view, flags) >= 0;
+            return new PyBuffer(view, handle);
```
filmor: Why isn't this part of `PyBuffer`'s constructor?

src/runtime/runtime.cs
```diff
@@ -959,6 +959,60 @@ internal static long PyObject_Size(IntPtr pointer)
...
+
+        [DllImport(_PythonDll, CallingConvention = CallingConvention.Cdecl)]
+        internal static extern void PyBuffer_FillContiguousStrides(int ndims, IntPtr shape, IntPtr strides, int itemsize, char order);
+
+        [DllImport(_PythonDll, CallingConvention = CallingConvention.Cdecl)]
+        internal static extern int PyBuffer_FillInfo(IntPtr view, IntPtr exporter, IntPtr buf, long len, int _readonly, int flags);
+
+        /* buffer interface */
+        [StructLayout(LayoutKind.Sequential, CharSet = CharSet.Auto)]
+        public struct Py_buffer
```
filmor: This shouldn't be public.

filmor:
By the way, the failing CI is probably since you didn't add `PyBuffer.cs` to the old project files.

SnGmng:
Fixed most of your comments. But i dont know why git marked almost all lines of the old csproject file as changed, even though i only added on line. I hope this is no problem :)

filmor:
You probably changed the line endings

src/runtime/pybuffer.cs
```diff
@@ -0,0 +1,167 @@
...
+
+        /// <summary>
+        /// Release the buffer view and decrement the reference count for view->obj. This function MUST be called when the buffer is no longer being used, otherwise reference leaks may occur.
+        /// It is an error to call this function on a buffer that was not obtained via <see cref="PyObject.GetBuffer()"/>.
+        /// </summary>
+        public void Release()
+        {
+            Runtime.PyBuffer_Release(_handle);
+            _gchandle.Free();
+        }
```
lostmsu: Can you implement `IDisposable` and a finalizer?

src/runtime/pybuffer.cs
```diff
@@ -0,0 +1,167 @@
...
+
+        internal PyBuffer(IntPtr obj, int flags, out bool success)
+        {
+            int size = Marshal.SizeOf(typeof(Runtime.Py_buffer));
+            byte[] rawData = new byte[size];
+            _gchandle = GCHandle.Alloc(rawData, GCHandleType.Pinned);
+            _handle = _gchandle.AddrOfPinnedObject();
+            _view = (Runtime.Py_buffer)Marshal.PtrToStructure(_handle, typeof(Runtime.Py_buffer));
+
+            success = Runtime.PyObject_GetBuffer(obj, _handle, flags) >= 0;
```
lostmsu: Why return success/failure via `out`, and not simply throw an exception?

src/runtime/pybuffer.cs
```diff
@@ -0,0 +1,167 @@
...
+
+            if (_view.shape != IntPtr.Zero)
+            {
+                Shape = new long[_view.ndim];
+                Marshal.Copy(_view.shape, Shape, 0, (int)_view.len * sizeof(long));
+            }
```
lostmsu: When the buffer failed to be retrieved, work on the object continues?

If this failed, IMHO, all operations with side effects from above need to be reversed.

src/runtime/pybuffer.cs
```diff
@@ -0,0 +1,167 @@
...
+        private Runtime.Py_buffer _view;
+        private GCHandle _gchandle;
+
+        internal PyBuffer(IntPtr obj, int flags, out bool success)
+        {
+            int size = Marshal.SizeOf(typeof(Runtime.Py_buffer));
+            byte[] rawData = new byte[size];
+            _gchandle = GCHandle.Alloc(rawData, GCHandleType.Pinned);
+            _handle = _gchandle.AddrOfPinnedObject();
+            _view = (Runtime.Py_buffer)Marshal.PtrToStructure(_handle, typeof(Runtime.Py_buffer));
```
lostmsu: This is the only line where `_view` is initialized, however, at this moment the memory `_handle` points to is empty (e.g. all zeroes). So the `_view` is also always empty.
None of the `if`s below ever succeed.

src/runtime/pybuffer.cs
```diff
@@ -0,0 +1,167 @@
...
+                Marshal.Copy(_view.strides, Strides, 0, (int)_view.len * sizeof(long));
+            }
+
+            if (_view.suboffsets != IntPtr.Zero) {
+                SubOffsets = new long[_view.ndim];
+                Marshal.Copy(_view.suboffsets, SubOffsets, 0, (int)_view.len * sizeof(long));
+            }
+        }
+
+        public PyObject Object => new PyObject(_view.obj);
```
lostmsu: Creating `PyObject` does not increment reference counter of the object. You have to manually do it in this getter.

src/runtime/pybuffer.cs
```diff
@@ -0,0 +1,167 @@
...
+        public long[] SubOffsets { get; }
+
+        /// <summary>
+        /// Release the buffer view and decrement the reference count for view->obj. This function MUST be called when the buffer is no longer being used, otherwise reference leaks may occur.
+        /// It is an error to call this function on a buffer that was not obtained via <see cref="PyObject.GetBuffer()"/>.
+        /// </summary>
+        public void Release()
+        {
+            Runtime.PyBuffer_Release(_handle);
+            _gchandle.Free();
```
lostmsu: It would be good to reset these handles to zero, and check, that the buffer has not been released yet.

src/runtime/runtime.cs
```diff
@@ -959,6 +959,58 @@ internal static long PyObject_Size(IntPtr pointer)
...
+        internal static extern int PyBuffer_FillInfo(IntPtr view, IntPtr exporter, IntPtr buf, long len, int _readonly, int flags);
+
+        /* buffer interface */
+        [StructLayout(LayoutKind.Sequential, CharSet = CharSet.Auto)]
+        internal struct Py_buffer
+        {
+            public IntPtr buf;
+            public IntPtr obj;        /* owned reference */
+            public long len;
+            public long itemsize;  /* This is Py_ssize_t so it can be
```
lostmsu: `Py_size_t` is not the same as `Int64` aka `long`. There is an ongoing discussion about sizes of Python types.

src/runtime/pybuffer.cs
filmor: ... in which case you also have to release resources.

src/runtime/pybuffer.cs
SnGmng: Your wrong. `_view` _should_ be empty, because PyObject_GetBuffer fills `_view` with data. Every `if` below works as expected

src/runtime/runtime.cs
SnGmng: Well it works with `long` and i didnt see a reason to use `Py_ssize_t` if it will get casted to long anyway. I looked it up in the python source and it litteraly `typdef`s `Py_ssize_t` as long. What do you recommend to do?

src/runtime/pybuffer.cs
SnGmng: But PyObject_GetBuffer() does. And PyBuffer_Release decrements the reference count... Just read the docs first. https://docs.python.org/3/c-api/buffer.html

src/runtime/pybuffer.cs
SnGmng: If GetBuffer fails, `Shape`, `SubOffsets` and `Strides` are `IntPtr.Zero` anyway.

src/runtime/pybuffer.cs
SnGmng: Your right. My bad.

src/runtime/pybuffer.cs
SnGmng: Yeah that would be better

SnGmng:
> This PR needs unit tests with good coverage. Currently, basic functionality is non-functional.
Yeah it needs unit tests. But why are you claiming its non functional? Did you even test this PR? It works as expected. Just try out one of the examples i provided, you will see.

Apart from that, im going to implement all of your other requested changes like implementing `IDisposable` later this day

src/runtime/pybuffer.cs
lostmsu: But in this can you actually get a new reference, so you have to increase the recount, or cache the instance of pyobject in constructor.

src/runtime/pybuffer.cs
lostmsu: @SnGmng you might be using some funky runtime, which makes this happen. `PyObject_GetBuffer` does not get a pointer to `_view`. It gets a pointer to the `rawData` instead. The call to `Marshal.PtrToStructure` simply creates a copy of memory it is pointing to.

SnGmng: Oh my bad i messed up the order of these two lines when i was moving the code into the constructor. The PR was working before this commit tho.

src/runtime/pybuffer.cs
lostmsu: @SnGmng you are right, that it usually takes a pre-increfed pointer. But each instance requires its own incref. Otherwise when the instance is disposed, all other instances of `PyObject` pointing to the same reference will simply crash on access.

src/runtime/runtime.cs
SnGmng: @lostmsu What about something like this:    

    [StructLayout(LayoutKind.Sequential, CharSet = CharSet.Auto)]
    internal struct Py_ssize_t
    {
        [MarshalAs(UnmanagedType.SysInt)]
        public IntPtr size;

        public long GetValue()
        {
            if (Is32Bit)
                return size.ToInt32();
            else
                return size.ToInt64();
        }
    }

SnGmng:
Ive implemented most of the requested changes, but i dont know what you want me to do with the `Py_ssize_t` fields? Leave them as `long`s? Or use a struct like i proposed?

src/runtime/runtime.cs
lostmsu: @SnGmng this struct might not work either, because on 64-bit windows `IntPtr` is 64-bit, but `Py_ssize_t` is probably 32-bit. https://stackoverflow.com/questions/384502/what-is-the-bit-size-of-long-on-64-bit-windows

And that problem will have to be solved, before PR is accepted.

lostmsu:
@SnGmng unfortunately, the only way I can think of so far is to have two struct types defined, and decide which one to use at runtime.

src/runtime/runtime.cs
filmor: long and thus Py_ssize_t is 4 bytes on Windows (32 and 64 bit) and 8 bytes on other 64 bit systems. Very annoying.

lostmsu:
BTW, there appears to be a number of warnings regarding XML doc comments on the new file in the build, I guess related to the recent refactoring: https://ci.appveyor.com/project/pythonnet/pythonnet/builds/28472200/job/3otegb2hqnq9f6on

src/runtime/pybuffer.cs
```diff
@@ -0,0 +1,228 @@
...
+
+        /// <summary>
+        /// Writes a managed byte array into the buffer of a python object. This can be used to pass data like images from managed to python.
+        /// </summary>
+        public void Write(byte[] buffer, int offset, int count)
+        {
+            if (disposedValue)
+                throw new ObjectDisposedException("PyBuffer");
+            if (ReadOnly)
+                throw new Exception("Buffer is read-only");
```
filmor: `InvalidOperationException`?

src/runtime/pybuffer.cs
```diff
@@ -0,0 +1,228 @@
...
+            return Runtime.PyBuffer_ToContiguous(buf, _handle, Length, order);
+        }
+
+        /// <summary>
+        /// Fill the strides array with byte-strides of a contiguous (C-style if order is 'C' or Fortran-style if order is 'F') array of the given shape with the given number of bytes per element.
+        /// </summary>
+        /// <param name="order">C-style if order is 'C' or Fortran-style if order is 'F'</param>
+        public static void FillContiguousStrides(int ndims, IntPtr shape, IntPtr strides, int itemsize, char order)
+        {
+            Runtime.PyBuffer_FillContiguousStrides(ndims, shape, strides, itemsize, order);
```
filmor: Missing check for `disposedValue`

src/runtime/pybuffer.cs
```diff
@@ -0,0 +1,228 @@
...
+        /// </summary>
+        public long[] SubOffsets { get; private set; }
+
+        /// <summary>
+        /// Return the implied itemsize from format. On error, raise an exception and return -1.
+        /// New in version 3.9.
+        /// </summary>
+        public static long SizeFromFormat(string format)
+        {
+            if (Runtime.pyversionnumber < 39)
```
filmor: Missing check for `disposedValue`

src/runtime/pybuffer.cs
```diff
@@ -0,0 +1,228 @@
...
+            int copylen = count < _view.len ? count : (int)_view.len;
+            Marshal.Copy(buffer, offset, _view.buf, copylen);
+        }
+
+        /// <summary>
+        /// Reads the buffer of a python object into a managed byte array. This can be used to pass data like images from python to managed.
+        /// </summary>
+        public int Read(byte[] buffer, int offset, int count) {
+            if (disposedValue)
+                throw new ObjectDisposedException("PyBuffer");
```
filmor: Use `nameof` for easier refactoring.

src/runtime/pybuffer.cs
```diff
@@ -0,0 +1,228 @@
...
+                throw new PythonException();
+            }
+
+            _exporter = exporter;
+            _view = (Runtime.Py_buffer)Marshal.PtrToStructure(_handle, typeof(Runtime.Py_buffer));
+
+            if (_view.shape != IntPtr.Zero)
+            {
+                Shape = new long[_view.ndim];
+                Marshal.Copy(_view.shape, Shape, 0, (int)_view.len * sizeof(long));
```
filmor: This is the same issue as noted below, on 32bit Windows (and actually also on 32bit Linux) this will fail, it might actually segfault.

filmor:
Sorry for the confusion, we had this discussion already before (in #531) and the conclusion is that `Py_ssize_t` is indeed `IntPtr` (not `long`, though!), so if you change the `long`s in the `Py_buffer` definition and ensure that the copy is done correctly this is fine. (Currently you do the copy to `Strides` etc. assuming you have 64bit data).

src/runtime/runtime.cs
filmor: See my comment below, I made this mistake (assuming `Py_ssize_t == long`) before ;)

`Py_ssize_t` is a typedef for `intptr_t` or `ssize_t` (on POSIX), so in all relevant cases it is equivalent in size to `IntPtr`.

src/runtime/pybuffer.cs
SnGmng: its static

src/runtime/pybuffer.cs
SnGmng: its static

src/runtime/pybuffer.cs
filmor: Ah sorry, missed that :)

lostmsu:
@filmor it might be fine for parameter passing which will occupy an entire register anyway, but for structs incorrect size will cause the following fields to be incorrectly placed, leading to wrong memory being accessed.
If our tests are run on Windows x86 (which we should do exactly for this kind of issue) they will likely catch the problem.

filmor:
Hmm, it still seems to not quite work on x86, I'll try to reproduce this locally.

lostmsu:
@filmor weird, should not be a problem on x86.

filmor:
> @filmor it might be fine for passing function parameters which will occupy an entire register anyway, but for structs incorrect size will cause the following fields to be incorrectly placed, leading to wrong memory being accessed.
> 
> If our tests are run on Windows x64 (which we should do exactly for this kind of issue) they will likely catch the problem.

I'm not sure what exactly this one was the answer to, but `IntPtr` *is* the correct type (on both 32 and 64 bit systems) and should lead to the correct layout.

Interestingly enough, non-xplat built Py3.7 ran through on x86...

lostmsu:
@filmor you are correct. @SnGmng needs to investigate why non-xplat x86 builds fail.

SnGmng:
Well, i tested Python 3.6 x86 + non-xplat. No errors at all.
Screenshot_111

Since i cant spot any error in the failing appveyor logs at all and my own test allways succeeds, Im beyond my capabilities. Maybe there is some config differences in the appveryor builds that cause the error? Maybe it randomly detects unmanaged memory copys and fails? I dont know man...

filmor:
Yeah, it doesn't really fail but times out. I'll try to reproduce it locally (running exactly what AppVeyor is supposed to run), but I don't use Windows as my primary dev platform, so I need a while to set everything up.

filmor:
@lostmsu Could you rereview already?

src/runtime/pybuffer.cs
```diff
@@ -0,0 +1,228 @@
...
+        public long[] SubOffsets { get; private set; }
+
+        /// <summary>
+        /// Return the implied itemsize from format. On error, raise an exception and return -1.
+        /// New in version 3.9.
+        /// </summary>
+        public static long SizeFromFormat(string format)
+        {
+            if (Runtime.pyversionnumber < 39)
+                throw new NotSupportedException("GetPointer requires at least Python 3.9");
```
lostmsu: Should the message say "SizeFromFormat ..." not "GetPointer ..."?

src/runtime/pybuffer.cs
```diff
@@ -0,0 +1,228 @@
...
+            return (long)Runtime.PyBuffer_SizeFromFormat(format);
+        }
+
+        /// <summary>
+        /// Returns true if the memory defined by the view is C-style (order is 'C') or Fortran-style (order is 'F') contiguous or either one (order is 'A'). Returns false otherwise.
+        /// </summary>
+        /// <param name="order">C-style (order is 'C') or Fortran-style (order is 'F') contiguous or either one (order is 'A')</param>
```
lostmsu: NIT: would be nice to have an enum

src/runtime/pybuffer.cs
```diff
@@ -0,0 +1,228 @@
...
+            if (Runtime.pyversionnumber < 37)
+                throw new NotSupportedException("GetPointer requires at least Python 3.7");
+            return Runtime.PyBuffer_GetPointer(_handle, indices.Select(x => (IntPtr)x).ToArray());
+        }
+
+        /// <summary>
+        /// Copy contiguous len bytes from buf to view. fort can be 'C' or 'F' (for C-style or Fortran-style ordering). 0 is returned on success, -1 on error.
+        /// </summary>
+        /// <returns>0 is returned on success, -1 on error.</returns>
+        public int FromContiguous(IntPtr buf, long len, char fort)
```
lostmsu: Since this is a public method, why not throw a `PythonException` on failure?

src/runtime/pybuffer.cs
```diff
@@ -0,0 +1,228 @@
...
+        /// <param name="order">order can be 'C' or 'F' or 'A' (for C-style or Fortran-style ordering or either one).</param>
+        /// <param name="buf">Buffer to copy to</param>
+        /// <returns>0 is returned on success, -1 on error.</returns>
+        public int ToContiguous(IntPtr buf, char order)
+        {
+            if (disposedValue)
+                throw new ObjectDisposedException(nameof(PyBuffer));
+            if (Runtime.pyversionnumber < 36)
+                throw new NotSupportedException("ToContiguous requires at least Python 3.6");
+            return Runtime.PyBuffer_ToContiguous(buf, _handle, _view.len, order);
```
lostmsu: Same, why not throw?

src/runtime/pybuffer.cs
```diff
@@ -0,0 +1,228 @@
...
+        /// The flags argument indicates the request type. This function always fills in view as specified by flags, unless buf has been designated as read-only and PyBUF_WRITABLE is set in flags.
+        /// On success, set view->obj to a new reference to exporter and return 0. Otherwise, raise PyExc_BufferError, set view->obj to NULL and return -1;
+        /// If this function is used as part of a getbufferproc, exporter MUST be set to the exporting object and flags must be passed unmodified.Otherwise, exporter MUST be NULL.
+        /// </remarks>
+        /// <returns>On success, set view->obj to a new reference to exporter and return 0. Otherwise, raise PyExc_BufferError, set view->obj to NULL and return -1;</returns>
+        public int FillInfo(IntPtr exporter, IntPtr buf, long len, int _readonly, int flags)
+        {
+            if (disposedValue)
+                throw new ObjectDisposedException(nameof(PyBuffer));
+            return Runtime.PyBuffer_FillInfo(_handle, exporter, buf, (IntPtr)len, _readonly, flags);
```
lostmsu: And here

src/runtime/pybuffer.cs
```diff
@@ -0,0 +1,228 @@
...
+        /// FillInfo Method
+        /// </summary>
+        /// <remarks>
+        /// Handle buffer requests for an exporter that wants to expose buf of size len with writability set according to readonly. buf is interpreted as a sequence of unsigned bytes.
+        /// The flags argument indicates the request type. This function always fills in view as specified by flags, unless buf has been designated as read-only and PyBUF_WRITABLE is set in flags.
+        /// On success, set view->obj to a new reference to exporter and return 0. Otherwise, raise PyExc_BufferError, set view->obj to NULL and return -1;
+        /// If this function is used as part of a getbufferproc, exporter MUST be set to the exporting object and flags must be passed unmodified.Otherwise, exporter MUST be NULL.
+        /// </remarks>
+        /// <returns>On success, set view->obj to a new reference to exporter and return 0. Otherwise, raise PyExc_BufferError, set view->obj to NULL and return -1;</returns>
+        public int FillInfo(IntPtr exporter, IntPtr buf, long len, int _readonly, int flags)
```
lostmsu: `_readonly` should be `bool`

src/runtime/pybuffer.cs
```diff
@@ -0,0 +1,228 @@
...
+        /// Writes a managed byte array into the buffer of a python object. This can be used to pass data like images from managed to python.
+        /// </summary>
+        public void Write(byte[] buffer, int offset, int count)
+        {
+            if (disposedValue)
+                throw new ObjectDisposedException(nameof(PyBuffer));
+            if (ReadOnly)
+                throw new InvalidOperationException("Buffer is read-only");
+            if (count > buffer.Length)
+                throw new ArgumentOutOfRangeException("count", "Count is bigger than the buffer.");
```
lostmsu: Might also want to check, that `offset` is non-negative, and that `offset + count`  is `<= buffer.Length`

src/runtime/pybuffer.cs
```diff
@@ -0,0 +1,228 @@
...
+        }
+
+        /// <summary>
+        /// Reads the buffer of a python object into a managed byte array. This can be used to pass data like images from python to managed.
+        /// </summary>
+        public int Read(byte[] buffer, int offset, int count) {
+            if (disposedValue)
+                throw new ObjectDisposedException(nameof(PyBuffer));
+            if (count > buffer.Length)
+                throw new ArgumentOutOfRangeException("count", "Count is bigger than the buffer.");
```
lostmsu: Same checks

src/runtime/pybuffer.cs
```diff
@@ -0,0 +1,228 @@
...
+                throw new ObjectDisposedException(nameof(PyBuffer));
+            if (ReadOnly)
+                throw new InvalidOperationException("Buffer is read-only");
+            if (count > buffer.Length)
+                throw new ArgumentOutOfRangeException("count", "Count is bigger than the buffer.");
+            // we dont support multidimensional arrays or scalars
+            if (_view.ndim != 1)
+                throw new NotSupportedException("Multidimensional arrays, scalars and objects without a buffer are not supported.");
+
+            int copylen = count < (int)_view.len ? count : (int)_view.len;
```
lostmsu: This cast to `int` should be `checked` to avoid data loss

src/runtime/pybuffer.cs
```diff
@@ -0,0 +1,228 @@
...
+        public int Read(byte[] buffer, int offset, int count) {
+            if (disposedValue)
+                throw new ObjectDisposedException(nameof(PyBuffer));
+            if (count > buffer.Length)
+                throw new ArgumentOutOfRangeException("count", "Count is bigger than the buffer.");
+            // we dont support multidimensional arrays or scalars
+            if (_view.ndim != 1)
+                throw new NotSupportedException("Multidimensional arrays, scalars and objects without a buffer are not supported.");
+
+            int copylen = count < (int)_view.len ? count : (int)_view.len;
```
lostmsu: This cast to `int` should be `checked` to avoid data loss

src/runtime/pybuffer.cs
```diff
@@ -0,0 +1,228 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Runtime.InteropServices;
+using System.Text;
+
+namespace Python.Runtime
+{
+    public class PyBuffer : IDisposable
```
lostmsu: I'd make this class `sealed`, since it does not provide any overridable behavior.

src/runtime/pybuffer.cs
```diff
@@ -0,0 +1,228 @@
...
+        private GCHandle _gchandle;
+
+        unsafe internal PyBuffer(PyObject exporter, int flags)
+        {
+            int size = Marshal.SizeOf(typeof(Runtime.Py_buffer));
+            byte[] rawData = new byte[size];
+            _gchandle = GCHandle.Alloc(rawData, GCHandleType.Pinned);
+            _handle = _gchandle.AddrOfPinnedObject();
+
+            if (Runtime.PyObject_GetBuffer(exporter.Handle, _handle, flags) < 0)
```
lostmsu: Why do yo need to keep `_handle` and `_gchandle`? Can't you simply pass address of `_view` where a pointer to `Py_buffer` is expected?

src/runtime/runtime.cs
```diff
@@ -959,6 +959,61 @@ internal static long PyObject_Size(IntPtr pointer)
...
+
+        [DllImport(_PythonDll, CallingConvention = CallingConvention.Cdecl)]
+        internal static extern void PyBuffer_FillContiguousStrides(int ndims, IntPtr shape, IntPtr strides, int itemsize, char order);
+
+        [DllImport(_PythonDll, CallingConvention = CallingConvention.Cdecl)]
+        internal static extern int PyBuffer_FillInfo(IntPtr view, IntPtr exporter, IntPtr buf, IntPtr len, int _readonly, int flags);
+
+        /* buffer interface */
+        [StructLayout(LayoutKind.Sequential, CharSet = CharSet.Auto)]
+        internal struct Py_buffer
```
lostmsu: NIT: move into its own file

lostmsu:
@SnGmng what's your solution architecture? VS has unexpected behavior when project and solution don't match. Also try running tests from command line like CI does it, e.g. with `nunit3-console.exe`

src/runtime/pybuffer.cs
lostmsu: Also, I think it is wrong to silently write `_view.len` instead of `count`, when `count` is `> _view.len`.
Return the number of bytes actually written, or throw an exception.

src/runtime/runtime.cs
SnGmng: Into `pybuffer.cs` or into its own new file?

lostmsu: An own file would not hurt definitely

SnGmng:
I run it through `nunit3-console` and it timeouts at the exact same line. Which is odd because the tests didnt even run...
The output is not very much and doesnt give any error.

```
PS D:\Users\user\Downloads\gitrepos\pythonnetsn\src\embed_tests\bin> nunit3-console Python.EmbeddingTest.dll --trace=Verbose
NUnit Console Runner 3.10.0 (.NET 2.0)
Copyright (c) 2019 Charlie Poole, Rob Prouse
Samstag, 2. November 2019 00:00:32

Runtime Environment
   OS Version: Microsoft Windows NT 10.0.18362.0
  CLR Version: 4.0.30319.42000

Test Files
    Python.EmbeddingTest.dll

=> Python.EmbeddingTest.TestDomainReload.DomainReloadAndGC
[Program.Main] === creating domain for assembly test1, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null
[Program.Main] Before Domain Unload on test1, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null
[Program.Main] After Domain Unload on test1, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null
[Program.Main] The Proxy object is not valid anymore, domain unload complete.
[Program.Main] === creating domain for assembly test2, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null
[Program.Main] Before Domain Unload on test2, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null
[Program.Main] After Domain Unload on test2, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null
[Program.Main] The Proxy object is not valid anymore, domain unload complete.
```

lostmsu:
@SnGmng you can try `--inprocess` parameter, with and without debugger. E.g. https://docs.microsoft.com/en-us/visualstudio/debugger/how-to-debug-from-a-dll-project?view=vs-2019

SnGmng:
> @SnGmng you can try `--inprocess` parameter, with and without debugger. E.g. https://docs.microsoft.com/en-us/visualstudio/debugger/how-to-debug-from-a-dll-project?view=vs-2019

If i try `--inprocess`:
```
NUnit.Engine.NUnitEngineException : Cannot run tests in process - a 32 bit process is required.

--NUnitEngineException
Cannot run tests in process - a 32 bit process is required.
   bei NUnit.Engine.Runners.MasterTestRunner.InitializePackage()
   bei NUnit.Engine.TestEngine.GetRunner(TestPackage package)
   bei NUnit.ConsoleRunner.ConsoleRunner.RunTests(TestPackage package, TestFilter filter)
```

lostmsu:
@SnGmng build fails because of the wrong casing in Py_buffer.cs in csproj file

lostmsu:
@SnGmng you can try https://stackoverflow.com/questions/1507268/force-x86-clr-on-an-any-cpu-net-assembly `corflags /32bit+ <PathToExe>` to force nunit-console to run as 32 bit process (note, this is persisted).

If the binary is not AnyCPU, you could try building a x86 binary.

src/runtime/pybuffer.cs
```diff
@@ -78,19 +85,28 @@ unsafe internal PyBuffer(PyObject exporter, int flags)
...
                 throw new ObjectDisposedException(nameof(PyBuffer));
-            return Convert.ToBoolean(Runtime.PyBuffer_IsContiguous(_handle, order));
+
+            char style = 'C';
+            if (order == BufferOrderStyle.C)
+                style = 'C';
+            else if (order == BufferOrderStyle.Fortran)
+                style = 'F';
+            else if (order == BufferOrderStyle.EitherOne)
+                style = 'A';
```
lostmsu: NIT: this could be a function

src/runtime/pybuffer.cs
```diff
@@ -78,19 +85,28 @@ unsafe internal PyBuffer(PyObject exporter, int flags)
...
+
+            return Convert.ToBoolean(Runtime.PyBuffer_IsContiguous(_gchandle.AddrOfPinnedObject(), style));
```
lostmsu: Again, why can't you use `&_view` or `ref _view` everywhere you call `_gchandle.AddrOfPinnedObject()`? You would not need the `_gchandle` in the first place.

SnGmng: Dont we need `gchandle` so it doesnt get garbage collected? And to garbage collect it in the `Dispose()` function by calling `gchandle.Free()`?

SnGmng: I mean, wouldnt the garbage collertor collect the `byte[] rawData` because it thinks its unused without pinning it through a `GCHandle`?

SnGmng:
> @SnGmng you can try https://stackoverflow.com/questions/1507268/force-x86-clr-on-an-any-cpu-net-assembly `corflags /32bit+ <PathToExe>` to force nunit-console to run as 32 bit process (note, this is persisted).
> 
> If the binary is not AnyCPU, you could try building a x86 binary.

@lostmsu I didnt compile as AnyCPU. I compiled as x86:
```
PS D:\Users\user\Downloads\gitrepos\pythonnetsn\src\embed_tests\bin\netcoreapp2.0> [reflection.assemblyname]::GetAssemblyName("D:\Users\user\Downloads\gitrepos\pythonnetsn\src\embed_tests\bin\Python.EmbeddingTest.dll") | fl


Name                  : Python.EmbeddingTest
Version               : 2.4.1.0
CultureInfo           :
CultureName           :
CodeBase              : file:///D:/Users/user/Downloads/gitrepos/pythonnetsn/src/embed_tests/bin/Python.EmbeddingTest.
                        dll
EscapedCodeBase       : file:///D:/Users/user/Downloads/gitrepos/pythonnetsn/src/embed_tests/bin/Python.EmbeddingTest.
                        dll
ProcessorArchitecture : X86
ContentType           : Default
Flags                 : None
HashAlgorithm         : SHA1
VersionCompatibility  : SameMachine
KeyPair               :
FullName              : Python.EmbeddingTest, Version=2.4.1.0, Culture=neutral, PublicKeyToken=null
```
Notice that `ProcessorArchitecture` is `X86` and not `MSIL` or `AMD64`
But i tried `corflags /32bit+ <PathToExe>` anyway. Same nunit3 error

SnGmng:
I just noticed you meant to set `corflags` on `nunit` and not `Python.Runtime.dll`.
After i did that, nunit had the following error:
```
PS D:\Users\user\Downloads\gitrepos\pythonnetsn\src\embed_tests\bin> nunit3-console .\Python.EmbeddingTest.dll --inprocess
NUnit Console Runner 3.10.0 (.NET 2.0)
Copyright (c) 2019 Charlie Poole, Rob Prouse
Samstag, 2. November 2019 14:46:38

Runtime Environment
   OS Version: Microsoft Windows NT 10.0.18362.0
  CLR Version: 4.0.30319.42000

Test Files
    .\Python.EmbeddingTest.dll

  File "<string>", line 1
    import 1234
              ^
SyntaxError: invalid syntax
[Proxy] Entering RunPython
[My Domain test1 in .NET] In PythonRunner.RunPython
[My Domain test1 in python] imported clr
[My Domain test1 in python] allocated a clr object
[My Domain test1 in python] collected garbage
[Proxy] Leaving RunPython
[My Domain test1 in .NET] unloading
[Proxy] Entering RunPython
[My Domain test2 in .NET] In PythonRunner.RunPython
[My Domain test2 in python] imported clr
[My Domain test2 in python] allocated a clr object
[My Domain test2 in python] collected garbage
[Proxy] Leaving RunPython
[My Domain test2 in .NET] unloading
=> Python.EmbeddingTest.TestDomainReload.DomainReloadAndGC
[Program.Main] === creating domain for assembly test1, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null
[Program.Main] Before Domain Unload on test1, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null
[Program.Main] After Domain Unload on test1, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null
[Program.Main] The Proxy object is not valid anymore, domain unload complete.
[Program.Main] === creating domain for assembly test2, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null
[Program.Main] Before Domain Unload on test2, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null
[Program.Main] After Domain Unload on test2, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null
[Program.Main] The Proxy object is not valid anymore, domain unload complete.
Enter
Enter
Exit

Unhandled Exception: System.Reflection.TargetInvocationException: Call target caused an exception. ---> System.AccessViolationException: Attempted to read or write to protected memory. This is often an indication that other memory is corrupted.
   in Python.Runtime.Runtime.PyObject_Call(IntPtr pointer, IntPtr args, IntPtr kw)
   in Python.Runtime.Runtime.InitializePlatformData() in D:\Users\sinan\Downloads\gitrepos\pythonnetsn\src\runtime\runtime.cs:Zeile 339.
   in Python.Runtime.Runtime.Initialize(Boolean initSigs) in D:\Users\sinan\Downloads\gitrepos\pythonnetsn\src\runtime\runtime.cs:Zeile 294.
   in Python.EmbeddingTest.TestTypeManager.Init() in D:\Users\sinan\Downloads\gitrepos\pythonnetsn\src\embed_tests\TestTypeManager.cs:Zeile 12.
   --- Ende der internen Ausnahmestapelüberwachung ---
   in System.RuntimeMethodHandle.InvokeMethod(Object target, Object[] arguments, Signature sig, Boolean constructor)
   in System.Reflection.RuntimeMethodInfo.UnsafeInvokeInternal(Object obj, Object[] parameters, Object[] arguments)
   in System.Reflection.RuntimeMethodInfo.Invoke(Object obj, BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture)
   in NUnit.Framework.Internal.Reflect.InvokeMethod(MethodInfo method, Object fixture, Object[] args)
   in NUnit.Framework.Internal.Commands.SetUpTearDownItem.RunSetUp(TestExecutionContext context)
   in NUnit.Framework.Internal.Commands.SetUpTearDownCommand.<>c__DisplayClass0_0.<.ctor>b__0(TestExecutionContext context)
   in NUnit.Framework.Internal.Commands.BeforeAndAfterTestCommand.Execute(TestExecutionContext context)
   in NUnit.Framework.Internal.Execution.SimpleWorkItem.PerformWork()
   in NUnit.Framework.Internal.Execution.WorkItem.RunOnCurrentThread()
   in NUnit.Framework.Internal.Execution.WorkItem.Execute()
   in NUnit.Framework.Internal.Execution.ParallelWorkItemDispatcher.Dispatch(WorkItem work, ExecutionStrategy strategy)
   in NUnit.Framework.Internal.Execution.ParallelWorkItemDispatcher.Dispatch(WorkItem work)
   in NUnit.Framework.Internal.Execution.CompositeWorkItem.RunChildren()
   in NUnit.Framework.Internal.Execution.CompositeWorkItem.PerformWork()
   in NUnit.Framework.Internal.Execution.WorkItem.RunOnCurrentThread()
   in NUnit.Framework.Internal.Execution.WorkItem.Execute()
   in NUnit.Framework.Internal.Execution.ParallelWorkItemDispatcher.Dispatch(WorkItem work, ExecutionStrategy strategy)
   in NUnit.Framework.Internal.Execution.ParallelWorkItemDispatcher.Dispatch(WorkItem work)
   in NUnit.Framework.Internal.Execution.CompositeWorkItem.RunChildren()
   in NUnit.Framework.Internal.Execution.CompositeWorkItem.PerformWork()
   in NUnit.Framework.Internal.Execution.WorkItem.RunOnCurrentThread()
   in NUnit.Framework.Internal.Execution.WorkItem.Execute()
   in NUnit.Framework.Internal.Execution.ParallelWorkItemDispatcher.Dispatch(WorkItem work, ExecutionStrategy strategy)
   in NUnit.Framework.Internal.Execution.ParallelWorkItemDispatcher.Dispatch(WorkItem work)
   in NUnit.Framework.Internal.Execution.CompositeWorkItem.RunChildren()
   in NUnit.Framework.Internal.Execution.CompositeWorkItem.PerformWork()
   in NUnit.Framework.Internal.Execution.WorkItem.RunOnCurrentThread()
   in NUnit.Framework.Internal.Execution.WorkItem.Execute()
   in NUnit.Framework.Internal.Execution.ParallelWorkItemDispatcher.Dispatch(WorkItem work, ExecutionStrategy strategy)
   in NUnit.Framework.Internal.Execution.ParallelWorkItemDispatcher.Dispatch(WorkItem work)
   in NUnit.Framework.Internal.Execution.CompositeWorkItem.RunChildren()
   in NUnit.Framework.Internal.Execution.CompositeWorkItem.PerformWork()
   in NUnit.Framework.Internal.Execution.WorkItem.RunOnCurrentThread()
   in NUnit.Framework.Internal.Execution.WorkItem.Execute()
   in NUnit.Framework.Internal.Execution.TestWorker.TestWorkerThreadProc()
   in System.Threading.ThreadHelper.ThreadStart_Context(Object state)
   in System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state, Boolean preserveSyncCtx)
   in System.Threading.ExecutionContext.Run(ExecutionContext executionContext, ContextCallback callback, Object state, Boolean preserveSyncCtx)
   in System.Threading.ExecutionContext.Run(ExecutionContext executionContext, ContextCallback callback, Object state)
   in System.Threading.ThreadHelper.ThreadStart()
```

src/runtime/pybuffer.cs
lostmsu: I am unsure why do you even need `rawData`. Why can't you just pass `&_view` or `ref _view`?

lostmsu:
@SnGmng yep, that looks like a memory corruption, that might have, or might not have been caused by your change. Can you debug it and check, that `Py_buffer` is filled in correctly?

SnGmng:
> @SnGmng yep, that looks like a memory corruption, that might have, or might not have been caused by your change. Can you debug it and check, that `Py_buffer` is filled in correctly?

@lostmsu  `Py_buffer` was always filled in correctly, and `TestRead` and `TestWrite` always succeed. But `TestTypeManager` fails (just like in the stacktrace i posted). This is strange. Maybe there these tests cant run in parralel without causing issuses?

lostmsu:
They should not be running in parallel. The test failure looks like it happens because of python engine reinitialization. @amos402 can you take a look at this PR to see if anything here could have corrupted memory on domain reinitialization?

@SnGmng , try removing finalizer on the buffer class. If that helps, we may have to take in PR without it, and reintroduce it later.

src/runtime/pybuffer.cs
SnGmng: @lostmsu Ok i tested it and `&_view` didnt work, but `ref _view` worked. I always thought `ref` and `&` work only for primitive and unmanaged types and not on managed types.

src/embed_tests/Python.EmbeddingTest.csproj
```diff
@@ -90,6 +90,7 @@
     <Compile Include="TestExample.cs" />
     <Compile Include="TestFinalizer.cs" />
     <Compile Include="TestPyAnsiString.cs" />
+	<Compile Include="TestPyBuffer.cs" />
```
lostmsu: NIT: a bit off

SnGmng: Hmm ive used TAB instead of spaces. I didnt know they have different length than 4 spaces on github.

src/runtime/bufferinterface.cs
```diff
@@ -0,0 +1,106 @@
...
+        /// </summary>
+        SIMPLE = 0,
+        /// <summary>
+        /// Controls the <see cref="PyBuffer.ReadOnly"/> field. If set, the exporter MUST provide a writable buffer or else report failure. Otherwise, the exporter MAY provide either a read-only or writable buffer, but the choice MUST be consistent for all consumers.
+        /// </summary>
+        WRITABLE = 0x0001,
+        /// <summary>
+        /// Controls the <see cref="PyBuffer.Format"/> field. If set, this field MUST be filled in correctly. Otherwise, this field MUST be NULL.
+        /// </summary>
+        FORMATS = 0x0004,
```
SnGmng: This is originally named `FORMAT` and not `FORMATS` but for some reason i get a "name not cls compilant" waring if i use `FORMAT`

lostmsu:
@SnGmng have you tried tests locally after removing finalizer? Do they pass?

SnGmng:
> @SnGmng have you tried tests locally after removing finalizer? Do they pass?
Visual studio fooled me that they pass but they didnt when i retested through `nunit3-console`. Weird.

lostmsu:
@SnGmng a sanity check: if you comment your tests out, do the tests pass?

SnGmng:
@lostmsu 
Sometimes errors seem weird but the answer couldnt be more obvious.
So i was comparing my unit tests with the others and noticed that mine was the only one using a `PyScope`. And then it hit me: If i used a `PyScope`, did i call `Py.GIL()` too?! Well it turned out i didnt.
Im surprised it took me a week to notice this obvious cause.😑

lostmsu:
@SnGmng hahaha, now I remember facing the same issue. Damn. Looks like we should add GIL checks to many entry points...

SnGmng:
@lostmsu yeah would be better hahaha. But the build seems to fail again... seems like `Py.GIL` wasnt the only cause.. 😭

filmor:
At least the failure is now less erratic in that it fails exactly for all Python 3 versions on Windows running on x86 :)

amos402:
> 
> 
> They should not be running in parallel. The test failure looks like it happens because of python engine reinitialization. @amos402 can you take a look at this PR to see if anything here could have corrupted memory on domain reinitialization? I don't see anything, but maybe you'll notice some issue.
> 
> @SnGmng , try removing finalizer on the buffer class. If that helps, we may have to take in PR without it, and reintroduce it later.

I'm not sure about that,  maybe the cause is calling `PyBuffer_Release` in `~PyBuffer`, it have not been got GIL before calling `PyBuffer_Release`(destructor  called in GC thread). Also the destructor may call after `Py_Finalize`.

btw. I wonder why the `PyBuffer` not making an implementation of `System.Stream`?

filmor:
A `Stream` implementation is (I think) out of scope here, but the GIL comment is valid.

lostmsu:
@SnGmng can you rebase/merge `master` and update the `PyBuffer` finalizer to behave similarly to the one in `PyObject` (see latest changes in `master`)?

SnGmng:
> @SnGmng can you rebase/merge `master` and update the `PyBuffer` finalizer to behave similarly to the one in `PyObject` (see latest changes in `master`)?

yeah sure

SnGmng:
Why is appveyor configured to do the conda build? I mean it fails everytime and is ignored anyway? And it takes 6min which is a lot. Removing these checks would improve the test speed to 1min which is a 6 times speed improvement. I am waiting multiple hours for appveyor because it is so slow, just because of the unnecessary conda build...

lostmsu:
Related: https://github.com/pythonnet/pythonnet/issues/1063

lostmsu:
@SnGmng the issue was just fixed in master by filmor.

lostmsu:
Given 2.5 is just around the corner, and in 3.0 I'd like to fix https://github.com/pythonnet/pythonnet/pull/980 , which might affect the interfaces proposed here, I suggest we get it after 2.5 is released (should be soon).

@SnGmng the issues you mentioned are fixed in `master`, you should be able to update the finalizer now.

SnGmng:
Be aware that the last commit to update the finalizer is still untested because i had some problems using nunit and couldnt run the tests.

lostmsu:
@SnGmng oh, your `Dispose` method was perfectly fine. It is the finalizer (e.g. `~PyBuffer()`), that needed to be updated.

lostmsu:
Somehow this is already in o-O

PR #988, opened 2019-11-14
Add CI support for py3.8.

matham:
This add CI support for py3.8 on Windows (appveyor).

I didn't add it for travis, because I'm not sure it is supported yet on travis (I recently switched all my CI to github so I'm not sure the status of travis).


matham:
Hmm, looks like it's failing [here](https://github.com/pythonnet/pythonnet/blob/master/setup.py#L302) with `pycparser` not being installed. Somehow adding it to `setup_requires` [here](https://github.com/pythonnet/pythonnet/blob/master/setup.py#L626) is too late.

I'm not sure how to explicitly add this for this project - because I don't see why it's only failing for 3.8?

filmor:
If one of the `interop*.cs` files doesn't exist, it is generated on build. Generate it manually once and commit the resulting file, then `pycparser` will not be invoked.

filmor:
Looks good, not quite sure what is going on on AppVeyor, though.

matham:
The error seems to be that `System.EntryPointNotFoundException : Unable to find an entry point named 'Py_CompileString' in DLL 'python38'.`. I traced it down to this bug in python: [Py_CompileString and PyParser_SimpleParseString not exported in python38.dll](https://bugs.python.org/issue37633).

Seems this will be fixed for the next release though [in mid December](https://www.python.org/dev/peps/pep-0569/#schedule). Not sure what to do about this PR in the meantime though.

filmor:
Add the Windows/Python 3.8 builds as `allow_failures` in the `appveyor.yaml` (https://www.appveyor.com/docs/build-configuration/#allow-failing-jobs), a note to the README and we'll merge. No need to wait until December for this.

matham:
Ok, added a note to the readme and `allow_failures`. Apologies, but you may want to squash the PR as I was doing it in the browser so there's a commit for each change in a file.

Would it be possible to make a wheel of the last release for 3.8 and upload to pypi (e.g. by backporting this PR)? I'm not sure how you generate the wheels for the release - I can't find it happening on the CI. In my projects, we have a branch for each release called `stable-x.y.z` so we can backport and have the CI run again generating the wheels etc. But here it is only a tag so I'm not sure that is possible?

filmor:
It's probably easier to just do a 2.4.1 now, I'll see to it next weekend.

m-rossi:
The CI pipelines at conda-forge just started the Python 3.8 migration for the pythonnet-feedstock and failed: https://github.com/conda-forge/pythonnet-feedstock/pull/3
So am I right we need a new version of pythonnet or Python 3.8.1 to get it working on Python 3.8?

filmor:
pythonnet from current master should work fine for a lot of things, it will just fail on functions that run python code as a string from .net on Windows until 3.8.1.

m-rossi:
conda-forge uses pypi uploads, so I will disable builds for Python 3.8 for now.

PR #991, opened 2019-11-20
Update mp_length to support all ICollection implementers

slide:

Currently, if you try and use `len` on anything other than Array subclasses, it fails. This adds the ability to use `len` on any .NET class that implements ICollection (List, Dictionary, etc).

No, there is no current issue that I know of.

### Any other comments?

Not sure if this is enhancement PR that requires a doc, but I am happy to add doc info if others think it is necessary.

### Checklist

### DIFF
--- a/src/runtime/slots/mp_length.cs
+++ b/src/runtime/slots/mp_length.cs
@@ -1,0 +1,31 @@
+using System;
+using System.Reflection;
+
+namespace Python.Runtime.Slots
+{
+    static class mp_length_slot
+    {
+        /// <summary>
+        /// Implements __len__ for classes that implement ICollection
+        /// (this includes any IList implementer or Array subclass)
+        /// </summary>
+        public static int mp_length(IntPtr ob)
+        {
+            var co = ManagedType.GetManagedObject(ob) as CLRObject;
+            if (co == null)
+            {
+                Exceptions.RaiseTypeError("invalid object");
+            }
+
+            Type t = co.inst.GetType();
+            PropertyInfo p = t.GetProperty("Count");
+            if (p != null)
+            {
+                return (int)p.GetValue(co.inst, null);
+            }
+
+            Exceptions.SetError(Exceptions.TypeError, $"object of type '{co.inst.GetType().Name}' has no len()");
+            return -1;
+        }
+    }
+}
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -12,6 +12,7 @@
 
 ## Contributors
 
+-   Alex Earl ([@slide](https://github.com/slide))
 -   Alex Helms ([@alexhelms](https://github.com/alexhelms))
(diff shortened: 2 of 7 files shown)


lostmsu:
Two problems:
- it does not work for `ICollection<T>` and `IReadOnlyCollection<T>`, which might be a surprise
- you should only assign `mp_length` slot in types, that actually implement those interfaces, because this assignment would override possibly existing inherited length implementations

slide:
That makes sense, I am still learning pythonnet. I really would like to have a `len` for IList implementations, is there a better way to do that?

lostmsu:
See this diff of typemanager.cs on how to fix the second comment: https://github.com/losttech/pythonnet/commit/6f85ed7aa4415943ea1d4c10d1aaadc3892cd503#diff-7cc8954e74bf64fe9d83afa409dc8c0d

slide:
@lostmsu Since that code is not in mainline, would you recommend I create a Slots.cs in my PR?

slide:
Also, it looks like the code is building with .NET 4.0 runtime, so `IReadOnlyCollection<T>` is not available.

lostmsu:
I actually think it would be nice to have folder `Slots` with a file per slot or per Python or C# protocol.

slide:
I think there is just something I am not understanding yet about how this works. I will close this for now and rereview.

src/runtime/slots/mp_length.cs
```diff
@@ -0,0 +1,31 @@
...
+        {
+            var co = ManagedType.GetManagedObject(ob) as CLRObject;
+            if (co == null)
+            {
+                Exceptions.RaiseTypeError("invalid object");
+            }
+
+            Type t = co.inst.GetType();
+            PropertyInfo p = t.GetProperty("Count");
+            if (p != null)
```
lostmsu: I'd suggest to start with `as ICollection` for performance reasons.

src/runtime/slots/mp_length.cs
```diff
@@ -0,0 +1,31 @@
...
+        public static int mp_length(IntPtr ob)
+        {
+            var co = ManagedType.GetManagedObject(ob) as CLRObject;
+            if (co == null)
+            {
+                Exceptions.RaiseTypeError("invalid object");
+            }
+
+            Type t = co.inst.GetType();
+            PropertyInfo p = t.GetProperty("Count");
```
lostmsu:
```csharp
> typeof(int[]).GetProperty("Count")
null
```

E.g. you need to look for explicit interface implementations.
An option is to loop through the implemented interfaces as you did in `CreateType`, and then get `Count` property from it instead.

lostmsu: `typeof(ICollection<int>).GetProperty("Count").GetValue(new int[10])`

Issue #886, opened 2019-06-19
Add microbenchmarking to CI


lostmsu:
Python.NET continuous integration needs microbenchmark tests to track changes in performance.

amos402:
I suggest separate the benchmark from the normal tests, it's merely a assistance running. And it even not running the the newest branch code for now, that may disturb the judgement of test correctness.

lostmsu:
@amos402 not sure what you mean by not running the newest branch code. It does run the newest Python.NET code, and compares it to the baseline.

amos402:
Oh yeah, it is. Seems just the misunderstanding of mine. But I still thought it's better to separate them.

lostmsu:
@amos402 why though? Having the benchmark run as part of CI (as long as it quick enough and reliable) has a major advantage: if your change fails it, you'll see, and will either have a chance to fix performance regression, or adjust the performance target explicitly in the test code, which would let reviewers immediately weight the performance impact against usefulness of the change.

amos402:
Jumble functional test and performance test isn't a good idea. They're two things. Just make some search you may find more articles like this https://www.neotys.com/blog/what-does-test-driven-development-mean-for-performance-testers/

lostmsu:
@amos402 the article you linked actually advocates perf tests alongside functional tests :-D

amos402:
Yeah, they're two things, don't mix them.

lostmsu:
@amos402 they are not mixed. They are two separate steps of the CI. The only problem is that the results report in the GitHub UI is mixed, but you can still drill down in AppVeyor. If it is a problem, feel free to separate the pipeline.

PR #992, opened 2019-11-21
CI for performance tests

lostmsu:

Runs performance tests in CI, comparing build with new changes against the latest (2.3 as of now) NuGet release.

#886

### Checklist

Nothing applicable.


lostmsu:
@filmor  if you are to merge this, I recommend squashing the commit, as there are many intermediate changes, which are insignificant on their own.

src/perf_tests/Python.PerformanceTests.csproj
```diff
@@ -3,12 +3,19 @@
...
+    <OutputPath>bin\</OutputPath>
 
     <IsPackable>false</IsPackable>
+
+    <Platforms>x64;x86</Platforms>
   </PropertyGroup>
 
   <ItemGroup>
     <PackageReference Include="BenchmarkDotNet" Version="0.11.5" />
+    <PackageReference Include="Microsoft.NETFramework.ReferenceAssemblies.net461" Version="1.0.0-preview.2">
```
filmor: Any particular reason for referencing net461 explicitly? I was under the impression that the package selects the correct one without a version.

lostmsu: I am unsure if selection happens before downloading or after. If it has to download all packages first, then it should shorten build time. I can look it up, but I hope we will get rid of this soon, as it is only needed for pythonnet-2.3 packages, that don't support .NET Standard.

lostmsu:
OK, test is passing, and I can see in [AppVeyor logs for 3.5 x64 xplat build](https://ci.appveyor.com/project/pythonnet/pythonnet/builds/29033240/job/535gmdinq8hnvbul#L2050), that performance is checked.

BTW, 3.5 x64 xplat is the only configuration where perf test runs, as it is the only configuration supported by the published NuGet package.

lostmsu:
I guess this boils down to how much flakiness we can tolerate in perf tests in CI. Letting a bit more wiggle room in the performance goal should make it very unlikely to fail randomly.

src/perf_tests/Python.PerformanceTests.csproj
lostmsu: Fixed

PR #993, opened 2019-11-21
Split from PR 958: restoring the __import__ after shutdown.

benoithudson:
When C# shuts down we should restore Python to its original state.

This is adapted from the large patch that amos put in PR #958. It restores the import hook.

Not yet.

### Any other comments?

I would have used the ShutdownHandler but that's only in the PythonEngine, whereas `__import__` is in the Runtime. Maybe we should move the ShutdownHandler mechanism to the runtime?

### Checklist

Can't really test this at the moment. After Py_Finalize (in Shutdown), we can't check the state of the `__import__` function until we call Py_Initialize. But Py_Initialize will create a new `__import__`, so we can't see whether our code ran. This change only becomes relevant when we stop calling Py_Finalize in Shutdown.

### DIFF
--- a/src/runtime/importhook.cs
+++ b/src/runtime/importhook.cs
@@ -26,25 +26,64 @@ namespace Python.Runtime
 #endif
 
         /// <summary>
+        /// Get a <i>New reference</i> to the builtins module.
+        /// </summary>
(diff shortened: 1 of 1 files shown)


filmor:
Nice, thank you very much :)

PR #994, opened 2019-11-21
Add mp_length slot for .NET classes implementing ICollection/ICollection<T>

slide:

Implements the mp_length slot for .NET classes that implement ICollection or ICollection<T>, either explicitly or non-explicitly.

There are no open issues relating to this as far as I know.

### Checklist


src/runtime/slots/mp_length.cs
```diff
@@ -0,0 +1,66 @@
...
+namespace Python.Runtime.Slots
+{
+    internal static class mp_length_slot
+    {
+        /// <summary>
+        /// Finds a get_Count (Count property) that is private and
+        /// is either ICollection or ICollection&lt;T&gt; explicit
+        /// implementation
+        /// </summary>
+        static MethodInfo GetCountGetter(Type t)
```
lostmsu: Instead of doing this (which looks rather obscure), you could simply do:

```csharp
var genericCollectionInterface = clrType.GetInterfaces()
  // this should return typeof(ICollection<int>)
  .FirstOrDefault(x => x.IsGenericType && x.GetGenericTypeDefinition() == typeof(ICollection<>));
if (genericCollectionInterface == null) { ...error... }

// this won't be null, as `ICollection<int>` has a "public" property `Count`
var countProperty = genericCollectionInterface.GetProperty(nameof(ICollection<int>.Count));
return countProperty.GetValue(co.inst);
...
```

src/runtime/slots/mp_length.cs
```diff
@@ -0,0 +1,66 @@
...
+            if (co.inst is ICollection c)
+            {
+                return c.Count;
+            }
+
+            Type clrType = co.inst.GetType();
+
+            // now look for things that implement ICollection<T> directly (non-explicitly)
+            PropertyInfo p = clrType.GetProperty("Count");
+            if (p != null && clrType.GetInterfaces().Any(x => x.IsGenericType && x.GetGenericTypeDefinition() == typeof(ICollection<>)))
```
amos402: Too many reflections every times, you can try like this.
```c#
    class Program
    {
        interface ICountGetter
        {
            int GetCount(object obj);
        }

        class CountGetter<Cls, T> : ICountGetter
        {
            private Func<Cls, int> _getter;

            public CountGetter(Func<Cls, int> getter)
            {
                _getter = getter;
            }

            public int GetCount(object obj)
            {
                return _getter((Cls)obj);
            }
        }

        {
            var obj = new MyClass();
            var type = obj.GetType();
            var declType = type.GetInterfaces().FirstOrDefault(
                t => t.IsGenericType
                && t.GetGenericTypeDefinition() == typeof(ICollection<>));
            if (declType == null)
            {
                // set error and return
            }
            var pi = declType.GetProperty("Count");
            var elemType = declType.GenericTypeArguments[0];
            var getterType = typeof(CountGetter<,>).MakeGenericType(declType, elemType);
            var getterFuncType = typeof(Func<,>).MakeGenericType(declType, typeof(int));
            var getterFunc = Delegate.CreateDelegate(getterFuncType, pi.GetMethod);
            var getter = (ICountGetter)Activator.CreateInstance(getterType, getterFunc);
            // You should cache the ([decltype] = getter), it makes you no reflection costs when you call GetCount next time.
            int count = getter.GetCount(obj);
            Console.WriteLine(count);
        }
    }
```

src/runtime/slots/mp_length.cs
slide: I tried doing something similar to that, but it said that the target type is not the same as the property type when I was running it. I'll revisit to make sure I didn't miss something.

lostmsu: It works for arrays in interactive:
```csharp
> typeof(int[]).GetInterfaces()
  .FirstOrDefault(x => x.IsGenericType && x.GetGenericTypeDefinition() == typeof(ICollection<>))
  .GetProperty(nameof(ICollection<int>.Count))
  .GetValue(new int[42])
42
```

src/runtime/slots/mp_length.cs
lostmsu: I agree, but I feel like it would be good to have a feature + tests first. It can be optimized later.

src/runtime/slots/mp_length.cs
```diff
@@ -0,0 +1,57 @@
...
+                return (int)p.GetValue(co.inst, null);
+            }
+
+            // finally look for things that implement the interfaces explicitly
+            var iface = clrType.GetInterface(nameof(ICollection));
+            if(iface != null)
+            {
+                p = iface.GetProperty(nameof(ICollection.Count));
+                return (int)p.GetValue(co.inst, null);
+            }
```
lostmsu: I don't think you needs this one for non-generic `ICollection`: no matter if the implementation is implicit or explicit `co.inst is ICollection c` will actually pass, and handle this case.
The generic one is necessary only because you can't do `is ICollection<*>`

filmor:
Thanks a lot for your contribution :)

PR #1002, opened 2019-11-30
Release method wrapper(split from #958)

amos402:

Release method wrapper and module definition(Python3) when shutdown

...

### Any other comments?

...

### Checklist


src/runtime/importhook.cs
```diff
@@ -112,13 +129,26 @@ internal static void Initialize()
...
+            }
 
-                Runtime.XDecref(py_clr_module);
-                Runtime.XDecref(root.pyHandle);
+            RestoreImport();
+
+            bool shouldFreeDef = Runtime.Refcount(py_clr_module) == 1;
+            Runtime.XDecref(py_clr_module);
+            py_clr_module = IntPtr.Zero;
+#if PYTHON3
```
lostmsu: Can we avoid introducing another compile-time check, and use version info from `Runtime` class instead?

src/runtime/methodwrapper.cs
```diff
@@ -31,5 +32,21 @@ public IntPtr Call(IntPtr args, IntPtr kw)
...
+                return;
+            }
+            _disposed = true;
+            bool freeDef = Runtime.Refcount(ptr) == 1;
+            Runtime.XDecref(ptr);
+            if (freeDef && mdef != IntPtr.Zero)
+            {
+                Runtime.PyMem_Free(mdef);
+                mdef = IntPtr.Zero;
+            }
```
lostmsu: If we call `Release`, and there are still references to the method, how is it going to be freed later?

src/runtime/importhook.cs
filmor: Actually, in the modernisation branch I gave up on the idea of having Python 2 and 3 in the same codebase, especially considering the imminent EoL of Python 2. I do that by defaulting to Python 3 and letting the user supply `PYTHON2` explicitly for the Py2 version. So here, to stay in line with that, it would be enough to change to `#if !PYTHON2`.

src/runtime/methodwrapper.cs
filmor: @amos402 Could you comment on this question?

amos402: It won't release anymore, here's the only chance for now. Assume the ptr deallocating in another domain in the future, the cod in this domain had already been invalid, unless we use a C extension take place the ptr's `tp_dealloc`.

PR #1003, opened 2019-11-30
Transfer the ownership of method's thunk to caller(split from #958)

amos402:
Let the caller handle the lifecycle for achieving the memory release when shutdown pythonnet regardless of normal shutdown or soft shutdown in #958.

...

### Any other comments?

...

### Checklist


lostmsu:
Is there a reason why we are not using `Marshal.GetFunctionPointerForDelegate` function?

amos402:
Usually I also use `Marshal.GetFunctionPointerForDelegate` too. If my suspicion corrected, this historic code was written for .net1.x when `Marshal.GetFunctionPointerForDelegate` doesn't exists.
As for now, we can simply cover it by `Marshal.GetFunctionPointerForDelegate`.

filmor:
@amos402 Do you think you can update this PR to make use of `Marshal.GetFunctionPointerForDelegate`? That would probably simplify the code quite a bit.

Issue #995, opened 2019-11-27
geninterop.py doesn't handle c_ast.FuncDecl; creates broken interop38.cs


Jeff17Robbins:
### Details

I'm concerned that the automated process pythonnet uses to generate the interop[version].cs file might be flawed.  This is the relevant snippet from `object.h`

```
    unsigned int tp_version_tag;
    destructor tp_finalize;
    vectorcallfunc tp_vectorcall;
    /* bpo-37250: kept for backwards compatibility in CPython 3.8 only */
    Py_DEPRECATED(3.8) int (*tp_print)(PyObject *, FILE *, int);
```

this is how I mapped those fields in my handmade interop38.cs

```
        public static int tp_version_tag = 0;
        public static int tp_finalize = 0;
        public static int tp_vectorcall = 0;
        public static int tp_print = 0;
        public static int am_await = 0;
```

but here's how the auto-generated interop38.cs maps them

```
        public static int tp_version_tag = 0;
        public static int tp_finalize = 0;
        public static int tp_vectorcall = 0;
        public static int am_await = 0;
```

Notice the missing **public static int tp_print = 0;**

I'm thinking that the `Py_DEPRECATED(3.8) int (*tp_print)(PyObject *, FILE *, int);` messed up the auto-gen technique?  Won't all fields following `tp_vectorcall` be offset incorrectly due to this?

filmor:
If I get https://docs.python.org/3/c-api/typeobj.html#c.PyVectorcall_Call correctly, `tp_vectorcall` replaces `tp_print`. I'll check what happens in this header.

Jeff17Robbins:
I think ```Py_ssize_t tp_vectorcall_offset;``` replaced the previous location of ```public static int tp_print = 0;`,  and then

Jeff17Robbins:
`typemanager.cs` assigns the type's name into `qualname` via this code


```
#if PYTHON3
            // For python3 we leak two objects. One for the ASCII representation
            // required for tp_name, and another for the Unicode representation
            // for ht_name.
            IntPtr temp = Runtime.PyBytes_FromString(name);
            IntPtr raw = Runtime.PyBytes_AS_STRING(temp);
            temp = Runtime.PyUnicode_FromString(name);
#elif PYTHON2
            IntPtr temp = Runtime.PyString_FromString(name);
            IntPtr raw = Runtime.PyString_AsString(temp);
#endif
[...]

PR #1015, opened 2019-12-16
Modernisation

filmor:
Ongoing development to get to a .NET Standard only, platform-independent and Python version agnostic version of Python.NET that runs on .NET Framework, .NET Core and Mono (and hopefully also the respective AoT variants).


Jeff17Robbins:
Is the intention to only support Python3.8?

filmor:
The intention is to support at least all Python 3 versions from 3.5 onwards. The Python (3) API and ABI is mostly backwards-compatible for our purposes. The one comment on Python 3.8 just refers to https://docs.python.org/3/whatsnew/3.8.html#debug-build-uses-the-same-abi-as-release-build, which allows us to use the exact same build even for Debug Pythons from 3.8 onwards.

Jeff17Robbins:
How can a single `interop3.cs` work across disparate Python 3 releases?  

As #995 demonstrated, the Python team continues to change the internals of `struct PyObject`, with potential impact on Python.NET, so I'm assuming we need to keep version-dependent `interop3[n].cs` files?

filmor:
@Jeff17Robbins No, it didn't change. The problem was that they changed the /meaning/ of one of the slots and it now used a different type (same size, though). The structure as such hasn't changed in the Python 3 line, just reusage of deprecated slots and addition of new ones at the end.

Jeff17Robbins:
@filmor I think it actually did change.  

You are correct that new field `Py_ssize_t tp_vectorcall_offset;` simply overwrote `public static int tp_print = 0;`

But, they also added two new fields right after `tp_finalize`:

```
vectorcallfunc tp_vectorcall;

/* bpo-37250: kept for backwards compatibility in CPython 3.8 only */
Py_DEPRECATED(3.8) int (*tp_print)(PyObject *, FILE *, int);
```

These two new fields change all the offsets of the fields that are defined after them.  The structure thus *is* a different size, and fields we care about, like `ht_qualname` (aka `qualname`) has a different offset in Python3.8 as compared to Python3.7.

If you differ `interop37.cs` with`interop38.cs`, you'll see these two new fields:


```
        public static int tp_vectorcall = 0;
        public static int tp_print = 0;
```

filmor:
Meh. Ah well, I'll add this back in due time, but probably in form of a hash-table that is stored in a resource stream instead of autogenerated code. Thanks for the hint :)

Jeff17Robbins:
Maybe we can create some kind of "sanity check" on the automated process that generates the hash-table to make sure that `geninterop.py` captured all the fields from the C struct definitions?  It is a crucial, yet somewhat fragile, step of the process.

zivillian:
I looked at this branch today and it is currently broken for VS2019.

Even after adding the missing projects to the .sln and fixing the Project SDK
```patch
-﻿<Project Sdk="Microsoft.NET.Test.Sdk">
+<Project Sdk="Microsoft.NET.Sdk">
```
it throws errors about missing properties and methods like
```cs
Python.Runtime.Runtime.pyversion
Python.Runtime.Runtime.Machine
Python.Runtime.Runtime.MachineName
Python.Runtime.Runtime.OperatingSystem
Python.Runtime.Runtime.OperatingSystemName
Python.Runtime.Runtime.Py_Initialize()
```
Am I looking at the incorrect source, or is there anything missing?

lostmsu:
@filmor btw, I think this could take advantage of https://github.com/pythonnet/pythonnet/pull/1043

PR #1016, opened 2019-12-17
Reset the type slots (split from #958)

BadSingleton:
On shutdown, clear the managed Types' slots.

This implements part of #958, where the managed type slots are reset on shutdown.

Not yet.

### Any other comments?

Most of the code was originally written by @amos402 

### Checklist

Not really testable at the moment. Once the soft shutdown is implemented, tests can be written.

### DIFF
--- a/src/runtime/interop.cs
+++ b/src/runtime/interop.cs
@@ -449,7 +449,7 @@ namespace Python.Runtime
             return pmap[name] as Type;
         }
 
-        internal static IntPtr GetThunk(MethodInfo method, string funcType = null)
+        internal static ThunkInfo GetThunk(MethodInfo method, string funcType = null)
         {
             Type dt;
             if (funcType != null)
@@ -457,18 +457,13 @@ namespace Python.Runtime
             else
                 dt = GetPrototype(method.Name);
 
-            if (dt != null)
+            if (dt == null)
             {
-                IntPtr tmp = Marshal.AllocHGlobal(IntPtr.Size);
-                Delegate d = Delegate.CreateDelegate(dt, method);
-                Thunk cb = new Thunk(d);
-                Marshal.StructureToPtr(cb, tmp, false);
-                IntPtr fp = Marshal.ReadIntPtr(tmp, 0);
-                Marshal.FreeHGlobal(tmp);
-                keepAlive.Add(d);
-                return fp;
+                return ThunkInfo.Empty;
             }
-            return IntPtr.Zero;
+            Delegate d = Delegate.CreateDelegate(dt, method);
+            IntPtr fp = Marshal.GetFunctionPointerForDelegate(d);
+            return new ThunkInfo(d, fp);
         }
 
         [UnmanagedFunctionPointer(CallingConvention.Cdecl)]
@@ -521,5 +516,28 @@ namespace Python.Runtime
         {
             fn = d;
         }
+    }
+
+    internal struct ThunkInfo
+    {
+        public Delegate Target;
+        public IntPtr Address;
+
+        public static readonly ThunkInfo Empty = new ThunkInfo(null, IntPtr.Zero);
+
+        public ThunkInfo(Delegate target, IntPtr addr)
+        {
+            Target = target;
+            Address = addr;
+        }
+    }
+    
+    [StructLayout(LayoutKind.Sequential, CharSet = CharSet.Ansi)]
+    struct PyMethodDef
+    {
+        public IntPtr ml_name;
+        public IntPtr ml_meth;
+        public int ml_flags;
+        public IntPtr ml_doc;
     }
 }
--- a/src/runtime/interop.cs
+++ b/src/runtime/interop.cs
@@ -509,7 +509,7 @@ namespace Python.Runtime
     }
 
 
-    [StructLayout(LayoutKind.Sequential, CharSet = CharSet.Ansi)]
+    [StructLayout(LayoutKind.Sequential)]
     internal struct Thunk
     {
         public Delegate fn;
@@ -536,5 +536,14 @@ namespace Python.Runtime
             Target = target;
             Address = Marshal.GetFunctionPointerForDelegate(target);
         }
+    }
+
+    [StructLayout(LayoutKind.Sequential)]
+    struct PyMethodDef
+    {
+        public IntPtr ml_name;
+        public IntPtr ml_meth;
+        public int ml_flags;
+        public IntPtr ml_doc;
     }
 }
--- a/src/runtime/interop.cs
+++ b/src/runtime/interop.cs
@@ -4,6 +4,7 @@ using System.Collections.Specialized;
 using System.Runtime.InteropServices;
 using System.Reflection;
 using System.Text;
+using System.Collections.Generic;
 
 namespace Python.Runtime
 {
@@ -227,7 +228,7 @@ namespace Python.Runtime
             byte[] ascii = Encoding.ASCII.GetBytes(modulename);
             int size = name + ascii.Length + 1;
             IntPtr ptr = Marshal.AllocHGlobal(size);
-            for (int i = 0; i < m_free; i += IntPtr.Size)
+            for (int i = 0; i <= m_free; i += IntPtr.Size)
                 Marshal.WriteIntPtr(ptr, i, IntPtr.Zero);
             Marshal.Copy(ascii, 0, (IntPtr)(ptr + name), ascii.Length);
             Marshal.WriteIntPtr(ptr, m_name, (IntPtr)(ptr + name));
@@ -334,7 +335,7 @@ namespace Python.Runtime
 
     internal class Interop
     {
-        private static ArrayList keepAlive;
+        private static List<ThunkInfo> keepAlive;
         private static Hashtable pmap;
 
         static Interop()
@@ -351,8 +352,7 @@ namespace Python.Runtime
                 p[item.Name] = item;
             }
 
-            keepAlive = new ArrayList();
-            Marshal.AllocHGlobal(IntPtr.Size);
+            keepAlive = new List<ThunkInfo>();
             pmap = new Hashtable();
 
             pmap["tp_dealloc"] = p["DestructorFunc"];
@@ -462,8 +462,10 @@ namespace Python.Runtime
                 return ThunkInfo.Empty;
             }
             Delegate d = Delegate.CreateDelegate(dt, method);
-            IntPtr fp = Marshal.GetFunctionPointerForDelegate(d);
-            return new ThunkInfo(d, fp);
+            var info = new ThunkInfo(d);
+            // TODO: remove keepAlive when #958 merged, let the lifecycle of ThunkInfo transfer to caller.
+            keepAlive.Add(info);
+            return info;
         }
 
         [UnmanagedFunctionPointer(CallingConvention.Cdecl)]
@@ -518,26 +520,21 @@ namespace Python.Runtime
         }
     }
 
-    internal struct ThunkInfo
+    internal class ThunkInfo
     {
-        public Delegate Target;
-        public IntPtr Address;
+        public readonly Delegate Target;
+        public readonly IntPtr Address;
 
-        public static readonly ThunkInfo Empty = new ThunkInfo(null, IntPtr.Zero);
+        public static readonly ThunkInfo Empty = new ThunkInfo(null);
 
-        public ThunkInfo(Delegate target, IntPtr addr)
+        public ThunkInfo(Delegate target)
         {
+            if (target == null)
+            {
+                return;
+            }
             Target = target;
-            Address = addr;
+            Address = Marshal.GetFunctionPointerForDelegate(target);
         }
-    }
-    
-    [StructLayout(LayoutKind.Sequential, CharSet = CharSet.Ansi)]
-    struct PyMethodDef
-    {
-        public IntPtr ml_name;
-        public IntPtr ml_meth;
-        public int ml_flags;
-        public IntPtr ml_doc;
     }
 }
--- a/src/runtime/metatype.cs
+++ b/src/runtime/metatype.cs
@@ -22,6 +22,15 @@ namespace Python.Runtime
             return PyCLRMetaType;
         }
 
+        public static void Release()
+        {
+            if (Runtime.Refcount(PyCLRMetaType) > 1)
+            {
+                SlotsHolder.ReleaseTypeSlots(PyCLRMetaType);
+            }
+            Runtime.XDecref(PyCLRMetaType);
+            PyCLRMetaType = IntPtr.Zero;
+        }
 
         /// <summary>
         /// Metatype __new__ implementation. This is called to create a new
--- a/src/runtime/metatype.cs
+++ b/src/runtime/metatype.cs
@@ -24,6 +24,10 @@ namespace Python.Runtime
 
         public static void Release()
         {
+            if (PyCLRMetaType == IntPtr.Zero)
+            {
+                throw new ObjectDisposedException("PyCLRMetaType");
+            }
             if (Runtime.Refcount(PyCLRMetaType) > 1)
             {
                 _metaSlotsHodler.ResetSlots();
--- a/src/runtime/metatype.cs
+++ b/src/runtime/metatype.cs
@@ -11,14 +11,14 @@ namespace Python.Runtime
     internal class MetaType : ManagedType
     {
         private static IntPtr PyCLRMetaType;
-
+        private static SlotsHolder _metaSlotsHodler;
 
         /// <summary>
         /// Metatype initialization. This bootstraps the CLR metatype to life.
         /// </summary>
         public static IntPtr Initialize()
         {
-            PyCLRMetaType = TypeManager.CreateMetaType(typeof(MetaType));
+            PyCLRMetaType = TypeManager.CreateMetaType(typeof(MetaType), out _metaSlotsHodler);
             return PyCLRMetaType;
         }
 
@@ -26,10 +26,10 @@ namespace Python.Runtime
         {
             if (Runtime.Refcount(PyCLRMetaType) > 1)
             {
-                SlotsHolder.ReleaseTypeSlots(PyCLRMetaType);
+                _metaSlotsHodler.ResetSlots();
             }
-            Runtime.XDecref(PyCLRMetaType);
-            PyCLRMetaType = IntPtr.Zero;
+            Runtime.Py_CLEAR(ref PyCLRMetaType);
+            _metaSlotsHodler = null;
         }
 
         /// <summary>
--- a/src/runtime/methodwrapper.cs
+++ b/src/runtime/methodwrapper.cs
@@ -12,25 +12,43 @@ namespace Python.Runtime
     {
         public IntPtr mdef;
         public IntPtr ptr;
-        private ThunkInfo thunk;
+        private bool _disposed = false;
+
+        private ThunkInfo _thunk;
 
         public MethodWrapper(Type type, string name, string funcType = null)
         {
             // Turn the managed method into a function pointer
 
-            thunk = Interop.GetThunk(type.GetMethod(name), funcType);
+            _thunk = Interop.GetThunk(type.GetMethod(name), funcType);
 
             // Allocate and initialize a PyMethodDef structure to represent
             // the managed method, then create a PyCFunction.
 
             mdef = Runtime.PyMem_Malloc(4 * IntPtr.Size);
-            TypeManager.WriteMethodDef(mdef, name, thunk.Address, 0x0003);
+            TypeManager.WriteMethodDef(mdef, name, _thunk.Address, 0x0003);
             ptr = Runtime.PyCFunction_NewEx(mdef, IntPtr.Zero, IntPtr.Zero);
         }
 
         public IntPtr Call(IntPtr args, IntPtr kw)
         {
             return Runtime.PyCFunction_Call(ptr, args, kw);
+        }
+
+        public void Release()
+        {
+            if (_disposed)
+            {
+                return;
+            }
+            _disposed = true;
+            bool freeDef = Runtime.Refcount(ptr) == 1;
+            Runtime.XDecref(ptr);
+            if (freeDef && mdef != IntPtr.Zero)
+            {
+                Runtime.PyMem_Free(mdef);
+                mdef = IntPtr.Zero;
+            }
         }
     }
 }
--- a/src/runtime/methodwrapper.cs
+++ b/src/runtime/methodwrapper.cs
@@ -13,7 +13,6 @@ namespace Python.Runtime
         public IntPtr mdef;
         public IntPtr ptr;
         private bool _disposed = false;
-
         private ThunkInfo _thunk;
 
         public MethodWrapper(Type type, string name, string funcType = null)
--- a/src/runtime/methodwrapper.cs
+++ b/src/runtime/methodwrapper.cs
@@ -12,18 +12,19 @@ namespace Python.Runtime
     {
         public IntPtr mdef;
         public IntPtr ptr;
+        private ThunkInfo thunk;
 
         public MethodWrapper(Type type, string name, string funcType = null)
         {
             // Turn the managed method into a function pointer
 
-            IntPtr fp = Interop.GetThunk(type.GetMethod(name), funcType);
+            thunk = Interop.GetThunk(type.GetMethod(name), funcType);
 
             // Allocate and initialize a PyMethodDef structure to represent
             // the managed method, then create a PyCFunction.
 
             mdef = Runtime.PyMem_Malloc(4 * IntPtr.Size);
-            TypeManager.WriteMethodDef(mdef, name, fp, 0x0003);
+            TypeManager.WriteMethodDef(mdef, name, thunk.Address, 0x0003);
             ptr = Runtime.PyCFunction_NewEx(mdef, IntPtr.Zero, IntPtr.Zero);
         }
 
--- a/src/runtime/typemanager.cs
+++ b/src/runtime/typemanager.cs
@@ -1,9 +1,11 @@
 using System;
 using System.Collections;
 using System.Collections.Generic;
+using System.Linq;
 using System.Reflection;
 using System.Runtime.InteropServices;
 using Python.Runtime.Platform;
+using Python.Runtime.Slots;
 
 namespace Python.Runtime
 {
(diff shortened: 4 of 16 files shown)


src/runtime/interop.cs
```diff
@@ -522,4 +517,27 @@ public Thunk(Delegate d)
...
+        public static readonly ThunkInfo Empty = new ThunkInfo(null, IntPtr.Zero);
+
+        public ThunkInfo(Delegate target, IntPtr addr)
+        {
+            Target = target;
+            Address = addr;
+        }
+    }
+    
+    [StructLayout(LayoutKind.Sequential, CharSet = CharSet.Ansi)]
```
lostmsu: NIT: none of the fields have strings, so why does `CharSet` matter?

src/runtime/metatype.cs
```diff
@@ -22,6 +22,15 @@ public static IntPtr Initialize()
...
         }
 
+        public static void Release()
+        {
+            if (Runtime.Refcount(PyCLRMetaType) > 1)
+            {
+                SlotsHolder.ReleaseTypeSlots(PyCLRMetaType);
+            }
+            Runtime.XDecref(PyCLRMetaType);
+            PyCLRMetaType = IntPtr.Zero;
```
lostmsu: Since you are setting `PyCLRMetaType` to 0, I think it would be a good idea to check for it in the beginning of the method, and throw something like `ObjectDisposedException`

src/runtime/metatype.cs
```diff
@@ -22,6 +22,15 @@ public static IntPtr Initialize()
             return PyCLRMetaType;
         }
 
+        public static void Release()
+        {
+            if (Runtime.Refcount(PyCLRMetaType) > 1)
```
lostmsu: Not sure I understand the reasoning behind `> 1` condition.
In fact, why even use `Refcount`? This class should have its own way to track if `PyCLRMetaType` is allocated.

src/runtime/typemanager.cs
```diff
@@ -739,13 +835,34 @@ internal static void InitializeSlots(IntPtr type, Type impl)
                 // load them into a separate code page that is leaked
                 // intentionally.
                 InitializeNativeCodePage();
-                ret1 = NativeCodePage + native.Return1;
-                ret0 = NativeCodePage + native.Return0;
+                IntPtr ret1 = NativeCodePage + native.Return1;
+                IntPtr ret0 = NativeCodePage + native.Return0;
+                InitializeSlot(type, ret0, "tp_traverse", false);
+                InitializeSlot(type, ret0, "tp_clear", false);
```
lostmsu: NIT: can you add parameter name to `false` here? (e.g. `paramName: false`) Otherwise it is unclear what it means.

src/runtime/typemanager.cs
```diff
@@ -761,14 +878,43 @@ internal static void InitializeSlots(IntPtr type, Type impl)
         /// </summary>
         /// <param name="type">Type being initialized.</param>
         /// <param name="slot">Function pointer.</param>
-        /// <param name="name">Name of the method.</param>
-        static void InitializeSlot(IntPtr type, IntPtr slot, string name)
+        /// <param name="canOverride">Can override the slot when it existed</param>
+        static void InitializeSlot(IntPtr type, IntPtr slot, string name, bool canOverride = true)
```
lostmsu: Can you explain why is it OK sometimes to not overwrite a slot?

I see this is used for `tp_traverse` and `tp_clear`. What are consequences of not overwriting this slot? Would not this cause our cleanup code to simply fail to be run?

src/runtime/typemanager.cs
```diff
@@ -817,4 +963,183 @@ internal static void CopySlot(IntPtr from, IntPtr to, int offset)
...
+        public const string HolderKeyName = "_slots_holder";
+        public delegate void Resetor(IntPtr type, int offset);
+
+        private GCHandle _handle;
+        private Interop.DestructorFunc _destructor;
+        private IntPtr _capsule;
+        private IntPtr _type;
+        private Dictionary<int, ThunkInfo> _slots = new Dictionary<int, ThunkInfo>();
+        private List<Delegate> _keepalive = new List<Delegate>();
+        private Dictionary<int, Resetor> _customRestors = new Dictionary<int, Resetor>();
```
lostmsu: NIT: `_customResetors`?

src/runtime/typemanager.cs
```diff
@@ -817,4 +963,183 @@ internal static void CopySlot(IntPtr from, IntPtr to, int offset)
             Marshal.WriteIntPtr(to, offset, fp);
         }
     }
+
+    class SlotsHolder
+    {
+        public const string HolderKeyName = "_slots_holder";
+        public delegate void Resetor(IntPtr type, int offset);
```
lostmsu: NIT: this delegate type looks like a generic action on a slot. Maybe `SlotAction` or `TypeSlotAction`?

src/runtime/typemanager.cs
```diff
@@ -817,4 +963,183 @@ internal static void CopySlot(IntPtr from, IntPtr to, int offset)
...
+        }
+
+        public void Add(int offset, ThunkInfo thunk)
+        {
+            _slots.Add(offset, thunk);
+        }
+
+        public void Add(int offset, Resetor resetor)
+        {
+            _customRestors[offset] = resetor;
```
lostmsu: This really overwrites the previous value without any warning. Probably should use `.Add`

src/runtime/typemanager.cs
```diff
@@ -817,4 +963,183 @@ internal static void CopySlot(IntPtr from, IntPtr to, int offset)
...
+            {
+                Runtime.XIncref(_capsule);
+                return _capsule;
+            }
+            _handle = GCHandle.Alloc(this);
+            _destructor = OnDestruct;
+            var fp = Marshal.GetFunctionPointerForDelegate(_destructor);
+            _capsule = Runtime.PyCapsule_New((IntPtr)_handle, null, fp);
+            return _capsule;
+        }
```
lostmsu: From a name `ToCapsule`, I would expect this method to behave similarly as `ToString`. E.g. first, I'd expect it to not modify the object state at all, second, I'd expect it to construct a new object every time it is called, and for it to be callable multiple times.

Overall, it would be good to either have a more detailed description of what this class does, and how capsules fit into that, or refactor code to it would be clear what it does.

lostmsu: It is also unclear why you can't just use `slotsHolder.ToPython()`.

src/runtime/typemanager.cs
```diff
@@ -817,4 +963,183 @@ internal static void CopySlot(IntPtr from, IntPtr to, int offset)
...
+
+        public static void ReleaseTypeSlots(IntPtr type)
+        {
+            IntPtr capsule = Runtime.PyObject_GetAttrString(type, HolderKeyName);
+            if (capsule == IntPtr.Zero)
+            {
+                return;
+            }
```
lostmsu: Is there a scenario, where calling `ReleaseTypeSlots` twice on the same type is meaningful? If not, this should throw `InvalidOperationException` instead.

src/runtime/typemanager.cs
```diff
@@ -817,4 +963,183 @@ internal static void CopySlot(IntPtr from, IntPtr to, int offset)
...
+
+        private void ResetSlots()
+        {
+            if (_alredyReset)
+            {
+                return;
+            }
+            _alredyReset = true;
+            IntPtr tp_name = Marshal.ReadIntPtr(_type, TypeOffset.tp_name);
+            string typeName = Marshal.PtrToStringAnsi(tp_name);
```
lostmsu: Judging by the code around typemanager.cs:430, the value of `tp_name` can well be a Unicode string. In Python3 they clearly use UTF8.

src/runtime/typemanager.cs
```diff
@@ -817,4 +963,183 @@ internal static void CopySlot(IntPtr from, IntPtr to, int offset)
...
+            _deallocators.Clear();
+
+            // Custom reset
+            IntPtr tp_base = Marshal.ReadIntPtr(_type, TypeOffset.tp_base);
+            Runtime.XDecref(tp_base);
+            Marshal.WriteIntPtr(_type, TypeOffset.tp_base, IntPtr.Zero);
+
+            IntPtr tp_bases = Marshal.ReadIntPtr(_type, TypeOffset.tp_bases);
+            Runtime.XDecref(tp_bases);
+            Marshal.WriteIntPtr(_type, TypeOffset.tp_bases, IntPtr.Zero);
```
lostmsu: Is overwriting base types in a type object a defined behavior in Python?
Why do we need this?

src/runtime/typemanager.cs
```diff
@@ -817,4 +963,183 @@ internal static void CopySlot(IntPtr from, IntPtr to, int offset)
...
+
+            // Because __len__ may have a "default" implementation, not registered to us
+            // always explicitly clobber it. See mp_lenth_slot class
+            Marshal.WriteIntPtr(_type, TypeOffset.mp_length, IntPtr.Zero);
+        }
+
+        private static void OnDestruct(IntPtr ob)
```
lostmsu: Is `ob` a Python type here? Perhaps then name as such.

src/runtime/typemanager.cs
```diff
@@ -817,4 +963,183 @@ internal static void CopySlot(IntPtr from, IntPtr to, int offset)
...
+            self.ResetSlots();
+        }
+
+        private static SlotsHolder RecoverFromCapsule(IntPtr ob)
+        {
+            var ptr = Runtime.PyCapsule_GetPointer(ob, null);
+            PythonException.ThrowIfIsNull(ptr);
+            GCHandle handle = GCHandle.FromIntPtr(ptr);
+            return (SlotsHolder)handle.Target;
+        }
```
lostmsu: This all would look much nicer, if all capsule-related stuff would be extracted into a separate class (e.g. like `PyCapsule<T>`)

src/runtime/typemanager.cs
```diff
@@ -817,4 +963,183 @@ internal static void CopySlot(IntPtr from, IntPtr to, int offset)
...
+
+        private static IntPtr GetDefaultSlot(int offset)
+        {
+            if (offset == TypeOffset.tp_clear
+                || offset == TypeOffset.tp_traverse)
+            {
+                return TypeManager.NativeCodePage + TypeManager.NativeCode.Active.Return0;
```
lostmsu: Maybe I missed a conversation. What is the story for ARM in this code?

benoithudson:
FYI: Felix is back from holiday on Monday.

src/runtime/interop.cs
BadSingleton: It doesn't, I'll remove it.

src/runtime/typemanager.cs
BadSingleton: >[...] I would expect this method to behave similarly as ToString
I agree, I'll change it.

From what I see the usage of the capsule fits the semantics they were created for. Maybe we  could also use a regular. @amos402 Any other reason to use a capsule that I'm not seeing?

BadSingleton: > [...] I would expect this method to behave similarly as ToString

I agree, I'll change it.

From what I see the usage of the capsule fits the semantics they were created for. Maybe we could also use a regular. @amos402 Any other reason to use a capsule that I'm not seeing?

src/runtime/interop.cs
BadSingleton: It doesn't, I'll remove it.

src/runtime/metatype.cs
BadSingleton: I think the reasoning is the same as with the similar pattern in typemanager.cs#38. The only way I can see this class tracking it's allocation is to look if the `PyCLRMetaType` is `IntPtr.Zero` or not.

src/runtime/typemanager.cs
BadSingleton: Yes, but the only action performed/context it is used for now is Reset.

src/runtime/typemanager.cs
BadSingleton: Agreed. But this may have been the original intention? @amos402 ?

src/runtime/metatype.cs
benoithudson: The logic is: If the refcount is 1, then at line 31 when decref, it'll go to zero and the entire type will be released. In that case we don't care about clearing the slots. It's a standard optimization in a ref-counting framework.

It's correct assuming that ReleaseTypeSlots doesn't also release refcounts that otherwise won't be released.

That said: Do we have coverage on line 29? I'm not sure I really understand under what circumstance it happens.

src/runtime/typemanager.cs
BadSingleton: There could be. I'll add the exception and see what happens.

src/runtime/typemanager.cs
BadSingleton: > Is overwriting base types in a type object a defined behavior in Python?

It should, both of these values are still valid if `NULL`.
As to why, this is the reason of this PR. On domain reload (and unload) we clobber the Python objects that points to C#, leaving empty, but still valid, Python objects that no longer refer to code that is now gone.

src/runtime/typemanager.cs
BadSingleton: I'm not sure if this is what you're looking for, but it seems that it was removed in #913

src/runtime/typemanager.cs
BadSingleton: Will do

src/runtime/typemanager.cs
```diff
@@ -311,6 +337,24 @@ internal static IntPtr WriteMethodDefSentinel(IntPtr mdef)
             return WriteMethodDef(mdef, IntPtr.Zero, IntPtr.Zero, 0, IntPtr.Zero);
         }
 
+        internal static void FreeMethodDef(IntPtr mdef)
+        {
+            unsafe
```
benoithudson: Have we verified this still works in Unity with default options? Default Unity settings block unsafe code. I think it's just at compile-time, so unsafe compiled code is fine, but I want to make sure.

src/runtime/typemanager.cs
```diff
@@ -343,30 +388,76 @@ internal static IntPtr CreateMetaType(Type impl)
...
+                IntPtr mdefAddr = mdef;
+                slotsHolder.AddDealloctor(() =>
+                {
+                    IntPtr t = type;
+                    IntPtr tp_dict = Marshal.ReadIntPtr(t, TypeOffset.tp_dict);
+                    if (Runtime.PyDict_DelItemString(tp_dict, "__instancecheck__") != 0)
+                    {
+                        Runtime.PyErr_Print();
+                    }
+                    FreeMethodDef(mdefAddr);
```
benoithudson: Why do we open a scope and create a new local variable that just copies a local variable that already exists?

benoithudson: Oh, I get it -- so that the lambda grabs the right value.

Feels like we could refactor these two scopes into a function, so it's a bit less mysterious.

src/runtime/typemanager.cs
```diff
@@ -739,13 +835,34 @@ internal static void InitializeSlots(IntPtr type, Type impl)
...
+                    if (slotsHolder != null)
+                    {
+                        slotsHolder.Add(offset, thunkRet0);
+                    }
+                }
             }
-
-            InitializeSlot(type, ret0, "tp_traverse");
-            InitializeSlot(type, ret0, "tp_clear");
-            InitializeSlot(type, ret1, "tp_is_gc");
```
benoithudson: Why does this work without `tp_is_gc`? It used to be required (I didn't really fully understand it, but leaving it out led to crashes).

If we actually don't need it then we can remove everything about Return1 from the code.

src/runtime/typemanager.cs
```diff
@@ -739,13 +835,34 @@ internal static void InitializeSlots(IntPtr type, Type impl)
...
+            {
+                if (!IsSlotSet(type, "tp_traverse"))
+                {
+                    var thunkRet0 = Interop.GetThunk(((Func<IntPtr, int>)Return0).Method);
+                    var offset = GetSlotOffset("tp_traverse");
+                    Marshal.WriteIntPtr(type, offset, thunkRet0.Address);
+                    if (slotsHolder != null)
+                    {
+                        slotsHolder.Add(offset, thunkRet0);
+                    }
```
benoithudson: Why inline InitializeSlot here?

src/runtime/typemanager.cs
BadSingleton: I think it relates to other parts of #958 where some classes define their own `tp_traverse` and `tp_clear`. This could be removed for now

src/runtime/typemanager.cs
```diff
@@ -761,14 +878,43 @@ internal static void InitializeSlots(IntPtr type, Type impl)
...
             FieldInfo fi = typeOffset.GetField(name);
             var offset = (int)fi.GetValue(typeOffset);
 
-            Marshal.WriteIntPtr(type, offset, slot);
+            if (!canOverride && Marshal.ReadIntPtr(type, offset) != IntPtr.Zero)
+            {
+                return;
+            }
+            Marshal.WriteIntPtr(type, offset, thunk.Address);
+            slotsHolder.Add(offset, thunk);
```
benoithudson: won't this throw with the default slotsHolder?

Seems like we should remove the function with one less arg, and handle null here.

src/runtime/typemanager.cs
benoithudson: ResetSlotAction?

Some documentation of what the function should do + what the arguments mean would be nice.

src/runtime/typemanager.cs
```diff
@@ -817,4 +963,183 @@ internal static void CopySlot(IntPtr from, IntPtr to, int offset)
             Marshal.WriteIntPtr(to, offset, fp);
         }
     }
+
+    class SlotsHolder
+    {
+        public const string HolderKeyName = "_slots_holder";
+        public delegate void Resetor(IntPtr type, int offset);
+
+        private GCHandle _handle;
```
benoithudson: The C# world tends to either use `m_` or nothing at all to mark instance variables (I think that's the python.net tradition) -- but not just a leading underscore.

src/runtime/typemanager.cs
```diff
@@ -817,4 +963,183 @@ internal static void CopySlot(IntPtr from, IntPtr to, int offset)
...
+        private Interop.DestructorFunc _destructor;
+        private IntPtr _capsule;
+        private IntPtr _type;
+        private Dictionary<int, ThunkInfo> _slots = new Dictionary<int, ThunkInfo>();
+        private List<Delegate> _keepalive = new List<Delegate>();
```
benoithudson: Does this actually get used anywhere? I see Add and Clear called on it but nothing else.

src/runtime/typemanager.cs
```diff
@@ -817,4 +963,183 @@ internal static void CopySlot(IntPtr from, IntPtr to, int offset)
...
+        {
+            _customRestors[offset] = resetor;
+        }
+
+        public void AddDealloctor(Action deallocate)
+        {
+            _deallocators.Add(deallocate);
+        }
+
+        public void KeeapAlive(Delegate d)
```
benoithudson: The spelling is a bit odd?

A comment about what this does would be good.

src/runtime/typemanager.cs
```diff
@@ -817,4 +963,183 @@ internal static void CopySlot(IntPtr from, IntPtr to, int offset)
...
+            IntPtr tp_dict = Marshal.ReadIntPtr(type, TypeOffset.tp_dict);
+            if (Runtime.PyDict_DelItemString(tp_dict, HolderKeyName) != 0)
+            {
+                throw new PythonException();
+            }
+        }
+
+        private void ResetSlots()
+        {
+            if (_alredyReset)
```
benoithudson: typoe: alre*a*dyReset

src/runtime/typemanager.cs
benoithudson: Then it should be called Set or Replace.

src/runtime/typemanager.cs
benoithudson: If it's just a place to park a pointer to avoid getting GC'd then the KeeapAlive (!) function should document it.

src/runtime/typemanager.cs
benoithudson: This variable isn't used. We should remove it (and tp_name), and also remove the commented-out use at line 1055.

src/runtime/typemanager.cs
```diff
@@ -817,4 +963,183 @@ internal static void CopySlot(IntPtr from, IntPtr to, int offset)
...
+            IntPtr tp_base = Marshal.ReadIntPtr(_type, TypeOffset.tp_base);
+            Runtime.XDecref(tp_base);
+            Marshal.WriteIntPtr(_type, TypeOffset.tp_base, IntPtr.Zero);
+
+            IntPtr tp_bases = Marshal.ReadIntPtr(_type, TypeOffset.tp_bases);
+            Runtime.XDecref(tp_bases);
+            Marshal.WriteIntPtr(_type, TypeOffset.tp_bases, IntPtr.Zero);
+
+            // Because __len__ may have a "default" implementation, not registered to us
+            // always explicitly clobber it. See mp_lenth_slot class
```
benoithudson: Is there a typo in mp_len*g*th_slot?

src/runtime/typemanager.cs
benoithudson: Here and below, `ob` should be `tp`

src/runtime/typemanager.cs
```diff
@@ -817,4 +963,183 @@ internal static void CopySlot(IntPtr from, IntPtr to, int offset)
...
+        {
+            if (offset == TypeOffset.tp_clear
+                || offset == TypeOffset.tp_traverse)
+            {
+                return TypeManager.NativeCodePage + TypeManager.NativeCode.Active.Return0;
+            }
+            else if (offset == TypeOffset.tp_dealloc)
+            {
+                // tp_free of PyTypeType is point to PyObejct_GC_Del.
+                return Marshal.ReadIntPtr(Runtime.PyTypeType, TypeOffset.tp_free);
```
benoithudson: PyObejct_GC_Del has a typo

src/runtime/typemanager.cs
```diff
@@ -817,4 +963,183 @@ internal static void CopySlot(IntPtr from, IntPtr to, int offset)
...
+
+        private static SlotsHolder RecoverFromCapsule(IntPtr ob)
+        {
+            var ptr = Runtime.PyCapsule_GetPointer(ob, null);
+            PythonException.ThrowIfIsNull(ptr);
+            GCHandle handle = GCHandle.FromIntPtr(ptr);
+            return (SlotsHolder)handle.Target;
+        }
+
+        private static IntPtr GetDefaultSlot(int offset)
```
benoithudson: I think the doc here should say that we're looking for the C function to call when resetting the slot at the given offset within the type.

We might want to eventually consider having those functions throw exceptions e.g. "the underlying .NET type has been deleted"

src/runtime/typemanager.cs
amos402: Since the  slot can be overwrite, they need to be overwrite able, so it may change to Set.

src/runtime/typemanager.cs
amos402: `ToPython` cannot be use here, some functions of pythonnet had been shutdown.
I used capsule as a simple mechanism to track the life cycle of type, so it can call a callback when then type deallocated, we can handle this manually as well if my conception correct. So my next step would be try to remove this mechanism.

src/runtime/typemanager.cs
amos402: Just some debug code I left...

src/runtime/typemanager.cs
amos402: No need to handle mp_length manually here, it already included in common case.

src/runtime/typemanager.cs
BadSingleton: Yes, it does, there are other uses of unsafe code in the project. (XDecref for instance)

src/runtime/typemanager.cs
BadSingleton: Can't find a good reason, will uninline

src/runtime/typemanager.cs
BadSingleton: It was probably an oversight, and I added it back.

src/runtime/typemanager.cs
BadSingleton: The capsules mechanism was removed by @amos402 in commit 433d0f6, partially cherry-picked for this branch

src/runtime/metatype.cs
```diff
@@ -11,17 +11,30 @@ namespace Python.Runtime
     internal class MetaType : ManagedType
     {
         private static IntPtr PyCLRMetaType;
-
+        private static SlotsHolder _metaSlotsHodler;
```
benoithudson: I just noticed the typo here. This is holding slots, not bitcoin.

src/runtime/methodwrapper.cs
```diff
@@ -13,7 +13,6 @@ internal class MethodWrapper
         public IntPtr mdef;
         public IntPtr ptr;
         private bool _disposed = false;
-
```
benoithudson: if you add this empty line back in you remove an entire file from the squash-commit.

src/runtime/typemanager.cs
```diff
@@ -819,5 +942,182 @@ internal static void CopySlot(IntPtr from, IntPtr to, int offset)
...
+
+        private static SlotsHolder CreateSlotsHolder(IntPtr type)
+        {
+            var holder = new SlotsHolder(type);
+            _slotsHolders.Add(type, holder);
+            return holder;
+        }
+    }
+
+    class SlotsHolder
```
lostmsu: Put this class into a separate file

src/runtime/typemanager.cs
lostmsu: Problem is: with domain gone, who will provide those functions?

src/runtime/typemanager.cs
```diff
@@ -819,5 +942,182 @@ internal static void CopySlot(IntPtr from, IntPtr to, int offset)
...
+                // PyObject_GenericSetAttr
+                return Marshal.ReadIntPtr(Runtime.PyBaseObjectType, TypeOffset.tp_setattro);
+            }
+
+            return Marshal.ReadIntPtr(Runtime.PyTypeType, offset);
+        }
+    }
+
+
+    static class SlotHelper
```
lostmsu: Does not seem to be used anywhere

BadSingleton:
Work has stopped on our side on this, so I'm closing the PR. We might get back to it eventually.

PR #1018, opened 2019-12-17
Allow formatting an exception with traceback.format_exception

slide:

This implements the ability to format an exception using traceback.format_exception. This is very useful for when you are embedding Python into an application and want to show the Python exception similar to how it would be printed in the Python console.

None that I know of.

### Checklist


slide:
From looking at the logs for the travis builds, it seems like the agent on travis has an issue rather than the test?

slide:
Close and reopen for another build try.

filmor:
@slide Really sorry for the mess, could you please resubmit the PR? I accidentally merged it when trying to update it to the current master and my browser got stuck ...

slide:
Sure, not a problem. I assume you will revert the commit?

filmor:
I immediately force-pushed the previous state.

PR #1019, opened 2019-12-18
Decref the members of Runtime(split from #958)

amos402:
* Add mechanism for calling XDecref and set member to null automately.
* Unified the GetBuiltins method

...

### Any other comments?

...

### Checklist

### DIFF
--- a/src/runtime/runtime.cs
+++ b/src/runtime/runtime.cs
@@ -198,20 +198,29 @@ namespace Python.Runtime
             IntPtr op;
             {
                 var builtins = GetBuiltins();
-                SetPyMember(ref PyNotImplemented, PyObject_GetAttrString(builtins, "NotImplemented"));
+                SetPyMember(ref PyNotImplemented, PyObject_GetAttrString(builtins, "NotImplemented"),
+                    () => PyNotImplemented = IntPtr.Zero);
 
-                SetPyMember(ref PyBaseObjectType, PyObject_GetAttrString(builtins, "object"));
+                SetPyMember(ref PyBaseObjectType, PyObject_GetAttrString(builtins, "object"),
+                    () => PyBaseObjectType = IntPtr.Zero);
 
-                SetPyMember(ref PyNone, PyObject_GetAttrString(builtins, "None"));
-                SetPyMember(ref PyTrue, PyObject_GetAttrString(builtins, "True"));
-                SetPyMember(ref PyFalse, PyObject_GetAttrString(builtins, "False"));
+                SetPyMember(ref PyNone, PyObject_GetAttrString(builtins, "None"),
+                    () => PyNone = IntPtr.Zero);
+                SetPyMember(ref PyTrue, PyObject_GetAttrString(builtins, "True"),
+                    () => PyTrue = IntPtr.Zero);
+                SetPyMember(ref PyFalse, PyObject_GetAttrString(builtins, "False"),
+                    () => PyFalse = IntPtr.Zero);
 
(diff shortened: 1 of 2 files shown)


src/runtime/runtime.cs
```diff
@@ -1945,5 +1962,54 @@ internal static void SetNoSiteFlag()
...
+            }
+        }
+
+        public void Release()
+        {
+            foreach (var objRef in _objects)
+            {
+                unsafe
+                {
+                    var p = (void**)objRef;
```
lostmsu: This looks overly complicated, and uses undefined behavior. There's no reason to assume CLR will never change addresses of those static fields, so trying to remember them for future access is a bad idea.

Also, it feels a bit overengineered. I understand the problem of accidentally forgetting to decref a field, and how zeroing fields could help debugging, but the implementation is too tricky and thus prone to errors on its own.

I think it would be quite more readable to use an ordinary `List<IntPtr>` for all objects. To zero fields you can make a holder struct for all of them, put an instance of it into `Runtime`, and simply call `Holder = default` at the end of `Shutdown`

amos402: I just check the ECMA, the static fields is indeed moveable variable. The motivation is I want to make the assignments and action that set the member to null  in a single short line without a another hard code region. Thus a holder records the pointers is still unacceptable.
I updated the code, using lambda callbacks to set them to IntPtr.Zero, well, although it became two lines and needs a explicit assignment, the operations still in one code region(still better than separate them I think).
Seems no other better ways to achieve the object.
Since there's the only use case here, I decide not to put the `op` as the action's parameter.

src/runtime/runtime.cs
```diff
@@ -393,6 +415,19 @@ internal static int AtExit()
             return 0;
         }
 
+        private static void SetPyMember(ref IntPtr obj, IntPtr value, Action onRelease)
```
lostmsu: NIT: `obj` can be `out IntPtr`

src/runtime/runtime.cs
```diff
@@ -1945,5 +1984,39 @@ internal static void SetNoSiteFlag()
...
+    {
+        private List<KeyValuePair<IntPtr, Action>> _actions = new List<KeyValuePair<IntPtr, Action>>();
+
+        /// <summary>
+        /// Record obj's address to release the obj in the future,
+        /// obj must alive before calling Release.
+        /// </summary>
+        public void Add(IntPtr ob, Action onRelease)
+        {
+            _actions.Add(new KeyValuePair<IntPtr, Action>(ob, onRelease));
```
lostmsu: NIT: use `Tuple<T1, T2>`, e.g. just `_actions.Add((ob, onRelease))`

src/runtime/runtime.cs
```diff
@@ -1945,5 +1984,39 @@ internal static void SetNoSiteFlag()
...
+        }
+
+        public void Release()
+        {
+            foreach (var item in _actions)
+            {
+                Runtime.XDecref(item.Key);
+                item.Value?.Invoke();
+            }
```
lostmsu: NIT: use `foreach(var (ob, onRelease) in _actions)` (requires modification from previous comment).

src/runtime/runtime.cs
amos402: Why using Tuple? As my experience, with two elements cases, KeyValue would always faster than tuple(unless the struct is too big).

lostmsu: Simply for readability. This code is not a hot path, so a little performance difference hardly matters.
Also, once we retarget to newer framework, C# will use `ValueTuple` anyway.

PR #1020, opened 2019-12-19
Add method to format exception traceback

slide:

This implements the ability to format an exception using traceback.format_exception. This is very useful for when you are embedding Python into an application and want to show the Python exception similar to how it would be printed in the Python console.

None that I know of.

### Checklist


slide:
I will look into the failures and see if I can reproduce them.

PR #1021, opened 2019-12-19
Added parameter validation to PyObject methods

lostmsu:

This adds basic validations to input parameters of various `PyObject` methods, which should help in debugging code, that uses Python.NET.

Also adds a few function overloads

No, but it reduces divergence with Lost Tech fork

### Checklist

### DIFF
--- a/src/runtime/pyobject.cs
+++ b/src/runtime/pyobject.cs
@@ -274,7 +274,6 @@ namespace Python.Runtime
         /// Returns the named attribute of the Python object, or the given
         /// default object if the attribute access fails.
         /// </remarks>
-        [Obsolete("This method ignores any Python exceptions, and should not be used. Use GetAttrOrElse.")]
         public PyObject GetAttr(string name, PyObject _default)
         {
             if (name == null) throw new ArgumentNullException(nameof(name));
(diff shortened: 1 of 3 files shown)


lostmsu:
@filmor any chance to look at this? This is a pretty light change.

src/runtime/pyobject.cs
```diff
@@ -257,17 +268,41 @@ public PyObject GetAttr(string name)
...
+                Runtime.PyErr_Clear();
+                return _default;
+            }
+            return new PyObject(op);
+        }
+
+        /// <summary>
+        /// Gets the specified attribute. If attribute is not found, returns the fallback value.
+        /// </summary>
+        public PyObject GetAttrOrElse(string name, PyObject _default)
```
filmor: How much would it hurt to make this behaviour the default? We are using `GetAttr` a lot, I think (even in this file).

filmor:
Almost completely fine, but I'm not quite sure about the `GetAttr` changes, in particular since `GetAttr(PyObject)` is unchanged AFAICT. Will have a closer look.

src/runtime/pyobject.cs
lostmsu: @filmor I don't know, I don't have any legacy code :)
But that would be a breaking change in contract.

lostmsu: I decided to take this change out, so the rest could be merged.

lostmsu:
@filmor I decided to take the `GetAttrOrElse` out for now so that the rest could be merged.

PR #1022, opened 2019-12-19
Codecs: customize set of Py<->C# conversions via PyObjectConversions

lostmsu:

This enables registering new automatic Py->C# and C#->Py conversions (codecs), that will run when Python calls .NET code and vice versa.

This is a new feature

### Any other comments?

...

### Checklist

-   [x] If an enhancement PR, please create docs and at best an example: see `TupleCodec`


src/embed_tests/Python.EmbeddingTest.15.csproj
```diff
@@ -23,7 +23,7 @@
     <SolutionDir Condition="$(SolutionDir) == '' Or $(SolutionDir) == '*Undefined*'">..\..\</SolutionDir>
     <PythonBuildDir Condition="'$(TargetFramework)'=='net40' AND '$(PythonBuildDir)' == ''">$(SolutionDir)\bin\</PythonBuildDir>
     <PublishDir Condition="'$(TargetFramework)'!='net40'">$(OutputPath)\$(TargetFramework)_publish</PublishDir>
-    <LangVersion>6</LangVersion>
+    <LangVersion>7.3</LangVersion>
```
koubaa: Is this intentional? @lostmsu

lostmsu: Yes, I think I am using a few features. Is there a C# compiler, that does not support 7.3 yet?

src/runtime/runtime.cs
```diff
@@ -567,6 +569,15 @@ internal static unsafe void XIncref(IntPtr op)
 #endif
         }
 
+        /// <summary>
+        /// Increase Python's ref counter for the given object, and get the object back.
+        /// </summary>
+        internal static IntPtr SelfIncRef(IntPtr op)
```
koubaa: This is neat and might simplify other code in the project

koubaa:
I like the conciseness of the terminology (encode/decode) but for me it isn't entirely obvious which direction (py <-> C#) is encoding and decoding.  It is ok with me as long as we document it well enough.

Which other encoders should exist inside of Python.Runtime aside from the Tuple one you've added?  Would you be OK with the List and Action ones I am interested in eventually being distributed by PythonNet?

I am writing the below snippet to document how it'd be available for external codecs to see how we like the syntax.  To me it is not too bad:

````
import clr
clr.AddReference("MyCodecs")
import MyCodecs
for encoder in MyCodecs.Encoders:
    Python.Runtime.PyObjectConversions.RegisterEncoder(encoder)
for decoder in MyCodecs.Decoders:
    Python.Runtime.PyObjectConversions.RegisterDecoder(decoder)
````

File: src/embed_tests/Python.EmbeddingTest.15.csproj (hunk 1 above)
Comment (koubaa) [reply src/embed_tests/Python.EmbeddingTest.15.csproj]: @lostmsu just wanted to make sure it wasn't added by mistake.

File: src/runtime/converter.cs [hunk 3]
```diff
@@ -135,8 +135,16 @@ internal static IntPtr ToPython(object value, Type type)
                 return result;
             }
 
-			if (value is IList && !(value is INotifyPropertyChanged) && value.GetType().IsGenericType)
-			{
+            if (Type.GetTypeCode(type) == TypeCode.Object && value.GetType() != typeof(object)) {
```
koubaa: glad this comes before other builtin conversions but ultimately the built-in conversions should probably use this codec system internally.

src/runtime/converterextensions.cs
```diff
@@ -0,0 +1,185 @@
...
+        static readonly MethodInfo genericDecode = typeof(IPyObjectDecoder).GetMethod(nameof(IPyObjectDecoder.TryDecode));
+
+        #endregion
+
+        internal static void Reset()
+        {
+            lock (encoders)
+                lock (decoders)
+                {
+                    clrToPython.Clear();
```
koubaa: why do we need clrToPython and pythonToClr if we already have decoders and encoders?

koubaa: Got it, a single instance of an encoder/decoder can handle many types so these containers lazily process the list of encoders and decoders on each type encountered.

lostmsu:
Pushed a fix to failing test

lostmsu:
Mark methods as obsolete until they are stable.

lostmsu:
@filmor I think this can be merged

lostmsu:
Either I got lucky in this run, or #1050 actually was the real cause for test failures in this branch, and so was fixed by #1062.

src/embed_tests/TestPyScope.cs
```diff
@@ -337,9 +338,12 @@ public void TestThread()
                     //add function to the scope 
                     //can be call many times, more efficient than ast 
                     ps.Exec(
+                        "import clr\n" +
+                        "from System.Threading import Thread\n" +
                         "def update():\n" +
                         "    global res, th_cnt\n" +
                         "    res += bb + 1\n" +
+                        "    Thread.MemoryBarrier()\n" +
```
amos402: @lostmsu  Not a good idea for using some clr code here, and the counter didn't matched were because `res += bb + 1` it's not a atomic operate, they had been separated to several op codes. So even make some barrier here it may still be failed(if pushing more theads), but decrease the probability.
thread 1: executing `bb + 1` before assign to `res`
thread 2: executing op `STORE_FAST` of `res += `
context changed between these two threads, thus res may only increase once.
this should be enough: https://github.com/pythonnet/pythonnet/pull/958/commits/da97502006791bb0597446766ad00a6f9d291895

lostmsu: Regarding CLR: Python does not have a defined memory model, and I could not find any primitive analogous `Thread.MemoryBarrier` in Python. `threading.Lock` that you suggested does not have any mentions of memory barriers, so I could not rely on it.

Regarding atomicity. I think your understanding of the issue is incorrect. All calls to `update` functions from .NET threads are performed while having GIL (they are wrapped into `using Py.GIL()`), so only 1 thread executes the entirety of `update` at once and from this standpoint of view it should be considered atomic.

The issue is that acquiring GIL (at least judging by Python 2.7 code) does not introduce any kind of memory barrier, so after .NET thread 1 finished executing `update`, and .NET thread 2 entered it, thread 2 might not see the changes thread 1 made, even though thread 1 already completed and exited the `update` function.

amos402: That sounds right, but even so, it still needs the lock, change code like this, you shall see.
图片

lostmsu: Looks like you are right, memory barrier is insufficient. I though GIL ensures only 1 Python thread can execute while it is held, but actually it just ensures, the the current thread can execute Python code, and more than 1 can do so simultaneously. Now I wonder if we should simply lock in this test on .NET side. But I guess it does not matter, if you do it in Python or in C#. Might still need a memory barrier if Python's lock does not insert one.

lostmsu: Opened https://github.com/pythonnet/pythonnet/issues/1067 to track

amos402:
The code style of this PR is totally different by the current code, is it really good for this?

src/embed_tests/TestPyScope.cs
ddneilson: Have you considered a race condition? The `PyScopeTest` is not thread safe, in general. It is set up to have a single private `PyScope` at the class level that is created/disposed during `SetUp`/`Dispose`, and referenced in each test. However, tests within the fixture can, and do, run in parallel. 
So, `ps` can be modified mid-test for any test in this fixture. `TestThread` is one of the longer tests in the fixture, so one might expect it to be more vulnerable to the data races expected of a shared state variable being replaced.
Have you considered setting either setting `NonParallelizable` on all of the tests in the fixture, or making the fixture itself threadsafe by making each test create its own, locally scoped, `PyScope`?

igiona:
The issue #823 has been marked as solved by this PR.
It's not clear to me which approach has been taken here (from the ones described in the mentioned bug)?
@lostmsu is the PR going to make Python.NET fully support .NET objects providing the `IDynamicMetaObjectProvider ` interface? I'm moving away from IronPython (which supports `IDynamicMetaObjectProvider` objects) and running into many issues...

From what I can see, this new feature didn't make it to the latest release, is there a plan for for a 2.5.3 or there will be only the v3.0.0?
Thanks

lostmsu:
There's no support for `IDynamicMetaObjectProvider`. Instead for custom types to be presented to Python (or vice versa) as something they are not, you need to implement encoders and/or decoders: [customizing object marshalling between .NET and Python](https://github.com/pythonnet/pythonnet/wiki/Codecs:-customizing-object-marshalling-between-.NET-and-Python)

igiona:
Thank you @lostmsu for the quick answer.
This implies that the PythonNet assemblies needs to imported in the .NET application, which is not always possible/desirable (not technically, but rather "_politically_"/by-design.

What's the rational behind the decision to not support `IDynamicMetaObjectProvider` ?
Is there some architectural constraint that prevent this, or it has a chance to get supported by Python.NET in the future?

lostmsu:
The rational is that it is an open source project, and nobody stepped up to implement this feature.

igiona:
> The rational is that pythonnet is an open source project, and nobody stepped up to implement this feature.
This makes sense 👍 

I did not checked yet the implementation above into details, but I assumed that the codecs have to implement the interfaces `IPyObjectDecoder ` and/or `IPyObjectEncoder` in order to be able to be registered via `RegisterEncoder/RegisterDecoder`.
Which means that my C#-application (which provides these codec) needs to import the PaythonNET assemblies as dependency in order to compile ("_which is not always possible/desirable_"). Unless your implementation looks for the interface method names by reflection and wraps/maps them to the PythonNET interfaces internally!?

lostmsu:
Again, the app does not have to import pythonnet assemblies. Codecs can be put into a separate library.

PR #1031, opened 2020-01-15
Add Format method to pythonexception

slide:

Adds a Format method to PythonException which calls traceback.format_exception to get a nice exception message similar to something you would see on the Python console.

No

### Any other comments?

N/A

### Checklist


src/runtime/pythonexception.cs
```diff
@@ -145,6 +146,42 @@ public string PythonTypeName
...
+                        buffer.Append(val.ToString());
+                    }
+                    res = buffer.ToString();
+                }
+            }
+            else
+            {
+                res = "Missing exception/traceback information";
+            }
+            PythonEngine.ReleaseLock(gs);
```
lostmsu: Add `try ... finally ...` for the lock

src/runtime/pythonexception.cs
```diff
@@ -145,6 +146,42 @@ public string PythonTypeName
...
+                Runtime.XIncref(_pyValue);
+                Runtime.XIncref(_pyTB);
+                using (PyObject pyType = new PyObject(_pyType))
+                using (PyObject pyValue = new PyObject(_pyValue))
+                using (PyObject pyTB = new PyObject(_pyTB))
+                using (PyObject tb_mod = PythonEngine.ImportModule("traceback"))
+                {
+                    var buffer = new StringBuilder();
+                    var values = tb_mod.InvokeMethod("format_exception", pyType, pyValue, pyTB);
+                    Runtime.PyErr_Clear();
```
lostmsu: Normally, this would not need to be here. Since you are using high-level API to call Python code, the API should do all the error handling.

src/embed_tests/TestPythonException.cs
```diff
@@ -54,5 +54,41 @@ public void TestPythonErrorTypeName()
...
+        public void TestPythonExceptionFormat()
+        {
+            try
+            {
+                PythonEngine.Exec("raise ValueError('Error!')");
+                Assert.Fail("Exception should have been raised");
+            }
+            catch (PythonException ex)
+            {
+                Assert.That(ex.Format(), Does.Contain("Traceback").And.Contains("(most recent call last):").And.Contains("ValueError: Error!"));
```
lostmsu: Not sure if Python exceptions messages are localized. If that is the case, you need to ensure the right locale is used before performing this test.

src/runtime/pythonexception.cs
slide: I thought that would be good, but there are other places in the same file that don't, so I was following those. I will update.

src/embed_tests/TestPythonException.cs
slide: I will try and figure out if exception messages are localized.

slide:
Not sure why the Travis CI build is not showing up, it is here: https://travis-ci.org/github/pythonnet/pythonnet/builds/672296231

Issue #1047, opened 2020-02-20
ModuleObject __getattribute__ doesn't treat exceptions ocurred during internal GetAttribute


andreydani:
### Environment

-   Pythonnet version: 2.4.0
-   Python version: 3.7.6
-   Operating System: Windows 10

### Details

From Python code, using clr, when trying to access a class 'Plugin' inside a ModuleObject an exception System.IO.FileNotFoundException is thrown (this isn't a bug on pythonnet) and is never caught, so it isn't converted into a python exception.
The program aborts abnormally, and the Python code never receives the exception.

```python
   module_clr = __import__(namespace_assembly)
   if 'ModuleObject' in self.modulo_clr.__class__.__name__:
      plugin = module_clr.Plugin
```

lostmsu:
Since #1048 is merged, I am closing this issue. Please, open if it resurfaces on `master`

PR #1048, opened 2020-02-20
#1047 ModuleObject __getattribute__ doesn't treat exceptions ocurred during internal GetAttribute

andreydani:
…during internal GetAttribute

It converts any exception retured from GetAttribute into a Python exception

Yes, Issue #1047 

### Checklist

### DIFF
--- a/src/runtime/moduleobject.cs
+++ b/src/runtime/moduleobject.cs
@@ -280,7 +280,18 @@ namespace Python.Runtime
                 return self.dict;
             }
 
-            ManagedType attr = self.GetAttribute(name, true);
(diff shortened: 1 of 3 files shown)


src/runtime/moduleobject.cs
```diff
@@ -280,7 +280,18 @@ public static IntPtr tp_getattro(IntPtr ob, IntPtr key)
                 return self.dict;
             }
 
-            ManagedType attr = self.GetAttribute(name, true);
+            ManagedType attr = null;
```
lostmsu: NIT: `= null` is not needed AFAIK

Issue #1050, opened 2020-02-20
AppVeyor tests failing with timeout


lostmsu:
Some recent changes increased duration of our tests, so they now randomly do not finish in the allotted 7 minutes.

This change is known to increase 3.5 x64 build time: https://github.com/pythonnet/pythonnet/pull/992

However other legs are failing too.

lostmsu:
Fixed in #1051

lostmsu:
#1051 reduced the perf tests time, so they are no longer a factor. However build still gets stuck on

> => Python.EmbeddingTest.TestDomainReload.DomainReloadAndGC
> [Program.Main] === creating domain for assembly test1, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null
> [Program.Main] Before Domain Unload on test1, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null
> [Program.Main] After Domain Unload on test1, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null
> [Program.Main] The Proxy object is not valid anymore, domain unload complete.
> [Program.Main] === creating domain for assembly test2, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null
> [Program.Main] Before Domain Unload on test2, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null
> [Program.Main] After Domain Unload on test2, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null
> [Program.Main] The Proxy object is not valid anymore, domain unload complete.

lostmsu:
Here's a memory dump I got with #1052
[memory.zip](https://github.com/pythonnet/pythonnet/files/4238937/memory.zip)

lostmsu:
Seems like #1062 fixed the issue.

PR #1051, opened 2020-02-20
Report AppVeyor build timings

lostmsu:
This is to diagnose #1050 [AppVeyor tests failing with timeout](https://github.com/pythonnet/pythonnet/issues/1050)


lostmsu:
I reduced the number of iterations in perf tests. This should lower be enough for them to pass all time in AppVeyor.

PR #1052, opened 2020-02-21
If tests are taking too long, create a memory dump and abort

lostmsu:
This change is to investigate [AppVeyor tests failing with timeout](https://github.com/pythonnet/pythonnet/issues/1050) #1050

### DIFF
--- a/src/embed_tests/TestDomainReload.cs
+++ b/src/embed_tests/TestDomainReload.cs
@@ -211,9 +211,9 @@ namespace Python.EmbeddingTest
             {
                 Console.WriteLine($"[Program.Main] The Proxy object is valid ({theProxy}). Unexpected domain unload behavior");
             }
-            catch (Exception)
+            catch (Exception exception)
             {
-                Console.WriteLine("[Program.Main] The Proxy object is not valid anymore, domain unload complete.");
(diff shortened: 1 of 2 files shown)


lostmsu:
@filmor this is still useful, as I experience a similar issue with tests hanging just after starting `DomainReloadAndGC`: https://ci.appveyor.com/project/pythonnet/pythonnet/builds/31086476/job/aigfnc5gv4ni8stk , which I can not reproduce locally.

src/embed_tests/TestDomainReload.cs
```diff
@@ -211,9 +211,9 @@ static void RunAssemblyAndUnload(Assembly assembly, string assemblyName)
             {
                 Console.WriteLine($"[Program.Main] The Proxy object is valid ({theProxy}). Unexpected domain unload behavior");
             }
-            catch (Exception)
+            catch (Exception exception)
```
amos402: You should only catch the `AppDomainUnloadedException` and make it fail it can read the object.

https://github.com/pythonnet/pythonnet/blob/35cbe55b285923bc7253d545d3ae6f530703f4b3/src/embed_tests/TestDomainReload.cs#L163-L171

lostmsu: Fixed.

filmor:
Since we haven't seen this in quite a while, I'm going to close this PR.

PR #1053, opened 2020-02-21
Reduce perf iterations further to keep up with BenchmarkDotNet autotuning

lostmsu:
BenchmarkDotNet tuned the number of runs up to keep the total perf tests run time high. This reduces the number of runs per test further, which allows BenchmarkDotNet to do more runs of a smaller test to collect more information for better statistical results.


lostmsu:
This still affects #1050

PR #1058, opened 2020-02-23
Disable 3.8 tests in AppVeyor, as they never pass yet anyway

lostmsu:
Subj

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -23,22 +23,16 @@ environment:
       BUILD_OPTS: --xplat
     - PYTHON_VERSION: 3.7
       BUILD_OPTS: --xplat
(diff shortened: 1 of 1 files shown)


filmor:
What for? As they are allowed failures, we can just keep them until we have actually fixed the tests.

lostmsu:
@filmor they take extra ~20 minutes to run.

filmor:
Hmm

lostmsu:
I am reopening this, as it also affects attempts to rerun CI where a single leg randomly failed, as it always reruns all 3.8 in addition to that, turning a 5 min job into 25 min job.

filmor:
Okay, let's do this for now. Linking #1066 for reference.

PR #1062, opened 2020-02-26
Fixed reference counting for exception objects in Py.With

lostmsu:

`PyObject`s constructed for `__exit__` method referenced existing Python objects without increasing refcount appropriately, which could lead to double-free.

Not sure, but this issue might have been one of the possible causes for #1050


alexhelms:
I just wanted to comment on here to tell you that I ran into this issue and this PR was the root cause. We had some python exceptions being fired within a `Py.With` and it was causing memory exceptions and blowing up.

Thanks!

PR #1068, opened 2020-02-27
Explorarion of the new *Reference types: reimplemented some of the PyList members using BorrowedReference

lostmsu:
Just playing with the new *Reference types to see what patterns might arise.

This adds:
- `PyObject.Reference` internal property, that exposes the same pointer as `obj`, but in a way, that it is clear, that the reference is borrowed
- `PyObject(BorrowedReference)` constructor, that extracts the address from the reference and `IncRef`s it before storing in the instance
- a few of the `PyList_*` methods now explicitly take `BorrowedReference` instead of raw pointers


lostmsu:
@amos402 I noticed you had to make code changes using `PyList` in #958 . If you review this, those changes might not be needed anymore.

amos402:
@lostmsu Still needs, because I need to put them into collection and return somewhere. That's why I suggest making them copyable before.

PR #1075, opened 2020-03-03
Check for exceptions on Py_Finalize on Python 3.6+

lostmsu:
Subj

### DIFF
--- a/src/runtime/runtime.cs
+++ b/src/runtime/runtime.cs
@@ -8,7 +8,6 @@ using Python.Runtime.Platform;
 
 namespace Python.Runtime
 {
-
     /// <summary>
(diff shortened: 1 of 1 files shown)


filmor:
Should we raise here? The docs are a bit light on what status the interpreter is in after running `Py_FinalizeEx`, sounds to me more like a log-worthy thing.

lostmsu:
@filmor just looked at the source: https://github.com/python/cpython/blob/6df421fe87a9418d6c59f89dbc5d5573b6826855/Python/pylifecycle.c#L1410

It does indeed simply continue finalization, so I guess if there's an error there, it is not related to our `Shutdown`, so change is unimportant.

PR #298, opened 2016-12-01
Addresses issue #261 with providing more detail information missing d…

fdanny:
This addresses issue #261 to provide more information when there is a missing assembly dependency.  The logic basically does the following
1.  After importing module check if an ImportError exception occured
2. If it does load the assembly dependencies and try to load each one
3. if I can't load that assembly I add to the list
4. At the end I display the list of missing assemblies.

Issue #261, opened 2016-09-07
Give more explicit ImportErrors


filmor:
Currently, an `ImportError` raised from Python.NET does not include information on the actual dependency that failed to load.

denfromufa:
related? https://github.com/pythonnet/pythonnet/issues/111

filmor:
Could be. I just recognised this today as a problem, when I (again) had to do an `clr.AddReference ... .GetDependentAssemblies`-dance to find out which library could not be loaded.

denfromufa:
@filmor the problem is that some third-party packages try to import some libraries and have backup logic if no success. So how would we distinguish between package -handled and -unhandled imports?

filmor:
We should still raise an import error, we should just add information to the message, which assembly could not be loaded and why, basically the information that the.NET exception contains.

denfromufa:
what I meant is actually cases like this and we would bug down the user with optional imports failing:

https://github.com/pydata/pandas/blob/master/pandas/computation/__init__.py

filmor:
That's not what I'm saying, I'm suggesting keeping the `ImportError` and just extending the information it carries, i.e. you get `ImportError: No module named 'bla', tried to import bla.dll but failed loading the dependency blubb.dll because of reasons`.

denfromufa:
@filmor agreed! we need to extend the passed information.

lostmsu:
@filmor the right way to do it is to enable Fusion Log on Windows or set Mono debug environment variables for Mono.

PR #1076, opened 2020-03-04
Provide more info about failuers to load CLR assemblies

lostmsu:

1. When trying to implicitly load assemblies, and that fails **not** because an assembly is missing, but because loading failed for some reason, emit Python warning.
2. When trying to import a module in our import hook, if the module name is an assembly name, and CLR fails to load it, and Python also fails to find a module with the same name, add the exceptions we got from both during the attempt into `__cause__` attribute of the final `ImportError`.

BREAKING: `clr.AddReference` will now throw exceptions besides `FileNotFoundException`.

Additional: a few uses of `BorrowedReference`

This addresses https://github.com/pythonnet/pythonnet/issues/261
It is an alternative to https://github.com/pythonnet/pythonnet/pull/298

### Checklist

### DIFF
--- a/src/runtime/assemblymanager.cs
+++ b/src/runtime/assemblymanager.cs
@@ -137,7 +137,7 @@ namespace Python.Runtime
         /// </summary>
         internal static void UpdatePath()
         {
-            IntPtr list = Runtime.PySys_GetObject("path");
+            BorrowedReference list = Runtime.PySys_GetObject("path");
             var count = Runtime.PyList_Size(list);
             if (count != pypath.Count)
             {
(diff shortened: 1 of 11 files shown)


src/runtime/assemblymanager.cs
```diff
@@ -242,25 +227,18 @@ public static Assembly LoadAssemblyPath(string name)
         /// <returns></returns>
         public static Assembly LoadAssemblyFullPath(string name)
         {
-            Assembly assembly = null;
             if (Path.IsPathRooted(name))
             {
-                if (!Path.HasExtension(name))
+                if (!Path.HasExtension(name) && Runtime.InteropVersion < new Version(3, 0))
```
filmor, line 232: If this was for compatibility, you can probably remove it.

filmor:
Fine by me, but please add a changelog entry for the breaking change.

lostmsu:
I dropped the update, that tried to remove legacy "CLR." module from this change, as it started growing in scope, and is not directly related to this change.

PR #1080, opened 2020-03-06
Add codecs for functions

koubaa:

...

This adds codecs to convert between python functions and actions.  The tests which I added pass but I have some concerns about the code which I will comment about inline.
### Checklist

### DIFF
--- a/src/embed_tests/Codecs.cs
+++ b/src/embed_tests/Codecs.cs
@@ -6,28 +6,34 @@ namespace Python.EmbeddingTest {
     using Python.Runtime;
     using Python.Runtime.Codecs;
 
-    public class Codecs {
+    public class Codecs
+    {
         [SetUp]
-        public void SetUp() {
+        public void SetUp()
+        {
             PythonEngine.Initialize();
         }
 
         [TearDown]
-        public void Dispose() {
+        public void Dispose()
+        {
             PythonEngine.Shutdown();
         }
 
         [Test]
-        public void ConversionsGeneric() {
+        public void ConversionsGeneric()
+        {
             ConversionsGeneric<ValueTuple<int, string, object>, ValueTuple>();
         }
 
-        static void ConversionsGeneric<T, TTuple>() {
+        static void ConversionsGeneric<T, TTuple>()
+        {
             TupleCodec<TTuple>.Register();
             var tuple = Activator.CreateInstance(typeof(T), 42, "42", new object());
             T restored = default;
             using (Py.GIL())
-            using (var scope = Py.CreateScope()) {
+            using (var scope = Py.CreateScope())
+            {
                 void Accept(T value) => restored = value;
                 var accept = new Action<T>(Accept).ToPython();
                 scope.Set(nameof(tuple), tuple);
@@ -38,15 +44,18 @@ namespace Python.EmbeddingTest {
         }
 
         [Test]
-        public void ConversionsObject() {
+        public void ConversionsObject()
+        {
             ConversionsObject<ValueTuple<int, string, object>, ValueTuple>();
         }
-        static void ConversionsObject<T, TTuple>() {
+        static void ConversionsObject<T, TTuple>()
+        {
             TupleCodec<TTuple>.Register();
             var tuple = Activator.CreateInstance(typeof(T), 42, "42", new object());
             T restored = default;
             using (Py.GIL())
-            using (var scope = Py.CreateScope()) {
+            using (var scope = Py.CreateScope())
+            {
                 void Accept(object value) => restored = (T)value;
                 var accept = new Action<object>(Accept).ToPython();
                 scope.Set(nameof(tuple), tuple);
@@ -57,12 +66,15 @@ namespace Python.EmbeddingTest {
         }
 
         [Test]
-        public void TupleRoundtripObject() {
+        public void TupleRoundtripObject()
+        {
             TupleRoundtripObject<ValueTuple<int, string, object>, ValueTuple>();
         }
-        static void TupleRoundtripObject<T, TTuple>() {
+        static void TupleRoundtripObject<T, TTuple>()
+        {
             var tuple = Activator.CreateInstance(typeof(T), 42, "42", new object());
-            using (Py.GIL()) {
+            using (Py.GIL())
+            {
                 var pyTuple = TupleCodec<TTuple>.Instance.TryEncode(tuple);
                 Assert.IsTrue(TupleCodec<TTuple>.Instance.TryDecode(pyTuple, out object restored));
                 Assert.AreEqual(expected: tuple, actual: restored);
@@ -70,17 +82,38 @@ namespace Python.EmbeddingTest {
         }
 
         [Test]
-        public void TupleRoundtripGeneric() {
+        public void TupleRoundtripGeneric()
+        {
             TupleRoundtripGeneric<ValueTuple<int, string, object>, ValueTuple>();
         }
 
-        static void TupleRoundtripGeneric<T, TTuple>() {
+        static void TupleRoundtripGeneric<T, TTuple>()
+        {
             var tuple = Activator.CreateInstance(typeof(T), 42, "42", new object());
-            using (Py.GIL()) {
+            using (Py.GIL())
+            {
                 var pyTuple = TupleCodec<TTuple>.Instance.TryEncode(tuple);
                 Assert.IsTrue(TupleCodec<TTuple>.Instance.TryDecode(pyTuple, out T restored));
                 Assert.AreEqual(expected: tuple, actual: restored);
             }
+        }
+
+        [Test]
+        public void Function()
+        {
+            FunctionCodec.Register();
+            var locals = new PyDict();
+
+            PythonEngine.Exec(@"
+def foo():
+    return 1
+", null, locals.Handle);
+
+            var func = locals.GetItem("foo");
+
+            Action action;
+            Assert.IsTrue(FunctionCodec.Instance.TryDecode(func, out action));
+            Assert.DoesNotThrow(() => action());
         }
     }
 }
--- a/src/embed_tests/Codecs.cs
+++ b/src/embed_tests/Codecs.cs
@@ -101,97 +101,89 @@ namespace Python.EmbeddingTest {
         [Test]
         public void FunctionAction()
         {
-            FunctionCodec.Register();
             var codec = FunctionCodec.Instance;
 
-            //decoding - python functions to C# actions
-            {
-                PyInt x = new PyInt(1);
-                PyDict y = new PyDict();
-                //non-callables can't be decoded into Action
-                Assert.IsFalse(codec.CanDecode(x, typeof(Action)));
-                Assert.IsFalse(codec.CanDecode(y, typeof(Action)));
+            PyInt x = new PyInt(1);
+            PyDict y = new PyDict();
+            //non-callables can't be decoded into Action
+            Assert.IsFalse(codec.CanDecode(x, typeof(Action)));
+            Assert.IsFalse(codec.CanDecode(y, typeof(Action)));
 
-                var locals = new PyDict();
-                PythonEngine.Exec(@"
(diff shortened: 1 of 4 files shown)


src/embed_tests/Codecs.cs
```diff
@@ -6,28 +6,34 @@ namespace Python.EmbeddingTest {
     using Python.Runtime;
     using Python.Runtime.Codecs;
 
-    public class Codecs {
+    public class Codecs
```
koubaa: my IDE keeps adding this automatically.  Is there something we can do in visual studio to prevent that?

src/embed_tests/Codecs.cs
```diff
@@ -57,30 +66,188 @@ public class Codecs {
...
+def bar(a):
+    return 2
+", null, locals.Handle);
+
+                //foo, the function with no arguments
+                var fooFunc = locals.GetItem("foo");
+                Assert.IsFalse(codec.CanDecode(fooFunc, typeof(bool)));
+
+                //CanDecode does not work for variadic actions
+                //Assert.IsFalse(codec.CanDecode(fooFunc, typeof(Action<object[]>)));
```
koubaa: @lostmsu because of the interface of CanDecode this doesn't work.  Since I defined functions in C# to use object[] I cannot determine the number of arguments without an instance.

src/runtime/Codecs/FunctionCodec.cs
```diff
@@ -0,0 +1,377 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+
+namespace Python.Runtime.Codecs
+{
+    //like MethodWrapper but not static, so we can throw some state into it.
```
koubaa: Would like to see this combined with the MethodWrapper in methodwrapper.cs if possible.  In my case I want to wrap an member function rather than a static function.  @amos402 @filmor @lostmsu

src/runtime/Codecs/FunctionCodec.cs
```diff
@@ -0,0 +1,377 @@
...
+                callable = new FunctionWrapper(value as Func<object[], object>);
+            }
+            else
+            {
+                throw new Exception("object cannot be encoded!");
+            }
+
+            var methodWrapper = new MethodWrapper2(callable, "RunAction", "BinaryFunc");
+
+            //TODO - lifetime??
```
koubaa: @amos402 @lostmsu @filmor my main concern with this change.  I have no idea how to manage the lifetime of the method wrapper and ICallable.

src/embed_tests/Codecs.cs
lostmsu: You can create a special delegate type just for variadic parameter handling, and check for it explicitly in the codec.
E.g. `delegate void VariadicAction<T>(params T[] args)`

src/embed_tests/Codecs.cs
lostmsu: Keep the new formatting. This code was written before I made .editorconfig with the style of braces the rest of the project uses.

src/runtime/Codecs/FunctionCodec.cs
```diff
@@ -0,0 +1,377 @@
...
+            var type = instance.GetType();
+            _thunk = GetThunk(instance, type.GetMethod(name), funcType);
+
+            // Allocate and initialize a PyMethodDef structure to represent
+            // the managed method, then create a PyCFunction.
+
+            mdef = Runtime.PyMem_Malloc(4 * IntPtr.Size);
+            TypeManager.WriteMethodDef(mdef, name, _thunk.Address, 0x0003);
+            ptr = Runtime.PyCFunction_NewEx(mdef, IntPtr.Zero, IntPtr.Zero);
+        }
```
lostmsu: I though we already have a delegate -> Python callable conversion. What is different about this one?
Same goes for method groups: you can just do `dotNetInstance.ToPython().GetAttr('NameOfGroupOfMethods')`.

src/runtime/Codecs/FunctionCodec.cs
```diff
@@ -0,0 +1,377 @@
...
+    class FunctionCodec : IPyObjectDecoder, IPyObjectEncoder
+    {
+        private static int GetNumArgs(PyObject pyCallable)
+        {
+            var locals = new PyDict();
+            locals.SetItem("f", pyCallable);
+            using (Py.GIL())
+                PythonEngine.Exec(@"
+from inspect import signature
+x = len(signature(f).parameters)
```
lostmsu: AFAIK, `signature` might fail.

src/runtime/Codecs/FunctionCodec.cs
```diff
@@ -0,0 +1,377 @@
...
+
+        public static FunctionCodec Instance { get; } = new FunctionCodec();
+        public bool CanDecode(PyObject objectType, Type targetType)
+        {
+            //python object must be callable
+            if (!objectType.IsCallable()) return false;
+
+            //C# object must be an Action
+            if (!IsCallable(targetType))
+                return false;
```
lostmsu: or a function?
Is there a reason you can't support arbitrary delegate type?

src/runtime/Codecs/FunctionCodec.cs
```diff
@@ -0,0 +1,377 @@
...
+            {
+                Runtime.XIncref(pyObj.Handle);
+                PyObject pyAction = new PyObject(pyObj.Handle);
+                var pyArgs = new PyObject[0];
+                using (Py.GIL())
+                {
+                    var pyResult = pyAction.Invoke(pyArgs);
+                    Runtime.XIncref(pyResult.Handle);
+                    object result;
+                    Converter.ToManaged(pyResult.Handle, typeof(object), out result, true);
```
lostmsu: NIT: `out var result`

src/runtime/Codecs/FunctionCodec.cs
```diff
@@ -0,0 +1,377 @@
...
+        public PyObject TryEncode(object value)
+        {
+            if (value == null) return null;
+
+            var targetType = value.GetType();
+            ICallable callable = null;
+            if (IsUnaryAction(targetType))
+            {
+                callable = new ActionWrapper(value as Action);
+
```
lostmsu: empty line

src/runtime/Codecs/FunctionCodec.cs
lostmsu: For example: 42.ToPython().GetAttr('ToString')

lostmsu:
Actually, I'd like to see a specific use case for this change. Either I don't understand something, or the only thing Python.NET does not support currently is passing Python callables to `Action<*>` parameters. Reverse works though.

src/runtime/Codecs/FunctionCodec.cs
amos402: You can implement an `ExtensionType`.

src/embed_tests/Codecs.cs
koubaa: @lostmsu The issue is with the number of arguments, not the type of the arguments

src/runtime/Codecs/FunctionCodec.cs
koubaa: @amos402 thank you.  I don't know what that is but I will look into it

src/runtime/Codecs/FunctionCodec.cs
koubaa: @lostmsu I didn't know about it, I looked for PyCFunction_* and this is how I ended up finding the original MethodWrapper.  I'll see if I can find out how to do what you describe

src/runtime/Codecs/FunctionCodec.cs
koubaa: @lostmsu i tried to find something in the C API but failed.  I can try putting this in a try/except and return 0 in the except block

src/runtime/Codecs/FunctionCodec.cs
koubaa: @lostmsu the main reason is my ignorance of C# delegate types

koubaa:
> Actually, I'd like to see a specific use case for this change. Either I don't understand something, or the only thing Python.NET does not support currently is passing Python callables to `Action<*>` parameters. Reverse works though.
> 
> So you only actually need to implement `IPyObjectDecoder`.

@lostmsu I need to write an embedding test to show what I want to do.

src/runtime/Codecs/FunctionCodec.cs
koubaa: @lostmsu ToPython() uses Python.Runtime.Converter which uses codecs but falls back to GetInstHandle for TypeCode.Object (delegate types use that TypeCode). I guess if this already works I do not need to implement IPyObjectEncoder in my codec.  Let me try and write an embedding test to see what the problem is if there is one.

src/runtime/Codecs/FunctionCodec.cs
koubaa: @lostmsu thanks, I didn't know about that

src/embed_tests/Codecs.cs
koubaa: @lostmsu I think I get your point.  Do you mean rather than Action<object[]> and Func<object[], object> I can simply use VariadicAction so if I have a situation like this:

C#
public int CallMe(Func<int, int, string, double, object> func) {
    return CallMe(1,2, "hello", 4.4, new SomeObject());
}

Python
def call_me(i, j, name, real_value, obj):
    return 1

You except that given typeof(func) you can convert it to VariadicAction and call Invoke on it dynamically?  I can try and I think this may work except it may have a restriction to use object in all of the argument slots.

src/runtime/Codecs/FunctionCodec.cs
```diff
@@ -0,0 +1,187 @@
...
+            else
+            {
+                return false;
+            }
+        }
+
+        public static void Register()
+        {
+            PyObjectConversions.RegisterDecoder(Instance);
+            PyObjectConversions.RegisterDecoder(Instance);
```
lostmsu: duplicate registration

koubaa:
@lostmsu  I found that getting the number of arguments in python using inspect signature is not reliable because in the CanDecode method the codec only has the PyObject_TYPE of the PyObject rather than the object itself.  Are you open to changing the contract of the codecs to pass in the object itself and let the codec get its type?

image

I used delegate reflection to get the number of arguments from the C# delegate.  But I don't see how that would help with binding to any delegate types.  I need to construct a delegate of the type given by the parameter argument and then return it from the decoder.  I guess I can use four delegate types:

delegate void UnaryActionDelegate();
delegate void VariadicActionDelegate(object[] args);
delegate object UnaryFuncDelegate();
delegate object VariadicFuncDelegate(object[] args);

which I can easily construct.  But you cannot even construct a System.Action with a UnaryActionDelegate instance.  So I am failing to see how the delegate reflection will help with supporting Action and Func.

That being said I think it makes sense to implement these delegates anyways so I can support delegate binding.

koubaa:
@lostmsu I am working  on moving this codec to the codecs repo.  While doing so I am encountering some build issues related to access levels in the core library:

- Runtime.XIncref is marked internal
- I can't use Converter.ToManaged or ToPython because the Converter class is internal

Should we set InternalsVisibleTo in Python.Runtime to the codecs assembly?

lostmsu:
@koubaa do we have to? Can you achieve the same using public `PyObject` class?

koubaa:
@lostmsu I don't see a way to increment a reference count using PyObject or to use Converter.ToManaged.

lostmsu:
@koubaa instead of incrementing refcount you can just keep the PyObject. `PyObject.As<T>` is a direct counterpart of `Converter.ToManaged`.

koubaa:
@lostmsu I'll try it.

koubaa:
@lostmsu that was helpful.  I can use As<> and even object.ToPython (a public extension method defined in python runtime).  Testing will show if I get the reference count working, will update the codecs repo soon

Issue #962, opened 2019-09-26
Type parameter works directly, but fails inside `List<>`


andy-ms:
### Environment

-   Pythonnet version: 2.4.0
-   Python version: Python 3.7.4
-   Operating System: Windows 10 (18362.356)

### Details

-   Describe what you were trying to get done.

Call a generic method taking a list parameter.

-   What commands did you run to trigger this issue? If you can provide a
    [Minimal, Complete, and Verifiable example](http://stackoverflow.com/help/mcve)
    this will help us understand the issue.

C# library:

```cs
using System.Collections.Generic;

namespace CSLib
{
    public static class CSLib
    {
        public static bool GetBool() =>
            true;

        public static void TestValue<T>(T t) {}

        public static List<bool> GetBoolList() =>
            new List<bool>() { true, false };

        public static void TestList<T>(List<T> list) {}

        public static void Test()
        {
            TestValue<bool>(GetBool());
            TestList<bool>(GetBoolList());
        }
    }
}
```

Python code:

```python
from clr import AddReference
from pathlib import Path

AddReference(str(Path.cwd() / "cslib/bin/Debug/netstandard2.0/cslib.dll"))

from System import Boolean
from CSLib import CSLib

CSLib.TestValue[Boolean](CSLib.GetBool())
CSLib.TestList[Boolean](CSLib.GetBoolList())
```

It works fine when the type argument is used directly, but fails when it is inside `List<>`.

-   If there was a crash, please include the traceback here.

```python
Traceback (most recent call last):
  File "a.py", line 11, in <module>
    CSLib.TestList[Boolean](CSLib.GetBoolList())
TypeError: No method matches given arguments for TestList
```

filmor:
This is due to the implicit conversion of .NET lists to Python lists again. Since we don't do it the other way round (and it would also be very tricky to get that right), when you get `CSLib.GetBoolList()` you end up with a Python list `[True]` which doesn't match the method signature.

I'm working on removing the implicit conversion in #963.

lostmsu:
This should now be fixed in 3.0

Issue #912, opened 2019-07-10
Automatic conversions


filmor:
This is a meta-issue to collect all of the existing automatic conversions, possible future conversions and discussion on how we want to go forward.

As a start, we definitely have (in some places, not all!) the following identification:

- `null` <=> `None`

Cronan:
Does anyone her have perms to mark issues with labels - in the minutes of 25-June project meeting it was suggested that proposed work to be done be marked with the tag epic - this isn't being done, probbaly because nobody has the permissions to do it, I guess ...

filmor:
I have, I just forgot :)

lostmsu:
This issue is not being used.

Issue #909, opened 2019-07-10
Inconsistent Conversion of Sequence Types


rmadsen-ks:
### Environment
-   Pythonnet version: 2.4 (build from df0574db552020397728648ee3ec576aa3a1c2a8)
-   Python version: 2.7, 3.6, 3.7
-   Operating System: Windows 10

### Details

I am trying to convert from List<int> to PyList and back without losing type information.

```C#
            // this works:
            int[] values2 = new int[] { 1, 2, 3 };
            var pyvalues2 = values2.ToPython();
            pyvalues2.As<int[]>();

            // this does not work.
            List<int> values = new List<int> { 1, 2, 3 };
            var pyvalues = values.ToPython();
            pyvalues.As<List<int>>(); // unable to convert back to List<int>
```

It seems ToPython will just take any IEnumerable or IList and turn it into a python list. Even though it is not generally possible to convert it back. In this case I'd rather keep it as the original type, because eventually I'll need to get it back to C#.

I understand why the current behavior makes things easier, but it would be great if it was possible to disable or alter this behavior.

filmor:
I agree, the whole conversion story of Python.NET is currently quite inconsistent. Before we apply more band-aids, I'd like to have an overall plan, though, s.t. if we have to break the API we make it count :)

Essentially, we need a set of wrapping/unwrapping functions and additionally a (separate) set of functions for conversion (for convenience).

rmadsen-ks:
I think this could be done without breaking the API. 

I was thinking something in the style of:
```cs
 values.ToPython(ImplicitConvert: false); 
```

lostmsu:
The particular issue with `List<T>` should have been fixed in 3.0

Issue #601, opened 2018-01-19
Converter works for IList<T> but not for T[]


jjaskulowski:
### Environment

-   Pythonnet version: 
-   Python version: all
-   Operating System: all

### Details

I lost hours on this.
converter.ToPython() does not enter condition "if (value is IList && value.GetType().IsGenericType)" for arrays so arrays that could be converted to python lists are not being converted.

I think it should also work for any generic and not generic ICollection and any Array.

Instead, it currently treats it as a generic PyObject.

Otherwise, great job!

denfromufa:
@jjaskulowski .NET arrays are not the same as Python lists, even .NET lists are not the same as Python lists.
So technically even what is currently done is not ideal. So what should you do for now? I suggest converting arrays between Python and .NET using some of techniques you can find in this issue tracker.

PR #1084, opened 2020-03-10
implement list codec

koubaa:
Implement conversions from python lists, iterables, and sequences to appropriate C# interfaces
...

Probably a few of them. I'll look for it later.

### Any other comments?
I've done the codec for three levels of collections.  roughly I map these:
| python | C# |
|---|---|
| iterable | IEnumerable |
| sequence | ICollection |
| list | IList |

The mapping isn't exactly perfect, naturally.  But this is the best I think I can do right now.

I need to add some more codec unit tests and functional embedding tests.

### Checklist


koubaa:
@amos402 @filmor @lostmsu please review.  I think this PR may help with our implicit conversions of lists.

lostmsu:
Haven't done a full pass yet, but instead of doing "levels" you can just have 3 decoders, and register them in order from highest priority to lowest priority. E.g.
```csharp
Register(ListEncoder.Instance);
Register(CollectionDecoder.Instance);
Register(IterableDecoder.Instance);
```

koubaa:
@lostmsu I can change the physical layout but wanted to make sure I was on the right track first.  I like the idea of having multiple codecs with priority, that way clients can choose which encoder they need.  I'll work on that.

src/embed_tests/Codecs.cs
```diff
@@ -82,5 +82,130 @@ public class Codecs {
...
+            var x = new PyList(items.ToArray());
+            Assert.IsTrue(codec.CanDecode(x, typeof(IList<bool>)));
+            Assert.IsTrue(codec.CanDecode(x, typeof(System.Collections.IEnumerable)));
+            Assert.IsTrue(codec.CanDecode(x, typeof(IEnumerable<int>)));
+            Assert.IsTrue(codec.CanDecode(x, typeof(ICollection<float>)));
+            Assert.IsFalse(codec.CanDecode(x, typeof(bool)));
+
+            //we'd have to copy into a list to do this.  not the best idea to support it.
+            //maybe there can be a flag on listcodec to allow it.
+            Assert.IsFalse(codec.CanDecode(x, typeof(List<int>)));
```
koubaa: @lostmsu @filmor With your idea of multiple codecs I can have a DeepCopyListCodec that takes any ienumerable and converts it by copying to a list.  This is what we do right now in master and if we get rid of that implicit conversion the user can just register this codec if that causes an undesirable behavior change for them.

lostmsu: I am thinking, that `Python.Runtime.dll` should only contain codecs, that perform lossless conversions, perhaps even by default. E.g. if a type is exactly `builtins.list` it is safe to convert it to `IList<T>` via some safe wrapper type `PythonList<T>`, that would contain a reference to the original object. Preferably, when this wrapper is passed back to Python it should either just pass the wrapped object, or behave exactly like a normal `list` would.

The codecs, that are not lossless, like converting a `builtins.list` to `T[]`, and codecs, that act on relatively uncommon types like `System.Uri` should go to a separate extra package (created yesterday after seeing your PR): https://github.com/pythonnet/codecs

koubaa: @lostmsu done.  I didn't go as far as to enable by default but that can be done as a followup commit.  Please look closely at the test cases as I tried to capture the behavior of each of the three codecs to make sure it meets your expectation.

koubaa:
This is related to #912

src/runtime/CollectionWrappers/IterableWrapper.cs
```diff
@@ -0,0 +1,52 @@
...
+{
+    internal class IterableWrapper<T> : IEnumerable<T>
+    {
+        protected PyObject iterObject;
+        protected PyObject pyObject;
+
+        public IterableWrapper(PyObject pyObj)
+        {
+            pyObject = pyObj;
+            iterObject = new PyObject(Runtime.PyObject_GetIter(pyObj.Handle));
```
lostmsu: `iterObject` should be local to `GetEnumerator` implementations, otherwise you won't be able to enumerate this instance twice.

src/runtime/CollectionWrappers/ListWrapper.cs
```diff
@@ -0,0 +1,66 @@
...
+        public ListWrapper(PyObject pyObj) : base(pyObj)
+        {
+
+        }
+
+        public T this[int index]
+        {
+            get
+            {
+                IntPtr item = Runtime.PySequence_GetItem(pyObject.Handle, index);
```
lostmsu: I wonder if there would be any difference between using `PySequence_GetItem` vs `PyList_GetItem` vs `PyObject_GetItem` here.

src/runtime/CollectionWrappers/IterableWrapper.cs
```diff
@@ -0,0 +1,52 @@
...
+        }
+
+        IEnumerator IEnumerable.GetEnumerator()
+        {
+            IntPtr item;
+            while ((item = Runtime.PyIter_Next(iterObject.Handle)) != IntPtr.Zero)
```
lostmsu: If `PyIter_Next` fails, this should check the reason. If it is not `StopIteration`, an exception should be passed to .NET

src/runtime/CollectionWrappers/ListWrapper.cs
```diff
@@ -0,0 +1,66 @@
...
+            set
+            {
+                IntPtr pyItem = Converter.ToPython(value, typeof(T));
+                if (pyItem == IntPtr.Zero)
+                    throw new Exception("failed to set item");
+
+                var result = Runtime.PySequence_SetItem(pyObject.Handle, index, pyItem);
+                Runtime.XDecref(pyItem);
+                if (result == -1)
+                    throw new Exception("failed to set item");
```
lostmsu: You should raise the actual Python exception.

src/runtime/CollectionWrappers/ListWrapper.cs
```diff
@@ -0,0 +1,66 @@
...
+        }
+
+        public void Insert(int index, T item)
+        {
+            if (IsReadOnly)
+                throw new NotImplementedException();
+
+            IntPtr pyItem = Converter.ToPython(item, typeof(T));
+            if (pyItem == IntPtr.Zero)
+                throw new Exception("failed to insert item");
```
lostmsu: Same, raise the actual Python exception

src/runtime/CollectionWrappers/ListWrapper.cs
```diff
@@ -0,0 +1,66 @@
...
+
+        public int IndexOf(T item)
+        {
+            return indexOf(item);
+        }
+
+        public void Insert(int index, T item)
+        {
+            if (IsReadOnly)
+                throw new NotImplementedException();
```
lostmsu: I think this should be `InvalidOperationException("Collection is read-only")`

src/runtime/CollectionWrappers/ListWrapper.cs
```diff
@@ -0,0 +1,66 @@
...
+
+            IntPtr pyItem = Converter.ToPython(item, typeof(T));
+            if (pyItem == IntPtr.Zero)
+                throw new Exception("failed to insert item");
+
+            var result = Runtime.PyList_Insert(pyObject.Reference, index, pyItem);
+            Runtime.XDecref(pyItem);
+            if (result == -1)
+                throw new Exception("failed to insert item");
```
lostmsu: Python exception

src/runtime/CollectionWrappers/SequenceWrapper.cs
```diff
@@ -0,0 +1,98 @@
...
+            }
+
+            return (T)obj;
+        }
+
+        public void CopyTo(T[] array, int arrayIndex)
+        {
+            for (int index = 0; index < Count; index++)
+            {
+                array[index + arrayIndex] = getItem(index);
```
lostmsu: Not sure, but it might be more effective to use iterator.

src/runtime/CollectionWrappers/IterableWrapper.cs
```diff
@@ -0,0 +1,52 @@
...
+namespace Python.Runtime.CollectionWrappers
+{
+    internal class IterableWrapper<T> : IEnumerable<T>
+    {
+        protected PyObject iterObject;
+        protected PyObject pyObject;
+
+        public IterableWrapper(PyObject pyObj)
+        {
+            pyObject = pyObj;
```
lostmsu: I recommend check the argument for `null`

src/runtime/CollectionWrappers/SequenceWrapper.cs
```diff
@@ -0,0 +1,98 @@
...
+        public SequenceWrapper(PyObject pyObj) : base(pyObj)
+        {
+
+        }
+
+        public int Count
+        {
+            get
+            {
+                return (int)Runtime.PySequence_Size(pyObject.Handle);
```
lostmsu: `PySequence_Size` might fail, according to documentation.

src/runtime/CollectionWrappers/SequenceWrapper.cs
```diff
@@ -0,0 +1,98 @@
...
+            throw new NotImplementedException();
+        }
+
+        public void Clear()
+        {
+            if (IsReadOnly)
+                throw new NotImplementedException();
+            var result = Runtime.PySequence_DelSlice(pyObject.Handle, 0, Count);
+            if (result == -1)
+                throw new Exception("failed to clear sequence");
```
lostmsu: Python exception

src/runtime/CollectionWrappers/ListWrapper.cs
koubaa: @lostmsu good point.  I should use PyList_GetItem in ListWrapper and PySequence_GetItem in SequenceWrapper.

src/runtime/CollectionWrappers/ListWrapper.cs
koubaa: @lostmsu Converter.ToPython registers a python exception if it fails?

src/runtime/CollectionWrappers/SequenceWrapper.cs
koubaa: @lostmsu true, but it doesn't say it throws, so I will throw a c# exception in this case

src/runtime/CollectionWrappers/SequenceWrapper.cs
koubaa: @lostmsu not disagreeing but please define effective.  Code clarity or runtime performance?

koubaa:
@lostmsu While the unit tests work when I tried it from python I realized (again) that CanDecode is based on the type rather than the instance, so I changed that.  Also I added some python tests

src/runtime/CollectionWrappers/SequenceWrapper.cs
lostmsu: @koubaa , I meant runtime performance. I am not sure if Python sequences must provide O(1) indexing.

src/runtime/CollectionWrappers/IterableWrapper.cs
```diff
@@ -0,0 +1,79 @@
...
+        }
+
+        private void propagateIterationException()
+        {
+            var err = Runtime.PyErr_Occurred();
+            if (err != null && err != Exceptions.StopIteration)
```
lostmsu, line 19: Documentation recommends using PyErr_ExceptionMatches

src/runtime/CollectionWrappers/IterableWrapper.cs
```diff
@@ -0,0 +1,79 @@
...
+    internal class IterableWrapper<T> : IEnumerable<T>
+    {
+        protected PyObject pyObject;
+
+        public IterableWrapper(PyObject pyObj)
+        {
+            pyObject = pyObj;
+        }
+
+        private void propagateIterationException()
```
lostmsu, line 16: I think this will keep `StopIteration` "raised". E.g. next check for unhandled Python exception will return it.

src/runtime/CollectionWrappers/IterableWrapper.cs
lostmsu: Why is this resolved? Is there a legitimate scenario for `IterableWrapper` wrapping `None`/`null`?

src/runtime/CollectionWrappers/IterableWrapper.cs
```diff
@@ -0,0 +1,79 @@
+using System;
+using System.Collections.Generic;
+using System.Collections;
+
+namespace Python.Runtime.CollectionWrappers
+{
+    internal class IterableWrapper<T> : IEnumerable<T>
+    {
+        protected PyObject pyObject;
```
lostmsu, line 9: This can be made `readonly`. Then you can drop null checks in `GetEnumerator`

src/runtime/CollectionWrappers/IterableWrapper.cs
```diff
@@ -0,0 +1,79 @@
...
+                yield return obj;
+            }
+
+            propagateIterationException();
+        }
+
+        public IEnumerator<T> GetEnumerator()
+        {
+            if (pyObject == null) yield break;
+            PyObject iterObject = null;
```
lostmsu, line 50: Ideally, you should dispose this object.

src/runtime/CollectionWrappers/ListWrapper.cs
```diff
@@ -0,0 +1,63 @@
...
+        {
+
+        }
+
+        public T this[int index]
+        {
+            get
+            {
+                var item = Runtime.PyList_GetItem(pyObject.Handle, index);
+                var pyItem = new PyObject(item);
```
lostmsu, line 18: Dispose this

src/runtime/CollectionWrappers/ListWrapper.cs
```diff
@@ -0,0 +1,63 @@
...
+                if (!Converter.ToManaged(pyItem.Handle, typeof(T), out object obj, true))
+                    Runtime.CheckExceptionOccurred();
+
+                return (T)obj;
+            }
+            set
+            {
+                IntPtr pyItem = Converter.ToPython(value, typeof(T));
+                if (pyItem == IntPtr.Zero)
+                    throw new Exception("failed to set item");
```
lostmsu, line 29: Perhaps `InvalidCastException`. Also, might need to supply the cause (e.g. underlying `PythonException`).

src/runtime/CollectionWrappers/ListWrapper.cs
```diff
@@ -0,0 +1,63 @@
...
+
+                var result = Runtime.PyList_SetItem(pyObject.Handle, index, pyItem);
+                Runtime.XDecref(pyItem);
```
lostmsu, line 32: According to docs https://docs.python.org/3.5/c-api/list.html#c.PyList_SetItem decrefs internally.

src/runtime/CollectionWrappers/ListWrapper.cs
lostmsu: @koubaa you'd have to take a look at it.

src/runtime/CollectionWrappers/SequenceWrapper.cs
```diff
@@ -0,0 +1,108 @@
...
+
+        public int Count
+        {
+            get
+            {
+                var size = Runtime.PySequence_Size(pyObject.Handle);
+                if (size == -1)
+                {
+                    Runtime.CheckExceptionOccurred();
+                    throw new Exception("Unable to get sequence size!");
```
lostmsu, line 21: `CheckExceptionOccured` should throw in the line above

src/runtime/CollectionWrappers/SequenceWrapper.cs
```diff
@@ -0,0 +1,108 @@
...
+        }
+
+        protected bool removeAt(int index)
+        {
+            if (IsReadOnly)
+                throw new NotImplementedException();
+            if (index >= Count || index < 0)
+                throw new IndexOutOfRangeException();
+
+            return Runtime.PySequence_DelItem(pyObject.Handle, index) != 0;
```
lostmsu, line 88: Failure of `PySequence_DelItem` is an exception too.

src/runtime/converterextensions.cs
```diff
@@ -156,7 +156,7 @@ bool TryDecode(IntPtr pyHandle, out object result)
 
         #endregion
 
-        internal static void Reset()
+        public static void Reset()
```
lostmsu, line 159: Is there a scenario when this should be done outside of runtime shutdown?

src/runtime/CollectionWrappers/IterableWrapper.cs
koubaa: @lostmsu This is how I interpreted your comment.  did you mean to say I should throw on a null input?

src/runtime/CollectionWrappers/IterableWrapper.cs
koubaa: @lostmsu shouldn't its finalizer do that?

src/runtime/CollectionWrappers/ListWrapper.cs
koubaa: @lostmsu again I thought the point of making it IDisposable was that the finalizer would take care of that

src/runtime/CollectionWrappers/SequenceWrapper.cs
koubaa: @lostmsu according to python docs for PySequenceDelSlice:
Delete the slice in sequence object o from i1 to i2. Returns -1 on failure.

It doesn't explicitly say it will set an error on failure.  Is this implicit anywhere I see the word failure in the python doc?

src/runtime/CollectionWrappers/SequenceWrapper.cs
koubaa: @lostmsu in that case I will clear the exception in Remove (In C# it is idiomatic to return false in Remove rather than to throw) but throw it in ListWrapper RemoveAt

src/runtime/converterextensions.cs
koubaa: @lostmsu Doubt it.  I just needed it to clear the state between unit test runs.

src/runtime/CollectionWrappers/ListWrapper.cs
amos402: Wrap it by using statement is better. If you knew it's no use anymore you should dispose it manually instead of waiting for the auto collection to do it.

koubaa: @lostmsu @amos402 sorry, I might be missing context (I rebased and I don't know if this comment lines up with the code it was pointing to initially).

What specifically should I be disposing or wrap in a using statement?

amos402:
```c#
using (var pyItem = new PyObject(item)) { }
```

amos402:
If you're considering the performance improvement, this branch may help you.
https://github.com/joonhwan/pythonnet/tree/fast-call
It's an ancient branch I deleted a long time ago, my current internal branch has far more than that.
You can do some performance tests, like method calls, properties get-set, etc, you will see that the branch above has a lot faster than the master now. The simple points are avoiding any `MethodInfo.Invoke`, use pre-created delegates instead, prevent any boxing or unboxing if you can. After that, if you want a higher than it, code generation is an option.

The `codec` has some design points can improve:
* `Decode`, `Encode` used `object`, which means it may occur boxing.
* When it `decode/encode` an object, it has iteration operations which can be avoided, or use some hash table for getting the coder directly.
* For achieving no reflection calling, no boxing or unboxing, you may need to redesign the method calling mechanism just like I did, you can check the `methodobject.cs` on the above branch.

koubaa (commented):
@lostmsu could you please take another look at this?

koubaa:
> If you're considering the performance improvement, this branch may help you.
> https://github.com/joonhwan/pythonnet/tree/fast-call
> It's an ancient branch I deleted a long time ago, my current internal branch has far more than that.
> You can do some performance tests, like method calls, properties get-set, etc, you will see that the branch above has a lot faster than the master now. The simple points are avoiding any `MethodInfo.Invoke`, use pre-created delegates instead, prevent any boxing or unboxing if you can. After that, if you want a higher than it, code generation is an option.
> 
> The `codec` has some design points can improve:
> 
> * `Decode`, `Encode` used `object`, which means it may occur boxing.
> * When it `decode/encode` an object, it has iteration operations which can be avoided, or use some hash table for getting the coder directly.
> * For achieving no reflection calling, no boxing or unboxing, you may need to redesign the method calling mechanism just like I did, you can check the `methodobject.cs` on the above branch.

Thank you for sharing!  I haven't measured performance issues but now I know what to try if I do.

src/runtime/CollectionWrappers/IterableWrapper.cs
lostmsu: @koubaa this is right. Though you should throw `ArgumentNullException`

src/runtime/CollectionWrappers/IterableWrapper.cs
```diff
@@ -0,0 +1,95 @@
...
+        {
+            if (pyObj == null)
+                throw new PythonException();
+            pyObject = pyObj;
+        }
+
+        private void propagateIterationException()
+        {
+            var err = Runtime.PyErr_Occurred();
+            if (err == null) return;
```
lostmsu, line 21: `PyErr_Occurred` returns `IntPtr`. You can't compare it to `null`, you have to compare it to `IntPtr.Zero`.

src/runtime/CollectionWrappers/IterableWrapper.cs
```diff
@@ -0,0 +1,95 @@
...
+                return;
+            }
+
+            Runtime.CheckExceptionOccurred();
+        }
+
+        IEnumerator IEnumerable.GetEnumerator()
+        {
+            PyObject iterObject = null;
+            using (Py.GIL())
```
lostmsu, line 36: Actually, it is a job of the caller of this method to hold GIL for any call on `GetEnumerator` or `MoveNext`

src/runtime/CollectionWrappers/IterableWrapper.cs
```diff
@@ -0,0 +1,95 @@
...
+            {
+                iterObject = new PyObject(Runtime.PyObject_GetIter(pyObject.Handle));
+            }
+
+            while (true)
+            {
+                IntPtr item = IntPtr.Zero;
+                using (Py.GIL())
+                {
```
lostmsu, line 45: Same: GIL has to be held by the caller.

src/runtime/CollectionWrappers/IterableWrapper.cs
```diff
@@ -0,0 +1,95 @@
...
+        {
+            var err = Runtime.PyErr_Occurred();
+            if (err == null) return;
+
+            //remove StopIteration exceptions
+            if (0 != Runtime.PyErr_ExceptionMatches(Exceptions.StopIteration))
+            {
+                Runtime.PyErr_Clear();
+                return;
+            }
```
lostmsu, line 28: According to the official documentation, `PyIter_Next` should not be setting `StopIteration` exception:
https://docs.python.org/3.5/c-api/iter.html#c.PyIter_Next
> If there are no remaining values, returns NULL with no exception set

src/runtime/CollectionWrappers/IterableWrapper.cs
```diff
@@ -0,0 +1,95 @@
...
+
+            Runtime.CheckExceptionOccurred();
+        }
+
+        IEnumerator IEnumerable.GetEnumerator()
```
lostmsu, line 33: Because `IEnumerator<T>` also implements `IEnumerator`, you don't actually need this method to have full body. You can just replace it with `IEnumerator IEnumerable.GetEnumerator() => this.GetEnumerator();`

src/runtime/CollectionWrappers/ListWrapper.cs
```diff
@@ -0,0 +1,71 @@
...
+        {
+            get
+            {
+                var item = Runtime.PyList_GetItem(pyObject.Reference, index);
+                var pyItem = new PyObject(item);
+
+                if (!Converter.ToManaged(pyItem.Handle, typeof(T), out object obj, true))
+                    Runtime.CheckExceptionOccurred();
+
+                return (T)obj;
```
lostmsu, line 23: This can be replaced with `return pyItem.As<T>();`

src/runtime/CollectionWrappers/ListWrapper.cs
```diff
@@ -0,0 +1,71 @@
...
+            }
+            set
+            {
+                IntPtr pyItem = Converter.ToPython(value, typeof(T));
+                if (pyItem == IntPtr.Zero)
+                {
+                    throw new InvalidCastException(
+                        "cannot cast " + value.ToString() + "to type: " + typeof(T).ToString(),
+                        new PythonException());
+                }
```
lostmsu, line 33: Why not just call `value.ToPython()` here?

src/runtime/CollectionWrappers/ListWrapper.cs
```diff
@@ -0,0 +1,71 @@
...
+        {
+            if (IsReadOnly)
+                throw new InvalidOperationException("Collection is read-only");
+
+            IntPtr pyItem = Converter.ToPython(item, typeof(T));
+            if (pyItem == IntPtr.Zero)
+                throw new PythonException();
+
+            var result = Runtime.PyList_Insert(pyObject.Reference, index, pyItem);
+            Runtime.XDecref(pyItem);
```
lostmsu, line 56: Doing a decref with error set might have unintended consequences.

lostmsu (changes requested):

src/runtime/CollectionWrappers/IterableWrapper.cs
koubaa: @lostmsu thanks.  I was trying to remember what the shorthand for this was when I was implementing it but couldn't and just duplicated it.  I knew it was something this simple..

src/runtime/CollectionWrappers/ListWrapper.cs
koubaa: @lostmsu with value.ToPython() I don't have to manually mess with the ref count

src/runtime/CollectionWrappers/IterableWrapper.cs
koubaa: @lostmsu I defer to you but I am not convinced.  If I am passing an instance of a python list using this wrapper (let's say as an IEnumerable<T>) to some 3rd party library which is going to iterate over it at some point in the future, I would have to worry about when that library actually does the enumeration and enclose that with holding the GIL.

Personally I think it makes more sense to consider that the contract of IEnumerable<T> is "An object which can be iterated".  I think adding holding the GIL to the expectation of whoever is given an IEnumerable<T> that happens to be an IterableWrapper<T> isn't idiomatic C#.

lostmsu: > it at some point in the future, I would have to worry about when that library actually does the enumeration and enclose that with holding the GIL.

I think you're right here.

src/runtime/CollectionWrappers/IterableWrapper.cs
lostmsu: When we know the lifetime, we should call `Dispose` at the end of it to reduce the load on GC

src/runtime/converterextensions.cs
lostmsu: Yes, I think you should be shutting down the runtime.

lostmsu (changes requested):

koubaa:
@lostmsu would you mind reviewing this again?  I resolved all of your previous comments

lostmsu:
@koubaa sure, but can't do it until Saturday.

src/runtime/CollectionWrappers/IterableWrapper.cs
```diff
@@ -0,0 +1,42 @@
...
+                iterObject = new PyObject(Runtime.PyObject_GetIter(pyObject.Handle));
+
+            while (true)
+            {
+                using (Py.GIL())
+                {
+                    var item = Runtime.PyIter_Next(iterObject.Handle);
+                    if (item == IntPtr.Zero)
+                    {
+                        iterObject.Dispose();
```
lostmsu, line 33: You must check if there was an exception here.

https://docs.python.org/3.8/c-api/iter.html#c.PyIter_Next

src/runtime/CollectionWrappers/SequenceWrapper.cs
```diff
@@ -0,0 +1,101 @@
...
+            //not sure if IEquatable is implemented and this will work!
+            foreach (var element in this)
+                if (element.Equals(item)) return true;
+
+            return false;
+        }
+
+        public void CopyTo(T[] array, int arrayIndex)
+        {
+            var index = 0;
```
lostmsu, line 59: Please, check inputs, and, if possible that the collection will fit before making an attempt to copy.

Generally for any operation it is highly preferred that it either completes successfully, or does not make any changes to its inputs in case of exception.

src/runtime/CollectionWrappers/SequenceWrapper.cs
lostmsu: That depends on the reason for the failure. It should not throw only if the item was not in the collection in the first place. If it was in the collection, but something prevented it from being removed, it must throw.

src/runtime/CollectionWrappers/IterableWrapper.cs
```diff
@@ -0,0 +1,42 @@
+using System;
+using System.Collections.Generic;
+using System.Collections;
+
+namespace Python.Runtime.CollectionWrappers
+{
+    internal class IterableWrapper<T> : IEnumerable<T>
```
lostmsu, line 7: To be a lossless conversion, when passed back to Python an instance of this class should pass the `pyObject` field. I think this behavior must be added along with decoders, that create it.

lostmsu (changes requested):

src/runtime/CollectionWrappers/IterableWrapper.cs
koubaa: @lostmsu Right now I only defined decoders.  I would think this is the responsibility of the code handling object conversion back to python (i.e. encoders) - not the collection wrappers themselves.

In this branch these decoders are used by default, but in a future PR which replaces implicit copy list conversion with them, we can make sure to handle this.

koubaa:
@lostmsu I'll look to enable these by default (and handle issues associated w/that) in a later PR

koubaa:
@lostmsu I rebased to fix the merge conflict.  Is there anything else needed before this can be merged?

src/embed_tests/Codecs.cs
```diff
@@ -57,31 +66,262 @@ public class Codecs {
...
+            var fooType = foo.GetPythonType();
+            Assert.IsFalse(codec.CanDecode(fooType, typeof(ICollection<int>)));
+
+            //python tuples do satisfy the python sequence protocol
+            var pyTuple = new PyObject(Runtime.PyTuple_New(3));
+            var pyTupleType = pyTuple.GetPythonType();
+
+            Runtime.PyTuple_SetItem(pyTuple.Handle, 0, items[0].Handle);
+            Runtime.PyTuple_SetItem(pyTuple.Handle, 1, items[1].Handle);
+            Runtime.PyTuple_SetItem(pyTuple.Handle, 2, items[2].Handle);
```
lostmsu, line 226: This will cause double-free with possible subsequent access violation in tests, as `PyTuple_SetItem` steals the item reference.

I suggest to use `pyTuple[i] = item` instead.

lostmsu (changes requested):

christabella:
@koubaa do you plan to finish this PR? The open issues that this would close include
#909 #601 #962

koubaa:
@christabella I have been waiting for @lostmsu to merge it

koubaa:
@lostmsu I rebased the changes to resolve the conflicts.  Is there anything needed before this can be merged?

lostmsu:
@koubaa @christabella once the CI tests pass, I can take this. Note, however, that I have a different approach in mind, based on [mixins and base class faking](https://github.com/losttech/pythonnet/tree/losttech/master/src/runtime/Mixins) (a feature I have in my fork, that is yet to be upstreamed, its implementation is relatively involved). So this will probably be replaced before 3.0 is released.

koubaa:
@lostmsu the CI is fixed

src/embed_tests/Codecs.cs
```diff
@@ -57,31 +66,258 @@ public class Codecs {
...
+        static PyObject GetPythonIterable()
+        {
+            var locals = new PyDict();
+            using (Py.GIL())
+            {
+                PythonEngine.Exec(GetIntIterableCommands("foo_instance"), null, locals.Handle);
+            }
+
+            return locals.GetItem("foo_instance");
+        }
```
lostmsu, line 129: I think you can simply do `Eval("map(lambda x: x, [1,2,3])")`.

src/runtime/Codecs/IterableDecoder.cs
```diff
@@ -0,0 +1,55 @@
...
+
+            if (!targetType.IsGenericType)
+                return false;
+
+            return targetType.GetGenericTypeDefinition() == typeof(IEnumerable<>);
+        }
+
+        internal static bool IsIterable(PyObject objectType)
+        {
+            return objectType.HasAttr("__iter__");
```
lostmsu, line 22: Can this be replaced with a call to `PyIter_Check`?

src/runtime/Codecs/ListDecoder.cs
```diff
@@ -0,0 +1,49 @@
...
+            return targetType.GetGenericTypeDefinition() == typeof(IList<>);
+        }
+
+        private static bool IsList(PyObject objectType)
+        {
+            //must implement sequence protocol to fully implement list protocol
+            if (!SequenceDecoder.IsSequence(objectType)) return false;
+
+            //returns wheter the type is a list.
+            return objectType.Handle == Runtime.PyListType;
```
lostmsu, line 22: Seems like this check is the only one needed.

src/runtime/CollectionWrappers/IterableWrapper.cs
```diff
@@ -0,0 +1,43 @@
...
+    {
+        protected readonly PyObject pyObject;
+
+        public IterableWrapper(PyObject pyObj)
+        {
+            if (pyObj == null)
+                throw new ArgumentNullException();
+            pyObject = pyObj;
```
lostmsu, line 15: I believe you should make a new `PyObject` here, pointing to the same object in Python (e.g. `pyObject = new PyObject(pyObj.Reference)`). The reason is that if the original `PyObject` is disposed (as suggested below in `ListWrapper.this[int index]`), this iterator will become useless.

lostmsu (changes requested):

src/embed_tests/Codecs.cs
koubaa: @lostmsu wow that is clever, nice!

src/runtime/Codecs/IterableDecoder.cs
koubaa: Yes it can

src/runtime/Codecs/ListDecoder.cs
koubaa: @lostmsu True, but eventually this should support any object that implements the list protocol (e.g. numpy arrays).

src/runtime/CollectionWrappers/IterableWrapper.cs
koubaa: @lostmsu thanks, I often forget that C# objects can be explicitly disposed before they get axed by the GC.

src/runtime/CollectionWrappers/ListWrapper.cs
lostmsu: @koubaa this still needs a fix. With C# 8+ which we use now, it should just be `using var pyItem = new PyObject(item);` so that the Python object, corresponding to the `item` would be disposed and not pollute finalizer thread unless duplicated.

src/runtime/Codecs/IterableDecoder.cs
koubaa: @lostmsu actually as it turns out, no it can't.  It leads to some test failures.
https://github.com/pythonnet/pythonnet/pull/1084/checks?check_run_id=1930584148

The reason is that the codec only has the *type*, not the *instance*.  PyIter_Check is meant to be called on the instance.

lostmsu: Right. OK then.

PR #1085, opened 2020-03-10
Codec groups: EncoderGroup and DecoderGroup

lostmsu:

These classes would help to manage codec layers. For example, a library could register its own codecs, but also allow anyone to inject their codecs before library's own:
```csharp
public static EncoderGroup BeforeLibraryEncoders { get; }
    = new EncoderGroup();

LibraryRegisterCodecs()
{
  PyObjectConversions.RegisterEncoder(BeforeLibraryEncoders);
  PyObjectConversions.RegisterEncoder(LibraryEncoder.Instance);
}
```
Then in a program using that library:
```csharp
Library.BeforeLibraryEncoders.Add(preencoder);
```

No

### Checklist

### DIFF
--- a/src/embed_tests/CodecGroups.cs
+++ b/src/embed_tests/CodecGroups.cs
@@ -1,0 +1,132 @@
+namespace Python.EmbeddingTest
+{
+    using System;
+    using System.Linq;
+    using NUnit.Framework;
+    using Python.Runtime;
+    using Python.Runtime.Codecs;
+
+    public class CodecGroups
+    {
+        [Test]
+        public void GetEncodersByType()
+        {
+            var encoder1 = new FakeEncoder<Uri>();
+            var encoder2 = new FakeEncoder<Uri>();
+            var group = new EncoderGroup {
+                new FakeEncoder<Tuple<int>>(),
+                encoder1,
+                encoder2,
+            };
+
+            var got = group.GetEncoders(typeof(Uri)).ToArray();
+            CollectionAssert.AreEqual(new[]{encoder1, encoder2}, got);
+        }
+
+        [Test]
+        public void CanEncode()
+        {
+            var group = new EncoderGroup {
+                new FakeEncoder<Tuple<int>>(),
+                new FakeEncoder<Uri>(),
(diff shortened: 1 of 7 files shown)


src/embed_tests/CodecGroups.cs
```diff
@@ -0,0 +1,132 @@
+namespace Python.EmbeddingTest
+{
+    using System;
+    using System.Linq;
+    using NUnit.Framework;
+    using Python.Runtime;
+    using Python.Runtime.Codecs;
```
filmor, line 7: While I don't have a particular problem with this, why are you using this style? It's not in use in the established codebase and I have never seen it outside either.

src/embed_tests/Codecs.cs
```diff
@@ -83,4 +83,33 @@ public class Codecs {
...
         }
     }
+
+    class FakeEncoder<T> : IPyObjectEncoder
+    {
+        public bool CanEncode(Type type) => type == typeof(T);
+        public PyObject TryEncode(object value) => this.GetRawPythonProxy();
+    }
+
+    class FakeDecoder<TTarget> : IPyObjectDecoder
```
filmor, line 93: Maybe give this a few lines of documentation to make it easier to follow the tests.

src/embed_tests/CodecGroups.cs
lostmsu: Imagine we use a `System.Uri` a lot. At some point in the future we decide to take a dependency (or introduce our own) on something, that has `Python.Uri` class, which we might not need ourselves, but it is public.

With

```csharp
using System;

namespace Python.Runtime {
  class SomeClass {
    Uri uriField;
  }
}
```

the `uriField` will silently change its type from `System.Uri` to `Python.Uri`.

If the `using System;` is inside `namespace Python.Runtime`, that won't happen.

src/embed_tests/Codecs.cs
lostmsu: @filmor renamed classes to make their behavior more clear, and added xmldoc.

PR #1088, opened 2020-03-11
Allow borrowing from NewReference

lostmsu:
implemented as an implicit conversion

To be honest, I would prefer an explicit `NewReference.Borrow()` method due to "explicit is better than implicit" motto. The only problem with implicit conversion that I see is `new PyObject(newRef)` will call an extra incref implicitly as opposed to `newRef.MoveToPyObject()`, and will go unnoticed.

Another thing to consider is having a `null`-check in the conversion. E.g. should we throw `NullReferenceException` if `NewReference` is `null`?

Related to #1087

### Checklist

### DIFF
--- a/src/embed_tests/References.cs
+++ b/src/embed_tests/References.cs
@@ -36,5 +36,20 @@ namespace Python.EmbeddingTest
                 reference.Dispose();
             }
         }
+
+        [Test]
+        public void CanBorrowFromNewReference()
+        {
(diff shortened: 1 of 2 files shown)


lostmsu:
@amos402 can you review this?

lostmsu:
@amos402 dang
https://github.com/ufcpp/NonCopyableAnalyzer/issues/14

PR #1089, opened 2020-03-13
The unicode error was resolved.

nowonbun:
Hello.
When using Korean or Japanese this pythonnet, I was found where the encoding type cannot be selected. When using the existing general string, the following error occurs.
image
So, I have commit the solution, please can you confirm it?

Thank you very much.

...

...

### Any other comments?

...

### Checklist


lostmsu:
This test passes in 3.0, so I believe this change is no longer needed.

```csharp
[Test]
public void Logograms()
{
    PythonEngine.RunSimpleString(
        "import sys\n" +
        "sys.__logograms__ = '안녕하세요'\n"
    );

    Assert.AreEqual(sys.__logograms__.ToString(), "안녕하세요");
}
```

Issue #1093, opened 2020-03-17
Finalizer calls PythonEngine.Initialize


lostmsu:
Here's a couple of concurrent stack traces I caught during today's debugging session (cleaned up a bit for readability):

### Thread pool thread (lambda in `AddPendingCollect`)

 	Debug.Assert(bool condition, string message, string detailMessage)	Unknown
 	DebugUtil.EnsureGIL() Line 145	C#
 	Runtime.XIncref(System.IntPtr op) Line 599	C#
 	Runtime.PyObject_Type(System.IntPtr op) Line 908	C#
 	Runtime.Initialize(bool initSigs) Line 243	C#
 	PythonEngine.Initialize(System.Collections.Generic.IEnumerable<string> args,
                            bool setSysArgv, bool initSigs) Line 174	C#
 	PythonEngine.Initialize(bool setSysArgv, bool initSigs) Line 151	C#
 	PythonEngine.Initialize() Line 146	C#
 	Py.GIL() Line 624	C#
	Finalizer.AddPendingCollect.AnonymousMethod__37_0() Line 147	C#
 	System.Threading.ExecutionContext.RunInternal(
                System.Threading.ExecutionContext executionContext,
                System.Threading.ContextCallback callback,
                object state)	Unknown
 	System.Threading.Tasks.Task.ExecuteWithThreadLocal(ref Task currentTaskSlot)	Unknown
 	System.Threading.ThreadPoolWorkQueue.Dispatch()	Unknown

### Main Thread

 	System.Threading.Tasks.Task.Wait()	Unknown
 	[Waiting on Async Operation]	
 	Finalizer.Collect(bool forceDispose) Line 84	C#
        Finalizer.Shutdown() Line 130	C#
 	Runtime.Shutdown() Line 410	C#
 	PythonEngine.ExecuteShutdownHandlers() Line 385	C#
 	PythonEngine.Shutdown() Line 317	C#
 	TestCustomMarshal.Dispose() Line 18	C#

It looks like when an object is collected while Python engine is shutting down, `Finalizer` might force-restart Python runtime, causing it to be initialized on a different thread (in this case - a thread pool thread).

This is from my branch, but I did not make any changes to disposal/finalization I will try to reproduce it in master. Do not feel familiar enough with shutdown and finalizer machinery to investigate. @amos402 , would you take a look?

amos402:
[...]

PR #1094, opened 2020-03-17
Simplified Finalizer: drain finalization queue when new objects are created.

lostmsu:

This is an attempt to make finalizer much simpler.

Instead of using .NET async tasks, `Py_AddPendingCall` or other synchronization mechanisms we simply move objects to be finalized into a `ConcurrentQueue`, which is periodically drained when a new `PyObject` is constructed.

At the very least this should be resolved, as the new finalization code never tries to acquire GIL on its own: #1093

### Performance

`Finalizer` is not in the baseline (e.g. Python.Runtime 2.3.0), but the following test drops from 98ms/op to 93ms/op on my machine:
```csharp
const int testObjectCount = 100000;
using (Py.GIL())
{
    for (int i = 0; i < testObjectCount; i++)
        new PyList();
}
GC.Collect();
GC.WaitForPendingFinalizers();
Finalizer.Instance.Collect();
```

### Checklist

### Suggested reviewers

@amos402 @Martin-Molinero @benoithudson

### DIFF
--- a/src/runtime/finalizer.cs
+++ b/src/runtime/finalizer.cs
@@ -28,9 +28,7 @@ namespace Python.Runtime
         public bool Enable { get; set; }
 
         private ConcurrentQueue<IPyDisposable> _objQueue = new ConcurrentQueue<IPyDisposable>();
-        private bool _pending = false;
-        private readonly object _collectingLock = new object();
-        private Task _finalizerTask;
+        private int _throttled;
 
         #region FINALIZER_CHECK
 
@@ -75,19 +73,16 @@ namespace Python.Runtime
             Threshold = 200;
         }
 
-        public void Collect(bool forceDispose = true)
+        [Obsolete("forceDispose parameter is unused. All objects are disposed regardless.")]
+        public void Collect(bool forceDispose) => this.DisposeAll();
+        public void Collect() => this.DisposeAll();
+
+        internal void ThrottledCollect()
         {
-            if (Instance._finalizerTask != null
-                && !Instance._finalizerTask.IsCompleted)
-            {
-                var ts = PythonEngine.BeginAllowThreads();
-                Instance._finalizerTask.Wait();
-                PythonEngine.EndAllowThreads(ts);
-            }
-            else if (forceDispose)
-            {
-                Instance.DisposeAll();
-            }
+            _throttled = unchecked(this._throttled + 1);
+            if (!Enable || _throttled < Threshold) return;
+            _throttled = 0;
+            this.Collect();
         }
 
         public List<WeakReference> GetCollectedObjects()
@@ -101,62 +96,18 @@ namespace Python.Runtime
             {
                 return;
             }
-            if (Runtime.Py_IsInitialized() == 0)
-            {
-                // XXX: Memory will leak if a PyObject finalized after Python shutdown,
-                // for avoiding that case, user should call GC.Collect manual before shutdown.
-                return;
-            }
+
 #if FINALIZER_CHECK
             lock (_queueLock)
 #endif
             {
-                _objQueue.Enqueue(obj);
-            }
-            GC.ReRegisterForFinalize(obj);
-            if (!_pending && _objQueue.Count >= Threshold)
-            {
-                AddPendingCollect();
+                this._objQueue.Enqueue(obj);
             }
         }
 
         internal static void Shutdown()
         {
-            if (Runtime.Py_IsInitialized() == 0)
-            {
-                Instance._objQueue = new ConcurrentQueue<IPyDisposable>();
-                return;
-            }
-            Instance.Collect(forceDispose: true);
-        }
-
-        private void AddPendingCollect()
-        {
-            if(Monitor.TryEnter(_collectingLock))
-            {
-                try
-                {
-                    if (!_pending)
-                    {
-                        _pending = true;
-                        // should already be complete but just in case
-                        _finalizerTask?.Wait();
-
-                        _finalizerTask = Task.Factory.StartNew(() =>
-                        {
-                            using (Py.GIL())
-                            {
-                                Instance.DisposeAll();
-                                _pending = false;
-                            }
-                        });
-                    }
-                }
-                finally
-                {
-                    Monitor.Exit(_collectingLock);
-                }
-            }
+            Instance.DisposeAll();
         }
 
         private void DisposeAll()
@@ -178,12 +129,18 @@ namespace Python.Runtime
                     try
                     {
                         obj.Dispose();
-                        Runtime.CheckExceptionOccurred();
                     }
                     catch (Exception e)
                     {
-                        // We should not bother the main thread
-                        ErrorHandler?.Invoke(this, new ErrorArgs()
+                        var handler = ErrorHandler;
+                        if (handler is null)
+                        {
+                            throw new FinalizationException(
+                                "Python object finalization failed",
+                                disposable: obj, innerException: e);
+                        }
+
+                        handler.Invoke(this, new ErrorArgs()
                         {
                             Error = e
                         });
@@ -266,5 +223,16 @@ namespace Python.Runtime
             indexer.Clear();
         }
 #endif
+    }
+
+    public class FinalizationException : Exception
+    {
+        public IPyDisposable Disposable { get; }
+
+        public FinalizationException(string message, IPyDisposable disposable, Exception innerException)
+            : base(message, innerException)
+        {
+            this.Disposable = disposable ?? throw new ArgumentNullException(nameof(disposable));
+        }
     }
 }
--- a/src/runtime/pyobject.cs
+++ b/src/runtime/pyobject.cs
@@ -187,7 +187,29 @@ namespace Python.Runtime
 
             if (!Runtime.IsFinalizing)
             {
-                Runtime.XDecref(this.obj);
+                long refcount = Runtime.Refcount(this.obj);
+                Debug.Assert(refcount > 0, "Object refcount is 0 or less");
+
+                if (refcount == 1)
+                {
+                    Runtime.PyErr_Fetch(out var errType, out var errVal, out var traceback);
+
+                    try
+                    {
+                        Runtime.XDecref(this.obj);
+                        Runtime.CheckExceptionOccurred();
+                    }
+                    finally
+                    {
+                        // Python requires finalizers to preserve exception:
+                        // https://docs.python.org/3/extending/newtypes.html#finalization-and-de-allocation
+                        Runtime.PyErr_Restore(errType, errVal, traceback);
+                    }
+                }
+                else
+                {
+                    Runtime.XDecref(this.obj);
+                }
             }
             this.obj = IntPtr.Zero;
         }
--- a/src/runtime/pyobject.cs
+++ b/src/runtime/pyobject.cs
@@ -30,8 +30,6 @@ namespace Python.Runtime
 #endif  
 
         protected internal IntPtr obj = IntPtr.Zero;
-        private bool disposed = false;
-        private bool _finalized = false;
 
         internal BorrowedReference Reference => new BorrowedReference(obj);
 
@@ -49,6 +47,7 @@ namespace Python.Runtime
             if (ptr == IntPtr.Zero) throw new ArgumentNullException(nameof(ptr));
 
             obj = ptr;
+            Finalizer.Instance.ThrottledCollect();
 #if TRACE_ALLOC
             Traceback = new StackTrace(1);
 #endif
@@ -64,6 +63,7 @@ namespace Python.Runtime
             if (reference.IsNull) throw new ArgumentNullException(nameof(reference));
 
             obj = Runtime.SelfIncRef(reference.DangerousGetAddress());
+            Finalizer.Instance.ThrottledCollect();
 #if TRACE_ALLOC
             Traceback = new StackTrace(1);
 #endif
@@ -74,6 +74,7 @@ namespace Python.Runtime
         [Obsolete("Please, always use PyObject(*Reference)")]
         protected PyObject()
         {
+            Finalizer.Instance.ThrottledCollect();
 #if TRACE_ALLOC
             Traceback = new StackTrace(1);
 #endif
@@ -87,12 +88,6 @@ namespace Python.Runtime
             {
                 return;
             }
-            if (_finalized || disposed)
-            {
-                return;
-            }
-            // Prevent a infinity loop by calling GC.WaitForPendingFinalizers
-            _finalized = true;
             Finalizer.Instance.AddFinalizedObject(this);
         }
 
@@ -182,17 +177,19 @@ namespace Python.Runtime
         /// </remarks>
         protected virtual void Dispose(bool disposing)
         {
-            if (!disposed)
+            if (this.obj == IntPtr.Zero)
             {
-                if (Runtime.Py_IsInitialized() > 0 && !Runtime.IsFinalizing)
-                {
-                    IntPtr gs = PythonEngine.AcquireLock();
-                    Runtime.XDecref(obj);
-                    obj = IntPtr.Zero;
-                    PythonEngine.ReleaseLock(gs);
-                }
-                disposed = true;
+                return;
             }
+
+            if (Runtime.Py_IsInitialized() == 0)
+                throw new InvalidOperationException("Python runtime must be initialized");
+
+            if (!Runtime.IsFinalizing)
+            {
+                Runtime.XDecref(this.obj);
+            }
+            this.obj = IntPtr.Zero;
         }
 
         public void Dispose()
--- a/src/runtime/runtime.cs
+++ b/src/runtime/runtime.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Diagnostics.Contracts;
 using System.Runtime.InteropServices;
 using System.Security;
 using System.Text;
@@ -8,7 +9,6 @@ using Python.Runtime.Platform;
 
 namespace Python.Runtime
 {
-
     /// <summary>
     /// Encapsulates the low-level Python C API. Note that it is
     /// the responsibility of the caller to have acquired the GIL
@@ -106,7 +106,7 @@ namespace Python.Runtime
         internal static object IsFinalizingLock = new object();
         internal static bool IsFinalizing;
 
-        internal static bool Is32Bit = IntPtr.Size == 4;
+        internal static bool Is32Bit => IntPtr.Size == 4;
 
         // .NET core: System.Runtime.InteropServices.RuntimeInformation.IsOSPlatform(OSPlatform.Windows)
         internal static bool IsWindows = Environment.OSVersion.Platform == PlatformID.Win32NT;
@@ -649,6 +649,7 @@ namespace Python.Runtime
 #endif
         }
 
+        [Pure]
         internal static unsafe long Refcount(IntPtr op)
         {
             var p = (void*)op;
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@ This document follows the conventions laid out in [Keep a CHANGELOG][].
 -   When calling C# from Python, enable passing argument of any type to a parameter of C# type `object` by wrapping it into `PyObject` instance. ([#881][i881])
 -   Added support for kwarg parameters when calling .NET methods from Python
 -   Changed method for finding MSBuild using vswhere
+-   Reworked `Finalizer`. Now objects drop into its queue upon finalization, which is periodically drained when new objects are created.
 
 ### Fixed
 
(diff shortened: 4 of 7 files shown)


src/runtime/finalizer.cs
```diff
@@ -178,12 +123,18 @@ private void DisposeAll()
                     try
                     {
                         obj.Dispose();
-                        Runtime.CheckExceptionOccurred();
```
amos402: Remove this can save the overhead, but it may ignore some exceptions.
When reference count become 0, tp_dealloc may call, some internal exceptions may cause in there.
A better way is check the reference count is 1 or not, if it's 1, check the exception after decref the obj, of course these should implement inside Dispose.

src/runtime/pyobject.cs
```diff
@@ -182,16 +177,16 @@ public T As<T>()
         /// </remarks>
         protected virtual void Dispose(bool disposing)
         {
-            if (!disposed)
+            if (this.obj != IntPtr.Zero)
```
amos402: Use `==` condition then `return` can save the indents, I prefer.😂

src/runtime/finalizer.cs
```diff
@@ -97,66 +93,15 @@ public List<WeakReference> GetCollectedObjects()
...
-                return;
-            }
-#if FINALIZER_CHECK
-            lock (_queueLock)
-#endif
+            if (Enable)
             {
-                _objQueue.Enqueue(obj);
-            }
-            GC.ReRegisterForFinalize(obj);
```
amos402: Ahaha, I didn't search this fully in the past, if clr won't release the memory of object after calling the finalizer and it's needless for resurrect it, it can be remove indeed, also including the `_finalized ` member. I need some confirmation.

src/runtime/finalizer.cs
```diff
@@ -25,12 +25,10 @@ public class ErrorArgs : EventArgs
         public event EventHandler<ErrorArgs> ErrorHandler;
 
         public int Threshold { get; set; }
+        [Obsolete("Finalizer will always be enabled in the future")]
```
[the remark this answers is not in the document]
lostmsu: How does the switch help to investigate that?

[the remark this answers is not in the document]
amos402: If on disabled state, you just put the object to the queue and never release them, obviously you can check all garbage at later.

src/runtime/finalizer.cs
lostmsu: It is hard to find documentation about this. According to [this SO answer](https://stackoverflow.com/questions/52468416/who-should-call-pyerr-fetch/52469036#52469036) `Py_DecRef` should preserve exception state.

src/runtime/finalizer.cs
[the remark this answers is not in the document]
lostmsu: @amos402 that is not the current behavior. Even prior to this change setting `Enabled` to `false` would just prevent objects from being put into the queue in the first place.

src/runtime/finalizer.cs
amos402: I encountered some scenes in the past: `tp_dealloc` raised an exception but no one catch it until back to the script then some other calls check and raise, if this happened, you can't locate the exception precisely at the short time. Thus I simply add this.
According to other C source within python, you can assume that `tp_dealloc` wouldn't raise any exceptions, so it's acceptable to remove it.
Nevertheless, `check the reference count is 1 or not, if it's 1, check the exception after decref the obj, of course these should implement inside Dispose` in the debug mode still can be a option to make it expose the bugs easily.

src/runtime/finalizer.cs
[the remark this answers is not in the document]
amos402: Yes, but with this update, you can change it into that behavior since it wouldn't change the mechanism too much.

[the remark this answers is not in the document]
lostmsu: @amos402 it can always be "undeprecated" later. Right now it serves no good purpose.

src/runtime/finalizer.cs
lostmsu: @amos402 problem is, our `Dispose` might be called during someone else's `tp_dealloc`, and raising that exception in C# will cause their `tp_dealloc` to fail with a different exception, thus violating the requirement for `tp_dealloc`  to preserve exceptions.

amos402: Then it should be `Py_FatalError` instead an exception throw in `Dispose`, of course since it would make some performance impact keep it in debug mode.

lostmsu: @amos402 I am still unconvinced by this. Problem is the Python documentation is very lacking on such important details.

Here's the scenario I have in mind:

```python
a = SomeClass()
try:
  foo()
finally:
  a = None # <- this calls __del__ and might eventually call our Dispose
```

I am unsure about Python behavior here, but if `a = None` is executed with an error flag set (because `foo()` threw an exception), then doing `Runtime.CheckExceptionOccurred()` in `Dispose` will cause the exception to be reraised inside `finally` block, which would be totally wrong.

Now that scenario might not behave as I described, but I am unsure if that scenario is impossible.

The reason I don't want to put this behind an `#if DEBUG` is that I don't want switching to Debug build to cause unrelated major change in behavior.

amos402: Not calling `Runtime.CheckExceptionOccurred()`, use `Assert` or `Py_FatalError` instead, the purpose the for exposing the error as soon as possible and make that locatable.

https://github.com/python/cpython/blob/3d53d8756f0403eec6a4e12f183103d651bed6c5/Python/ceval.c#L3691-L3699
Like this, it can help developer find out the problem quickly. Not the same case, I just give an example.

Reraise an exception wouldn't be ignore by Python3, it will throw an exception chain.
Try this:
```python
try:
    rasie Exception(1)
finally:
    raise Exception(2)
```
Following the C sources, not only depending the document should be helpful.

lostmsu: @amos402 as I said, I was not implying this exact example breaks. Forget it. I found a relevant piece in [the documentation](https://docs.python.org/3/extending/newtypes.html#finalization-and-de-allocation):

> One important requirement of the deallocator function is that it leaves any pending exceptions alone. This is important since deallocators are frequently called as the interpreter unwinds the Python stack; when the stack is unwound due to an exception (rather than normal returns), nothing is done to protect the deallocators from seeing that an exception has already been set. Any actions which a deallocator performs which may cause additional Python code to be executed may detect that an exception has been set. This can lead to misleading errors from the interpreter.

amos402: `Any actions which a deallocator performs which may cause additional Python code to be executed may detect that an exception has been set. This can lead to misleading errors from the interpreter.`
That's what I said before. That's why you shouldn't ignore the exception, especially on debug build.

lostmsu: @amos402 this is exactly why we must ignore the exception! Our `Dispose` can be called during Python stack unwinding as described in this quote, and it should totally ignore the exception, that caused the unwinding, otherwise it will cause double fault.

amos402: https://gist.github.com/amos402/5eb5db264bf63414db395fdc4fd178e1
Try it on both Debug and Release of CPython.

lostmsu: @amos402 I am still unconvinced, but I feel like this change might be important for general stability, so I've added the check to debug builds, as there's little harm in doing so.

src/runtime/pyobject.cs
```diff
@@ -182,17 +177,47 @@ public T As<T>()
...
+                {
+                    if (errType != IntPtr.Zero || errVal != IntPtr.Zero || traceback != IntPtr.Zero)
+                    {
+                        // Python requires finalizers to preserve exception:
+                        // https://docs.python.org/3/extending/newtypes.html#finalization-and-de-allocation
+                        Runtime.PyErr_Restore(errType, errVal, traceback);
+                    }
+                    else
+                    {
+                        Runtime.CheckExceptionOccurred();
```
amos402, line 215: Not a good practice for using Fetch and Restore in this case.
It makes a delay raise if some exceptions occurred in `this.obj.ob_type.tp_dealloc` on here.
case 1: Dispose of object A raises an exception, then it pass to next object B till here to raise.
case 2: Dispose of object A raises an exception, but it's the last object in finalize queue, it will be raise on a unpredictable position.
We can assume(or expect) there's no exception before calling `Decref(obj)`, setting an assert check before starting collect may help for debug build.

amos402:
```c#
bool checkError =  Runtime.Refcount(this.obj) == 1;
Runtime.XDecref(this.obj);
if (checkError)
{
    Runtime.CheckExceptionOccurred();
}
```
Just enough for here.

lostmsu: > Dispose of object A raises an exception, then it pass to next object B till here to raise.

But that can't happen in this code. Object's A error will be thrown in line 215 before object B's decref will even happen.

Just doing `Runtime.CheckExceptionOccurred()` is not enough due to the requirement to preserve set exception I mentioned before.

amos402: No, it won't, if the error occurred in `Runtime.XDecref(this.obj);` and no errors before calling this.

lostmsu: @amos402 can you re-read the code? Either I don't understand your comment, or you've read the code incorrectly.

Assuming object A has refcount 1, then `checkError == true` for object A:

- line 198 `Runtime.PyErr_Fetch` will fill all error variables with `IntPtr.Zero`, now no error is set
- line 202 will call decref on A
- line 207 `if` condition `== false`, because all error variable are `IntPtr.Zero` from line 198, so it will go to `else` branch
- line 215, if decref from line 202 raised an error, it will be thrown here immediately

That happens when there's no pending error upon entering `Dispose`. If there's one, line 207 condition will instead be `true`, the potential error from decrefing A will be ignored, and the original error restored instead.

amos402:
```c#
#if DEBUG
                long refcount = Runtime.Refcount(this.obj);
                Debug.Assert(refcount > 0);

                bool checkError = refcount == 1; // refcoucnt == 1
                IntPtr errType = IntPtr.Zero, errVal = IntPtr.Zero, traceback = IntPtr.Zero;
                if (checkError) // true
                {
                    Runtime.PyErr_Fetch(out errType, out errVal, out traceback);
                    // errType != IntPtr.Zero
                }
                disposed = true;
#endif
                Runtime.XDecref(this.obj);
                // Exception raised in Runtime.XDecref(this.obj), or add following code for simulate
                Runtime.PyErr_SetString(Exceptions.Exception, "dealloc exception");
#if DEBUG
                if (checkError) // true
                {
                    // errType != IntPtr.Zero
                    if (errType != IntPtr.Zero || errVal != IntPtr.Zero || traceback != IntPtr.Zero)
                    {
                        // Enter here, exception will be override and it will raise next time.

                        // Python requires finalizers to preserve exception:
                        // https://docs.python.org/3/extending/newtypes.html#finalization-and-de-allocation
                        Runtime.PyErr_Restore(errType, errVal, traceback);
                    }
                    else
                    {
                        Runtime.CheckExceptionOccurred();
                    }
                }
```

lostmsu: @amos402 but `errType != IntPtr.Zero` can only be when some exception is already set. In which case simply doing `Runtime.CheckExceptionOccurred` will just raise that exception, which we should not do because "Python requires finalizers to preserve exception"                     https://docs.python.org/3/extending/newtypes.html#finalization-and-de-allocation

amos402: Not for here, exception may set before the first Dispose. That's why I said `We can assume(or expect) there's no exception before calling Decref(obj), setting an assert check before starting collect may help for debug build.`
And if with that precondition, fetch and restore become meaningless.

lostmsu: @amos402 it is unclear why can we assume there's no exception before calling `Decref`.

lostmsu: @amos402 in fact, I can see we could face a problem with that assumption. Suppose you wants to mix high- and low-level wrappers, and make something like:

```csharp
using (dynamic uri = Py.Import("uri")){
  if (uri.fetch() == null) { Exceptions.SetError(NotFound, "address not found"); }
}
DoSomethingElse();
```
Without `PyErr_Fetch` in `Dispose` this code will raise `PythonException` when exiting `using` block, so `DoSomethingElse` will never be called.

amos402: If we preserved exception and give it to Python till Python call some byte code or Python check the exception set or not, the exception may raise in the wrong place. My previous sample(https://gist.github.com/amos402/5eb5db264bf63414db395fdc4fd178e1) code shown that. Thus, we must handle these exceptions immediately, raise them or make a PyFatalError abort the program.

Why "Python requires finalizers to preserve exception" didn't match here?
Because we custom handle the object's lifecycle by `Runtime.PyObject`, yet not in Python interpreter in `ceval.c`, it will check the exception set on every byte code boundary(not 100% correct, because internal function in CPython may check and raise) and raise them to the user code.

When should we accept "Python requires finalizers to preserve exception"?
After we handled the exception.
You can see the example code from document:
```c
cbresult = PyObject_CallObject(self->my_callback, NULL);
        if (cbresult == NULL)
            PyErr_WriteUnraisable(self->my_callback);
```
it handled by print the exception information. Well, you can say it ignore the exception too, but at least it left some information not just ignore totally.
If 

Sample:
```c#
IntPtr result = Runtime.PyDict_GetItemWithError();
// error occurred, result == IntPtr.Zero, exception set;
// developer ignore the result and continue.
var xxx = new PyObejct(); // just exactly the timing finalizer call `Collect`,
// finish create `xxx`, the exception still remain, when it would revealed to user? Hard to say, maybe after this scope program return to the Python interpreter and Python raise it, or maybe after calling some C function it check the exception, like `assert(PyErr_Occurred());` or `if (PyErr_Occurred()) { // some stuff }`, no matter which kind, the true exception position is just far from the raise position, that's what we want to see.
```

lostmsu: @amos402 I can add PyErr_WriteUnraisable, but I think my example shows why we have to do fetch and restore in `Dispose`

amos402: > 
> 
> @amos402 in fact, I can see we could face a problem with that assumption. Suppose you want to mix high- and low-level wrappers, and make something like:
> 
> ```cs
> using (dynamic uri = Py.Import("uri")){
>   if (uri.fetch() == null) { Exceptions.SetError(NotFound, "address not found"); }
> }
> DoSomethingElse();
> ```
> 
> Without `PyErr_Fetch` and `Restore` in `Dispose` this code will raise `PythonException` when exiting `using` block, so `DoSomethingElse` will never be called.

Yes, that assumption should build on a program which is bug free for this part. So `setting an assert check before starting collect may help for debug build.` is as a requirement.

lostmsu: @amos402 the sample that I posted has no bugs. In fact, I think because of that sample we should do fetch and restore in release mode too.

amos402: > 
> 
> @amos402 I can add `PyErr_WriteUnraisable`, but I think my example shows why we have to do fetch and restore in `Dispose`

If the exception handled, fetch and restore is good.

> 
> 
> @amos402 the sample that I posted has no bugs. In fact, I think because of that sample we should do fetch and restore in release mode too.

Not really, no bugs if we handle the exception, and fetch & restore can guard the exception not be affected, but the raise position of r the exception may still far from it's exactly position. Plus every `Dispose` plus two C calls. Wouldn't it be simple to check exception and raise at start before we extract the object from queue? Then we can assume it's crystal clear before all `Dispose`. 😆I have say that I changed my mind for putting the check on debug build only, it should also check on release build.

lostmsu: @amos402 instead of using `PyErr_WriteUnraisable` I just throw .NET exception, but preserve the Python one.

amos402: > If user got an exception set somehow, why does it have to be `Finalizer` to reveal it? This is a completely unrelated problem, and should not be our concern in the first place. If really necessary, user should be able to call `PyErr_Fetch` or a higher level wrapper at any time they want.

It may not related, but it may help. You can search the usages of `assert(!PyErr_Occurred());` on CPython code, same principle. Well, I always trend to reduce the PInvoke calls because of their costs. So it's better if we can avoid to call fetch & restore.

If just keep the `PyErr_Fetch` and call `CheckExceptionOccurred`, when exception occurred in `Decref`, the exception will lost, it will not preserve them to Python. Exceptions left to the local variables of this frame.

src/runtime/runtime.cs
```diff
@@ -106,7 +106,7 @@ public class Runtime
         internal static object IsFinalizingLock = new object();
         internal static bool IsFinalizing;
 
-        internal static bool Is32Bit = IntPtr.Size == 4;
+        internal static bool Is32Bit => IntPtr.Size == 4;
```
amos402, line 109: Its' better to change it into readonly field, using property may make it check the condition everytimes.

src/runtime/finalizer.cs
```diff
@@ -75,19 +72,17 @@ private Finalizer()
             Threshold = 200;
         }
 
-        public void Collect(bool forceDispose = true)
+        [Obsolete("forceDispose parameter is unused. All objects are disposed regardless.")]
+        public void Collect(bool forceDispose) => this.DisposeAll();
+        public void Collect() => this.DisposeAll();
+
+        int _throttled;
```
amos402, line 79: Is it better to move it to the fields region than isolate it on here?

src/runtime/runtime.cs
lostmsu: https://gist.github.com/lostmsu/ed31ec13815e2af674abd96d0a4fc8ae
|         Method |           Job |       Runtime |      Mean |    Error |   StdDev |
|--------------- |-------------- |-------------- |----------:|---------:|---------:|
| StaticReadonly |      .NET 4.8 |      .NET 4.8 |  93.69 us | 1.812 us | 2.541 us |
|         Inline |      .NET 4.8 |      .NET 4.8 |  92.48 us | 1.498 us | 1.402 us |
| StaticReadonly | .NET Core 3.1 | .NET Core 3.1 | 188.60 us | 3.608 us | 4.155 us |
|         Inline | .NET Core 3.1 | .NET Core 3.1 |  94.08 us | 1.868 us | 1.919 us |
| StaticReadonly |          Mono |          Mono |  92.82 us | 1.824 us | 2.172 us |
|         Inline |          Mono |          Mono |  93.29 us | 1.831 us | 2.035 us |

src/runtime/finalizer.cs
lostmsu: 1. There's no fields region in this class.
2. This field is not used outside of the `ThrottledCollect` function.

src/runtime/pyobject.cs
lostmsu: > So it's better if we can avoid to call fetch & restore.

We can't due to the example I provided:
```csharp
using (dynamic uri = Py.Import("uri")){
  if (uri.fetch() == null) { Exceptions.SetError(NotFound, "address not found"); }
}
DoSomethingElse();
```

> If just keep the `PyErr_Fetch` and call `CheckExceptionOccurred`, when exception occurred in `Decref`, the exception will lost, it will not preserve them to Python

Why are you saying that? The original Python exception is restored for Python in `finally` block. And the exception from `Decref` is raised as a C# exception, which the caller of `Dispose` can handle in whatever way they want.

src/runtime/runtime.cs
amos402: https://gist.github.com/amos402/7bd6adab1fe489c0552a109a8b3dff28
Haven't been use that benchmark framework before so I wrote it by raw.
Forget the errors, seems they were same after optimized. Haven't dig in their JITed code, I'm not sure either the usage striped by optimization or just generated the same machine code.
Never mind, it won't be big deal always.

Debug:
```
.NET Framework 4.8.4150.0
is32bit: False
Is32Bit: False
StaticReadonly best: 00:00:00.1345072
Inline best: 00:00:00.2713410
Press any key to continue . . .
```

Release:
```
.NET Framework 4.8.4150.0
is32bit: False
Is32Bit: False
StaticReadonly best: 00:00:00.0660235
Inline best: 00:00:00.0660266
Press any key to continue . . .
```

src/runtime/pyobject.cs
amos402: Aha, just too for this conversation, didn't notice that you wrap it by finally, now it should be fine.

src/runtime/finalizer.cs
amos402: I mean put into here for example.
```
         private ConcurrentQueue<IPyDisposable> _objQueue = new ConcurrentQueue<IPyDisposable>();
         private int _throttled;
```

Issue #943, opened 2019-08-20
calling C# method that have Params argument , throwing TypeError when not using that argument 


tbeidas:
### Environment

-   Pythonnet version: 2.4.0 
-   Python version: 3.6.8
-   Operating System: Windows Server 2016

### Details

After upgrade from Pythonnet 2.2.2 to 2.4.0, I get the following error:

> No method matches given arguments for GetConfigInterface

C# method:

    public static IConfig GetConfigInterface (ConfigType type , params string [] param)    

### What commands did you run to trigger this issue? 

    GetConfigInterface(self.pdcuConfigType) 

This worked without any issue using pythonnet 2.2.2 while it fails at version 2.4.0.

It seems to be related to the reported issue https://github.com/pythonnet/pythonnet/issues/916, but it's a different use case.

PR #1106, opened 2020-04-08
Params arg no params

slide:

Check the last parameter to see if it is params (and not supplied). By setting the `paramsArray` variable, the binding code does the right thing and passes an empty array.

#943 

### Checklist


src/runtime/methodbinder.cs
```diff
@@ -415,11 +416,32 @@ internal Binding Bind(IntPtr inst, IntPtr args, IntPtr kw, MethodBase info, Meth
...
-                        // the managed function accepts it - hopefully :]
-                        ? Runtime.PyTuple_GetSlice(args, arrayStart, pyArgCount)
-                        : Runtime.PyTuple_GetItem(args, paramIndex);
+                    if(arrayStart == paramIndex)
+                    {
+                        // for a params method, we may have a sequence or single/multiple items
+                        // here we look to see if the item at the paramIndex is there or not
+                        // and then if it is a sequence itself.
+                        IntPtr item = Runtime.PyTuple_GetItem(args, paramIndex);
+                        Runtime.PyErr_Clear();
```
lostmsu, line 425: That would just eat all exceptions. Should add proper handling.

src/runtime/methodbinder.cs
```diff
@@ -415,11 +416,32 @@ internal Binding Bind(IntPtr inst, IntPtr args, IntPtr kw, MethodBase info, Meth
...
+                        if(item != IntPtr.Zero && !Runtime.PyString_Check(item) && Runtime.PySequence_Check(item))
```
lostmsu, line 426: Please, add a test for `void Foo(params object[] args)` called from Python like this: `Foo([], [])`. In this case `args` should become an array of two arrays. It is unclear if this situation is handled.

src/runtime/methodbinder.cs
slide: Ok, I will revisit.

src/runtime/methodbinder.cs
slide: Will do!

src/runtime/methodbinder.cs
slide: I removed this completely based up updates.

src/runtime/methodbinder.cs
slide: Added test and fixed up code when it was not passing.

slide:
Close and reopen to see if the build is intermittent.

src/runtime/methodbinder.cs
```diff
@@ -415,11 +416,42 @@ internal Binding Bind(IntPtr inst, IntPtr args, IntPtr kw, MethodBase info, Meth
...
+                                {
+                                    Runtime.XDecref(item);
+                                }
+                            }
+                        }
+                        else
+                        {
+                            doDecref = true;
+                            op = Runtime.PyTuple_GetSlice(args, arrayStart, pyArgCount);
+                        }
```
lostmsu, line 448: Just a suggestion: maybe move params argument handling to a separate function for clarity

slide: That's a good idea. Would you like it done as part of this PR, or a separate PR?

slide:
Thanks for reopening, got pulled into a meeting and didn't get a chance to reopen.

lostmsu:
@slide better this one. Then I can squash.

src/runtime/methodbinder.cs
```diff
@@ -397,8 +432,9 @@ internal Binding Bind(IntPtr inst, IntPtr args, IntPtr kw, MethodBase info, Meth
             {
                 var parameter = pi[paramIndex];
                 bool hasNamedParam = kwargDict.ContainsKey(parameter.Name);
+                bool doDecref = false;
```
lostmsu, line 435: NIT: in python docs this situation is called `isNewReference`

lostmsu:
@slide Travis passed. Need to investigate why is it not being picked up.

slide:
Close and reopen to try and get a Travis build responding.

filmor:
Thank you very much :)

PR #1109, opened 2020-04-10
First chunk of modernisation changes

filmor:
- Move projects around and add clean project/solution files
- Drop monoclr and clrmodule
- Add stub clr module
- Drop nuget config
- Add test helper project
- Drop PYTHON3 flag (default now)
- Remove unneeded additional interop files
- Make NETSTANDARD the default
    Drops also
    - Custom XDecref/XIncref (should be readded)
    - Remote object handling for .NET Framework
- Remove Runtime._UCS dependency from CustomMarshaler
- Stop using Runtime.IsPython2/3

...

...

### Any other comments?

...

### Checklist


filmor:
@koubaa @lostmsu I've gotten quite a bit further. The current state of this PR implements embedding in .NET Core and Mono on Linux (haven't tested yet on Windows, will not be able to test on OS X), I tested with the following script:

```python
import os
#os.environ["MONO_LOG_LEVEL"] = "debug"
#os.environ["MONO_LOG_MASK"] = "cfg,dll"
#os.environ["COREHOST_DEBUG"] = "1"
#os.environ["COREHOST_TRACE"] = "1"
#os.environ["COREHOST_TRACE_VERBOSITY"] = "4"

import pythonnet, clr_loader

import sys

if sys.argv[1] == "mono":
    mono = clr_loader.get_mono()
    rt = mono
elif sys.argv[1] == "core":
    core = clr_loader.get_coreclr("/home/benedikt/git/clr-loader/example/out/example.runtimeconfig.json")
    rt = core

pythonnet.set_runtime(rt)
pythonnet.load()

print("Loaded pythonnet")

import clr
from System import Console
Console.WriteLine("Success")
```

Still crashes when I try to do this from IPython, but we're getting somewhere :)

Next steps are:
- Make a `clr.py` that handles these initial load steps like they worked before and make the test-suite work
- See how this works for the other direction (embedding Python in C#)
- Test on Windows in all variants
- Allow for a simpler way to configure .NET Core (currently requires a `runtimeconfig.json`)

koubaa:
@filmor this PR seems to accomplish a lot, I wish it could be staged so it would be easier to follow.  For instance, does this also fix 857 and drop support for python2?

filmor:
This doesn't drop Python 2 support, yet, it merely defaults to Python 3 unless `PYTHON2` is defined. I don't think moving around the files and sticking to the old csproj file format makes sense. Regarding stages, be my guest, I just don't want to redo all parts of this multiple times. Moving to the new project format only really works while simultaneously dropping the old Mono CLR loader, which then requires the new one to be in place for tests to be able to run.

What I want to accomplish in the end here is:
1. New project format
2. Use of `clr-loader`
3. .NET Standard only
4. Reasonable structure for the `interop*.cs` files

koubaa:
@filmor I'll push up a PR with some things staged.  I think the smaller and more incremental we can make this branch the easier it'll be to land (not to mention to allow users to bisect if it breaks anyone!)

My first PR will be:

- default to python 3 
- drop custom xincref/xdecref & remoting
- Stop using Runtime.IsPython2/3

Which variants do we need to test on windows?  I can also work on that + embedding.

filmor:
@koubaa Have a look at the other PR (#1015), it has much smaller grained commits. The reason I went for this larger variant is that I did some steps in the "wrong" order (first moving the files around, then modifying them).

The parts you are suggesting may safely go against master, all steps that change the project structure can't. Please ensure that they are compatible with this branch, though.

I don't understand what you mean by "variants".

koubaa:
@filmor I am referring to a task in your comment 3 days ago: "Test on Windows in all variants"

filmor:
On Windows we can use all .NET runtimes, i.e. Framework, Core and Mono, so all of those should be tested.

koubaa:
@filmor understood.  I think I can manage that.  In the meantime I pushed the first stage here: https://github.com/pythonnet/pythonnet/pull/1117

PR #1117, opened 2020-04-17
Modernization stage1

koubaa:
pulls out a first stage from this PR:
https://github.com/pythonnet/pythonnet/pull/1109/files
...

no
...

### Any other comments?

...

### Checklist

-   [N/A] Make sure to include one or more tests for your change
-   [N/A] If an enhancement PR, please create docs and at best an example

### DIFF
--- a/src/runtime/runtime.cs
+++ b/src/runtime/runtime.cs
@@ -22,7 +22,7 @@ namespace Python.Runtime
 
         public static int UCS => _UCS;
 
-#if UCS2
+#if !UCS2
         internal const int _UCS = 4;
 
         /// <summary>
--- a/src/runtime/runtime.cs
+++ b/src/runtime/runtime.cs
@@ -22,7 +22,7 @@ namespace Python.Runtime
 
         public static int UCS => _UCS;
 
-#if UCS4
+#if UCS2
         internal const int _UCS = 4;
 
         /// <summary>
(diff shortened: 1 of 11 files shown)


koubaa:
@filmor this is the first stage I described (although I couldn't figure out what you mean by custom IncRef and DecRef)

src/runtime/runtime.cs
```diff
@@ -22,85 +22,25 @@ public class Runtime
 
         public static int UCS => _UCS;
 
-#if UCS4
+#if UCS2
```
filmor, line 25: I corrected  this later, should be `!UCS2`

lostmsu:
@filmor if this is targeting Python.NET 3+, I think we should just drop Python 2 support altogether with all related code.

Keeping those `#if`s, but not making newly added code Python 2 compatible, and not running CI tests won't help people trying to backport, and will only get in the way.

We can just remove all Python 2-related code in a single commit, including the CI checks. Whoever needs it then can just revert it and try to work from there (which I doubt will ever happen).

@koubaa @filmor in my branch I simply use a static check using [RuntimeInformation.IsOSPlatform](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.runtimeinformation.isosplatform) to decide between UCS2 and UCS4, then just cache boolean flag. I don't think it has any significant impact on performance as opposed to using `#if`s.

koubaa:
@lostmsu @filmor this changeset is targeting the 2.x branch.  I totally agree with dropping python2 support in the 3.x branch.

PR #1118, opened 2020-04-17
Change as

koubaa:
Improve behavior of PyObject.As<>.  This is needed to make external function codec work:
https://github.com/pythonnet/codecs/pull/1
...

...

### Any other comments?

...

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,7 +26,7 @@ This document follows the conventions laid out in [Keep a CHANGELOG][].
 -   When calling C# from Python, enable passing argument of any type to a parameter of C# type `object` by wrapping it into `PyObject` instance. ([#881][i881])
 -   Added support for kwarg parameters when calling .NET methods from Python
 -   Changed method for finding MSBuild using vswhere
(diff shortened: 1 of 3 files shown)


filmor:
Hmm, I think in general this is fine because we don't lose anything (`As<PyObject>` will still work), but it's nevertheless a silent breaking change. This is possible in the future, but I wouldn't want to do it during a point-release. Can you introduce a preliminary new function for this behaviour?

koubaa:
@filmor I can do that.  I will also create a new issue to address this in a major release.

lostmsu:
@koubaa if this is specifically targeted to conversion to `object`, why does it have to be generic?

koubaa:
> @koubaa if this is specifically targeted to conversion to `object`, why does it have to be generic?

@lostmsu good point.  It would be more clear to just call it AsObject

lostmsu:
@koubaa there is already `AsManagedObject(Type)`. Does it not work for you? It appears to be doing the same when `Type == object`.

koubaa:
@lostmsu didn't know about that.  I don't need this change anymore

PR #1131, opened 2020-05-01
Allow excluding public .NET types from being exposed to Python via [PyExport(false)]

lostmsu:

Introduced [`PyExportAttribute`](https://github.com/pythonnet/pythonnet/wiki/PyExport), that can be applied to individual types or the entire assembly, and controls visibility of these types/assemblies to Python. This is useful in case .NET type has a name conflict with Python type.

The attribute in checked in `ScanAssembly`.

### Checklist

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -306,7 +306,6 @@ class BuildExtPythonnet(build_ext.build_ext):
             _config = "{0}Win".format(CONFIG)
             _solution_file = "pythonnet.sln"
             _custom_define_constants = False
-            defines.append("NET40")
         elif DEVTOOLS == "MsDev15":
             _xbuild = '"{0}"'.format(self._find_msbuild_tool_15())
             _config = "{0}Win".format(CONFIG)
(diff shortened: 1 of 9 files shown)


lostmsu:
My sanity is on the brink. This change does not even touch any Python objects, there are not changes in C# tests at all, and yet they start segfaulting. UGHR

setup.py
```diff
@@ -316,6 +317,7 @@ def build_extension(self, ext):
             _config = "{0}Mono".format(CONFIG)
             _solution_file = "pythonnet.sln"
             _custom_define_constants = False
+            defines.append("NET40")
```
filmor, line 320: This is new, is this intentional?

filmor:
You are unconditionally setting `NET40` now, even on Mono, maybe that's the problem?

setup.py
lostmsu: @filmor yes, it is. This is used in the same commit.

lostmsu:
@filmor we don't have any other `NET40` checks in the code, just the one I've added in this change. If this check would not have worked properly, the code would not compile due to a conflict with existing `GetCustomAttribute` extension method.

lostmsu:
Temporarily removed `NET40` check to ensure it is required.

lostmsu:
@filmor it seems to be passing without `NET40`, but I am still sad that I have no idea why it failed with it.

Issue #1098, opened 2020-03-29
Implement Python/CLR exception chaining (e.g. InnerException and __cause__)


lostmsu:
Some info for the Python side: https://stackoverflow.com/questions/51030659/modifying-or-reraising-python-error-in-c-api

Issue #893, opened 2019-06-24
Populate .NET's InnerException from exc.__cause__ when available


lostmsu:
`PythonException` objects should provide `InnerException` property when `__cause__` is set in the original Python exception. This is a reverse-direction of feature #223

Cronan:
Makes sense

PR #1134, opened 2020-05-05
Improve Python <-> .NET exception integration

lostmsu:

1. .NET exceptions bubbled through Python stack back into .NET are preserved (before: `PythonException` was thrown instead).
2. Codecs can now encode and decode exceptions
3. .NET `InnerException` property is populated from Python's `__cause__` attribute.
4. Python exception stack is added to .NET exception stack
5. Internal: simplified memory management in `PythonException` class. Instead of working with raw pointers, uses `PyObject` instances.

### Breaking

1. `Traceback`, `PyVal` and `PyType` are now stored and returned as `PyObject`s.
2. `PythonException.Restore` no longer nulls out `PythonException` instance.
3. `PythonException` instances are no longer disposable. You can still dispose `.Type`, `.Value` and `.Traceback`, but it is not recommended, as they may be shared with other instances.

#893, #1098

### Checklist


lostmsu:
Damn CI failures do not reproduce locally :/

Felk:
Hello, given that `Exceptions` is not public anymore, what is the recommended way of catching specific python exceptions? Previously I used this:
```cs
catch (PythonException ex) when (ex.PyType == Exceptions.KeyError)
```

lostmsu:
You can get `KeyError` from `builtins` module.

Felk:
Thank you, this is how I got it to work now:
```cs
dynamic builtins = Py.Import("builtins");
try
{
    // ...
}
catch (PythonException ex) when (ex.Type.Equals(builtins.KeyError))
{
    // ...
}
```

lostmsu:
@Felk comparing type handles will be slightly faster

Felk:
@lostmsu can you give me an example of what you mean?

lostmsu:
@Felk `ex.Type.Handle == builtins.KeyError.Handle`

Issue #1097, opened 2020-03-26
Incorrect function resolution when calling overloads from python with keyword args


jmlidbetter:
### Environment

-   Pythonnet version: 2.4.0
-   Python version: 3.6.5
-   Operating System: Centos

### Details

-   Describe what you were trying to get done.

    When calling .NET functions with overloads from python with keyword args, the MethodBinder.Bind method will return the first method it finds which might not be the actual method the user is trying to call. Take the following simple C# code:

```C#
public static string OptionalArgumentsMethod(int arg1, int arg2)
{ 
    return $"{arg1}{arg2}XX";
}

public static string OptionalArgumentsMethod(int arg1, int arg2, int arg3 = 3)
{
    return $"{arg1}{arg2}{arg3}X";
}

public static string OptionalArgumentsMethod(int arg1, int arg2, int arg3 = 4, int arg4 = 5)
{
    return $"{arg1}{arg2}{arg3}{arg4}";
}
```

Calling, for example, `OptionalArgumentsMethod(1, 2, arg3=5)` we expect to get back `125X` but instead we see:
```python
>>> NamedArgumentsTest.OptionalArgumentsMethod(1, 2)
'12XX'
>>> NamedArgumentsTest.OptionalArgumentsMethod(1, 2, 3)
'123X'
>>> NamedArgumentsTest.OptionalArgumentsMethod(1, 2, arg3=5)
'12XX'
>>> NamedArgumentsTest.OptionalArgumentsMethod(1, 2, 3, 5)
'1235'
>>> NamedArgumentsTest.OptionalArgumentsMethod(1, 2, 3, arg4=5)
'123X'
>>> NamedArgumentsTest.OptionalArgumentsMethod(1, 2, arg4=0)
'12XX'
```

jmlidbetter:
Also forgot to point out that we have merged commit [1bcbeb5d](https://github.com/pythonnet/pythonnet/commit/1bcbeb5d0e9bcc7f0304994a44a7485d6126873b) into our internal fork.

danabr:
Note that this was at least partially fixed by #1242 .

PR #1136, opened 2020-05-07
Fix kwarg func resolution

jmlidbetter:

A first run at fixing issue #1097. Currently, the method for selecting which method to use is quite simple and just takes the first function it finds to match. The problem is that the current picking algorithm has no knowledge of kwargs and how many have been supplied or used, i.e. MatchesArgumentCount will return true even if we've supplied some kwargs but the current overload does not require them. Ultimately, this means the first matched method might not be the best.

This small patch aims to rectify this issue by collecting together all the methods for which MatchesArgumentCount returns true and then from these selecting the best match based on the following criteria

- Maximizing the number of supplied kwargs that have been used
- Minimizing the number of default arguments needed

It also has some logic to abort in the case where

- The above criteria are met by more than one overload
- We need at least one default argument to make the call

This stops users in python making ambiguous function calls - see the tests for an example.

#1097 

### Any other comments?

Feedback please!

### Checklist


DrNickClarke:
Just to note that we have been using this fix for some time and it does appear to work. It is a welcome change from needing to specify long list of args you don't care about.

src/runtime/methodbinder.cs
```diff
@@ -335,6 +342,57 @@ internal Binding Bind(IntPtr inst, IntPtr args, IntPtr kw, MethodBase info, Meth
...
+                    if (argMatchedTuple.Item1 == bestKwargMatchCount)
+                    {
+                        bestKwargMatches.Add(argMatchedTuple);
+                    }
+                }
+
+                // Order by the number of defaults required and find the smallest
+                var fewestDefaultsRequired  = bestKwargMatches.OrderBy((x) => x.Item2).ToArray()[0].Item2;
+
+                List<Tuple<int, int, object[], int, MethodBase>> bestDefaultsMatches =
```
lostmsu, line 367: Can you make named structs from the tuples? It's hard to read with `Item1`-like references.

jmlidbetter:
> Just FYI, there's a much better way to do method resolution (which I hope we will adopt at some point) but it requires lots of rework: [C# Binder](https://docs.microsoft.com/en-us/dotnet/api/microsoft.csharp.runtimebinder.binder). I did a simple experiment here: [losttech@605c576](https://github.com/losttech/pythonnet/commit/605c576a1a5b333c76b4e92af56c4a977c44c3b2)
> 
> We can still take this change in meanwhile.

Yes this wasn't supposed to change the world, simply just to get something that worked :smile:

src/runtime/methodbinder.cs
```diff
@@ -335,6 +356,55 @@ internal Binding Bind(IntPtr inst, IntPtr args, IntPtr kw, MethodBase info, Meth
...
                 }
 
+                var matchedMethod = new MatchedMethod(kwargsMatched, defaultsNeeded, margs, outs, mi);
+                argMatchedMethods.Add(matchedMethod);
+            }
+            if (argMatchedMethods.Count() > 0)
+            {
+                // Order matched methods by number of kwargs matched and get the max possible number
+                // of kwargs matched
+                var bestKwargMatchCount = argMatchedMethods.OrderBy((x) => x.KwargsMatched).Reverse().ToArray()[0].KwargsMatched;
```
lostmsu, line 366: You can use `.Max` or `.Min` here.

src/runtime/methodbinder.cs
```diff
@@ -335,6 +356,55 @@ internal Binding Bind(IntPtr inst, IntPtr args, IntPtr kw, MethodBase info, Meth
...
+                foreach (MatchedMethod testMatch in argMatchedMethods)
+                {
+                    if (testMatch.KwargsMatched == bestKwargMatchCount)
+                    {
+                        bestKwargMatches.Add(testMatch);
+                    }
+                }
+
+                // Order by the number of defaults required and find the smallest
+                var fewestDefaultsRequired  = bestKwargMatches.OrderBy((x) => x.DefaultsNeeded).ToArray()[0].DefaultsNeeded;
```
lostmsu, line 378: Same, `.Max` or `.Min`

src/runtime/methodbinder.cs
```diff
@@ -335,6 +356,55 @@ internal Binding Bind(IntPtr inst, IntPtr args, IntPtr kw, MethodBase info, Meth
...
+                }
+
+                // If we're here either:
+                //      (a) There is only one best match
+                //      (b) There are multiple best matches but none of them require
+                //          default arguments
+                // in the case of (a) we're done by default. For (b) regardless of which
+                // method we choose, all arguments are specified _and_ can be converted
+                // from python to C# so picking any will suffice
+                MatchedMethod bestMatch = bestDefaultsMatches.ToArray()[0];
```
lostmsu, line 403: If you only ever need the first item from the list, why fill an entire list in the first place?

src/runtime/methodbinder.cs
```diff
@@ -335,6 +356,58 @@ internal Binding Bind(IntPtr inst, IntPtr args, IntPtr kw, MethodBase info, Meth
...
+                var fewestDefaultsRequired  = bestKwargMatches.Min(x => x.DefaultsNeeded);
+                int bestCount = 0;
+                int bestMatchIndex = -1;
+
+                foreach (MatchedMethod testMatch in bestKwargMatches)
+                {
+                    if (testMatch.DefaultsNeeded == fewestDefaultsRequired)
+                    {
+                        bestCount++;
+                        if (bestMatchIndex == -1)
```
jmlidbetter, line 387: Not having this check will cause tests to fail because a different overload is picked- e.g. MethodTest.TestOverloadedObjectTwo(5, 5) will get the (Object, Object) overload rather than the (Int, Int) overload

src/runtime/methodbinder.cs
```diff
@@ -335,6 +356,46 @@ internal Binding Bind(IntPtr inst, IntPtr args, IntPtr kw, MethodBase info, Meth
                     continue;
                 }
 
+                var matchedMethod = new MatchedMethod(kwargsMatched, defaultsNeeded, margs, outs, mi);
+                argMatchedMethods.Add(matchedMethod);
+            }
+            if (argMatchedMethods.Count() > 0)
```
lostmsu, line 362: Please use `Count` property. `Count()` method might require a run through the entire collection.

src/runtime/methodbinder.cs
```diff
@@ -335,6 +356,46 @@ internal Binding Bind(IntPtr inst, IntPtr args, IntPtr kw, MethodBase info, Meth
...
+                            bestMatchIndex = argMatchedMethods.IndexOf(testMatch);
+                    }
+                }
+
+                if (bestCount > 1 && fewestDefaultsRequired > 0)
+                {
+                    // Best effort for determining method to match on gives multiple possible
+                    // matches and we need at least one default argument - bail from this point
+                    return null;
+                }
```
lostmsu, line 385: I think this makes this change breaking for Python.NET 2.x branch. Currently the first match is returned irrespective if there are other matches. No need to fix it, just means might have to be merged later.

src/runtime/methodbinder.cs
```diff
@@ -335,6 +356,46 @@ internal Binding Bind(IntPtr inst, IntPtr args, IntPtr kw, MethodBase info, Meth
...
+
+                // If we're here either:
+                //      (a) There is only one best match
+                //      (b) There are multiple best matches but none of them require
+                //          default arguments
+                // in the case of (a) we're done by default. For (b) regardless of which
+                // method we choose, all arguments are specified _and_ can be converted
+                // from python to C# so picking any will suffice
+                MatchedMethod bestMatch = argMatchedMethods.ElementAt(bestMatchIndex);
```
lostmsu, line 394: `argMatchedMethods[bestMatchIndex]`

src/runtime/methodbinder.cs
```diff
@@ -539,11 +600,16 @@ static Type TryComputeClrArgumentType(Type parameterType, IntPtr argument, bool
...
             out bool paramsArray,
-            out ArrayList defaultArgList)
+            out ArrayList defaultArgList,
+            out int kwargsMatched,
+            out int defaultsNeeded)
         {
             defaultArgList = null;
             var match = false;
             paramsArray = false;
+            var kwargCount = kwargDict.Count();
```
lostmsu, line 610: Same: use `.Count` property instead of the method.

src/runtime/methodbinder.cs
```diff
@@ -335,6 +356,46 @@ internal Binding Bind(IntPtr inst, IntPtr args, IntPtr kw, MethodBase info, Meth
...
+                int bestCount = 0;
+                int bestMatchIndex = -1;
+
+                foreach (MatchedMethod testMatch in argMatchedMethods)
+                {
+                    if (testMatch.DefaultsNeeded == fewestDefaultsRequired && testMatch.KwargsMatched == bestKwargMatchCount)
+                    {
+                        bestCount++;
+                        if (bestMatchIndex == -1)
+                            bestMatchIndex = argMatchedMethods.IndexOf(testMatch);
```
lostmsu, line 376: Instead of doing `IndexOf` here, which might be expensive (results in quadratic complexity over number of overloads), change the loop to be a `for(int index = 0; index < argMatchedMethods.Count; index++) { ... }` loop, and use `index`.

src/runtime/methodbinder.cs
```diff
@@ -310,6 +327,8 @@ internal Binding Bind(IntPtr inst, IntPtr args, IntPtr kw, MethodBase info, Meth
                 _methods = GetMethods();
             }
 
+            List<MatchedMethod> argMatchedMethods = new List<MatchedMethod>(_methods.Length);
```
lostmsu, line 330: NIT: `var argMatchedMethods = ...`

lostmsu:
@jmlidbetter we have released 2.5 and `master` is now working on 3.x. Can you update this PR, so we could merge it?

jmlidbetter:
> @jmlidbetter we have released 2.5 and `master` is now working on 3.x. Can you update this PR, so we could merge it?

Sure. I assume just rebasing on top of master is sufficient or is there something else you'd want?

lostmsu:
@jmlidbetter it should be. I'll re-review just in case.

jmlidbetter:
@lostmsu Ok I have rebased, feel free to check and merge (as appropriate)

lostmsu:
@jmlidbetter related tests seem to be failing.

jmlidbetter:
Hmm that is strange - when I ran the tests on my machine they all passed.... I will have a look tomorrow.

jmlidbetter:
Seems like the Linux tests have passed ok but the windows ones are broken - perhaps there is some slight difference in behaviour on windows. I have limited ability test on windows but I'll see what I can do.

jmlidbetter:
@lostmsu I have not been able to reproduce the test failures. Building on Windows I get the following output:

image

image

Are you able to repro this issue on windows?

lostmsu:
I am not sure what causes the discrepancy, but judging by the code, it seems like the CI build messes up with named parameters. I will try to see if that is correct.

However, this also uncovers a bug: if Python passed a named parameter not defined in C#, the overload resolution behavior is incorrect. I suggest to handle that and add a test too.

```csharp
void NoOverloads(int a = 0) {}
```

```python
NoOverloads(c=1) # should fail, because c does not match a
```

jmlidbetter:
I believe I had originally implemented that behaviour (not submitted onto this branch) but it broke some other tests. I may be misremembering though, I will check tomorrow.

jmlidbetter:
@lostmsu here is the breaking test if you fail when a kwarg is supplied that doesn't appear in the method signature:

```
____________________________________________ test_namespace_and_init _____________________________________________

    def test_namespace_and_init():
        calls = []
        class TestX(System.Object):
            __namespace__ = "test_clr_subclass_with_init_args"
            def __init__(self, *args, **kwargs):
                calls.append((args, kwargs))
>       t = TestX(1,2,3,foo="bar")
E       TypeError: No constructor matches given arguments: (<class 'int'>, <class 'int'>, <class 'int'>)

src/tests/test_subclass.py:229: TypeError
```
I haven't done any further investigation into this. I guess there are two ways to proceed:
(1) Log this as an issue and proceed as is
(2) Dig in and fix the broken test

Its up to you, but personally I'm in favour of (1) because (a) having working kwargs (except in the case where you supply extra ones ;) ) is quite a big improvement and (b) python is fairly loose anyway - it might be possible to throw a warning when extraneous kwargs are given, I don't know.

jmlidbetter:
Are we good to merge this change now once the tests have passed?

filmor:
This looks good to me, @lostmsu are you also fine with merging this?

PR #1137, opened 2020-05-10
Added IsNone() and GetPyNone()

henon:

IsNone() and GetPyNone() were needed for Numpy.NET to be able to find out if a PyObject is none. 

No.

### Any other comments?

These are in use for a long time now, I just now figured out how to PR them. I'll commit the necessary things from the checklist below.

### Checklist

### DIFF
--- a/src/runtime/runtime.cs
+++ b/src/runtime/runtime.cs
@@ -473,6 +473,19 @@ namespace Python.Runtime
         internal static IntPtr PyNone;
         internal static IntPtr Error;
 
+        public static PyObject _none;
+        public static PyObject GetPyNone()
+        {
+            if (_none == null)
+            {
+                var result = Runtime.PyNone;
+                Runtime.XIncref(result);
+                _none= new PyObject(result);
+            }
+
+            return _none;
+        }
+
         /// <summary>
         /// Check if any Python Exceptions occurred.
         /// If any exist throw new PythonException.
--- a/src/runtime/runtime.cs
+++ b/src/runtime/runtime.cs
@@ -473,17 +473,14 @@ namespace Python.Runtime
         internal static IntPtr PyNone;
         internal static IntPtr Error;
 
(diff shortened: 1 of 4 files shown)


henon:
I have no idea how to add a test and the change is so small that it doesn't really need documentation.

filmor:
I completely get the `IsNone`, but what do you need the `GetPyNone` for? It's a bit arbitrary here, isn't it?

lostmsu:
@filmor re:`GetPyNone` it is actually useful. I have it in my branch too. That's the easiest way to pass `None`.

Though as a static property (and I don't see why it should not be a property).

src/runtime/runtime.cs
```diff
@@ -473,6 +473,19 @@ private static void ResetPyMembers()
         internal static IntPtr PyNone;
         internal static IntPtr Error;
 
+        public static PyObject _none;
+        public static PyObject GetPyNone()
```
lostmsu, line 477: A property would seem more natural.

src/runtime/runtime.cs
```diff
@@ -473,6 +473,19 @@ private static void ResetPyMembers()
...
+        {
+            if (_none == null)
+            {
+                var result = Runtime.PyNone;
+                Runtime.XIncref(result);
+                _none= new PyObject(result);
```
lostmsu, line 483: I do not think there's much point in caching result.

But if you do, you must deal with:
1. When runtime is shutdown and restarted, this value should be reinitialized.
2. On the .NET side writing to `_none` must be properly synchronized. E.g. you need a lock or an atomic swap.

Again, however, I think you can just not cache the value.

henon: ok, I best remove the caching

henon:
You might wanna squash this PR into a single commit.

filmor:
Thank you :)

PR #1138, opened 2020-05-10
Python 3.8

filmor:
Adds Python 3.8 to the CI and fixes the remaining issue by assigning `__classcell__` the created class if it exists. This is required as for the case of subclassing a C# class in Python, we (currently) don't initialise the type via `type_new` (`== PyType_Type.tp_new`) but instead generate it manually. The added code is done at the end of `type_new` and is required from Python 3.8 onwards (see command below).

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -23,10 +23,13 @@ environment:
       BUILD_OPTS: --xplat
     - PYTHON_VERSION: 3.7
       BUILD_OPTS: --xplat
+    - PYTHON_VERSION: 3.8
+      BUILD_OPTS: --xplat
     - PYTHON_VERSION: 2.7
     - PYTHON_VERSION: 3.5
     - PYTHON_VERSION: 3.6
     - PYTHON_VERSION: 3.7
+    - PYTHON_VERSION: 3.8
 
 matrix:
     allow_failures:
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -15,27 +15,21 @@ environment:
     CODECOV_ENV: PYTHON_VERSION, PLATFORM
 
   matrix:
+    - PYTHON_VERSION: 3.8
+      BUILD_OPTS: --xplat
+    - PYTHON_VERSION: 3.7
+      BUILD_OPTS: --xplat
+    - PYTHON_VERSION: 3.6
+      BUILD_OPTS: --xplat
+    - PYTHON_VERSION: 3.5
+      BUILD_OPTS: --xplat
     - PYTHON_VERSION: 2.7 
       BUILD_OPTS: --xplat
-    - PYTHON_VERSION: 3.5
-      BUILD_OPTS: --xplat
-    - PYTHON_VERSION: 3.6
-      BUILD_OPTS: --xplat
-    - PYTHON_VERSION: 3.7
-      BUILD_OPTS: --xplat
     - PYTHON_VERSION: 3.8
-      BUILD_OPTS: --xplat
+    - PYTHON_VERSION: 3.7
+    - PYTHON_VERSION: 3.6
+    - PYTHON_VERSION: 3.5
     - PYTHON_VERSION: 2.7
-    - PYTHON_VERSION: 3.5
-    - PYTHON_VERSION: 3.6
-    - PYTHON_VERSION: 3.7
-    - PYTHON_VERSION: 3.8
-
-matrix:
-    allow_failures:
-      - PYTHON_VERSION: 3.4
-        BUILD_OPTS: --xplat
-      - PYTHON_VERSION: 3.4
 
 init:
   # Update Environment Variables based on matrix/platform
--- a/src/runtime/runtime.cs
+++ b/src/runtime/runtime.cs
@@ -1966,6 +1966,15 @@ namespace Python.Runtime
         [DllImport(_PythonDll, CallingConvention = CallingConvention.Cdecl)]
         internal static extern void PyErr_Print();
 
+        //====================================================================
+        // Cell API
+        //====================================================================
+
+        [DllImport(_PythonDll, CallingConvention = CallingConvention.Cdecl)]
+        internal static extern NewReference PyCell_Get(IntPtr cell);
+
+        [DllImport(_PythonDll, CallingConvention = CallingConvention.Cdecl)]
+        internal static extern int PyCell_Set(IntPtr cell, IntPtr value);
 
         //====================================================================
         // Miscellaneous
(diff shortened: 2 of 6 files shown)


filmor:
The actual problem to fix is this (from https://docs.python.org/3/reference/datamodel.html#creating-the-class-object):

> CPython implementation detail: In CPython 3.6 and later, the __class__ cell is passed to the metaclass as a __classcell__ entry in the class namespace. If present, this must be propagated up to the type.__new__ call in order for the class to be initialised correctly. Failing to do so will result in a RuntimeError in Python 3.8.

I'll look into it tomorrow.

filmor:
Okay, so one bug left to fix: On Linux with .NET Core and Python 3.8 some of the embedding tests fail, will have a look.

filmor:
@lostmsu Tests are now passing for all cases as far as I can see, can you have a look? I'll update the Changelog and then this is good to go.

appveyor.yml
```diff
@@ -15,24 +15,21 @@ environment:
...
+    - PYTHON_VERSION: 3.7
       BUILD_OPTS: --xplat
     - PYTHON_VERSION: 3.6
       BUILD_OPTS: --xplat
-    - PYTHON_VERSION: 3.7
-      BUILD_OPTS: --xplat
-    - PYTHON_VERSION: 2.7
     - PYTHON_VERSION: 3.5
-    - PYTHON_VERSION: 3.6
+      BUILD_OPTS: --xplat
```
lostmsu, line 25: NIT: can you minimize changes here and in .travis.yml if possible?

src/runtime/runtime.cs
```diff
@@ -154,11 +148,6 @@ public class Runtime
         /// </summary>
         public static MachineType Machine { get; private set; }/* set in Initialize using python's platform.machine */
 
-        /// <summary>
-        /// Gets the machine architecture as reported by python's platform.machine().
-        /// </summary>
-        public static string MachineName { get; private set; }
-
         internal static bool IsPython2 = pyversionnumber < 30;
         internal static bool IsPython3 = pyversionnumber >= 30;
```
lostmsu, line 152: If you are removing public members of Runtime, this makes this change breaking.

src/runtime/runtime.cs
```diff
@@ -1966,6 +1985,15 @@ internal static IntPtr PyMem_Realloc(IntPtr ptr, long size)
...
 
+        //====================================================================
+        // Cell API
+        //====================================================================
+
+        [DllImport(_PythonDll, CallingConvention = CallingConvention.Cdecl)]
+        internal static extern NewReference PyCell_Get(IntPtr cell);
+
+        [DllImport(_PythonDll, CallingConvention = CallingConvention.Cdecl)]
+        internal static extern int PyCell_Set(IntPtr cell, IntPtr value);
```
lostmsu, line 1996: NIT: you can use `BorrowedReference` for arguments, that take a `PyObject*` and don't steal a reference.

src/runtime/typemanager.cs
```diff
@@ -617,7 +625,7 @@ int MAP_ANONYMOUS
                         case OperatingSystemType.Linux:
                             return 0x20;
                         default:
-                            throw new NotImplementedException($"mmap is not supported on {Runtime.OperatingSystemName}");
+                            throw new NotImplementedException($"mmap is not supported on this operating system");
```
lostmsu, line 628: NIT: here and below you can pass the `Runtime.OperatingSystem`, since that is the value used in `switch`

src/runtime/runtime.cs
filmor: Meh, I'll include them (with an `Obsolete` flag and just as a stringified version of the enum, absolutely not worth the effort exposing the actual `platform` values).

lostmsu: @filmor should be fine as long as the values are compatible.

appveyor.yml
filmor: I can move them to a single separate commit, but the changes are important for my sanity. I don't want to see old, unsupported versions fly by in CI while I wait 15min for the actually interesting thing. The only change here is that I reordered the build matrix to list the versions in descending order.

filmor: Actually, they are already in different commits (adding Py3.8 and reordering).

src/runtime/runtime.cs
filmor: They are not. But the values returned from Python in that regard are also not very pretty, just have a look at our dictionaries used to map the to something sensible. I highly doubt it's worth the effort.

src/runtime/runtime.cs
```diff
@@ -118,16 +119,16 @@ public class Runtime
             { "Linux", OperatingSystemType.Linux },
         };
 
-        /// <summary>
-        /// Gets the operating system as reported by python's platform.system().
-        /// </summary>
-        public static OperatingSystemType OperatingSystem { get; private set; }
+        [Obsolete]
+        public static string OperatingSystemName => Enum.GetName(typeof(OperatingSystemType), OperatingSystem);
```
lostmsu, line 123: NIT: Is there a reason why you can't just do NIT: is there a reason why can't you just do `OperatingSystem.ToString()`?

src/runtime/interop37m.cs
```diff
@@ -0,0 +1,150 @@
+
+// Auto-generated by geninterop.py.
+// DO NOT MODIFY BY HAND.
+
+
+#if PYTHON37 && PYTHON_WITH_PYMALLOC
```
lostmsu, line 6: Was this file added by accident?

filmor: Yep, removed now.

src/runtime/runtime.cs
filmor: Just incompetence, I simplified it.

src/runtime/typemanager.cs
```diff
@@ -280,6 +280,14 @@ internal static IntPtr CreateSubType(IntPtr py_name, IntPtr py_base_type, IntPtr
...
                 Runtime.PyDict_Update(cls_dict, py_dict);
 
+                // Update the __classcell__ if it exists
+                IntPtr cell = Runtime.PyDict_GetItemString(cls_dict, "__classcell__");
+                if (cell != IntPtr.Zero)
+                {
+                    Runtime.PyCell_Set(cell, py_type);
+                    Runtime.PyDict_DelItemString(cls_dict, "__classcell__");
+                }
+
```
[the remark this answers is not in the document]
lostmsu: NVM, I found the code to be replica of https://github.com/python/cpython/blob/145bf269df3530176f6ebeab1324890ef7070bf8/Objects/typeobject.c#L2844

Issue #1116, opened 2020-04-16
Copy .NET to numpy array: python 3 breaks constructor IntPtr.__overloads__[int](np_arr.__array_interface__['data'][0])


greg814:
### Environment

Pythonnet version: 2.4.0 (in both python versions)

Python version:
- code working in Python 2.7.17 (v2.7.17:c2f86d86e6, Oct 19 2019, 21:01:17) [MSC v.1500 64 bit (AMD64)] on win32
- code not working in Python 3.7.7 (tags/v3.7.7:d7c567b08f, Mar 10 2020, 10:41:24) [MSC v.1900 64 bit (AMD64)] on win32

-   Operating System: Windows 10 Professional 64-bit

### Details

Trying to copy a .NET array to a numpy array using Marshal as in the example on the [mailing list](https://mail.python.org/pipermail/pythondotnet/2014-May/001525.html):

```
import clr
from System import Array, Double, IntPtr, Random
from System.Runtime.InteropServices import Marshal
import numpy as np

r = Random()
src = Array.CreateInstance(Double, 10000000)
for i in xrange(len(src)): src[i] = r.NextDouble()

from System.Runtime.InteropServices import Marshal
dest = np.empty(len(src))
dest_ptr = IntPtr.__overloads__[int](dest.__array_interface__['data'][0])
Marshal.Copy(src, 0, dest_ptr, len(src))
```

was working fine in python 2 but in 3 results in

```
Traceback (most recent call last):
  File "test.py", line 12, in <module>
    dest_ptr = IntPtr.__overloads__[int](dest.__array_interface__['data'][0])
TypeError: no constructor matches given arguments
```

TODO
How can the changed [python 3 int](https://docs.python.org/3.0/whatsnew/3.0.html#integers) be used in this constructor?

greg814:
Did you see this [parsing workaround on stackoverflow](https://stackoverflow.com/a/72912639/13305034)? Funny you would comment here 2 hrs after the so post...

filmor:
It's related, @jhe-pf commented on gitter and I started to look into it :)

I have identified the problem, I just need to find a fix.

lostmsu:
@filmor as mentioned on StackOverflow, `__overloads__[Int64]` should work. The overloads fail with `int` because `int` is not a .NET type.

filmor:
No, check the linked PR. The issue is that primitive types are initialised differently, which in our case is implemented as a conversion, but not for the ptr types.

[...]

PR #1143, opened 2020-05-14
Improve "No constructor matches given arguments" message with more details

lostmsu:
Similar to #900, but for constructors (reuses the same code moved around)

This adds more details to "No constructor matches given arguments" message.

#### Was

> No constructor matches given arguments

#### Now

> No constructor matches given arguments for IntPtr: (int, list)

Does not close any issue, but is related to: #811, #265, #1116

### DIFF
--- a/src/runtime/methodbinder.cs
+++ b/src/runtime/methodbinder.cs
@@ -599,6 +599,40 @@ namespace Python.Runtime
             return Invoke(inst, args, kw, info, null);
         }
 
+        protected static void AppendArgumentTypes(StringBuilder to, IntPtr args)
+        {
+            long argCount = Runtime.PyTuple_Size(args);
+            to.Append("(");
+            for (long argIndex = 0; argIndex < argCount; argIndex++)
+            {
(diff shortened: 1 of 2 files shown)


src/runtime/methodbinder.cs
```diff
@@ -599,6 +599,40 @@ internal virtual IntPtr Invoke(IntPtr inst, IntPtr args, IntPtr kw, MethodBase i
...
+                {
+                    var type = Runtime.PyObject_Type(arg);
+                    if (type != IntPtr.Zero)
+                    {
+                        try
+                        {
+                            var description = Runtime.PyObject_Unicode(type);
+                            if (description != IntPtr.Zero)
+                            {
+                                to.Append(Runtime.GetManagedString(description));
```
filmor, line 619: If `GetManagedString` fails, you leak `description`

filmor: Ah, but I see that is not your doing but was already in the code...

lostmsu: @filmor it is still my doing :-D

PR #1158, opened 2020-06-14
Drop Python 2 support

filmor:

Drops support for and all references to Python 2.

Can't find it right now, but I think we had an issue discussing this.

### Any other comments?

-/-

### Checklist


src/runtime/runtime.cs
```diff
@@ -43,39 +43,24 @@ public class Runtime
 #error You must define either UCS2 or UCS4!
 #endif
 
-        // C# compiler copies constants to the assemblies that references this library.
-        // We needs to replace all public constants to static readonly fields to allow
-        // binary substitution of different Python.Runtime.dll builds in a target application.
-
-        public static string pyversion => _pyversion;
```
koubaa, line 50: @filmor I think a unit test depends on these public properties

koubaa:
@filmor have you tried just changing the runtime and leaving the tests as-is for a first pass?  I looked through the runtime and didn't see a mistake so I think I would start by ruling out problems with the test updating (which can easily be done in a followon branch)

src/runtime/runtime.cs
filmor: Which one? I was used in a few places, but only to check for Python 2 vs 3.

filmor:
Not needed. The failures before were due to one place where I forgot to replace `text_type` by `str` and instead removed it entirely, so instead `repr` was called which was buggy at the time. I fixed that (and a few other missing replacements of `long` and a missing import of `operator`) and now tests are passing. I'm not a big fan of doing multiple PRs for the same "change".

src/runtime/runtime.cs
koubaa: @filmor you already removed them actually from TestPythonEngineProperties.cs
Seems like this might require an entry in the changelog since its a breaking change technically

setup.py
```diff
@@ -255,17 +255,11 @@ def build_extension(self, ext):
 
         # Up to Python 3.2 sys.maxunicode is used to determine the size of
         # Py_UNICODE, but from 3.3 onwards Py_UNICODE is a typedef of wchar_t.
-        # TODO: Is this doing the right check for Py27?
```
koubaa, line 258: the check for python2 on line 153 can also be removed

src/runtime/runtime.cs
filmor: Yeah, I could do that, but `pyversion` etc. where never really consumables in the first place. I'd like to move the whole `Runtime` to `internal` as a single Changelog entry, no need to do this for every single component.

lostmsu: @filmor moving `Runtime` to `internal` deserves its own tracking issue.

src/tests/test_exceptions.py
```diff
@@ -278,7 +277,7 @@ def test_python_compat_of_managed_exceptions():
 
     e = OverflowException(msg)
     assert str(e) == msg
-    assert text_type(e) == msg
+    assert str(e) == msg
```
lostmsu, line 280: This line is identical to the one above it

src/tests/_compat.py
```diff
@@ -5,70 +5,11 @@
...
-    from itertools import izip, imap
-    cmp = cmp
-    map = imap
-    range = xrange
-    zip = izip
-
+import _thread as thread
+DictProxyType = type(object.__dict__)
 
 def check_output(*args, **kwargs):
```
lostmsu, line 12: Can we inline this function and get rid of the file entirely?

src/runtime/runtime.cs
filmor: Agreed: https://github.com/pythonnet/pythonnet/issues/1167

filmor:
The coverage is always against master, it will not repeatedly complain.

PR #1170, opened 2020-06-23
remove remoting block for .NET standard

koubaa:

Remove remoting support

Removes some dependency on .NET framework

### Any other comments?

...

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -57,7 +57,6 @@ This version improves performance on benchmarks significantly compared to 2.3.
     `Obsolete`, should never have been `public` in the first place. They also
     don't necessarily return a result that matches the `platform` module's.
 -   Unconditionally depend on `pycparser` for the interop module generation
--   Removed .NET remoting support
 
 ### Fixed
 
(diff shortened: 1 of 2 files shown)


lostmsu:
If this is the only line where it is used, can we replace it with runtime feature detection instead of completely removing a feature?

E.g.
```csharp
static readonly Func<object, bool> IsTransparentProxy = GetIsTransparentProxy();

static bool Never(object _) => false;

static Func<object, bool> GetIsTransparentProxy() {
  var remoting = typeof(int).Assembly.GetType("System.Runtime.Remoting.RemotingServices");
  if (remoting is null) return Never;

  var isProxy = remoting.GetMethod("IsTransparentProxy", new []{typeof(object)});
  if (isProxy is null) return Never;

  return (Func<object, bool>)Delegate.CreateDelegate(
    typeof(Func<object, bool>), isProxy,
    throwOnBindFailure: true);
}
```

P.S. @filmor Travis CI results are not being picked up...

koubaa:
@lostmsu its a good idea but this isn't a tested feature and I don't want to build a test for it.  Are you confident your code will work without tests?

lostmsu:
@koubaa the calls to this method probably have some coverage, so if it crashes, we should see in CI. The feature itself might be broken with this, but IMO it is better, than removing it completely.

koubaa:
@lostmsu as long as I'm not signing up for testing remoting, I am ok with that.  I'll push an update with your change and remove the changelog entry

lostmsu:
@koubaa , yeah, let's do that.

Issue #935, opened 2019-08-06
TypeError when calling a method with an enum argument


vladt3:
### Environment

-   Pythonnet version:  2.4.0
-   Python version: 3.6
-   Operating System: Windows10

### Details

-   calling a method that uses C# enum as an argument with the enum fails with TypeError: No method matches given arguments for XXX

The enum comes in Python as an int, and since the method signature is not int, it fails. 

I cannot see how to force the typecast, or whether I could somehow force the method call directly (__overloads__ lists the method, but I can't figure out whether I can call directly).

Searching google did not return anything recent, although some older comments (2017/18 IIRC) indicated there was a fix for this?

Cronan:
This kind of thing should work, I believe ...

```python
from Tests import TestClass, TestEnum
a = TestClass()
a.EnumMethod[TestEnum](TestEnum.FirstValue)
```

vladt3:
Thanks, it does indeed. Maybe there should be some FAQ that would cover things like these?

Cronan:
> Thanks, it does indeed. Maybe there should be some FAQ that would cover things like these?

Thanks, that's a good idea

PR #1179, opened 2020-07-06
Allow conversion of number to enum type when underlying type is assignable

slide:

In .NET you can do the following:

```
enum FooEnum {
    One = 1,
    Three = 3,
    Four = 4,
}

FooEnum x = (FooEnum)2;
```

This patch allows this from Python code as well.

Possibly #935

### Any other comments?

I would definitely like eyes on this, I am possibly missing something :-)

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ This document follows the conventions laid out in [Keep a CHANGELOG][].
 
 ### Changed
 -   Drop support for Python 2
+-   Added support for conversion from numerical types to enum in argument passing
 
 ### Fixed
 
--- a/src/runtime/converter.cs
+++ b/src/runtime/converter.cs
@@ -893,6 +893,16 @@ namespace Python.Runtime
                 return true;
             }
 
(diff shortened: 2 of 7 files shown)


lostmsu:
@slide in .NET you can't just assign a numeric value to an enum. You have to make an explicit cast for a good reason. The most pythonic way I see this could work is `FooEnum(42)`. Implicit conversion is a bad idea.

slide:
That's true, it does require the typecast. This only allows assignment/coercion when the underlying type is assignable, is that still too lenient?

lostmsu:
@slide oh, I checked the code of `ToEnum` and it seems problematic on its own. I think what you will find is that `IsAssignableFrom` will succeed in most cases. I actually think Python.NET should have required an explicit cast from the beginning, and should not have converted enum values to Python `int`, which we have an opportunity to fix given 3.0 is a major version bump. I'd definitely take a PR, that would make that change.

Summoning @filmor into conversation.

slide:
Ok, thanks, I'll see if I can work on something like that.

PR #1191, opened 2020-07-30
match python type T to Nullable<T> in C#

hatami57:
When a method parameter in C# is defined as `Nullable` like `int?` and you call it in python by like a literal int number, it does not match with `int?` and could not bind the method, at least in .NETStandard2.0. So this update is telling the `MethodBinder` class that a primitive value (with the type of `T`) in python can be matched with the corresponding primitive type (`T`) in C# or the `Nullable<T>` type too.

...

...

### Any other comments?

...

### Checklist


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/pythonnet/pythonnet?pullRequest=1191) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/pythonnet/pythonnet?pullRequest=1191) before we can accept your contribution.<br/><br/>:x: hatami57 [sign now](https://cla.dotnetfoundation.org/pythonnet/pythonnet?pullRequest=1191)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/pythonnet/pythonnet?pullRequest=1191) it.</sub>

lostmsu:
Honestly, I am not even sure we need `TryComputeArgumentType`. The only thing it provides right now is resolving between `Foo(int)` and `Foo(double)`, which should be handled in a different way altogether: the current version looks like it won't let you call `Foo(double, string)` doing `Foo(1, "42")` in Python.

hatami57:
Anyway I couldn't call a C# class constructor say `Foo(int, int, double?)` by calling `Foo(2, 5, 0.25)` in Python. So with this minor change in the code, it just works.

For now, I think `TryComputeArgumentType` do an important job in the library and it needs some improvements. If somehow the logic needed to be moved in other places of the code, still this minor change needed to be applied there...

lostmsu:
@hatami57 can you try passing a `Nullable(0.25)` or `Nullable[Double](0.25)` as a workaround?

hatami57:
> @hatami57 can you try passing a `Nullable(0.25)` or `Nullable[Double](0.25)` as a workaround?

As far as I remember, I did try it but it didn't work...

filmor:
@hatami57 Did you try what lostmsu suggested again on 3.0?

PR #1195, opened 2020-08-01
Pybuffer

koubaa:
Rebase of https://github.com/pythonnet/pythonnet/pull/980/files 
...

...

### Any other comments?

...

### Checklist

### DIFF
--- a/src/runtime/pybuffer.cs
+++ b/src/runtime/pybuffer.cs
@@ -88,7 +88,7 @@ namespace Python.Runtime
         /// </summary>
         public static long SizeFromFormat(string format)
         {
-            if (Runtime.PyVersion.Minor < 9)
+            if (Runtime.PyVersion < new Version(3,9))
                 throw new NotSupportedException("SizeFromFormat requires at least Python 3.9");
             return (long)Runtime.PyBuffer_SizeFromFormat(format);
         }
@@ -111,7 +111,7 @@ namespace Python.Runtime
         {
             if (disposedValue)
                 throw new ObjectDisposedException(nameof(PyBuffer));
-            if (Runtime.PyVersion.Minor < 7)
+            if (Runtime.PyVersion < new Version(3, 7))
                 throw new NotSupportedException("GetPointer requires at least Python 3.7");
             return Runtime.PyBuffer_GetPointer(ref _view, indices.Select(x => (IntPtr)x).ToArray());
         }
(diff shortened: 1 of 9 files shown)


koubaa:
@filmor I took this one

src/runtime/runtime.cs
```diff
@@ -147,6 +147,19 @@ public class Runtime
...
 
+        internal static int pyversionnumber
+        {
+            get
+            {
+                var versionString = PythonEngine.Version;
+                //"3.0a5+ (py3k:63103M, May 12 2008, 00:53:55) \n[GCC 4.2.3]"
+
+                //we only support python 3 so we just need to check the third character
+                return 30 + Convert.ToInt32(versionString[2]);
```
filmor, line 158: This seems very brittle to me, there must be a better solution. Also, we don't support any Python version before 3.5, so checks for that can go.

koubaa: @filmor I could use sys.version_info but I'd rather not run python code.  Let me look for a C API equivalent

koubaa: @filmor see now

src/runtime/runtime.cs
```diff
@@ -147,6 +147,18 @@ public class Runtime
 
         private static PyReferenceCollection _pyRefs = new PyReferenceCollection();
 
+        internal static int pyversionnumber
```
lostmsu, line 150: Can you use `System.Version` here? `major * 10 + minor` led to compatibility issue in many products before. This is not going to work for 3.10

koubaa: @lostmsu great point.  Updated

koubaa:
@lostmsu please review

src/runtime/pybuffer.cs
```diff
@@ -88,7 +88,7 @@ private static char OrderStyleToChar(BufferOrderStyle order, bool eitherOneValid
         /// </summary>
         public static long SizeFromFormat(string format)
         {
-            if (Runtime.pyversionnumber < 39)
+            if (Runtime.PyVersion.Minor < 9)
```
lostmsu, line 91: You must compare the full version. E.g. `Runtime.PyVersion < new Version(3, 9)`

koubaa: @lostmsu done

Issue #1033, opened 2020-01-22
Use PyType_FromSpec instead of explicit PyType building


filmor:
Python 3 (from 3.3 onwards) supports a version-independent style to create type objects, namely [PyType_FromSpec](https://docs.python.org/3/c-api/type.html#c.PyType_FromSpec). Since we routinely have to update the `interop*.cs` files right now, and this is our only version-dependency once the new DLL loading lands, we should refactor things to take advantage of this.

PR #1196, opened 2020-08-01
Start to implement PyType_FromSpec type approach

koubaa:
Begin to address #1033.  Just beginning with this but wanted to submit a PR to solicit input about direction/approach.
...

#1033 
...

### Any other comments?

...

### Checklist

### DIFF
--- a/src/runtime/typemanager.cs
+++ b/src/runtime/typemanager.cs
@@ -533,17 +533,6 @@ namespace Python.Runtime
         [StructLayout(LayoutKind.Sequential, CharSet = CharSet.Ansi)]
         internal class PyTypeSpecOffset
         {
-            static PyTypeSpecOffset()
-            {
-                Type type = typeof(PyTypeSpecOffset);
-                FieldInfo[] fi = type.GetFields();
-                int size = IntPtr.Size;
-                for (int i = 0; i < fi.Length; i++)
-                {
-                    fi[i].SetValue(null, i * size + TypeOffset.ob_size);
-                }
-            }
-
             public static IntPtr AllocPyTypeSpec(string typename, int obSize, int obFlags, IntPtr slotsPtr)
             {
                 byte[] ascii = System.Text.Encoding.ASCII.GetBytes(typename);
--- a/src/runtime/typemanager.cs
+++ b/src/runtime/typemanager.cs
@@ -202,6 +202,17 @@ namespace Python.Runtime
             return type;
         }
 
+        static PY_TYPE_SLOT InitializeSlot(int slotNumber, MethodInfo method)
+        {
+            var thunk = Interop.GetThunk(method);
+            return new PY_TYPE_SLOT { slot = slotNumber, func = thunk.Address};
+        }
+
+        static PY_TYPE_SLOT InitializeSlot(int slotNumber, IntPtr thunk)
+        {
+            return new PY_TYPE_SLOT { slot = slotNumber, func = thunk.Address };
+        }
+
         static void InitializeSlot(IntPtr type, int slotOffset, MethodInfo method)
         {
             var thunk = Interop.GetThunk(method);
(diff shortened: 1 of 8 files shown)


koubaa:
@filmor FYI.  My next step is to try to change TypeManager.CreateType.

filmor:
Thanks for working on this! This could simplify compatibility with different Python versions a lot :+1:

src/runtime/typemanager.cs
```diff
@@ -422,6 +434,198 @@ internal static IntPtr BasicSubType(string name, IntPtr base_, Type impl)
             return type;
         }
 
+        internal class TypeSlots
```
lostmsu, line 437: This should be an `enum`? Enums also have a built-in conversion from `string`.

src/runtime/typemanager.cs
```diff
@@ -422,6 +434,198 @@ internal static IntPtr BasicSubType(string name, IntPtr base_, Type impl)
...
+
+        [StructLayout(LayoutKind.Sequential)]
+        internal struct PY_TYPE_SLOT
+        {
+            public long slot; //slot id, from typeslots.h
+            public IntPtr func; //function pointer of the function implementing the slot
+        }
+
+        [StructLayout(LayoutKind.Sequential, CharSet = CharSet.Ansi)]
+        internal class PyTypeSpecOffset
```
lostmsu, line 537: Ideally, these extra types should go into their own files.

src/runtime/typemanager.cs
```diff
@@ -422,6 +434,198 @@ internal static IntPtr BasicSubType(string name, IntPtr base_, Type impl)
...
+                int size = IntPtr.Size;
+                for (int i = 0; i < fi.Length; i++)
+                {
+                    fi[i].SetValue(null, i * size + TypeOffset.ob_size);
+                }
+            }
+
+            public static IntPtr AllocPyTypeSpec(string typename, int obSize, int obFlags, IntPtr slotsPtr)
+            {
+                byte[] ascii = System.Text.Encoding.ASCII.GetBytes(typename);
```
lostmsu, line 552: .NET supports Unicode type names. Not sure about Python here.

src/runtime/typemanager.cs
```diff
@@ -422,6 +434,198 @@ internal static IntPtr BasicSubType(string name, IntPtr base_, Type impl)
...
+            static PyTypeSpecOffset()
+            {
+                Type type = typeof(PyTypeSpecOffset);
+                FieldInfo[] fi = type.GetFields();
+                int size = IntPtr.Size;
+                for (int i = 0; i < fi.Length; i++)
+                {
+                    fi[i].SetValue(null, i * size + TypeOffset.ob_size);
+                }
+            }
```
lostmsu, line 548: As long as this structure does not differ between Python versions (which I am hoping it is not designed to), you can just set these fields to be constant or `static readonly` values, and not use reflection at all.

src/runtime/typemanager.cs
```diff
@@ -422,6 +434,198 @@ internal static IntPtr BasicSubType(string name, IntPtr base_, Type impl)
...
+            {
+                byte[] ascii = System.Text.Encoding.ASCII.GetBytes(typename);
+
+                //This approach is the same as the one in interop.cs for AllocModuleDef
+                //allocate the size of the struct (which is given by the value of the last
+                //static member and enough space to hold to typename as a char buffer.  The
+                //amount of space needed is the length of the string and the null terminator
+                //char* name member will simply point to the position of the buffer.
+                int size = name_value + ascii.Length + 1;
+                IntPtr ptr = Marshal.AllocHGlobal(size);
```
lostmsu, line 560: NIT: name this `specPtr`?

src/runtime/typemanager.cs
```diff
@@ -422,6 +434,198 @@ internal static IntPtr BasicSubType(string name, IntPtr base_, Type impl)
...
+            public static int basicsize = 0;
+            public static int itemsize = 0;
+            public static int flags = 0;
+            public static int slots = 0;
+
+            private static int name_value = 0;
+        }
+
+
+        internal static IntPtr AllocateTypeObject(string name, PY_TYPE_SLOT[] type_slots)
```
lostmsu, line 584: NIT: this does not just allocate. It actually creates the instance representing the type.

src/runtime/typemanager.cs
```diff
@@ -680,6 +884,79 @@ internal static void InitializeNativeCodePage()
...
+                        continue;
+                    }
+
+                    if (seen.Contains(name))
+                    {
+                        continue;
+                    }
+
+                    typeslots.Add(InitializeSlot(TypeSlots.getSlotNumber(name), method));
+                    seen.Add(name);
```
lostmsu, line 921: I vaguely remember seeing this code elsewhere. You should remove it from the original location.

src/runtime/typemanager.cs
koubaa: @lostmsu I think its possible.  Maybe we should use UTF8 here - I should add a test to see for sure

src/runtime/typemanager.cs
koubaa: @lostmsu great point.  I changed it.

src/runtime/typemanager.cs
koubaa: @lostmsu its in this same file.  I'll cleanup once I get things working

src/runtime/typemanager.cs
koubaa: @lostmsu I think IntPtr.Size depends on the platform.  I'm getting crashes in PyType_FromSpec and I suspect the integer fields are only 4 bytes so using IntPtr.Size to increment each field in this struct would be wrong.  If my suspicion turns out to be correct then I wonder how things are working in the other struct layout objects which assume each field is a pointer.  

Either way my code should change..  I'll keep you posted

src/runtime/typemanager.cs
koubaa: @lostmsu I intend to do it after I get things working as part of the final cleanup

src/runtime/typemanager.cs
lostmsu: @koubaa you can use `IntPtr.Size` in `static readonly` fields.

BTW, does type spec allow us to drop `TypeOffset` class?

src/runtime/typemanager.cs
```diff
@@ -422,6 +434,195 @@ internal static IntPtr BasicSubType(string name, IntPtr base_, Type impl)
...
+
+        private static int getSlotNumber(string methodName)
+        {
+            return (int)Enum.Parse(typeof(TypeSlots), methodName);
+        }
+
+        [StructLayout(LayoutKind.Sequential)]
+        internal struct PY_TYPE_SLOT
+        {
+            public long slot; //slot id, from typeslots.h
```
lostmsu, line 529: Should be of type `TypeSlots` here. Correspondingly, `TypeSlots` should be based on `long`.

src/runtime/typemanager.cs
koubaa: @lostmsu Good point.  If the arithmetic isn't too obtuse I might go that way

Yes, we can drop TypeOffset once all usages of it has migrated to PyType_FromSpec, which is my intention before landing this change

src/runtime/typemanager.cs
koubaa: @lostmsu good point, I'll change it

src/runtime/typemanager.cs
```diff
@@ -434,7 +434,7 @@ internal static IntPtr BasicSubType(string name, IntPtr base_, Type impl)
             return type;
         }
 
-        enum TypeSlots
+        internal enum TypeSlots : long
```
lostmsu, line 437: Actually, according to [C API docs](https://docs.python.org/3/c-api/type.html#c.PyType_Slot) this type is C's `int` type, which is 32 bit on [most systems](https://www.ibm.com/support/knowledgecenter/en/SSFKSJ_9.0.0/com.ibm.mq.ref.dev.doc/q104610_.htm).

src/runtime/typemanager.cs
```diff
@@ -554,27 +554,30 @@ public static IntPtr AllocPyTypeSpec(string typename, int obSize, int obFlags, I
...
+                return specPtr;
             }
 
-            public static int name = 0;
-            public static int basicsize = 0;
-            public static int itemsize = 0;
-            public static int flags = 0;
-            public static int slots = 0;
+            public static int name = 16;
+            public static int basicsize = 24;
```
lostmsu, line 575: Hint: you can do `static basicsize = name + IntPtr.Size;`

src/runtime/typemanager.cs
```diff
@@ -422,6 +434,198 @@ internal static IntPtr BasicSubType(string name, IntPtr base_, Type impl)
...
+            int structSize = Marshal.SizeOf(typeof(PY_TYPE_SLOT));
+            GCHandle pinnedArray = GCHandle.Alloc(type_slots, GCHandleType.Pinned);
+
+            //Well, this will leak.  Maybe pinnedArray should be added as a member to managedtype.
+            IntPtr slotsPtr = pinnedArray.AddrOfPinnedObject();
+            //pinnedArray.Free(); //at some point
+
+            //create a type from the spec and return it.
+            IntPtr specPtr = PyTypeSpecOffset.AllocPyTypeSpec(name, obSize, obFlags, slotsPtr);
+            IntPtr typePtr = Runtime.PyType_FromSpec(specPtr);
```
koubaa, line 598: @filmor @lostmsu this part is driving me crazy. I fixed some offset issues but this one remains.  In my debugging session ( I figured out how to debug both CPython C code and pythonnet in the same session!)  I observe this:
image

which seems like the slotPtr is correctly assigned in the right offset. stepping into python code, I then see this:
image

All of the values are right except for slots.  somehow it gets a totally different value in slots.  I built/am debugging python 3.8 where the definition of PyType_Spec is:
image

If I manually modify the value of spec->slots in the immediate window before CPython uses spec it seems to be fine.

Any tips would be very welcome here.

lostmsu: Well, seeing the CPython code here reminds me, that since they intend this to be a stable API, you don't need offsets at all. Why don't you just make a C# struct with the same layout and use it instead of untyped pointers and `Marshal` stuff?

E.g.
```csharp
[StructLayout(LayoutKind.Sequential)]
unsafe struct PyTypeSpec {
  public IntPtr Name;
  public int BasicSize;
  public int ItemSize;
  public FlagsEnum Flags;
  public PyTypeSlot* Slots;
}
```

lostmsu: BTW, `PyType_FromSpec` should return `NewReference`

koubaa: @lostmsu I didn't know I had this option, it looks much easier.  I'll try it.

koubaa: @lostmsu the compiler doesn't allow me to pass ref spec into PyType_FromSpec

lostmsu: @koubaa you have to change the PInvoke signature to have `ref` too.

lostmsu: @koubaa actually, if `PyType_FromSpec` copies the contents of the spec, you can even change `Name` to be `string` and `Slots` to be an array instead of a  pointer, adding corresponding `[MarshalAs(...)]`. Probably also drop `ref` on the parameter, and change the `TypeSpec` to be a class.

koubaa: @lostmsu I changed it to use a struct and for the PInvoke signature to use ref.  I didn't go as far as MarshalAs but that should be doable too.  Now I no longer see a crash here.  Thanks!

koubaa:
@filmor I don't think this alone will rid us of TypeOffset.  tp_dict for instance does not appear to be covered in PyType_FromSpec and we use that offset in many places.

filmor:
This is nevertheless a huge step forward. If we can minimise the offsets in active use, we can probably maintain the respective offset info with a small script. By the way, this PEP tracks more efforts that will help us in this regard: https://www.python.org/dev/peps/pep-0620/

koubaa:
@filmor @lostmsu I'm getting an error that I am puzzled by.

When I run this code
```
import clr
import System
class Sub(System.Exception):
 pass

x=Sub()
```
or even replace the last line by
`x = Sub.__new__(Sub)`

It crashes inside of CPython with this message:
image

The last place in pythonnet when it reaches this point is in the CLRObject constructor at this line 
`IntPtr dict = Marshal.ReadIntPtr(py, ObjectOffset.TypeDictOffset(tp));`

I've been debugging this for several hours now and I wish I could say I gained some more insight but I really haven't.  What I did find is that the System.Exception MetaType construction seems to not go through the code path which I've modified.

Perhaps a more minor issue is this one - I am not sure if it is related:
```
>>> import System
__main__:1: DeprecationWarning: builtin type PropertyObject has no __module__ attribute
__main__:1: DeprecationWarning: builtin type EventObject has no __module__ attribute
__main__:1: DeprecationWarning: builtin type FieldObject has no __module__ attribute
```

koubaa:
> This is nevertheless a huge step forward. If we can minimise the offsets in active use, we can probably maintain the respective offset info with a small script. By the way, this PEP tracks more efforts that will help us in this regard: https://www.python.org/dev/peps/pep-0620/

@filmor I'll watch that PEP.  It looks like its for 3.10 which isn't that far away

lostmsu:
@koubaa the assert actually looks like it comes from .NET `Debug.Assert` call. Are you sure this is actually from CPython?

koubaa:
@lostmsu yes I'm sure.  I added `Debug.Asset` to try to better understand what's happening.  In fact, the dialog which comes from the latter looks different.

lostmsu:
@koubaa have you tried this code (`Sub` class) on `master` without your change? Does it work? If it does not, can you contribute a test (mark it disabled/inconclusive)? Then I can try to work on it outside of this PR.

The overload of `CreateType` you modified only works for metatypes defined in `Python.Runtime` (e.g. children of `ManagedType`). All other CLR types are handled via https://github.com/pythonnet/pythonnet/blob/61d1b7c22eb7cd68f991d35fcafbdf4dd90b7b19/src/runtime/typemanager.cs#L111 and that includes `System.Exception`, `System.Object` and all derived types.

It is probably related to my [shenanigans](https://stackoverflow.com/questions/60748800/deriving-from-arbitrary-python-class-with-c-api-and-tp-basicsize-of-object) with a similar problem. You might want to ask a similar question on StackOverflow about the relationship between `PyType_FromSpec` and `tp_basicsize`. The documentation is surprisingly scarce.

I just checked on Win64 Python's reflection of `System.Exception` takes 80 bytes == `ExceptionOffset.Size()`, which is as expected. However, the `Sub` class takes 88 bytes, which is not. We need an explanation as to why Python adds extra bytes when deriving from `System.Exception`. Depending on it we should either change the `Assert` condition, or change our code to ensure Python does not need to add anything when deriving from `System.Exception`.

I just checked and the reflected `System.Exception` type has `tp_dictoffset > 0`, so that's not it.

src/runtime/typemanager.cs
```diff
@@ -422,6 +434,163 @@ internal static IntPtr BasicSubType(string name, IntPtr base_, Type impl)
...
+            am_await = 77,
+            am_aiter = 78,
+            am_anext = 79,
+            tp_finalize = 80,
+        }
+
+        private static TypeSlots getSlotNumber(string methodName)
+        {
+            return (TypeSlots)Enum.Parse(typeof(TypeSlots), methodName);
+        }
```
amos402, line 524: I don't like to use string to get some constant information if it's not necessary, not to mention it use reflection for just getting the enum value.

koubaa:
@lostmsu Actually I discovered this issue from an existing test case: `test_constructors.py test_subclass_constructor()`

I thought I could do one overload at a time so it would be easier to bisect issues, do you think that it is wrong to expect this to work?

I'll look into `tp_basicsize`.  I have recently been poking around in CPython source so I'll try to see if I can figure out what the relationship is.

src/runtime/typemanager.cs
koubaa: @amos402 what do you propose?

amos402: Just use the enum value directly.

lostmsu:
The larger type object gets created for `Sub` here: https://github.com/pythonnet/pythonnet/blob/61d1b7c22eb7cd68f991d35fcafbdf4dd90b7b19/src/runtime/metatype.cs#L96

amos402:
> . We need an explanation as to why Python adds extra bytes

/* PyException_HEAD defines the initial segment of every exception class. */
#define PyException_HEAD PyObject_HEAD PyObject *dict;\
             PyObject *args; PyObject *traceback;\
             PyObject *context; PyObject *cause;\
             char suppress_context;

lostmsu:
@amos402 my understanding is that the class hierarchy looks like this: `Sub` -> `System.Exception` -> `builtins.BaseException`. The question is why the first `->` adds 8 bytes, not the last one. The last one works as intended.

I just checked on `master`, and the same assert actually fails there, so this is not caused by the @koubaa 's change in this PR, and can be evaluated separately.

koubaa:
@lostmsu on master, its a pythonnet assertion (you can ignore the dialogs and the python interpreter is still up).  With my branch, its a crash in native code.  Perhaps @amos402 's fix will also fix things in my branch, I can try to cherry-pick the relevant changes interop.cs and see what happens

koubaa:
@lostmsu @amos402 I confirm that with Amos's interop changes I do not have any crash anymore.   I submitted a PR to land amos' change separate from his soft shutdown branch and have that commit cherry-picked in this branch so that I can continue to work on this.

koubaa:
@lostmsu @amos402 I am looking at this again this weekend.  I need time to understand the changes from soft-shutdown here...

lostmsu:
@koubaa , if you are not actively working on this, I might pick it up over the holidays.

koubaa:
@lostmsu no I'm not actively working on it anymore.  I resolved some of the merge conflicts locally from the soft shutdown PR but haven't had much time to figure out how to get this working.

lostmsu:
Actually, I also decided against it for now, as I don't have any idea how to clear slots for types creates with `FromSpec`, if we remove `TypeOffset`.

lostmsu:
Closing for now, as this has been temporarily abandoned

PR #1205, opened 2020-08-19
Remove non-existent PInvoke functions

amos402:
Deprecated Python2 API:
* PyInstance_New
* PyInstance_NewRaw

Mere macros on C side:
* PyObject_CheckBuffer
* PyObject_GC_New (refer to `_PyObject_GC_New`)

Check method:
https://gist.github.com/amos402/fb41571856bc77a2bada8056e429b312

...

### Any other comments?

...

### Checklist

### DIFF
--- a/src/runtime/runtime.cs
+++ b/src/runtime/runtime.cs
@@ -847,12 +847,6 @@ namespace Python.Runtime
         internal static extern IntPtr PyCFunction_Call(IntPtr func, IntPtr args, IntPtr kw);
 
         [DllImport(_PythonDll, CallingConvention = CallingConvention.Cdecl)]
-        internal static extern IntPtr PyInstance_New(IntPtr cls, IntPtr args, IntPtr kw);
-
-        [DllImport(_PythonDll, CallingConvention = CallingConvention.Cdecl)]
-        internal static extern IntPtr PyInstance_NewRaw(IntPtr cls, IntPtr dict);
-
-        [DllImport(_PythonDll, CallingConvention = CallingConvention.Cdecl)]
         internal static extern IntPtr PyMethod_New(IntPtr func, IntPtr self, IntPtr cls);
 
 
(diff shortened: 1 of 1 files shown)


src/runtime/runtime.cs
```diff
@@ -1793,7 +1784,7 @@ internal static IntPtr PyType_GenericAlloc(IntPtr type, long n)
         [DllImport(_PythonDll, CallingConvention = CallingConvention.Cdecl)]
         internal static extern IntPtr _PyObject_GetDictPtr(IntPtr obj);
 
-        [DllImport(_PythonDll, CallingConvention = CallingConvention.Cdecl)]
+        [DllImport(_PythonDll, CallingConvention = CallingConvention.Cdecl, EntryPoint = "_PyObject_GC_New")]
         internal static extern IntPtr PyObject_GC_New(IntPtr tp);
```
filmor, line 1788: Is this one used anywhere?

amos402: None. 
Its implemetantion is 
```c
#define PyObject_GC_New(type, typeobj) \
                ( (type *) _PyObject_GC_New(typeobj) )
```
Just compare to the `PyObject_CheckBuffer`, it's easy to make it be a PInvoke function by specific the entry point.

filmor: If it's not used, just remove it.

amos402: Well, it's a not big deal though, but what's bad for declaring a valid api?

filmor: If we don't use it, it should not be included. The fewer functions we refer to, the easier it will be to keep compatible with different Python versions from the same DLL.

amos402: That's a point, although I don't think this API will be deleted in the future.
Removed it.

PR #1209, opened 2020-08-24
Convert projects to SDK style

filmor:
Another attempt, splitting up the conversion to a more modern style even further. This is just supposed to convert the project files and keep the existing loading for now (leaving the new loader for a separate PR based on #1109).


lostmsu:
Let's see the build results. If we take this, would need a tracking task for reestablishing testing on Mono.

setup.py
```diff
@@ -86,6 +112,13 @@ def run(self):
         dotnet_modules = [lib for lib in orig_modules if isinstance(lib, DotnetLib)]
         other_modules = [lib for lib in orig_modules if not isinstance(lib, DotnetLib)]
 
+        if dotnet_modules:
+            if os.path.isfile(CONFIGURED_PROPS):
+                self.announce("Already configured", level=distutils.log.INFO)
+            else:
+                self.announce("Writing configured.props...", level=distutils.log.INFO)
+                _write_configure_props()
```
lostmsu: Is this taking a lot of time? I feel like check for existing file might cause surprising behavior during repeated builds on the same machine.

filmor: It's not, I just want to add configuration options that I don't want to have to repeat all the time. If those are not needed, this step can also be simplified.

filmor:
There are two tests failing right now, both variants of `Python.EmbeddingTest.TestFinalizer.CollectBasicObject`. Any idea what could be the reason?

setup.py
filmor: Ah, now I understand that comment. Yeah, we should just always do this step. I'm going to work more on our build system once this lands.

filmor:
With the deactivated test, it seems to work now. Not entirely reliably, though. I had some failures in the finalisation.

pythonnet.15.sln
```diff
@@ -1,403 +0,0 @@
...
-		conda.recipe\bld.bat = conda.recipe\bld.bat
-		conda.recipe\meta.yaml = conda.recipe\meta.yaml
-		conda.recipe\README.md = conda.recipe\README.md
-	EndProjectSection
-EndProject
-Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Tools", "Tools", "{BC426F42-8494-4AA5-82C9-5109ACD97BD1}"
-	ProjectSection(SolutionItems) = preProject
-		tools\geninterop\geninterop.py = tools\geninterop\geninterop.py
-	EndProjectSection
-EndProject
```
lostmsu, line 49: Could you please keep these in the new .SLN?

pythonnet/.gitignore
```diff
@@ -0,0 +1,3 @@
+mono/
+netfx/
+runtime/
```
lostmsu, line 3: Wouldn't `runtime/` pattern generate a match for `/src/runtime/`?

Overall, I think all 3 are too generic.

setup.py
```diff
@@ -165,466 +26,206 @@ def _get_interop_filename():
...
-def _update_xlat_devtools():
-    global DEVTOOLS
-    if DEVTOOLS == "MsDev":
-        DEVTOOLS = "MsDev15"
-    elif DEVTOOLS == "Mono":
-        DEVTOOLS = "dotnet"
+    c = ET.SubElement(props, "ConfiguredConstants")
+    c.text = " ".join(defines)
 
+    ET.ElementTree(proj).write(CONFIGURED_PROPS)
```
lostmsu, line 60: Is it actually OK to create a new file in the current directory when installing a Python package?

setup.py
```diff
@@ -165,466 +26,206 @@ def _get_interop_filename():
...
-        for i, data_files in enumerate(self.data_files):
-            if isinstance(data_files, str):
-                self.data_files[i] = data_files[i].format(build_lib=build_lib)
+        orig_modules = self.distribution.ext_modules
+        dotnet_modules = [lib for lib in orig_modules if isinstance(lib, DotnetLib)]
+        other_modules = [lib for lib in orig_modules if not isinstance(lib, DotnetLib)]
+
+        if dotnet_modules:
+            if os.path.isfile(CONFIGURED_PROPS):
+                self.announce("Already configured", level=distutils.log.INFO)
```
lostmsu, line 110: Do these ever run in parallel? Couldn't there be a race condition around existence of this file?

setup.py
```diff
@@ -165,466 +26,206 @@ def _get_interop_filename():
...
-            bdist_wheel.bdist_wheel.finalize_options(self)
 
-        def run(self):
-            if self.xplat:
-                _update_xlat_devtools()
-            return bdist_wheel.bdist_wheel.run(self)
+class bdist_wheel_patched(bdist_wheel):
+    def finalize_options(self):
+        # Monkey patch bdist_wheel to think the package is pure even though we
+        # include DLLs
```
lostmsu, line 157: What is the purpose of this? "pure" in this context means arch-independent? Better state so in the comment then.

src/embed_tests/TestFinalizer.cs
```diff
@@ -47,6 +48,7 @@ private static bool FullGCCollect()
         }
 
         [Test]
+        [Ignore("Ignore temporarily")]
```
lostmsu, line 51: Please create a work item for newly disabled tests and assign to me.

src/monoclr/clrmod.c
```diff
@@ -1,70 +1,223 @@
...
+     */
+    mono_config_parse(NULL);
+
+    /* I can't use this call to run the main_thread_handler. The function
+     * runs it in another thread but *this* thread holds the Python
+     * import lock -> DEAD LOCK.
+     *
+     * mono_runtime_exec_managed_code(pn_args->domain, main_thread_handler,
+     *                                pn_args);
+     */
```
lostmsu, line 110: I think you can safely remove this comment. The official Mono doc says directly executing code is the way forward: http://docs.go-mono.com/index.aspx?link=xhtml%3adeploy%2fmono-api-embedding.html (search for `mono_runtime_exec_managed_code` on the page)

.github/workflows/main.yml
```diff
@@ -5,47 +5,23 @@ on: [ pull_request, push ]
 jobs:
   build-test:
     name: Build and Test
-    runs-on: ubuntu-16.04
+    runs-on: ubuntu-latest
```
lostmsu, line 8: I think we should explicitly specify version here. I suggest 20.04 for the latest LTS. Then we don't need to care about it until 2025.

pythonnet/.gitignore
filmor: This directory is a build directory, it doesn't contain source code. The `.gitignore` file only applies to the directories within this one.

src/monoclr/clrmod.c
filmor: Yeah, I was planning on switching to clr-loader as the next step, so this one will be obsolete. I'll remove it anyhow.

.github/workflows/main.yml
filmor: Why?

setup.py
filmor: I think so. Modern Python packaging (PEP517/518) runs in isolation in a temporary directory, if I understood things correctly.

setup.py
filmor: Not really, but as said, I'll just make these always run. The time to do this is completely dwarfed by the actual calls to `dotnet build`.

setup.py
lostmsu: If that would be so, I would not expect the need for `configured.props` in `.gitignore`. Something seems amiss.

.github/workflows/main.yml
lostmsu: It is a rare scenario, but I imagine if tests stop passing due to `latest` changing, we won't think of it as a possible cause until after a significant time spent on investigations.

setup.py
filmor: That is for development.

setup.py
filmor: I have removed this for now, I copied it over from the `clr-loader` `setup.py`. We will need this once our build is platform independent, as `setuptools` tries to be smart and always appends the full platform abi by default if DLLs show up (or if `build_ext` is used, could be that we just need to add a new command instead of abusing this one).

.github/workflows/main.yml
filmor: Using `latest` simplifies the matrix when we extend it to Windows and macOS :-)

We should also have weekly builds of master in place in the future, that would catch issues early enough.

src/embed_tests/TestFinalizer.cs
filmor: Done, #1309.

src/embed_tests/TestFinalizer.cs
```diff
@@ -38,6 +38,7 @@ private static bool FullGCCollect()
             catch (NotImplementedException)
             {
                 // Some clr runtime didn't implement GC.WaitForFullGCComplete yet.
+                Thread.Sleep(300);
```
filmor, line 41: I tried to introduce a small pause here to give the GC some time, didn't help, though. Should I remove this change?

filmor:
This is done from my side.

src/embed_tests/TestFinalizer.cs
lostmsu: Yep.

filmor: Doh, sorry. I'll do that directly on master.

lostmsu: No worries. I think it's good you unblocked merging of https://github.com/pythonnet/pythonnet/pull/1287

filmor: Yeah, that's why I sped this up a bit ;). I've undone the change on master now.

Issue #1211, opened 2020-08-29
Incorrect overloaded method got precedence while invoking .Net functions/methods from Python


partha-srivi:
### Environment

-   Pythonnet version: 2.5.1
-   Python version: 3.8.2
-   Operating System: Windows 10

### Details
In PythonNet 2.5.1, I found correct/appropriate overloaded .Net methods are not invoked from Python while executing .Net code statements 

for example please find the below code throws an error while trying to execute System.Data.DataRowCollection.Add(System.Data.DataRow) method.

Reference: https://docs.microsoft.com/en-us/dotnet/api/system.data.datarowcollection.add?view=netframework-4.8

Code:
```python
#PYTHON 3.8.x PYTHONNET 2.5.1
import clr
clr.AddReference("System")
clr.AddReference("System.Data")
import System
import System.Data

dt = System.Data.DataTable()
col = System.Data.DataColumn()
col.ColumnName = "Id"
col.DataType = System.Type.GetType("System.Int32")
dt.Columns.Add(col)
dr = dt.NewRow()
dr["Id"] = 10
#Throws error while invoking void System.Data.DataRowCollection.Add(System.Data.DataRow row)
try:
    dt.Rows.Add(dr)
except Exception as e:
    print(e)

#Working as expected while invoking System.Data.DataRow Add (params object[] values)
dr = dt.Rows.Add(20)
print(dr[0])
```
[params_overloaded_function_precedence_issue.py.txt](https://github.com/pythonnet/pythonnet/files/5145942/params_overloaded_function_precedence_issue.py.txt)

Unable to cast object of type 'System.Data.DataRow' to type 'System.IConvertible'.Couldn't store <System.Data.DataRow> in Id Column.  Expected type is Int32.
   at System.Data.DataColumn.set_Item(Int32 record, Object value)
   at System.Data.DataTable.NewRecordFromArray(Object[] value)
   at System.Data.DataRowCollection.Add(Object[] values)

I could fix this issue in local repo by patching Python.Runtime.MethodBinder.GetPrecedence method, I set lower precedence to overloaded methods with Params array parameter

Please have a look and let me know your suggestions

-Parthasarathy K

lostmsu:
Should now be fixed in master and the upcoming 2.5.2

filmor:
The fix for this should've been deployed with 2.5.2. If that is not the case, please reopen.

PR #1213, opened 2020-08-29
Bad params object[] precedence

lostmsu:

A bug in `ArgsPrecedence` caused parameters of type `object[]` to have the highest precedence, which affected `params object[]` handling, and also preferred `params object[]` to `params string[]` when a `string` argument was given.

#1211

### Any other comments?

...

### Checklist


lostmsu:
@filmor can you review this?

filmor (approved):

PR #1219, opened 2020-09-13
apply amos interop changes from soft shutdown

### LINKED ISSUE #1218: Multiple assertion failures for Python classes derived from System.Exception in debug builds
### Environment

-   Pythonnet version: 2.5.1
-   Python version: 3.8
-   Operating System: Win 10 64

### Details

In debug build (`DebugWinPY3`) this test:
https://github.com/pythonnet/pythonnet/blob/61d1b7c22eb7cd68f991d35fcafbdf4dd90b7b19/src/tests/test_constructors.py#L41

fails this assertion:
https://github.com/pythonnet/pythonnet/blob/61d1b7c22eb7cd68f991d35fcafbdf4dd90b7b19/src/runtime/interop.cs#L92

Python type object for `Sub` class gets an unknown extra field of size `IntPtr.Size`, causing it to be larger than its base type object (reflected) `System.Exception`.

See also https://github.com/pythonnet/pythonnet/pull/1196#issuecomment-690750845 for previous discussion.

@koubaa @amos402

koubaa:
Fixes #1218 

Changes taken from @amos402 soft shutdown branch

### DIFF
--- a/src/runtime/clrobject.cs
+++ b/src/runtime/clrobject.cs
@@ -11,8 +11,8 @@ namespace Python.Runtime
         {
             IntPtr py = Runtime.PyType_GenericAlloc(tp, 0);
 
-            long flags = Util.ReadCLong(tp, TypeOffset.tp_flags);
-            if ((flags & (int)TypeFlags.Subclass) != 0)
+            var flags = (TypeFlags)Util.ReadCLong(tp, TypeOffset.tp_flags);
+            if (flags.HasFlag(TypeFlags.Subclass))
             {
                 IntPtr dict = Marshal.ReadIntPtr(py, ObjectOffset.TypeDictOffset(tp));
                 if (dict == IntPtr.Zero)
--- a/src/runtime/clrobject.cs
+++ b/src/runtime/clrobject.cs
@@ -12,7 +12,7 @@ namespace Python.Runtime
             IntPtr py = Runtime.PyType_GenericAlloc(tp, 0);
 
             long flags = Util.ReadCLong(tp, TypeOffset.tp_flags);
-            if ((flags & TypeFlags.Subclass) != 0)
+            if ((flags & (int)TypeFlags.Subclass) != 0)
             {
                 IntPtr dict = Marshal.ReadIntPtr(py, ObjectOffset.TypeDictOffset(tp));
                 if (dict == IntPtr.Zero)
--- a/src/runtime/interop.cs
+++ b/src/runtime/interop.cs
@@ -118,16 +118,11 @@ namespace Python.Runtime
             size = fields.Length * IntPtr.Size;
         }
 
-        public static int GetSlotOffset(string name)
-        {
-            return NameMapping[name];
-        }
-
         private static int BaseOffset(IntPtr type)
         {
             Debug.Assert(type != IntPtr.Zero);
             int typeSize = Marshal.ReadInt32(type, TypeOffset.tp_basicsize);
-            Debug.Assert(typeSize > 0);
+            Debug.Assert(typeSize > 0 && typeSize <= ExceptionOffset.Size());
             return typeSize;
         }
         public static int DataOffset(IntPtr type)
@@ -358,38 +353,39 @@ namespace Python.Runtime
     /// Note that the two values reserved for stackless have been put
     /// to good use as PythonNet specific flags (Managed and Subclass)
     /// </summary>
-    internal class TypeFlags
+    [Flags]
+    internal enum TypeFlags
     {
-        public const int HeapType = (1 << 9);
-        public const int BaseType = (1 << 10);
-        public const int Ready = (1 << 12);
-        public const int Readying = (1 << 13);
-        public const int HaveGC = (1 << 14);
+        HeapType = (1 << 9),
+        BaseType = (1 << 10),
+        Ready = (1 << 12),
+        Readying = (1 << 13),
+        HaveGC = (1 << 14),
         // 15 and 16 are reserved for stackless
-        public const int HaveStacklessExtension = 0;
+        HaveStacklessExtension = 0,
         /* XXX Reusing reserved constants */
-        public const int Managed = (1 << 15); // PythonNet specific
-        public const int Subclass = (1 << 16); // PythonNet specific
-        public const int HaveIndex = (1 << 17);
+        Managed = (1 << 15), // PythonNet specific
+        Subclass = (1 << 16), // PythonNet specific
+        HaveIndex = (1 << 17),
         /* Objects support nb_index in PyNumberMethods */
-        public const int HaveVersionTag = (1 << 18);
-        public const int ValidVersionTag = (1 << 19);
-        public const int IsAbstract = (1 << 20);
-        public const int HaveNewBuffer = (1 << 21);
+        HaveVersionTag = (1 << 18),
+        ValidVersionTag = (1 << 19),
+        IsAbstract = (1 << 20),
+        HaveNewBuffer = (1 << 21),
         // TODO: Implement FastSubclass functions
-        public const int IntSubclass = (1 << 23);
-        public const int LongSubclass = (1 << 24);
-        public const int ListSubclass = (1 << 25);
-        public const int TupleSubclass = (1 << 26);
-        public const int StringSubclass = (1 << 27);
-        public const int UnicodeSubclass = (1 << 28);
-        public const int DictSubclass = (1 << 29);
-        public const int BaseExceptionSubclass = (1 << 30);
-        public const int TypeSubclass = (1 << 31);
+        IntSubclass = (1 << 23),
+        LongSubclass = (1 << 24),
+        ListSubclass = (1 << 25),
+        TupleSubclass = (1 << 26),
+        StringSubclass = (1 << 27),
+        UnicodeSubclass = (1 << 28),
+        DictSubclass = (1 << 29),
+        BaseExceptionSubclass = (1 << 30),
+        TypeSubclass = (1 << 31),
 
-        public const int Default = (
+        Default = (
             HaveStacklessExtension |
-            HaveVersionTag);
+            HaveVersionTag)
     }
 
 
--- a/src/runtime/interop.cs
+++ b/src/runtime/interop.cs
@@ -89,7 +89,6 @@ namespace Python.Runtime
     internal static class ManagedDataOffsets
     {
         public static int Magic { get; private set; }
-        public static readonly Dictionary<string, int> NameMapping = new Dictionary<string, int>();
 
         static class DataOffsets
         {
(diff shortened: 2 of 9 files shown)


src/runtime/interop.cs
```diff
@@ -69,27 +68,66 @@ public ModulePropertyAttribute()
...
+        {
+            return NameMapping[name];
+        }
 
         private static int BaseOffset(IntPtr type)
         {
             Debug.Assert(type != IntPtr.Zero);
             int typeSize = Marshal.ReadInt32(type, TypeOffset.tp_basicsize);
-            Debug.Assert(typeSize > 0 && typeSize <= ExceptionOffset.Size());
+            Debug.Assert(typeSize > 0);
```
lostmsu, line 130: @amos402 @koubaa 
As far as I can see, this is not actually fixing or explaining anything. It just removes the assert, that caused the problem.

lostmsu (changes requested):

src/runtime/interop.cs
```diff
@@ -69,27 +68,66 @@ public ModulePropertyAttribute()
...
+            int size = IntPtr.Size;
+            for (int i = 0; i < fields.Length; i++)
+            {
+                int offset = i * size;
+                FieldInfo fi = fields[i];
+                fi.SetValue(null, offset);
+            }
+        }
+
+        public static int magic() => ManagedDataOffsets.Magic;
```
lostmsu, line 86: @amos402 actually, this seems to be the line, that contains the actual fix.

Most of other changes are not necessary, and should be excluded from the PR.

I think it is fine to move `TypeOffset` sconstructor from generated files to `interop.cs`, but the template for the `interop*.cs` should also be updated.

lostmsu (changes requested):

src/runtime/interop.cs
lostmsu: IMHO, the assert should stay. This PR does not address it, but probably solves the problem @koubaa has.
I have more thorough fix in my fork, but it is quite extensive, and this will do for now.

src/runtime/interop.cs
```diff
@@ -69,27 +68,66 @@ public ModulePropertyAttribute()
         }
     }
 
+    internal static partial class TypeOffset
+    {
+        static TypeOffset()
+        {
+            Type type = typeof(TypeOffset);
+            FieldInfo[] fields = type.GetFields();
+            int size = IntPtr.Size;
```
lostmsu, line 77: I think it is fine to move `TypeOffset` sconstructor from generated files to `interop.cs`, but the template for the `interop*.cs` should also be updated.

src/runtime/interop.cs
```diff
@@ -320,34 +360,34 @@ public static void FreeModuleDef(IntPtr ptr)
     /// </summary>
     internal class TypeFlags
     {
-        public static int HeapType = (1 << 9);
-        public static int BaseType = (1 << 10);
-        public static int Ready = (1 << 12);
-        public static int Readying = (1 << 13);
-        public static int HaveGC = (1 << 14);
+        public const int HeapType = (1 << 9);
+        public const int BaseType = (1 << 10);
```
lostmsu, line 364: This really should be an enum

src/runtime/interop.cs
koubaa: @lostmsu I figured I'd keep all of the changes so that it'd make the soft shutdown PR smaller and easier to merge.  If you pushback I'll go ahead and trim it down

src/runtime/interop.cs
koubaa: @lostmsu Ah right, good point.  Side topic - python 3.5 was EOL today, can I submit an issue to drop 3.4 and 3.5?

lostmsu: Yep, that's a good idea.

src/runtime/interop.cs
lostmsu: @koubaa let's keep these changes.

There are a few _new_ functions, that are not used anywhere outside of the rest of `soft-shutdown` PR. These should be removed.

Also, we should turn `TypeFlags` into `[Flags]` enum. (at least I do not see any reason not to)

src/runtime/interop.cs
koubaa: @lostmsu https://github.com/pythonnet/pythonnet/issues/1225

src/runtime/interop.cs
koubaa: @lostmsu done.  It did require more changes elsewhere but nothing too drastic.  Good thing TypeFlags is internal or we'd have to document it as a breaking change.

setup.py
```diff
@@ -25,7 +25,7 @@
 
 # Allow config/verbosity to be set from cli
 # http://stackoverflow.com/a/4792601/5208670
-CONFIG = "Release"  # Release or Debug
+CONFIG = "Debug"  # Release or Debug
```
lostmsu, line 28: This seems to be a mistake

src/runtime/clrobject.cs
```diff
@@ -12,7 +12,7 @@ internal CLRObject(object ob, IntPtr tp)
             IntPtr py = Runtime.PyType_GenericAlloc(tp, 0);
 
             long flags = Util.ReadCLong(tp, TypeOffset.tp_flags);
-            if ((flags & TypeFlags.Subclass) != 0)
+            if ((flags & (int)TypeFlags.Subclass) != 0)
```
lostmsu, line 15: Should use `Enum.HasFlag`

src/runtime/managedtype.cs
```diff
@@ -29,7 +29,7 @@ internal static ManagedType GetManagedObject(IntPtr ob)
                 }
 
                 var flags = Util.ReadCLong(tp, TypeOffset.tp_flags);
-                if ((flags & TypeFlags.Managed) != 0)
+                if ((flags & (int)TypeFlags.Managed) != 0)
```
lostmsu, line 32: Same, `HasFlag`

pythonnet.15.sln
```diff
@@ -1,398 +1,347 @@
...
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Python.Runtime.15", "src\runtime\Python.Runtime.15.csproj", "{2759F4FF-716B-4828-916F-50FA86613DFC}"
+EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Python.EmbeddingTest.15", "src\embed_tests\Python.EmbeddingTest.15.csproj", "{66B8D01A-9906-452A-B09E-BF75EA76468F}"
+EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "clrmodule.15", "src\clrmodule\clrmodule.15.csproj", "{E08678D4-9A52-4AD5-B63D-8EBC7399981B}"
+EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Console.15", "src\console\Console.15.csproj", "{CDAD305F-8E72-492C-A314-64CF58D472A0}"
+EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Python.Test.15", "src\testing\Python.Test.15.csproj", "{F94B547A-E97E-4500-8D53-B4D64D076E5F}"
+EndProject
```
lostmsu, line 14: This file was changed by an accident?

BTW, you can unload a project in Visual Studio, if you don't need its tests for example.

lostmsu (changes requested):

pythonnet.15.sln
koubaa: @lostmsu Good point, actually for some reason this project fails to build on my setup and it breaks when I do setup.py.  Rather than figure it out I was lazy and removed it from the sln.

src/runtime/managedtype.cs
filmor: @koubaa This is not really resolved, is it?

src/runtime/interop.cs
```diff
@@ -69,21 +68,55 @@ public ModulePropertyAttribute()
...
             }
+        }
+
+        public static int magic() => ManagedDataOffsets.Magic;
+    }
+
+    internal static class ManagedDataOffsets
+    {
+        public static int Magic { get; private set; }
+        public static readonly Dictionary<string, int> NameMapping = new Dictionary<string, int>();
```
lostmsu, line 92: `NameMapping` seems unused.

src/runtime/interop.cs
amos402: The checking of `typeSize <= ExceptionOffset.Size()` is totally a wrong idea, there's no connection between them. Why an other object size whould relates to an exception object size.
You can simply make it lager than that size by
 ```python
class A(B, C, D, E, F, G): pass
```

lostmsu: @amos402 if I remember correctly, `BaseOffset` is only used for reflections of .NET classes created by Python.NET. Right now there are only two kinds of them: `System.Exception` and everything derived from it, that end up deriving Python's `builtins.BaseException` or `builtins.Exception`, and every other .NET type, which end up deriving from `builtins.object`. Within those two categories each of the types has identical structure, and the ones derived from `Exception` just happen to be larger.

When Python class is inherited from a reflected .NET class via single inheritance, and has no extra low-level PyType stuff defined explicitly, I see no reason for its objects to be larger, than corresponding base class objects. Hence all such types should have either the same size as reflected `System.Object` or as reflected `System.Exception`.

koubaa: @lostmsu @amos402  if the check needs to be removed, it can be done in a follow-on PR.  I don't think this discussion should hold this PR.

lostmsu: @koubaa  I am fine with this PR without the check removal.

lostmsu:
@amos402 @koubaa the fix is still incomplete. It places `GCHandle` past the last field of `PyType` (or Python class `type`), it never allocates space for it, hence it still writes to an offset it does not own.

For example, on Windows x64 Python 3.8 an instance of PyType is 880 bytes long. An instance of `CLR Metatype`, the type of all reflected classes, is also 880 bytes long. After this fix `class Sub(System.Exception)` invokes `MetaType.tp_new`, that has this line:

https://github.com/pythonnet/pythonnet/blob/022be190b6adbc46a67c67abd28e78dc95616f08/src/runtime/metatype.cs#L122

Where it writes the `GCHandle` immediately past the last byte of the new type object (an instance of `CLR Metatype`), because `TypeOffset.members` is also 880.

To fix this it might be sufficient to increase `tp_basicsize` of `CLR Metatype` by `IntPtr.Size` to allocate space for `GCHandle`.

amos402:
It allocated the space, you can check `PyType_GenericAlloc`. It has an extra space sizeof(PyMemberDef), as long as long you only use the first pointer which is points to the `name` and not to use wearef to CLR types, it's fine. If you want use more space roughtly, you should skip the memory of `type`, because the typeobject may check it.
If you want a more complete general solution, you can check this https://github.com/amos402/pythonnet/pull/5/files

PR #1226, opened 2020-09-16
fix compiler warning

koubaa:


koubaa:
@lostmsu please review

filmor (approved):
Looks good to me, can we make `PyObject.obj` `readonly` now?

lostmsu (changes requested):

lostmsu:
@koubaa sorry, reviewed, but forgot to submit :-)

src/runtime/pydict.cs
```diff
@@ -47,14 +46,13 @@ public PyDict()
...
         /// Python dictionary object.
         /// </remarks>
-        public PyDict(PyObject o)
+        public PyDict(PyObject o) : base(o.obj)
         {
             if (!IsDictType(o))
             {
                 throw new ArgumentException("object is not a dict");
             }
             Runtime.XIncref(o.obj);
```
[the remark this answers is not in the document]
koubaa: @lostmsu sure.  It seems like this is pre-existing though

src/runtime/pyiter.cs
```diff
@@ -25,19 +25,24 @@ public PyIter(IntPtr ptr) : base(ptr)
...
+        }
+
         /// <summary>
         /// PyIter Constructor
         /// </summary>
         /// <remarks>
         /// Creates a Python iterator from an iterable. Like doing "iter(iterable)" in python.
         /// </remarks>
-        public PyIter(PyObject iterable)
+        public PyIter(PyObject iterable) : base(FromObject(iterable))
```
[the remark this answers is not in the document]
koubaa: @lostmsu I'll add this to the changelog too

koubaa:
> Looks good to me, can we make `PyObject.obj` `readonly` now?

It's a good thought!  I tried it, but PyObject.Dispose sets it to IntPtr.Zero and that is a compiler error.

src/runtime/pydict.cs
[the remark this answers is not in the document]
lostmsu: Not really, as before this change `obj` field was set after incref.

src/runtime/pyiter.cs
```diff
@@ -25,19 +25,41 @@ public PyIter(IntPtr ptr) : base(ptr)
         {
         }
 
+        private static IntPtr FromObject(PyObject iterable)
+        {
+            if (iterable == null)
+            {
+                throw new NullReferenceException();
```
lostmsu, line 32: This should be `ArgumentNullException`

src/runtime/pyiter.cs
```diff
@@ -25,19 +25,41 @@ public PyIter(IntPtr ptr) : base(ptr)
...
+            }
+            IntPtr val = Runtime.PyObject_GetIter(iterable.obj);
+            if (val == IntPtr.Zero)
+            {
+                throw new PythonException();
+            }
```
lostmsu, line 38: NIT: you could use `PythonException.ThrowIfIsNull(val);` in all similar cases.

src/runtime/pytuple.cs
```diff
@@ -57,12 +60,24 @@ public PyTuple(PyObject o)
...
+        private static IntPtr FromArray(PyObject[] items)
+        {
+            int count = items.Length;
+            IntPtr val = Runtime.PyTuple_New(count);
+            for (var i = 0; i < count; i++)
+            {
+                IntPtr ptr = items[i].obj;
+                Runtime.XIncref(ptr);
+                int res = Runtime.PyTuple_SetItem(val, i, ptr);
+                PythonException.ThrowIfIsNotZero(res);
```
lostmsu, line 77: Same as above - we need to decref `val` on error.

lostmsu (changes requested):

src/runtime/pythonexception.cs
```diff
@@ -261,10 +261,11 @@ internal static void ThrowIfIsNull(BorrowedReference reference)
             }
         }
 
-        public static void ThrowIfIsNotZero(int value)
+        public static void ThrowIfIsNotZero(int value, Action onBeforeThrow = null)
         {
             if (value != 0)
             {
+                onBeforeThrow?.Invoke();
```
lostmsu, line 268: That's overengineered.

If you have
```csharp
if (value != 0)
{
  doStuff();
  throw;
}
```
pattern, it is better to spell it out (IMHO), than pass a lambda to a call. Besides lambdas have relatively significant cost to construct, whilst being totally unnecessary in 99% of cases when the call does not fail.

lostmsu (changes requested):

src/runtime/pythonexception.cs
koubaa: @lostmsu Fair point, I thought it wouldn't hurt to "propose" it.  Since the call usually doesn't fail I figured the branch predictor would optimize it well enough.

src/runtime/pytuple.cs
```diff
@@ -74,7 +74,11 @@ private static IntPtr FromArray(PyObject[] items)
                 IntPtr ptr = items[i].obj;
                 Runtime.XIncref(ptr);
                 int res = Runtime.PyTuple_SetItem(val, i, ptr);
-                PythonException.ThrowIfIsNotZero(res, () => Runtime.Py_DecRef(val));
+                if (res == 0)
+                {
+                    Runtime.Py_DecRef(val);
+                    PythonException.ThrowIfIsNotZero(res);
```
lostmsu, line 80: NIT: since you made corresponding check already, you can just `throw new PythonException()`

lostmsu (changes requested):

src/runtime/pytuple.cs
```diff
@@ -74,7 +74,11 @@ private static IntPtr FromArray(PyObject[] items)
                 IntPtr ptr = items[i].obj;
                 Runtime.XIncref(ptr);
                 int res = Runtime.PyTuple_SetItem(val, i, ptr);
-                PythonException.ThrowIfIsNotZero(res, () => Runtime.Py_DecRef(val));
+                if (res == 0)
```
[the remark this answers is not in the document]
koubaa: oops

lostmsu (approved):

Issue #1225, opened 2020-09-15
Drop support for python 3.4 and 3.5


koubaa:
python 3.5 is EOL

PR #1230, opened 2020-09-18
really remove old versions

koubaa:
More for #1225


koubaa:
@lostmsu  please review

PR #1233, opened 2020-09-23
Explicit interface implementations

danabr:

Previously, if a class implemented an interface explicitly, those
methods and properties become invisible to Python. To use them, you had
to make an explicit cast to the interface type.

No

### Checklist

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -28,6 +28,7 @@
 -   Christoph Gohlke ([@cgohlke](https://github.com/cgohlke))
 -   Christopher Bremner ([@chrisjbremner](https://github.com/chrisjbremner))
 -   Christopher Pow ([@christopherpow](https://github.com/christopherpow))
+-   Daniel Abrahamsson ([@danabr](https://github.com/danabr))
 -   Daniel Fernandez ([@fdanny](https://github.com/fdanny))
 -   Daniel Santana ([@dgsantana](https://github.com/dgsantana))
 -   Dave Hirschfeld ([@dhirschfeld](https://github.com/dhirschfeld))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ details about the cause of the failure
 -    Fix incorrect dereference of wrapper object in `tp_repr`, which may result in a program crash
 -    Fix incorrect dereference in params array handling
 -    Fix `object[]` parameters taking precedence when should not in overload resolution
+-    Fix methods and properties of explictly implemented interfaces being invisible
 
 ## [2.5.0][] - 2020-06-14
 
--- a/src/runtime/classmanager.cs
+++ b/src/runtime/classmanager.cs
@@ -241,32 +241,29 @@ namespace Python.Runtime
                 }
             }
 
-            if (type.IsInterface)
+            // Interface inheritance seems to be a different animal:
+            // more contractual, less structural.  Thus, a Type that
+            // represents an interface that inherits from another
+            // interface does not return the inherited interface's
+            // methods in GetMembers. For example ICollection inherits
+            // from IEnumerable, but ICollection's GetMemebers does not
+            // return GetEnumerator.
+            //
+            // Not sure if this is the correct way to fix this, but it
+            // seems to work. Thanks to Bruce Dodson for the fix.
+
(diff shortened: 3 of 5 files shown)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/pythonnet/pythonnet?pullRequest=1233) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/pythonnet/pythonnet?pullRequest=1233) before we can accept your contribution.<br/><br/>:x: danabr [sign now](https://cla.dotnetfoundation.org/pythonnet/pythonnet?pullRequest=1233)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/pythonnet/pythonnet?pullRequest=1233) it.</sub>

lostmsu:
I am not convinced this should be the expected behavior. Technically explicit interface implementations are private methods, and Python.NET does not expose them by design.

Is there a scenario, where you can not cast your instance to the desired interface? Or do you just want duck typing for some reason?

danabr:
Ah, that explains it. I ran into this when trying to switch out IronPython in a project. Since it worked there I got the impression that it would work with Python.NET also. I did not know that external interface implementations are private. Thanks for the explanation! I'll see if there is a way I can do the casting automatically in our module loader, so that end users don't have to worry about it.

filmor:
@lostmsu Can you link some docs regarding the privacy of explicit interface implementations? I thought they were mainly used to get around name and signature conflicts, I wasn't aware of a semantic significance.

danabr:
@lostmsu :  There's still an issue when a method returns an interface. If the returned object implements its interface explicitly, its methods becomes invisible to Python:

```C#
    public interface ISayHello1
    {
        string SayHello();
    }

    public class InterfaceTest2 : ISayHello1
    {
        public ISayHello1 GetISayHello1()
        {
            return this;
        }

        string ISayHello1.SayHello()
        {
            return "ISayHello1";
        }
    }
````

```python
 def test_interface_object_returned_through_method():
    """Test explicitly implemented methods are visible"""
    from Python.Test import InterfaceTest2

    ob = InterfaceTest2()
    hello1 = ob.GetISayHello1()
    assert hello1.SayHello() == 'ISayHello1'
```

This test fails with:
```
>       assert hello1.SayHello() == 'ISayHello1'
E       AttributeError: 'InterfaceTest2' object has no attribute 'SayHello'
```

The caller of GetISayHello1 shouldn't really need to cast the returned object to `ISayHello1`. The method is already declared to returned that type! It also adds to the confusion that casting is only sometimes necessary (namely when the interface has been explicitly implemented, something the caller should not need to know about).

I'd be happy for advise on how to handle this.

lostmsu:
@danabr this actually seems a legitimate reason for a change. When method explicitly returns an interface type, Python should see the object as an instance of that interface with all its methods available. @filmor thoughts?

filmor:
As said, I'm not aware of any semantic difference between the "usual" and explicit implementation of an interface, so IMO this change makes sense.

danabr:
The C# documentation ([1]) does say that:

> It's possible to implement an interface member explicitly—creating a class member that is only called through the interface, and is specific to that interface.
[...]
The class member IControl.Paint is only available through the IControl interface, and ISurface.Paint is only available through ISurface. Both method implementations are separate, and neither are available directly on the class.

So behaving differently in Python.NET might not be  a good idea, although perhaps never a problem in practice. Could the issue with returning an interface from a method be solved some other way? e.g. by doing the same thing that we do when casting to the interface in Python when we convert the return value of a method to a Python value?

[1] https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/interfaces/explicit-interface-implementation

lostmsu:
@danabr you would need to postprocess here: https://github.com/pythonnet/pythonnet/blob/451fae6bf496f15b581e228fe1582b4979aebe25/src/runtime/methodbinder.cs#L698

And maybe in some other places.

danabr:
@lostmsu , @filmor : An approach "casting" the object to the interface type before returning it has been implemented in #1240. It comes with its own sets of things to consider. I'd be happy to hear your thoughts on that approach.

PR #1240, opened 2020-09-25
Wrap returned objects in interface if method return type is interface

danabr:

This allows callers to call all methods of an interface, regardless of
whether the method was implemented implicitly or explicitly. Before this
change, you had to make an explicit cast to the interface to be able to
call the explicitly implemented method. Consider the following code:

```C#
namespace Python.Test {
    public interface ITestInterface
    {
        void Foo();
        void Bar();
    }

    public class TestImpl : ITestInterface
    {
        public void Foo() { };
        public void ITestInterface.Bar() { };
        public void Baz() { };

        public static ITestInterface GetInterface()
        {
            return new TestImpl();
        }
    }
}
```

And the following Python code, demonstrating the behavior before this
change:

```python
from Python.Test import TestImpl, ITestInterface

test = TestImpl.GetInterface()
test.Foo() # works
test.Bar() # AttributeError: 'TestImpl' object has no attribute 'Bar'
test.Baz() # works! - baz
```

After this change, the behavior is as follows:
```
test = TestImpl.GetInterface()
test.Foo() # works
test.Bar() # works
test.Baz() # AttributeError: 'ITestInterface' object has no attribute 'Baz'
```

This is a breaking change due to that `Baz` is no longer visible in
Python.

No.

### Any other comments?

See #1233 for an alternative approach, that exposes methods of explictly implemented interfaces without need to cast the object first.

### Checklist

### DIFF
--- a/src/runtime/interfaceobject.cs
+++ b/src/runtime/interfaceobject.cs
@@ -71,7 +71,43 @@ namespace Python.Runtime
                 return IntPtr.Zero;
             }
 
-            return CLRObject.GetInstHandle(obj, self.pyHandle);
+            return self.WrapObject(obj);
+        }
+
+        /// <summary>
+        /// Wrap the given object in an interface object, so that only methods
+        /// of the interface are available.
+        /// </summary>
+        public IntPtr WrapObject(object impl)
+        {
+            var objPtr = CLRObject.GetInstHandle(impl, pyHandle);
+            return objPtr;
+        }
+
+        /// <summary>
+        /// Expose the wrapped implementation through attributes in both
+        /// converted/encoded (__implementation__) and raw (__raw_implementation__) form.
+        /// </summary>
+        public static IntPtr tp_getattro(IntPtr ob, IntPtr key)
+        {
+            var clrObj = (CLRObject)GetManagedObject(ob);
+
+            if (!Runtime.PyString_Check(key))
+            {
+                return Exceptions.RaiseTypeError("string expected");
+            }
+
+            string name = Runtime.GetManagedString(key);
+            if (name == "__implementation__")
+            {
+                return Converter.ToPython(clrObj.inst);
+            }
+            else if (name == "__raw_implementation__")
+            {
+                return CLRObject.GetInstHandle(clrObj.inst);
+            }
+
+            return Runtime.PyObject_GenericGetAttr(ob, key);
         }
     }
 }
--- a/src/runtime/methodbinder.cs
+++ b/src/runtime/methodbinder.cs
@@ -744,7 +744,7 @@ namespace Python.Runtime
                     Type pt = pi[i].ParameterType;
                     if (pi[i].IsOut || pt.IsByRef)
                     {
-                        v = Converter.ToPython(binding.args[i], pt);
+                        v = Converter.ToPython(binding.args[i], pt.GetElementType());
                         Runtime.PyTuple_SetItem(t, n, v);
                         n++;
                     }
--- a/src/runtime/typemanager.cs
+++ b/src/runtime/typemanager.cs
@@ -164,12 +164,12 @@ namespace Python.Runtime
             // we want to do this after the slot stuff above in case the class itself implements a slot method
             InitializeSlots(type, impl.GetType());
 
-            if (!clrType.GetInterfaces().Any(ifc => ifc == typeof(IEnumerable) || ifc == typeof(IEnumerator)))
+            if (!typeof(IEnumerable).IsAssignableFrom(clrType) &&
+                !typeof(IEnumerator).IsAssignableFrom(clrType))
             {
                 // The tp_iter slot should only be set for enumerable types.
                 Marshal.WriteIntPtr(type, TypeOffset.tp_iter, IntPtr.Zero);
             }
-
 
             if (base_ != IntPtr.Zero)
             {
--- a/src/tests/test_subclass.py
+++ b/src/tests/test_subclass.py
@@ -104,8 +104,10 @@ def test_interface():
     assert ob.bar("bar", 2) == "bar/bar"
     assert FunctionsTest.test_bar(ob, "bar", 2) == "bar/bar"
 
-    x = FunctionsTest.pass_through(ob)
-    assert id(x) == id(ob)
+    # pass_through will convert from InterfaceTestClass -> IInterfaceTest,
+    # causing a new wrapper object to be created. Hence id will differ.
+    x = FunctionsTest.pass_through_interface(ob)
+    assert id(x) != id(ob)
 
 
 def test_derived_class():
@@ -173,14 +175,14 @@ def test_create_instance():
     assert id(x) == id(ob)
 
     InterfaceTestClass = interface_test_class_fixture(test_create_instance.__name__)
-    ob2 = FunctionsTest.create_instance(InterfaceTestClass)
+    ob2 = FunctionsTest.create_instance_interface(InterfaceTestClass)
     assert ob2.foo() == "InterfaceTestClass"
     assert FunctionsTest.test_foo(ob2) == "InterfaceTestClass"
     assert ob2.bar("bar", 2) == "bar/bar"
     assert FunctionsTest.test_bar(ob2, "bar", 2) == "bar/bar"
 
-    y = FunctionsTest.pass_through(ob2)
-    assert id(y) == id(ob2)
+    y = FunctionsTest.pass_through_interface(ob2)
+    assert id(y) != id(ob2)
 
 
 def test_events():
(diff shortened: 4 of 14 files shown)


src/tests/test_subclass.py
```diff
@@ -104,8 +104,10 @@ def test_interface():
     assert ob.bar("bar", 2) == "bar/bar"
     assert FunctionsTest.test_bar(ob, "bar", 2) == "bar/bar"
 
-    x = FunctionsTest.pass_through(ob)
-    assert id(x) == id(ob)
+    # pass_through will convert from InterfaceTestClass -> IInterfaceTest,
+    # causing a new wrapper object to be created. Hence id will differ.
+    x = FunctionsTest.pass_through_interface(ob)
+    assert id(x) != id(ob)
```
danabr, line 110: I don't know whether this could cause any surprises.

We still have that `x == y` and that `len(set([x,y]) == 1`, so the objects can still be used as dict keys etc with the same behavior as before. But perhaps there are other effects I have not thought of.

src/runtime/methodbinder.cs
```diff
@@ -761,7 +761,24 @@ internal virtual IntPtr Invoke(IntPtr inst, IntPtr args, IntPtr kw, MethodBase i
...
+        /// <summary>
+        /// Takes care of wrapping an object in an interface object
+        /// in case the method returns a value of a interface type.
+        /// </summary>
+        private IntPtr MakeReturnValue(object retVal, Type retType)
+        {
+            if (retVal != null && retType.IsInterface)
+            {
+                var ifaceObj = (InterfaceObject)ClassManager.GetClass(retType);
+                return CLRObject.GetInstHandle(retVal, ifaceObj.pyHandle);
```
danabr, line 776: One thing we lose out on here is that you cannot register custom converters for interfaces. I should probably restore that ability.

lostmsu: Yeah, can't you just set `retVal` in this branch to `ifaceObj` and pass it to the `Converter.ToPython` as usual?

lostmsu (changes requested):
I am actually unfamiliar with that side of Python.NET, but if we make this change, can users then somehow cast the interface back to the class? If that is not so yet, such capability must be added (perhaps as a Python property?)

@filmor , do you know if you can unwrap `InterfaceObject` from Python?

src/runtime/methodbinder.cs
danabr: Not really. Implementations of `IPyObjectEncoder` will expect to get the underlying CLR object, not an `InterfaceObject`.

I should probably move the logic into `Converter.ToPython` instead. The users will have the ability to encode interface instances any way they like if they want to.

danabr:
I also found that I have failed to handle array subscription. If a method returns a `ISomething[]` , then the individual elements will not be wrapped in the interface when you access them (e.g. `arr[0]`).

src/runtime/methodbinder.cs
danabr: I moved the logic, and now user defined encoders should once again be respected.

danabr:
I've addressed the issue with returning arrays of interface objects.

What remains is the ability to "downcast"/unwrap objects, as @lostmsu mentioned. As far as I have been able to see, there is no way of doing that today. One could add a property as @lostmsu suggests - the question then becomes how to name that so that the risk of colliding with one of the interface members is minimized. Perhaps there is a subset of names that would be valid in Python, but not in .NET so one can avoid collisions completely?

danabr:
Another approach would be to provide a service class that can be used for downcasting/unwrapping. A simple implementation of that today is:

```C#
namespace Python.Test {
    public class Cast
    {
        public static object Downcast(object input) // note the return type!
        {
            return input;
        }
    }
}
```

Which you then can use like this:
```python
from Python.Test import Cast
ob = Cast.Downcast(iface_obj)
```

The reason I bring this to attention is that it highlights the incompleteness of my approach in this PR. I have tried to make sure that whenever a .NET method returns an interface, the interface is all that Python sees (proper implementation hiding). However, I have done nothing about the corresponding case for classes. If a method is declared to return a value of type `ParentClass` and returns a `ChildClass`, Python-land is free to access any member of `ChildClass`. Having different behavior for classes and interfaces does not seem like a good idea.

Either we should apply implementation hiding everywhere (like done for interfaces in this PR), or nowhere (which is the case today). The former is more in line with what you would get in C#. The latter is perhaps what users of Python.NET wants (?), since that is the behavior that is currently there.

I'd be happy to hear your thoughts on what is the right way forward.

lostmsu:
@danabr completeness is not necessary. If there's a bug around C# `new` methods (not) hiding base class methods with the same name, let us track it separately.

The problem with no hiding is that it makes calling certain overloads potentially impossible due to name conflicts.

@filmor do you know if we have anything like `Downcast` methods already available?

If not, I suggest `InterfaceObject` to expose `__implementation__` property pointing to the unwrapped instance (with converters applied) and `__raw_implementation__` without any conversions.

danabr:
I've added the requested properties `__implementation__` and `__raw_implementation__`. Alas, I failed to come up with a good way to test the latter (namely that no conversions are applied). I was hoping I could make some clever use of custom encoders to test it.

What remains now is properly documenting that the properties are available. Should that be done somewhere in this repo, or on the wiki, or on pythonnet.github.io?

I also need to make a note of the changes in the CHANGELOG, and make it clear that this is a breaking change.

lostmsu:
@danabr to test `__raw_implementation__` you can do `IComparable` with `Int32`, and ensure, that type of the raw object is actually `System.Int32`. Currently Python.NET always converts primitive .NET int types to Python's `int`.

pythonnet.github.io seems to have only basic example, so you might want to create a new wiki page for the interface objects.

Yeah, a note in CHANGELOG would be helpful.

danabr:
>  to test __raw_implementation__ you can do IComparable with Int32, and ensure, that type of the raw object is actually System.Int32.

Great idea! I used that to add a test. I've also updated the CHANGELOG. I'll have a look at updating the Wiki if this gets merged.

Thanks for all the help and guidance, @lostmsu!

lostmsu:
@danabr seems like a few tests are failing because the `InterfaceObject` has no `tp_iter`.

danabr:
Argh, I wonder how come I did not notice before. I do run the tests locally. But now I can reproduce it. I'll look into it.

danabr:
It was an unexpected interaction with the work in one of my other PRs. I must have forgotten to run the tests after rebasing on master. Now all tests pass locally. Hopefully they will pass in CI as well.

src/runtime/interfaceobject.cs
```diff
@@ -71,7 +71,21 @@ public static IntPtr tp_new(IntPtr tp, IntPtr args, IntPtr kw)
...
+        /// Wrap the given object in an interface object, so that only methods
+        /// of the interface are available. The wrapped object is available
+        /// through properties in both converted/encoded (__implementation__)
+        /// and raw form (__raw_implementation__).
+        /// </summary>
+        public IntPtr WrapObject(object impl)
+        {
+            var objPtr = CLRObject.GetInstHandle(impl, pyHandle);
+            Runtime.PyObject_SetAttrString(objPtr, "__implementation__", Converter.ToPython(impl));
+            Runtime.PyObject_SetAttrString(objPtr, "__raw_implementation__", CLRObject.GetInstHandle(impl));
```
lostmsu, line 87: These two must be computed only on demand.

src/runtime/typemanager.cs
```diff
@@ -164,13 +164,13 @@ internal static IntPtr CreateType(ManagedType impl, Type clrType)
             // we want to do this after the slot stuff above in case the class itself implements a slot method
             InitializeSlots(type, impl.GetType());
 
-            if (!clrType.GetInterfaces().Any(ifc => ifc == typeof(IEnumerable) || ifc == typeof(IEnumerator)))
+            if (!(clrType.IsInterface && (clrType == typeof(IEnumerable) || clrType == typeof(IEnumerator))) &&
+                !clrType.GetInterfaces().Any(ifc => ifc == typeof(IEnumerable) || ifc == typeof(IEnumerator)))
```
lostmsu, line 168: Hint: you can use `typeof(I).IsAssignableFrom(type)`

lostmsu (changes requested):

src/runtime/interfaceobject.cs
danabr: OK. I fixed it by implementing `tp_getattro`. Note that the properties show up when you do `dir` on an interface object. If that is desired I should probably go the long route via `tp_getset`.

src/runtime/typemanager.cs
danabr: Thanks, that became a lot cleaner.

lostmsu (approved):

PR #1241, opened 2020-09-26
Ensure only implementers of IEnumerable or IEnumerator are considered Iterable

### LINKED ISSUE #1234: .NET class instances considered Iterable
### Environment

-   Pythonnet version: 2.5.0
-   Python version: 3.7
-   Operating System: Windows

### Details

All .NET class instances are considered instances of `collections.abc.Iterable`.

```C#

namespace Python.Test {
  public class Empty { }
}
```

```python

def test_not_Iterable():
    """Test regular .NET objects are not iterable"""
    from Python.Test import Empty
    from collections.abc import Iterable

    assert (not isinstance(Empty(), Iterable))
```

The test is expected to pass, but fails:

```
    def test_not_Iterable():
        """Test regular .NET objects are not iterable"""
        from Python.Test import Empty
        from collections.abc import Iterable

>       assert (not isinstance(Empty(), Iterable))
E       AssertionError: assert not True
E        +  where True = isinstance(<Python.Test.Empty object at 0x0000022D3E6DE5C8>, <class 'collections.abc.Iterable'>)
E        +    where <Python.Test.Empty object at 0x0000022D3E6DE5C8> = <class 'Python.Test.Empty'>()
```

danabr:

Fixes #1234.

### Checklist

### DIFF
--- a/src/runtime/typemanager.cs
+++ b/src/runtime/typemanager.cs
@@ -164,6 +164,13 @@ namespace Python.Runtime
             // we want to do this after the slot stuff above in case the class itself implements a slot method
             InitializeSlots(type, impl.GetType());
 
+            if (!clrType.GetInterfaces().Any(ifc => ifc == typeof(IEnumerable) || ifc == typeof(IEnumerator)))
+            {
+                // The tp_iter slot should only be set for enumerable types.
+                Marshal.WriteIntPtr(type, TypeOffset.tp_iter, IntPtr.Zero);
+            }
+
+
             if (base_ != IntPtr.Zero)
             {
                 Marshal.WriteIntPtr(type, TypeOffset.tp_base, base_);
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -28,6 +28,7 @@
 -   Christoph Gohlke ([@cgohlke](https://github.com/cgohlke))
 -   Christopher Bremner ([@chrisjbremner](https://github.com/chrisjbremner))
 -   Christopher Pow ([@christopherpow](https://github.com/christopherpow))
+-   Daniel Abrahamsson ([@danabr](https://github.com/danabr))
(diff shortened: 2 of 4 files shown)


filmor (approved):
Thank you very much, looks good to me.

@lostmsu Can you have a look as well?

Mid-term we should probably just fill the slots explicitly instead of picking up methods by name.

src/runtime/typemanager.cs
```diff
@@ -713,6 +713,13 @@ internal static void InitializeSlots(IntPtr type, Type impl)
                         continue;
                     }
 
+                    if (name == "tp_iter" && clrType != null &&
+                        !clrType.GetInterfaces().Any(i => i == typeof(IEnumerable) || i == typeof(IEnumerator)))
+                    {
+                        continue;
+                    }
+
+
```
lostmsu, line 722: This logic should be in the caller (I guess `CreateType` in this case), and `tp_iter` should be either be made private to not be handled by generic code here, or the slot should be overwritten after `InitializeSlots` is called.

A few reasons:
1) different slots might require different information to figure out if and how they need to be initialized. If in the future we would need to add custom initialization for a different method, the parameter list of this function would grow.
2) if each slot will require own handling, this method would grow too.

lostmsu (changes requested):

src/runtime/typemanager.cs
danabr: Yes, my first approach was too hackish.

I went with the second approach of overwriting the slot after `InitializrSlots`. For some reason, I couldn't get moving the handling of tp_iter out of the generic code to work.

lostmsu: @danabr its because it is already hackish, as it simply matches public methods to slots by name. I hope we will get rid of it...

lostmsu (approved):

lostmsu:
Thanks for the contribution!

PR #1242, opened 2020-09-30
Select correct method to invoke when keyword arguments are used

### LINKED ISSUE #1235: Wrong method chosen if argument is passed as keyword argument
### Environment

-   Pythonnet version: 2.5.0
-   Python version: 3.7

### Details

```C#
namespace Python.Test {
    public class MethodArityTest
    {
        public string Foo(int a) { return "Arity 1"; }
        public string Foo(int a, int b) { return "Arity 2"; }
    }
}
```

```python
def test_keyword_arg_method_resolution():
    """Test correct method is chosen using keyword arguments"""
    from Python.Test import MethodArityTest

    ob = MethodArityTest()
    assert ob.Foo(1, b=2) == "Arity 2" 
```

The test above fails with:

```
    def test_keyword_arg_method_resolution():
        """Test correct method is chosen using keyword arguments"""
        from Python.Test import MethodArityTest

        ob = MethodArityTest()
>       assert ob.Foo(1, b=2) == "Arity 2"
E       AssertionError: assert 'Arity 1' == 'Arity 2'
E         - Arity 1
E         ?       ^
E         + Arity 2
E         ?       ^
```

danabr:

If there were two methods with the same name, but with different arity,
`Foo(a)` and `Foo(a,b)`, and the latter was called with a keyword
argument `some.Foo(0, b=1)`, `Foo(0)` would be called instead of
`Foo(0,1)`.

Fixes #1235.

### Checklist

-   [] If an enhancement PR, please create docs and at best an example
-   [] Add yourself to [`AUTHORS`](../blob/master/AUTHORS.md)

### DIFF
--- a/src/runtime/constructorbinder.cs
+++ b/src/runtime/constructorbinder.cs
@@ -89,7 +89,7 @@ namespace Python.Runtime
                 // any extra args are intended for the subclass' __init__.
 
                 IntPtr eargs = Runtime.PyTuple_New(0);
-                binding = Bind(inst, eargs, kw);
+                binding = Bind(inst, eargs, IntPtr.Zero);
                 Runtime.XDecref(eargs);
 
                 if (binding == null)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ details about the cause of the failure
 -    Fix incorrect dereference in params array handling
 -    Fix `object[]` parameters taking precedence when should not in overload resolution
 -    Fixed a bug where all .NET class instances were considered Iterable
+-    Fix incorrect choice of method to invoke when using keyword arguments.
 
 ## [2.5.0][] - 2020-06-14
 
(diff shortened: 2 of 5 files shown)


src/runtime/constructorbinder.cs
```diff
@@ -89,7 +89,7 @@ internal object InvokeRaw(IntPtr inst, IntPtr args, IntPtr kw, MethodBase info)
                 // any extra args are intended for the subclass' __init__.
 
                 IntPtr eargs = Runtime.PyTuple_New(0);
-                binding = Bind(inst, eargs, kw);
+                binding = Bind(inst, eargs, IntPtr.Zero);
```
filmor, line 92: Why is this change necessary?

danabr: The old code was actually incorrect, but it was not exposed until I made the bugfix in `Bind`. We are looking for a zero argument constructor here, so no kwargs should be passed in. As the comment above says: "any extra args are intended for the subclass' __init__."

filmor (approved):

filmor:
Great thanks :)

PR #1246, opened 2020-10-05
Make indexers work for interface objects

danabr:

Makes the following work instead of throwing an exception:

```python
from System.Collections.Generic import Dictionary, IDictionary
d = IDictionary[str, str](Dictionary[str, str]())
d["one"] = "1"
assert d["one"] == "1"
```

No

### Checklist


lostmsu (changes requested):
Two things related to each other:
1. Please, add a negative test (if one does not exist yet). E.g. construct an empty `System.Object`, and ensure Python can't index it for read or write.
2. Consider only assigning `mp*_subscript` slots only when the type actually supports indexing like you did with `tp_next`.

danabr:
I've added the two things you asked for.

The erasing of the slots is ugly. We should consider moving `InitializeSlots` into `ManagedObject` and its descendants instead, so that you would just do `impl.InitializeSlots(type)`, and each type could handle its own initialization.

lostmsu (approved):

lostmsu:
Great, thanks!

I agree on `InitializeSlots`. If you'd like to work on that, you can start with a draft request demonstrating an approach for any one kind of `ManagedObject` (it does not have to pass tests or compile). Otherwise we'll just keep that in mind.

Overall it might make sense to move the majority of `TypeManager` type creation functionality into `ManagedObject` and deriving classes, including `PyType` construction. But we better do that along with the migration to `PyType_FromSpec`.

Issue #256, opened 2016-09-01
Cannot access indexer from .NET derived class


Thraka:
If you create a .NET base class that has an indexer, and then create a .NET derived class, that derived class used in Python cannot access the indexer. You can access existing properties though just fine.

``` python
import clr

clr.AddReference("PythonTest")

import PythonTest

a = PythonTest.IndexerBase()
b = PythonTest.Indexer()

print("count of items from base class: %i" % a.Count)
print("count of items from derived class: %i" % b.Count)
print("printing base class value: %s" % a[1])
print("printing derived class value: %s" % b[1]) #ERROR HERE
```

[PythonTest Library.zip](https://github.com/pythonnet/pythonnet/files/448914/PythonTest.Library.zip)

denfromufa:
This was brought up 2 times before but never resolved. There is mention of special syntax `.get_Item` that may be available.

https://mail.python.org/pipermail/pythondotnet/2011-September/001173.html

https://mail.python.org/pipermail/pythondotnet/2016-February/001720.html

There are some tests for indexers, not sure why your case and others above are not covered.

denfromufa:
https://github.com/pythonnet/pythonnet/blob/master/src/tests/test_indexer.py
https://github.com/pythonnet/pythonnet/blob/master/src/testing/indexertest.cs

denfromufa:
Oh, I just noticed the key word here - "in .NET derived class".

Thraka:
Yes, I tried to make sure I specified where the classes come from, otherwise it would be confusing 😄. The python code I specified above is the entire test using the library in the zip file.

abab987789:
For this bug I wonder if there is a good solution?

denfromufa:
@abab987789 this place needs to be modified in classmanager.cs to look for parent classes:
```
private static void InitClassBase(Type type, ClassBase impl)
        {
            // First, we introspect the managed type and build some class
            // information, including generating the member descriptors
            // that we'll be putting in the Python class __dict__.

            ClassInfo info = GetClassInfo(type);

            impl.indexer = info.indexer;
```

denfromufa:
@Thraka @abab987789 as a workaround you can cast to base class to access the indexer for now. For example:

```
PythonTest.IndexerBase(b)[1]
u'test2'
```

abab987789:
@denfromufa Now I solved the problem in a way,Check all property indexer
`// Loop through once to find out which names are declared
[...]

PR #1248, opened 2020-10-07
Make it possible to use inherited indexers

danabr:

If a class A had indexer, and class B derived from it, Python.NET would
not consider class B to have any indexer.

No

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,7 @@ details about the cause of the failure
 -    Fix incorrect choice of method to invoke when using keyword arguments.
 -    Fix non-delegate types incorrectly appearing as callable.
 -    Indexers can now be used with interface objects
+-    Fixed a bug where indexers could not be used if they were inherited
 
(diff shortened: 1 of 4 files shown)


lostmsu:
What about interfaces?

danabr:
> What about interfaces?

That is already handled by https://github.com/pythonnet/pythonnet/pull/1246. The calculation of whether something has an indexer or not is done the same way for classes and interfaces. It is all based on the implementation class.

lostmsu:
@danabr just to clarify. So if I have `interface IExtra: IIndexed` with indexer defined in `IIndexed`, and in Python do `extra = IExtra(extra_impl)` (e.g. explicit cast to `IExtra`), I would be able to do `extra[1234]`?

If so, I think a test for that scenario should be added to `test_indexer_accessed_through_interface` from your previous PR.

danabr:
Yes. The requested test has been added.

christabella:
This closes https://github.com/pythonnet/pythonnet/issues/256

PR #1254, opened 2020-10-09
Intern string

amos402:
Use the intern strings map to const strings instead of creating temporary strings everytimes.

### Checklist

### DIFF
--- a/src/runtime/intern.cs
+++ b/src/runtime/intern.cs
@@ -1,0 +1,74 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Python.Runtime
+{
+    static partial class InternString
+    {
+        private static Dictionary<string, IntPtr> _string2interns;
+        private static Dictionary<IntPtr, string> _intern2strings;
+
+        static InternString()
+        {
+            var identifierNames = typeof(PyIdentifier).GetFields().Select(fi => fi.Name);
+            var validNames = new HashSet<string>(identifierNames);
+            if (validNames.Count != _builtinNames.Length)
+            {
+                throw new InvalidOperationException("Identifiers args not matching");
+            }
+            foreach (var name in _builtinNames)
+            {
+                if (!validNames.Contains(name))
+                {
+                    throw new InvalidOperationException($"{name} is not declared");
+                }
+            }
+        }
+
+        public static void Initialize()
+        {
+            _string2interns = new Dictionary<string, IntPtr>();
+            _intern2strings = new Dictionary<IntPtr, string>();
+
+            Type type = typeof(PyIdentifier);
+            foreach (string name in _builtinNames)
+            {
+                IntPtr op = Runtime.PyUnicode_InternFromString(name);
+                SetIntern(name, op);
+                type.GetField(name).SetValue(null, op);
+            }
+        }
+
+        public static void Shutdown()
+        {
+            foreach (var ptr in _intern2strings.Keys)
+            {
+                Runtime.XDecref(ptr);
+            }
+            _string2interns = null;
+            _intern2strings = null;
+        }
+
+        public static string GetManagedString(IntPtr op)
+        {
+            string s;
+            if (TryGetInterned(op, out s))
+            {
+                return s;
+            }
+            return Runtime.GetManagedString(op);
+        }
+
+        public static bool TryGetInterned(IntPtr op, out string s)
+        {
+            return _intern2strings.TryGetValue(op, out s);
+        }
+
+        private static void SetIntern(string s, IntPtr op)
+        {
+            _string2interns.Add(s, op);
+            _intern2strings.Add(op, s);
+        }
+    }
+}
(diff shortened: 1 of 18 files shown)


src/runtime/intern.cs
```diff
@@ -0,0 +1,84 @@
...
+    {
+        class PyIdentifierDefinition
+        {
+            public string Name { get; private set; }
+
+            public PyIdentifierDefinition(string name)
+            {
+                Name = name;
+            }
+        }
```
lostmsu, line 17: This class does not appear to be used anywhere

src/runtime/intern_.tt
```diff
@@ -0,0 +1,58 @@
+<#@ template debug="true" hostSpecific="true" #>
+<#@ output extension=".cs" #>
+<#
+    string[] internNames = new string[]
```
lostmsu, line 4: Why use a template generator, when you can simply get `internNames` from reflection? It looks like unnecessary complication.

src/runtime/intern.cs
```diff
@@ -0,0 +1,84 @@
...
+
+        public static string GetManagedString(IntPtr op)
+        {
+            string s;
+            if (TryGetInterned(op, out s))
+            {
+                return s;
+            }
+            return Runtime.GetManagedString(op);
+        }
```
lostmsu, line 71: Having pointers ready for method names Python.NET uses internally is a good idea. E.g. `PyIdentifier.__class__` we might call very often.

I am not so sure about having a separate bidirectional mapping just for them though. A generic cache for conversions might be good, but not when it just caches a few specific strings.

I think the `GetManagedString`, `SetIntern` and `TryGetInterned` should all be removed.

lostmsu (changes requested):

src/runtime/intern.cs
amos402: Indeed, I shall remove it and push again.

src/runtime/intern_.tt
amos402: Two reasons make this scenario happened.
1. You need to make sure the string is a c# intern string too to make code like this work without extra checking
```c#
const string s = "1";
switch (
    case "1":
        // both "1" point to the same string
       break;
```
2. If you need (1), you have to make the `internNames` set on compile.
Of course, if you don't (1), internNames can just set by using reflection from reading `PyIdentities`.

src/runtime/intern.cs
amos402: `GetManagedString` is use for 
```c#
string name = InternString.GetManagedString(key);
switch (name)
{
    case "__doc__":
        // ...
}
```

lostmsu: @amos402 I meant the new one removed, and the old one kept.

amos402: You mean replace the old one with the new one?
It's better to keep both. Calling `Runtime.GetManagedString` inside the `Intern.GetManagedString` just a fallback if that's not an interned string.
Like the exmple above, the `key` may not be an intern string.
```python
obj.__doc__ # intern string
getattr(obj, "__doc__") # intern string
getattr(obj, "__doc" + "__") # new string
```
You can use `Runtime.GetManagedString` directly if you certainly know that string is not an interned string so it can skip the hash checking.

lostmsu: I mean completely remove this new one. I believe it has very low performance benefit, and may even harm user code performance. And the code is also very complicated, which does not seem to be justified by the potential benefit.

lostmsu (changes requested):

src/runtime/intern.cs
amos402: On the contrary it improve the performance, you should check the usages.

lostmsu: @amos402 the only usage I see is the one in `ModuleObject`, and I don't think accessing module object members is a hot path, that needs special optimization (again, considering the complexity).

amos402: A path is not hot means it can squander computation randomly? I object to this kind of point resolutely. That's the primary cause of why the master branch has a low performance.
Besides it's not a specialized function, you can use it in other places. It used one place here doesn't mean other branches would use one place too, doesn't mean in the future it would always has one place too.

lostmsu:
@amos402 did you have a chance to compare benchmarks on your branch with `InteropString.GetManagedString` replaced by `Runtime.GetManagedString`?

amos402:
https://gist.github.com/amos402/8decd54dee57fe654e098bc0b15c061a

CPU: Intel(R) Core(TM) i5-3470 CPU @ 3.20GHz
Python Version: 3.8.6

| function | origin | intern-string | without Intern.GetManagedString |
| ---- | ---- | ---- | ---- |
| get_doc | 1.30101 (100.00%%) | 0.89291 (68.63%%) | 1.12968 (86.83%%) |
| get_attr_doc | 1.32372 (100.00%%) | 1.0253 (77.46%%) | 1.25143 (94.54%%) |
| get_attr_doc_concat | 1.32726 (100.00%%) | 1.02239 (77.03%%) | 1.24972 (94.16%%) |

lostmsu (approved):

PR #1255, opened 2020-10-09
Use Type.DefaultBinder for overload resolution

lostmsu:

This replaces our custom overload binding code with the .NET default method binder.

Currently, default binder does not support generic type parameter inference.

This could let us remove large chunks of code, that duplicates standard .NET functionality.

### Checklist

### DIFF
--- a/src/runtime/BorrowedReference.cs
+++ b/src/runtime/BorrowedReference.cs
@@ -22,5 +22,7 @@ namespace Python.Runtime
         {
             this.pointer = pointer;
         }
+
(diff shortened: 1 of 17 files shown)


lostmsu:
`DefaultBinder` turned out insufficient for the purposes of Python.NET, as it does not support implicit conversions. As such, it can not allow `void accept(byte arg)` to be called with `PyInt` argument, even if `PyInt` would have an implicit conversion to `byte`.

This means we will have to try using more complicated (but also much more powerful) `Binder` from `Microsoft.CSharp`.

Issue #1256, opened 2020-10-12
Exception in finalizer of custom encoder


danabr:
### Environment

-   Pythonnet version: master
-   Python version: 3.7
-   Operating System: Windows 10

### Details

I am using a custom encoder defined in Python to work around the conversion of IList to Python lists (see https://github.com/pythonnet/pythonnet/pull/963#issuecomment-642938541):

```python
# Prevent Python.NET from converting IList to Python lists
from Python.Runtime import PyObjectConversions
from Python.Runtime.Codecs import RawProxyEncoder
class  ListAsRawEncoder(RawProxyEncoder):
  __namespace__ = 'Python.Runtime.Codecs'
  def CanEncode(self, clr_type):
    return clr_type.Name == 'IList`1' and clr_type.Namespace == 'System.Collections.Generic'

list_encoder = ListAsRawEncoder()
PyObjectConversions.RegisterEncoder(list_encoder)
```

This works fine until I have called `Python.Runtime.PythonEngine.Shutdown`, and GC kicks in. Then there is this exception:
```
Unhandled exception
Exception thrown:
System.InvalidOperationException: Handle is not initialized.
   at System.Runtime.InteropServices.GCHandle.Free()
   at Python.Runtime.PythonDerivedType.Finalize(IPythonDerivedType obj)
   at Python.Runtime.Codecs.ListAsRawEncoder.Finalize()
```

The crash started happening after #958 was merged.

lostmsu:
@danabr if you can reproduce this reliably, can you try to figure out the stack trace for the preceding `GCHandle.Free` call? Or maybe make a test case?

lostmsu:
@danabr nvm, I am able to reproduce the issue.

lostmsu:
@amos402 this appears to be because `MoveClrInstancesOnwershipToPython` calls `Free` on `GCHandle`, and does not prevent the object from being subsequently finalized.

amos402:
`MoveClrInstancesOnwershipToPython` didn't deal with the `IPythonDerivedType`, the `gcHandle` shouldn't be free twice, it only is called in `Finalize`.
I encountered some kinds of errors by subtypes too when I writing https://github.com/pythonnet/pythonnet/pull/1267, but I can't remember is this or not, if you can reproduce it reliably, can you test on it?
But this reminds me, currently, all the shutdown modes won't able to handle the `IPythonDerivedType`, because it would create dynamic CLR types. This is a defect.
[...]

PR #1260, opened 2020-10-13
Fixed crash in finalizer of CLR types defined in Python, that survive  engine shutdown

lostmsu:

During engine shutdown all links from Python to .NET instances are severed. If an instance of CLR class defined in Python survives the shutdown (for example, a reference is stored in static field) and later gets finalized, it will attempt to severe link again, which is an invalid operation.

The fix is to check if the link has already been severed and skip that step during finalization.

Additionally in this change: refactored `MoveClrInstancesOnwershipToPython`

https://github.com/pythonnet/pythonnet/issues/1256

### Checklist

### DIFF
--- a/src/runtime/classderived.cs
+++ b/src/runtime/classderived.cs
@@ -857,7 +857,7 @@ namespace Python.Runtime
             {
                 if (0 == Runtime.Py_IsInitialized() || Runtime.IsFinalizing)
                 {
-                    self.gcHandle.Free();
+                    if (self.gcHandle.IsAllocated) self.gcHandle.Free();
                     return;
                 }
             }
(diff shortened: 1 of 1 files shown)


lostmsu:
Interesting. It appears, that the fix is legitimate, but the test for it causes segfault in soft shutdown mode with xplat.

lostmsu:
@amos402 I could see it being freed by one of `tp_clear` or `tp_dealloc` and then again by finalizer.

amos402:
https://github.com/pythonnet/pythonnet/blob/927f980264eda7cc8b173461c9777c525dd01179/src/runtime/classderived.cs#L77-L80
It set by a weak GCHandle instead after it freed.

tminka:
I have successfully avoided the problem mentioned by @amos402 [above](https://github.com/pythonnet/pythonnet/pull/1260#issuecomment-726719595) via the following [commit](https://github.com/pythonnet/pythonnet/commit/de7e1ac776545d5834f5d9cc85480d60d8604860).

Issue #1190, opened 2020-07-25
PythonException.Format() throws another PythonException.


tetsushmz:
### Environment

-   Pythonnet version: V2.5.1.
-   Python version: V3.8.4 64 bit
-   Operating System: Windows 10 64 bit

### Details
1. Run the following code in a console application with Visual Studio.

```csharp
using (Py.GIL())
{

    var script = "a=b\n";
    try
    {
        PythonEngine.Exec(script);
    }
    catch (PythonException ex)
    {
        try
        {
            var m = ex.Format();
        }
        catch (Exception exception)
        {
            var t = exception.GetType();
            var s = exception.Message;
        }
    }
}
```

Since "a=b" is invalid, I expected an error message like so returned by `ex.Format()`:

```
Traceback (most recent call last):
  File "<string>", line 1, in <module>
NameError: name 'b' is not defined
```

However, Format() throws another PythonException - "AttributeError : 'str' object has no attribute '__cause__'" with the following stack trace - list(TracebackException(\n', '  File "C:\\Python38\\Lib\\traceback.py", line 479, in __init__\n    if (exc_value and exc_value.__cause__ is not None\n'] .

PR #1265, opened 2020-10-15
Call PyErr_NormalizeException for exceptions

slide:
See https://docs.python.org/3/c-api/exceptions.html#c.PyErr_NormalizeException

This fixes cases where some exceptions are "unnormalized" (see https://docs.python.org/3/c-api/exceptions.html#c.PyErr_NormalizeException). Calling PyErr_NormalizeException will normalize the exception so that any calls to the traceback module functions correctly.

#1190 

### Checklist


lostmsu:
@slide tbh, I think it would be better to be fixed by checking if `__cause__` attribute exists or available via some other means.

I am not sure if the `PyErr_NormalizeException` always does the right thing. I have two issues with it:

1. What if the exception class has `__init__` method, that performs something totally not expected when a value is passed to it? For example, tries to do `int(val)`, and the exception value happens to be something, that causes that to raise another exception?
2. If we do `PyErr_NormalizeException`, and completely discard the original values, subsequent `PythonException.Restore` would not return the error state to be exactly the same as before `PythonException`.

slide:
This function is called in PyErr_Print, which is why I thought it would be safe. Would you rather the call be moved to the Format method itself?

lostmsu:
Can you clarify on behavior of `PyErr_NormalizeException` in regards to reference counts? Does it decref objects passed to it? Do the objects it returns have their refcount > 0? If it can under some circumstances decref the objects passed to it, then you have to ensure to incref them, otherwise pointers in `PythonException` will become invalid.

A test for scenario 1 is still required (e.g. when exception class has `__init__` that throws an exception of its own). If that throws in `Format`, what should we do in that case?

slide:
The reference count stuff is something I tend to struggle with in regards to C# vs. Python stuff. :-) From looking at the source it looks like things would only be DECREF'd in case of an error or if the value is updated to be the correct value (https://github.com/python/cpython/blob/98c4433a81a4cd88c7438adbee1f2aa486188ca3/Python/errors.c#L353 and https://github.com/python/cpython/blob/98c4433a81a4cd88c7438adbee1f2aa486188ca3/Python/errors.c#L336). 

In the case of an error, it looks like PyErr_NormalizeException will chain the new exception and try and normalize it. https://github.com/python/cpython/blob/98c4433a81a4cd88c7438adbee1f2aa486188ca3/Python/errors.c#L366. So, it seems like case 1 is covered internally in the normalization process. I can add something though just to make sure.

lostmsu:
From my reading the code you linked, it looks like you need to pointers before sending them to `_PyErr_NormalizeException`, and then after the function you will own whatever is returned, so no need to incref after.

slide:
I have NOT abandoned this PR, just had some work stuff come up. I will revisit for item 1 above shortly.

slide:
I've been trying to come up with a way to test #1 above and I haven't come up with a good way to do it. I'll continue looking.

slide:
I ran the same code on C Python 3.7.7 interpreter and got this, so the behavior matches:

```
Python 3.7.7 (tags/v3.7.7:d7c567b08f, Mar 10 2020, 10:41:24) [MSC v.1900 64 bit (AMD64)] on win32
>>> class TestException(NameError):
...     def __init__(self, val):
...             super().__init__(val)
...             x = int(val)
...
>>> raise TestException('foo')
  File "<stdin>", line 4, in __init__
ValueError: invalid literal for int() with base 10: 'foo'
```

lostmsu:
@slide I don't think the new test with `__init__` actually uses normalization. What you need is to do `PyErr_NormalizeException(TestException, "hello", null)`.

slide:
You're absolutely right, I was thinking of something else, this makes complete sense.

src/embed_tests/TestPythonException.cs
```diff
@@ -86,9 +86,60 @@ public void TestPythonExceptionFormatNoTraceback()
...
+            }
+        }
+
+        [Test]
+        public void TestPythonException_PyErr_NormalizeException()
+        {
+            using (Py.GIL())
+            {
+                using (PythonEngine engine = new PythonEngine())
+                {
```
lostmsu, line 113: `PythonEngine` has a constructor and a `Dispose` method? o-O

Actually, this might be the reason for the test failures. You should not need neither `Py.GIL`, nor `PythonEngine`. And I believe the tests fail because `PythonEngine.Dispose` shuts Python down, so it crashes trying to release GIL when `Py.GIL()` goes out of scope.

src/embed_tests/TestPythonException.cs
```diff
@@ -86,9 +86,60 @@ public void TestPythonExceptionFormatNoTraceback()
...
+                    var scope = Py.CreateScope();
```
lostmsu, line 114: You should dispose `scope`.

lostmsu (changes requested):

slide:
closing and reopening to get a new build

lostmsu (approved):

slide:
Thanks!

PR #1267, opened 2020-10-22
Weakref support

amos402:
1. A new memory model for holding the CLR handle
2. Unified the meaning of `pyHandle` and `tpHandle` in `ManagedType`, the usage of `obj.tpHandle == obj.tpHandle` represents the `obj` as a `ClassBase` itself is no need anymore
3. `weakref` can be used for CLR object
```python
weakref.ref(clr_obj)
```
Solved the problem that the code like as above would occur a crash because the slot of weakref and the slot of clr handle was overlapping.
4. Minor refactor of `SlotsHolder`

https://github.com/pythonnet/pythonnet/pull/1219#issuecomment-698164386

### Any other comments?

...

### Checklist


src/runtime/managedtype.cs
```diff
@@ -80,48 +79,41 @@ internal void FreeGCHandle()
...
-                {
-                    tp = Marshal.ReadIntPtr(tp, TypeOffset.magic());
-                    var gc = (GCHandle)tp;
-                    return (ManagedType)gc.Target;
-                }
+                return null;
             }
-            return null;
+            IntPtr tp = Runtime.PyObject_TYPE(ob);
+            return GetManagedObject(tp);
```
lostmsu, line 116: I think this code is not equivalent to the original code. The original code did `ob.type.tp_flags.HasFlag(Managed)`, and the new code only does `ob.type.type.tp_flags.HasFlag(Managed)`.
I believe the old check might also need to be necessary.

src/runtime/metatype.cs
```diff
@@ -147,37 +148,26 @@ public static IntPtr tp_new(IntPtr tp, IntPtr args, IntPtr kw)
...
 
             // Hmm - the standard subtype_traverse, clear look at ob_size to
             // do things, so to allow gc to work correctly we need to move
             // our hidden handle out of ob_size. Then, in theory we can
             // comment this out and still not crash.
-            TypeManager.CopySlot(base_type, type, TypeOffset.tp_traverse);
-            TypeManager.CopySlot(base_type, type, TypeOffset.tp_clear);
-
+            //TypeManager.CopySlot(base_type, type, TypeOffset.tp_traverse);
+            //TypeManager.CopySlot(base_type, type, TypeOffset.tp_clear);
```
lostmsu, line 158: Please, remove dead code entirely.

Can you also comment why is this no longer necessary?

src/runtime/runtime.cs
```diff
@@ -1991,6 +1996,9 @@ internal static IntPtr PyType_GenericAlloc(IntPtr type, long n)
         [DllImport(_PythonDll, CallingConvention = CallingConvention.Cdecl)]
         internal static extern void PyObject_GC_UnTrack(IntPtr tp);
 
+        [DllImport(_PythonDll, CallingConvention = CallingConvention.Cdecl)]
+        internal static extern void PyObject_ClearWeakRefs(IntPtr obj);
```
lostmsu, line 2000: We should try to use `*Reference` types for clarity for all new PInvoke functions.

src/runtime/runtime.cs
```diff
@@ -1982,6 +1984,9 @@ internal static IntPtr PyType_GenericAlloc(IntPtr type, long n)
         [DllImport(_PythonDll, CallingConvention = CallingConvention.Cdecl)]
         internal static extern IntPtr _PyObject_GetDictPtr(IntPtr obj);
 
+        [DllImport(_PythonDll, CallingConvention = CallingConvention.Cdecl)]
+        internal static extern IntPtr _PyObject_GC_Calloc(IntPtr basicsize);
```
lostmsu, line 1988: It is a very bad idea to use private function. Why is it required? Can the same be done with only public functions?

src/runtime/typemanager.cs
```diff
@@ -436,11 +456,19 @@ internal static IntPtr CreateMetaType(Type impl, out SlotsHolder slotsHolder)
...
             // certain functions in the C runtime to work correctly with it.
 
-            IntPtr type = AllocateTypeObject("CLR Metatype");
             IntPtr py_type = Runtime.PyTypeType;
+            var heapTypeSize = (int)Marshal.ReadIntPtr(py_type, TypeOffset.tp_basicsize);
+            Debug.Assert(heapTypeSize == TypeOffset.members);
+            int metaSize = heapTypeSize + Marshal.SizeOf(typeof(ClrMetaTypeEx));
 
-            Marshal.WriteIntPtr(type, TypeOffset.tp_base, py_type);
+            IntPtr type = Runtime._PyObject_GC_Calloc(new IntPtr(metaSize));
```
lostmsu, line 464: Specifically, can we use `PyType_FromSpec` here?

src/runtime/typemanager.cs
```diff
@@ -784,21 +816,25 @@ public SlotsHolder(IntPtr type)
 
         public void Set(int offset, ThunkInfo thunk)
         {
+            if (_slots == null) _slots = new Dictionary<int, ThunkInfo>();
```
lostmsu, line 819: This is not thread safe. I am unsure holding GIL ensures this is properly synchronized.

lostmsu (changes requested):

src/runtime/classbase.cs
```diff
@@ -303,44 +301,40 @@ public static IntPtr tp_repr(IntPtr ob)
...
             base.OnSave(context);
-            if (pyHandle != tpHandle)
-            {
-                IntPtr dict = GetObjectDict(pyHandle);
-                Runtime.XIncref(dict);
-                context.Storage.AddValue("dict", dict);
-            }
+            IntPtr dict = GetObjectDict(pyHandle);
+            Runtime.XIncref(dict);
+            Runtime.XIncref(tpHandle);
```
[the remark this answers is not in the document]
amos402: That's the Save for domain reload, after `OnLoad`, the ref count would be balanced.

src/runtime/classmanager.cs
```diff
@@ -443,6 +443,7 @@ private static ClassInfo GetClassInfo(Type type)
                         }
                         // Note the given instance might be uninitialized
                         ob = GetClass(tp);
+                        ob.IncrRefCount();
```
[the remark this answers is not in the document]
amos402: `GetClass` returned a borrowed reference, memers in `ci.members` would decref at once, thus add a reference count here make the reference balance.

src/runtime/exceptions.cs
```diff
@@ -82,6 +82,13 @@ public new static IntPtr tp_str(IntPtr ob)
             }
             return Runtime.PyUnicode_FromString(message);
         }
+
+        public static int tp_init(IntPtr ob, IntPtr args, IntPtr kwds)
+        {
+            Exceptions.SetArgsAndCause(ob);
```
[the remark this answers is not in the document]
amos402: What comment?

src/runtime/typemanager.cs
amos402: You can assume it's safe when you holding the GIL, all these slots stuff are linear behaviors for the type initialization.

src/runtime/typemanager.cs
amos402: Maybe not, `PyType_FromSpec` still use `PyType_Type` to allocate memory, which means it would occur the same problem as before.

src/runtime/typemanager.cs
```diff
@@ -289,10 +303,16 @@ internal static IntPtr CreateType(ManagedType impl, Type clrType)
...
+                var typePtr = ClrMetaTypeEx.FromType(type);
+                typePtr->ClrHandle = (IntPtr)gc;
+                typePtr->ClrHandleOffset = (IntPtr)magicOffset;
+            }
 
             // Set the handle attributes on the implementing instance.
-            impl.tpHandle = type;
+            impl.tpHandle = Runtime.PyCLRMetaType;
+            Runtime.XIncref(type);
             impl.pyHandle = type;
```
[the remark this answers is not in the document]
amos402: Actually, we can't rely on GC for such low-level objects, it may cause troubles on Shutdown, so it's not much too different.

src/runtime/typemanager.cs
```diff
@@ -189,17 +194,25 @@ internal static IntPtr CreateType(ManagedType impl, Type clrType)
...
+                tp_dictoffset = ManagedExceptionOffset.ob_dict;
+                tp_weaklistoffset = 0;
+                ob_size = ManagedExceptionOffset.Size;
+                magicOffset = ManagedExceptionOffset.ob_data;
+            }
+            else
+            {
+                tp_dictoffset = baseOffset + ManagedDataOffsets.ob_dict;
+                tp_weaklistoffset = baseOffset + ManagedDataOffsets.ob_weaklist;
+                ob_size = baseOffset + ManagedDataOffsets.Size;
```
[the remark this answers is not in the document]
amos402: Of course it's the size of CLR wrapped type. It corresponds to`tp_basicsize` which means it represents the size of a C Python object that included clr information.

src/runtime/runtime.cs
```diff
@@ -2123,6 +2131,16 @@ internal static void Py_CLEAR(ref IntPtr ob)
             ob = IntPtr.Zero;
         }
 
+        internal static unsafe void Py_SETREF(IntPtr ob, int offset, IntPtr target)
```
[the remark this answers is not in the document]
amos402: The original version is the C macro, the implementation used pointer offset, maybe is not suited for such a high-level type to be involved. I want to keep such things simple, If you need that, maybe create a high-level wrapper.😓

src/runtime/runtime.cs
amos402: That makes sense, but can we just replace them all at once.😅

src/runtime/runtime.cs
amos402: I'm afraid there's no such public function now, it's required to allocate a custom size Python GC object.

src/runtime/metatype.cs
amos402: These slots can inherit from the base type automatically. Emmm, I can add some comments here, but it still needs the dead code comment as an explanation.

src/runtime/managedtype.cs
amos402: Yes, it's not equivalent. It becomes simple since the memory model been unified.
You can treat the old code was a trick: ob may be a type or an object. You can also see I removed those code `tpHandle == pyHandle  => means it's a ClassBase object`. For now, if the type of `ob` is managed, then the ob must be a managed object, it's obvious on the logic.

src/runtime/managedtype.cs
```diff
@@ -80,48 +79,41 @@ internal void FreeGCHandle()
...
+                return null;
+            }
+            IntPtr tp = Runtime.PyObject_TYPE(ob);
+            var flags = Util.ReadCLong(tp, TypeOffset.tp_flags);
+            if ((flags & TypeFlags.Managed) == 0)
+            {
+                return null;
+            }
+            int offset = ObjectOffset.magic(tp);
+            if (offset == 0)
```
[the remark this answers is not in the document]
amos402: A managed wrapped type which released the corresponded GCHandle, that would be 0.

src/runtime/interop.cs
```diff
@@ -182,6 +159,26 @@ static OriginalObjectOffsets()
...
+        public static int ob_weaklist { get; private set; }
+
+        public static int Size { get; private set; }
+
+        static ManagedExceptionOffset()
+        {
+            int baseOffset = OriginalObjectOffsets.Size + ExceptionOffset.Size();
+            ob_data = baseOffset + ManagedDataOffsets.ob_data;
+            ob_dict = baseOffset + ManagedDataOffsets.ob_dict;
+            ob_weaklist = baseOffset + ManagedDataOffsets.ob_weaklist;
```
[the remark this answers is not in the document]
amos402: It could be, this offset corresponded to the `ClrPyObjectEx`. But I can't see the point, since `ExceptionOffset` is fixed, it corresponded to `PyBaseExceptionObject` in the `pyerrors.h`.

src/runtime/managedtype.cs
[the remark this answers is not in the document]
lostmsu: But `magic` is the offset of the handle, not the handle itself.

src/runtime/typemanager.cs
lostmsu: Actually, I think I solved this one. Please see https://github.com/pythonnet/pythonnet/commit/9682dc6e3b3f9231d7a801ae3202a65ed68fbcca

src/runtime/classmanager.cs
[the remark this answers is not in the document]
lostmsu: > memers in ci.members would decref

Where do they do it? I think this only happens during engine shutdown

src/runtime/exceptions.cs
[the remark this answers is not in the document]
lostmsu: NIT: You seem to have moved this from `CLRObject`, and removed
```
// Fix the BaseException args (and __cause__ in case of Python 3)
// slot if wrapping a CLR exception
```

src/runtime/metatype.cs
lostmsu: The comment above also seems dangerously wrong.

src/runtime/runtime.cs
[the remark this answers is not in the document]
lostmsu: `NewReference` is not high-level. It is just a struct.

src/runtime/native/GeneratedTypeOffsets.cs
```diff
@@ -7,6 +7,9 @@ namespace Python.Runtime.Native
     [StructLayout(LayoutKind.Sequential)]
     abstract class GeneratedTypeOffsets
     {
+        // XXX: Use field to trick the public property finding
+        public readonly int Size;
+
```
lostmsu, line 12: Why do you think this is better, than the single line you removed from ABI.cs? This is hacky and uses reflection.

src/runtime/native/TypeOffset.cs
```diff
@@ -76,7 +81,7 @@ internal static void Use(ITypeOffsets offsets)
                 int value = (int)sourceProperty.GetValue(offsets, null);
                 offsetProperty.SetValue(obj: null, value: value, index: null);
             }
-
+            Size = ((GeneratedTypeOffsets)offsets).Size;
```
lostmsu, line 84: Definitely should not cast there. The whole point of the interface is that we eventually want users to be able to supply their own `ITypeOffsets` if their Python build is very special.

src/runtime/classmanager.cs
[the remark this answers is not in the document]
amos402: https://github.com/pythonnet/pythonnet/blob/a0d1a3dbd6d57103f4494468512e268e570344bb/src/runtime/classmanager.cs#L219
It's not related to the shutdown.

src/runtime/exceptions.cs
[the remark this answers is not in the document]
amos402: The comment on `SetArgsAndCause` won't be enough?

src/runtime/runtime.cs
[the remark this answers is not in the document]
amos402: Then you have to pull the pointer and add an offset, what a tortuous path.

src/runtime/typemanager.cs
amos402: No, it didn't, see https://github.com/python/cpython/blob/fbc7723778be01b8f3bb72d2dcac15ab9fbb9923/Objects/typeobject.c#L1051-L1082

src/runtime/native/GeneratedTypeOffsets.cs
amos402: It's not related to that line, the reason for removing that line just because was because that's not the fixed slot anymore.
Well, it was just need something to store the size, tp_basicsize might capable do that too, this part can be improved.

src/runtime/managedtype.cs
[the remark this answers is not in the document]
amos402: Ah, my bad, it should represent a managed type without GCHandle mapped, like `CLR Metatype`. You can think the new module of extension contains such a slot, but it might be 0.

src/runtime/runtime.cs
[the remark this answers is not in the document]
lostmsu: It is better, than accidentally forgetting to incref the `target` in the caller.

src/runtime/managedtype.cs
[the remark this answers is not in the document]
lostmsu: > it might be 0

what is the scenario for that?

src/runtime/typemanager.cs
lostmsu: Yes, it did, see https://github.com/python/cpython/blob/8b3414818f5289eac530bf38bcfbd7b2b851805c/Include/cpython/objimpl.h#L21 :-P

src/runtime/native/GeneratedTypeOffsets.cs
lostmsu: This field has the same purpose and value as the one you removed from ABI - `sizeof(PyTypeObject)`

src/runtime/classbase.cs
```diff
@@ -303,44 +301,40 @@ public static IntPtr tp_repr(IntPtr ob)
         public static void tp_dealloc(IntPtr ob)
         {
             ManagedType self = GetManagedObject(ob);
+            if (Runtime.PyType_SUPPORTS_WEAKREFS(Runtime.PyObject_TYPE(ob)))
```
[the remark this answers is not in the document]
amos402: I rather split up the Runtime feature and PInvoke Calls. All interop calls as another file. As for the customized interop methods, use the partial class to split them, maybe? Since they all start with PyXXX_ and all from CPython, it would be convenient to use them by typing xxx.PyXXX_XXXX. PyXXX is enough to distinguish their category, disperse them may not good for use.

src/runtime/typemanager.cs
amos402: It's not about the type size which the MetaType can generate. It's about the size of MetaType itself. You can't create the MetaType by using `PyType_GenericAlloc(PyTypeType, 0)`, unless you
```c
old_size = PyTypeType.tp_basicsize;
PyTypeType.tp_basicsize += x;
PyType_GenericAlloc(PyTypeType, 0);
PyTypeType.tp_basicsize = old_size;
``` 
As long as it claimed itself as a managed type, it should have a slot indicate the place of its instances where the related GCHandle would be stored. As for now, the MetaType doesn't attach any C# object, so it remains 0. Look back to the master, if you put the PyCLRMetaType to `GetManagedObject`, actually it would read the memory it should not read(the part of the reserved `tp_itemsize`). The code doesn't explicitly call something like that, but something may trigger it does, like domain reload may visit it through PyVisit. Even it won't crash right now, but it's hard to say that's a kind of sanity code. 
Of course, you can just remove the Manged flag from PyCLRMetaType to avoid that. But I suggest following the old rule: the type of a CLR type is a managed type.
Another approach is to allocate with PyObject_Malloc, but that means it has to be initialized manually, e.g. call `_Py_NewReference` manually on debug version, add it to GCTrack.
Nevertheless, the point is, the type must clarify itself what's itself and what it can do, that commit avoided some problems, but didn't solve completely.

src/runtime/metatype.cs
amos402: Hmmm, that was an old comment, that time before we unify the shutdown progresses, it can't and won't use tp_traverse and tp_clear. Seems it can be removed at one time.

src/runtime/native/TypeOffset.cs
amos402: Use PyTypeType.tp_basicsize as size of PyHeapTypeObject directly instead of calculate it now.

src/embed_tests/TestClass.cs
```diff
@@ -0,0 +1,82 @@
...
+        public void Dispose()
+        {
+            PythonEngine.Shutdown();
+        }
+
+        [Test]
+        public void WeakRefForClrObject()
+        {
+            var obj = new MyClass();
+            using (var scope = Py.CreateScope())
```
lostmsu, line 34: NIT: use C# 9 using syntax

src/runtime/classbase.cs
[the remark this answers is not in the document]
lostmsu: @amos402 I like the idea. Can you move this to a class `CPythonMacros`? This would make it absolutely clear, that the method is a direct port of CPython code, and make it easy to find the original

src/runtime/metatype.cs
lostmsu: Any dead code should be removed immediately.

src/runtime/metatype.cs
```diff
@@ -360,4 +356,18 @@ public static IntPtr __subclasscheck__(IntPtr tp, IntPtr args)
             return DoInstanceCheck(tp, args, true);
         }
     }
+
+
+    [StructLayout(LayoutKind.Sequential)]
+    struct ClrMetaTypeEx
+    {
+        public IntPtr ClrHandleOffset;
+        public IntPtr ClrHandle;
```
lostmsu, line 365: These two refer to different handles, yet are named similarly, which is confusing.

Besides, in C# 9 we can start using `nint` instead of `IntPtr` for non-pointers.

src/runtime/native/TypeOffset.cs
```diff
@@ -88,7 +94,7 @@ internal static void Use(ITypeOffsets offsets)
                 int value = (int)sourceProperty.GetValue(offsets, null);
                 offsetProperty.SetValue(obj: null, value: value, index: null);
             }
-
+            Size = (int)Marshal.ReadIntPtr(Runtime.PyTypeType, tp_basicsize); ;
```
lostmsu, line 97: NIT: Probably not an issue, but this assumes `Runtime` is already initialized, and the rest of the code in this file does not.
I'd rather you move this `Size` elsewhere, where it is known for sure `Runtime` is initialized.

Also: extra semicolon.

src/runtime/pyobject.cs
```diff
@@ -119,7 +119,9 @@ public static PyObject FromManagedObject(object ob)
                 Runtime.XIncref(Runtime.PyNone);
                 return new PyObject(Runtime.PyNone);
             }
-            IntPtr op = CLRObject.GetInstHandle(ob);
+            IntPtr op = typeof(Exception).IsAssignableFrom(ob.GetType()) ?
+                Exceptions.GetExceptHandle((Exception)ob)
+                : CLRObject.GetInstHandle(ob);
```
lostmsu, line 124: I feel like `Exceptions.GetExceptionHandle` should be built into `CLRObject.GetInstHandle`.

src/runtime/runtime.cs
```diff
@@ -522,11 +526,21 @@ private static void MoveClrInstancesOnwershipToPython()
...
                 {
-                    obj.gcHandle.Free();
+                    System.Diagnostics.Debug.Assert(obj.gcHandle == default);
+                    continue;
+                }
+                if (obj.RefCount > 1)
+                {
+                    obj.FreeGCHandle();
+                    Marshal.WriteIntPtr(obj.pyHandle, ObjectOffset.magic(obj.tpHandle), IntPtr.Zero);
                 }
```
lostmsu, line 542: Why can't you do this unconditionally? This condition makes reasoning about the state much harder.

src/runtime/typemanager.cs
lostmsu: As far as I can see, CLR Metatype does not contain `GCHandle`, so it does not need `GetManagedObject` to work on its instances. So I think my implementation linked above is correct. Perhaps `PyCLRMetaType` should not be a managed object then?

lostmsu (changes requested):

src/runtime/runtime.cs
amos402: If `obj.RefCount == 1`, the next `XDecref` would dealloc the obj, thus it's no need to call these things.

src/runtime/metatype.cs
amos402: What name do you prefer?

src/runtime/pyobject.cs
amos402: The exception object is a special case, we should keep it outside from the generic case `GetInstHandle`, for example, if it has multiple special cases instead of one, you won't want to contaminate the CLRObject methods, right? Just like I removed the special case code from the CLRObject's constructor.

src/runtime/typemanager.cs
amos402: It indeed has a bit of struggle with that, the PyCLRMetaType it's the ancestor of all managed types, if it claimed it's not a managed type, it against the commonsense. Currently, it doesn't need to be bound with a clr instance even it can. Well, just an identity problem.

src/runtime/native/TypeOffset.cs
amos402: `ABI.Initialize` called after all  Runtime.PyXXXX initialized, so it may not be a problem, if it does, bugs...

src/runtime/classbase.cs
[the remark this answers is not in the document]
amos402: Another name of the class? Won't it better just use a partial class of Runtime into another file?

src/runtime/metatype.cs
lostmsu: This is just a NIT. Maybe `TypeClrHandle` and `InstanceClrHandleOffset`?

src/runtime/runtime.cs
lostmsu: Is this similar to `tp_clear` in any way? I would suggest to extract these two updates (`FreeGCHandle` and zeroing out) into a method.

The point being `tp_clear` has certain semantics, so it is easier to understand code containing a call to it, rather than has same actions inline.

Also, if `RefCount` will correctly return 0, I'd move this `if` below.

src/runtime/classbase.cs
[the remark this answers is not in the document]
lostmsu: Actually, you are probably right. It will work as is.

Ideally, we'd split `Runtime` into classes the same way they are grouped in CPython, so that would no longer be a single giant file. E.g. `Runtime.PyObject_Str` -> `PyObject.Str`, `Runtime.PyDict_GetItem` -> `PyDict.GetItem`, etc.

src/runtime/typemanager.cs
lostmsu: I think it makes sense, that CLR metatype is **not** a managed type. In my mind managed type == type, whose implementation is managed == has a managed object, that implements functionality == has `GCHandle` pointing to that object. CLR metatype does not have a `GCHandle`.

lostmsu (changes requested):
I think it is good, except the `_PyObject_GC_Calloc`, which should be replaced as suggested.

src/runtime/runtime.cs
lostmsu: @amos402 are you prepared to make this happen? )

src/runtime/runtime.cs
amos402: Not exactly, `tp_clear` is used to clear the Python object reference in order to break reference cycles. These codes are used to disconnect the relation between Python and clr, if the RefCount reached 0, their relation would naturally be gone.
Their relation needs to be kept if  RefCount > 1 because in `tp_dealloc`, it needs to obtain the bound object.
`tp_dealloc` didn't zero out the handle because this chunk of memory would be freed soon after `tp_dealloc`, extract these into a method is not really necessary.

lostmsu:
I believe this has been superseded by one of the major changes in 3.0.

This works on the latest master:
```python
>>> import clr
>>> from System import Uri
>>> u = Uri('http://ddg.com')
>>> import weakref
>>> weak = weakref.ref(u)
>>> u
<System.Uri object at 0x000001ED729D8240>
>>> u.ToString()
'http://ddg.com/'
>>> weak().ToString()
'http://ddg.com/'
```

Issue #1099, opened 2020-04-03
Disambiguating an overloaded method with multiple parameters


tengounapregunta:
### Environment

-   Pythonnet version: 2.3.0 and 2.4.0
-   Python version: 3.6.5
-   Operating System: W10

### Details

- I'm having issues disambiguating an overloaded C# method that has multiple parameters. I've found #935 which is a very similar issue but can't get it to work. I'm following the same steps but I still get the same error.

The C# side
```C#
    public enum MyEnum: byte
    {
        Success= (byte)'S',
        Error = (byte)'E'
    }

    public class MyClass
    {
        public Dictionary<Enum, BlockingCollection<ISomeInterface>> MyDictionary { get; }

        public MyClass()
        {
            MyDictionary = new Dictionary<Enum, BlockingCollection<ISomeInterface>>(); 
        }
    }
```

On the python side, I began with this code
```python
    from System.Collections.Concurrent import BlockingCollection
    from MyModule import MyClass, ISomeInterface, MyEnum

    obj = MyClass()
    collection = BlockingCollection[ISomeInterface]()
    obj.MyDictionary.Add(MyEnum.Success, collection)
```
This fails with `TypeError: No method matches given arguments for Add`. I've tried what seemed to solve it for #935 but get the same results

```python
    from System.Collections.Concurrent import BlockingCollection
    from System import Enum
    from MyModule import MyClass, ISomeInterface, MyEnum

    obj = MyClass()
    collection = BlockingCollection[ISomeInterface]()
    obj.MyDictionary.Add[Enum, BlockingCollection[ISomeInterface]](MyEnum.Success, collection)
```

This fails with `TypeError: No match found for given type params`

-  The output of `print(obj.MyDictionary.Add.__overloads__)` is

```python
    Void Add(System.Enum, System.Collections.Concurrent.BlockingCollection`1[MyModule.ISomeInterface])
```

**UPDATE**

Running this code alone actually works, `add` shows as a bound method in the debugger:

```python
add = obj.MyDictionary.Add.Overloads[Enum, BlockingCollection[ISomeInterface]]
```

However, attempting to call the function always fails. I've attempted passing System.byte and 0, instead of the enum, but none of them work.
```python
add(MyEnum.Success, collection)
obj.MyDictionary.Add[Enum, BlockingCollection[ISomeInterface]](MyEnum.Success, collection)
[...]

PR #1278, opened 2020-11-12
DLR-based overload resolution

lostmsu:

This is an attempt to replace our custom (and in many cases borked) overload resolution with C# [`RuntimeBinder`](https://docs.microsoft.com/en-us/dotnet/api/microsoft.csharp.runtimebinder.binder?view=netframework-4.8), that is used to implement C#'s `dynamic` keyword.

This should make our overload resolution nearly identical to C# overload resolution. It has full handling for: `params`, implicit conversions, type hierarchy, generics, overload prioritization, keyword arguments.

A bonus feature is that supporting newer C# features in method binding should be as easy as simply updating the `Microsoft.CSharp` NuGet.

### Major consequences and further TODO

1. During overload selection, there will be no more multiple attempts to marshal Python objects to C# (was: one per attempted overload). All parameters will be decoded exactly once, and then C# binder will be used to pick the correct overload.
2. Because of 1, decoders should return object instances, that are compatible with expected .NET types. For example: decode of `builtins.int` should return an instance of `PyInt`, which then should implement static or dynamic (e.g. [`TryConvert`](https://docs.microsoft.com/en-us/dotnet/api/system.dynamic.dynamicobject.tryconvert)) conversion to `System.Int32` and/or `System.Int64` in order to be used with `void Foo(int arg)`.
3. The mechanism to select a specific overload from Python would have to be reworked. We need a clear separation between explicitly specifying generic parameters vs explicitly specifying parameter types.

https://github.com/pythonnet/pythonnet/issues/1099 , specifically https://github.com/pythonnet/pythonnet/issues/1099#issuecomment-640605647

### Other comments

I am not sure how we could approach getting this done. Until the entirety of binding features are re-implemented with DLR, current tests will be failing. Perhaps a separate branch would be a good approach, where changes could be reviewed incrementally. @amos402 , @filmor , thoughts?

### Checklist

### DIFF
--- a/src/embed_tests/TestDomainReload.cs
+++ b/src/embed_tests/TestDomainReload.cs
@@ -94,7 +94,7 @@ clr.AddReference('{0}')
 from Python.EmbeddingTest.Domain import MyClass
 obj = MyClass()
 obj.Method()
-obj.StaticMethod()
+MyClass.StaticMethod()
 obj.Property = 1
 obj.Field = 10
 ", Assembly.GetExecutingAssembly().FullName);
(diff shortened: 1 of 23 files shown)


lostmsu:
Closing for now as it is still in development

filmor:
@lostmsu Do you plan to pick this one up again?

lostmsu:
@filmor not any time soon

PR #1279, opened 2020-11-12
Use .NET Core 3.1 LTS for tests (instead of 2.0)

lostmsu:

.NET Core 2.0 is no longer supported. This moves tests to .NET Core 3.1 LTS, which should be supported [until Dec 2022](https://dotnet.microsoft.com/platform/support/policy/dotnet-core).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,9 +8,9 @@ python:
 
 env:
   matrix:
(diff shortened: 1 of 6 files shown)


lostmsu:
Travis actually passed, but for some reason was not recorded correctly: https://travis-ci.org/github/pythonnet/pythonnet/builds/743309873

Since there are no functional changes, and tests still pass, I am going to merge this.

PR #1284, opened 2020-11-17
Ensure methods of Object are also available on interface objects

danabr:

Allows users to call `ToString`, `GetHashCode` and `GetType` on objects wrapped in interface objects.

No

### Any other comments?

No

### Checklist


filmor (approved):

filmor:
Looks good to me, would you mind adding a Changelog entry for this?

danabr:
I was thinking no change log entry was needed, as this was mostly a fallout from #1240 . But now I realize it was actually the case before that change as well for interface objects that had been explicitly created. I'll write something.

(Not sure what is up with the codecov though, it says coverage became worse for a file not touched by the PR?)

lostmsu (approved):

PR #1275, opened 2020-11-09
Adds test cases for member changes during a domain reload

BadSingleton:
The tests are marked as expected failures for now.

### What does this implement/fix? Explain your changes.

This adds a test harness/framework for renaming/removing members during domain reload. Actual fixes for these tests will com in other pull request(s).

### Does this close any currently open issues?

This will help start addressing #1250

### Any other comments?

This supersedes, in part, #1269 

### Checklist

Check all those that are applicable and complete.

-   [x] Make sure to include one or more tests for your change
-   [ ] If an enhancement PR, please create docs and at best an example
-   [x] Add yourself to [`AUTHORS`](../blob/master/AUTHORS.md)
-   [ ] Updated the [`CHANGELOG`](../blob/master/CHANGELOG.md)

Issue #1268, opened 2020-10-22
What should reload reload?


benoithudson:
**Background:**
Amos created the soft reload branch which is now merged to master. It's a large amount of work.

We found #1250 as an issue in that code. Now the question is: disable some functionality or fix it?

The setting is as follows: I have code in Python. It stores references to objects in C#. That means there are proxies in Python pointing to objects in C#.

When C# unloads the domain, those proxies are now pointing to garbage. In the setting where this happens (Unity domain reload), C# will soon load a new domain. The new domain will be a lot like the old one.

Amos implemented a mechanism in `runtime_data.cs` to reconnect the proxies. The idea is we know all the proxies that were ever build (and not yet deleted). Serialize that C# data using `System.Runtime.Serialization`. When the new domain arrives, deserialize.

**Problem:**
Deserialization may fail. See #1250 for one example.

This happens in practice when the Unity user renames a field, a class, a namespace, or an assembly -- and on the Python side we had a reference to one of those. E.g. if we handed a C# static function to a python function that expects a callable, we'd have a FieldObject proxy in the code. If the user renames or deletes the C# static function in their code, when Unity reloads the domain, that FieldObject won't be able to reload.

This will cause an exception when trying to reinitialize Python for .NET on the C# side, and therefore the system is now in an undefined state.

The `System.Runtime.Serialization` is not fault tolerant. It's not possible to provide a callback if deserialization fails for a certain object. If you control a class you can specify custom deserialization (and serialization) for that class, and you can specifically code your stuff to be fault tolerant, but if you recursively serialize any other classes, upon deserialization everything that got serialized gets deserialized as a flat list; any unhandled exceptions there will stop the deserialization.

**What are the user requirements anyway?**

For a Unity user, the requirement is that when I change my C# code and Unity automatically reloads, the Python side doesn't throw a fit no matter how mean I've been to it.

This is mainly two things:
1. If in Python I `import MyModule` where MyModule is in C# (this includes `UnityEngine`), I need some way to be able to use that even after a domain reload. Ideally the module reloads itself.
2. If in Python I somehow store a reference to a proxy to C#, I need that the system is able to survive domain reload: it does not crash and it does not leave the Python for .NET data structures in an unstable state.

**Possible solution 1: fix the bugs**
[...]

Issue #1250, opened 2020-10-09
Soft reload throws in deserialization in a common workflow


benoithudson:
### Environment

-   Pythonnet version: current master (with soft reload).

This is in context of Unity Editor.

### Details

0. Use the soft-reload branch of pythonnet within Unity.
1. Create an asmdef called MyAssembly in Unity.
2. In the same folder, create a file Foo.cs
```
namespace Foo
{
  public class Bar
  {
    public static void Quux() { UnityEngine.Debug.Log("quux"); }
  }
}
```
3. In the Unity Python console:
```
import clr
clr.AddReference("MyAssembly")
import Foo
Foo.Bar.Quux()
```
4. This will print `quux` in the Unity console.
5. In Unity, modify Foo.cs e.g. add a space in it.
6. Allow Unity to reload the domain.
7. This works!
8. In Unity, modify Foo.cs and rename Quux to Quux2.

Expected: Python for .NET is available after reload. Running Python code again will raise AttributeError.

Actual: Python for .NET throws `SerializationException: Cannot get the member 'Quux'.` I can't even run the Python code again; it fails on `import clr`. I must restart Unity.

Similarly if I rename the class, the namespace, or the assembly.

The problem appears to be in ManagedType subclasses. These are `[Serializable]` but they point to things like `Type` whose deserialization may fail, destroying everything.

I'm working on a proof of concept so that we control the deserialization and we can replace the failed deserialize with a dummy or a `null`, which only fails if this eventually gets used. I'll post results tomorrow hopefully on a first attempt.

benoithudson:
Proof of concept:
[1250.txt](https://github.com/pythonnet/pythonnet/files/5356573/1250.txt)

This fixes this particular bug. I can reload, no error. This isn't a PR yet since there's no testing; and I haven't even thought about performance.

We'll have to audit all the ManagedType subclasses and fix them in this style.

benoithudson:
#1287 fixes much more than this test case.

Issue #957, opened 2019-09-17
Python can't reliably re-initialize; new proposal for Unity's domain reload


benoithudson:
**Problem**: [bpo-37931](https://bugs.python.org/issue37931) demonstrates that even Python itself can't re-initialize on all platforms without crashing. Major third-party packages including numpy and PySide also crash when you Py_Initialize after a Py_Finalize and reimport the package.

The reason we Py_Finalize in Shutdown is because otherwise, the python side has live objects that have function pointers into C#. If the C# domain is reloaded (which frequently happens in the Unity3d editor) then those pointers are pointing to garbage. So, calling any C#-implemented function on those python objects causes the PC to jump to garbage and thus "undefined behaviour" (usually a crash).

New proposed **solution**:
1. Don't Py_Finalize in Shutdown.
2. Find all the function pointers from Python to C# and set them to a reasonable value so that they have a reasonably pythonic exception rather than jumping to garbage.

@BadSingleton on our team has prototyped this. It seems the only place we jump to C# is in the `tp_*` slots of a type. Setting them to null, as most slots are set in Python's noddy example, seems to work fine (except for the slots for garbage collection, which must point to something, so we still need that silly native code page for now).

Example: we have a proxy to a C# object in python and we call `ToString()` on it. Before domain reload, Python calls the tp_getattro slot which is implemented in C# and it does the reflection to find ToString, and then it calls tp_call which is implemented in C# and it calls the function. After domain reload, Python sees that tp_getattro is null so it uses the default getattr implementation, which has never heard of ToString, so it throws an AttributeError. If we cached ToString on the python side, then Python looks at the tp_call slot which is null and throws a TypeError because it's not callable.

benoithudson:
The MVP here is just: don't crash.

A slight improvement is to see if it's easy to get an "underlying C# object has been deleted" exception, so the PySide crew feels right at home.

benoithudson:
I'm putting this up as a point of discussion for next week's meeting.

I want to make sure there's no other ways to jump from python into C#.

BadSingleton:
Speaking of prototype: https://github.com/BadSingleton/pythonnet/tree/issue-957-prototype
It is a fairly simple one; a proof of concept for a simple repro case to see if it was doable

benoithudson:
Compare link (for ease of reading):
https://github.com/pythonnet/pythonnet/compare/master...BadSingleton:issue-957-prototype

Not ready to make it a PR, we need much more testing.

[...]

PR #1287, opened 2020-11-19
Domain reload test cases fixes

BadSingleton:

This PR adds the fixes for the xfail tests in #1275 .

Serialization of System.Type, MemberInfo and MethodBase is now string based. At de-serialization, use reflection to attempt to recreate the object, which now fails safely instead of throwing a SerializationException during the de-serialization of the whole data stream. Appropriate Exceptions will now be raised when the Maybe*'s Value property is accessed.

This addresses #957 and #1250 and offers an answer to #1268 .

### Any other comments?

TODO: add changelog entries once the fixes are approved. Also, add docs on best practices with domain reloads.

### Checklist


src/runtime/moduleobject.cs
```diff
@@ -341,6 +341,17 @@ protected override void OnSave(InterDomainContext context)
...
             Runtime.XIncref(dict);
+            // destroy the cache(s)
+            foreach (var pair in cache)
+            {
+                Runtime.PyDict_DelItemString(dict, pair.Key);
+                pair.Value.DecrRefCount();
+            }
+            // Trying to remove a key that's not in the dictionary may 
+            // raise an error. We don't care about it.
+            Runtime.PyErr_Clear();
```
lostmsu, line 352: You must check for the specific error type after each call. See `PyObject.GetAttrOrDefault` for example.

src/runtime/typemanager.cs
```diff
@@ -76,11 +77,22 @@ internal static void RestoreRuntimeData(RuntimeDataStorage storage)
         {
             Debug.Assert(cache == null || cache.Count == 0);
             storage.GetValue("slots", out _slotsImpls);
-            storage.GetValue("cache", out cache);
-            foreach (var entry in cache)
+            var _cache = new Dictionary<MaybeType, IntPtr>();
+            storage.GetValue("cache", out _cache);
```
lostmsu, line 81: the value you put into `_cache` here will always be overwritten.

You can (and should) use `out var _cache`.

src/runtime/typemanager.cs
```diff
@@ -76,11 +77,22 @@ internal static void RestoreRuntimeData(RuntimeDataStorage storage)
...
+                Type type = null;
+                try
+                {
+                    type = entry.Key.Value;
+                }
+                catch
+                {
+                    Runtime.XDecref(entry.Value);
+                    continue;
+                }
```
lostmsu, line 93: Is there a specific exception we could handle?
Can the exception be avoided in the first place by adding a precheck?

src/runtime/metatype.cs
```diff
@@ -103,9 +104,16 @@ public static IntPtr tp_new(IntPtr tp, IntPtr args, IntPtr kw)
             var cb = GetManagedObject(base_type) as ClassBase;
             if (cb != null)
             {
-                if (!cb.CanSubclass())
+                try
                 {
-                    return Exceptions.RaiseTypeError("delegates, enums and array types cannot be subclassed");
+                    if (!cb.CanSubclass())
```
lostmsu, line 109: If this is the only place where `CanSubclass` is used, can it be inlined here, and raising an exception bypassed?

lostmsu (changes requested):

src/runtime/Python.Runtime.csproj
```diff
@@ -1,191 +1,192 @@
...
-  <PropertyGroup>
-    <TargetAssembly>$(TargetPath)</TargetAssembly>
-    <TargetAssemblyPdb>$(TargetDir)$(TargetName).pdb</TargetAssemblyPdb>
-  </PropertyGroup>
-  <Target Name="AfterBuild">
-    <Copy SourceFiles="$(TargetAssembly)" DestinationFolder="$(PythonBuildDir)" />
-    <!--Copy SourceFiles="$(TargetAssemblyPdb)" Condition="Exists('$(TargetAssemblyPdb)')" DestinationFolder="$(PythonBuildDir)" /-->
-  </Target>
-</Project>
+<?xml version="1.0" encoding="utf-8"?>
```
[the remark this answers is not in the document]
BadSingleton: Yes sorry, that should have been a one line change. My diff editor didn't show me the line endings changes

src/runtime/maybeserialize.cs
```diff
@@ -0,0 +1,227 @@
...
+                return m_type;
+            }
+        }
+
+        public string Name {get{return m_name;}}
+        public bool Valid => m_type != null;
+
+        public override string ToString()
+        {
+            return (m_type != null ? m_type.ToString() : $"missing type: {m_name}") + Valid.ToString();
```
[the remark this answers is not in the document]
BadSingleton: Yes, that was for debugging porposes. I'll remove

src/runtime/maybeserialize.cs
```diff
@@ -0,0 +1,227 @@
...
+            info.AddValue("s", m_name);
+            if (Valid)
+            {
+                info.AddValue("f", m_info.Name);
+                info.AddValue("t", m_info.ReflectedType.AssemblyQualifiedName);
+                var p = m_info.GetParameters();
+                string[] types = new string[p.Length];
+                for (int i = 0; i < p.Length; i++)
+                {
+                    types[i] = p[i].ParameterType.AssemblyQualifiedName;
```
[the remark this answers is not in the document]
BadSingleton: I'll have to test

src/runtime/typemanager.cs
BadSingleton: > ... adding a precheck?

Yes

src/runtime/metatype.cs
BadSingleton: Yes, and maybe. `DelegateObject` and `ArrayObject` override the method (always return `false`), we could do a type check here, but I don't think it will improve anything.

src/runtime/StateSerialization/MaybeMemberInfo.cs
```diff
@@ -0,0 +1,113 @@
...
+                deserializationException = e;
+            }
+        }
+
+        // This is complicated because we bind fields 
+        // based on the visibility of the field, properties 
+        // based on it's setter/getter (which is a method 
+        //  info) visibility and events based on their
+        // AddMethod visibility.
+        static bool ShouldBindMember(MemberInfo mi)
```
BadSingleton, line 85: I'm not totally satisfied with this solution, but it's the best I could find. If anyone has a better idea..

src/runtime/maybeserialize.cs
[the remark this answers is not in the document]
BadSingleton: Did a quick check and `out` and `ref` parameters both reflect to `void function (SomeType ByRef)` with the type argument being of type `(SomeType&, ...)`. Haven't been able to test with `in` (couldn't get to set the language version high enough in `TestRunner.cs`), but it looks like it's doing the correct thing. Will add a test case for it.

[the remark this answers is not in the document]
BadSingleton: Ah, found an edge case where changing `ref` to `out` will still "resolve" the function even though the semantics are now different

lostmsu:
Travis build fails: https://travis-ci.org/github/pythonnet/pythonnet/jobs/746288589
You need to explicitly reference https://www.nuget.org/packages/Microsoft.NETFramework.ReferenceAssemblies

src/runtime/classmanager.cs
```diff
@@ -209,11 +267,16 @@ private static void InitClassBase(Type type, ClassBase impl)
             IntPtr dict = Marshal.ReadIntPtr(tp, TypeOffset.tp_dict);
 
 
+            if (impl.dotNetMembers == null)
+            {
+                impl.dotNetMembers = new List<string>();
+            }
```
lostmsu, line 273: I am uncertain about thread safety here.

src/runtime/metatype.cs
lostmsu: Just FYI, the reason I am looking for this stuff, is that having exceptions raised and caught interferes with the debugging scenario, when you are debugging a user-visible exception `X`, and to catch its origin set the "Break on all exceptions of type `X`" flag, you will get this line come up.

This particular code is unlikely affect users in this way though.

lostmsu (changes requested):

pythonnet.sln
```diff
@@ -184,6 +186,30 @@ Global
...
+		{7539EBD5-7A15-4513-BCB0-1D9BEF112BC2}.DebugWin|x86.ActiveCfg = Debug|x86
+		{7539EBD5-7A15-4513-BCB0-1D9BEF112BC2}.DebugWin|x86.Build.0 = Debug|x86
+		{7539EBD5-7A15-4513-BCB0-1D9BEF112BC2}.DebugWinPY3|x64.ActiveCfg = Debug|x64
+		{7539EBD5-7A15-4513-BCB0-1D9BEF112BC2}.DebugWinPY3|x64.Build.0 = Debug|x64
+		{7539EBD5-7A15-4513-BCB0-1D9BEF112BC2}.DebugWinPY3|x86.ActiveCfg = Debug|x86
+		{7539EBD5-7A15-4513-BCB0-1D9BEF112BC2}.DebugWinPY3|x86.Build.0 = Debug|x86
+		{7539EBD5-7A15-4513-BCB0-1D9BEF112BC2}.ReleaseMono|x64.ActiveCfg = Release|x64
+		{7539EBD5-7A15-4513-BCB0-1D9BEF112BC2}.ReleaseMono|x86.ActiveCfg = Release|x86
+		{7539EBD5-7A15-4513-BCB0-1D9BEF112BC2}.ReleaseMonoPY3|x64.ActiveCfg = ReleaseMon|x64
+		{7539EBD5-7A15-4513-BCB0-1D9BEF112BC2}.ReleaseMonoPY3|x86.ActiveCfg = ReleaseMon|x86
```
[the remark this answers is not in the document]
BadSingleton: That's an accident

src/runtime/arrayobject.cs
```diff
@@ -30,15 +30,20 @@ public static IntPtr tp_new(IntPtr tpRaw, IntPtr args, IntPtr kw)
...
             var self = GetManagedObject(tp) as ArrayObject;
+            if (!self.type.Valid)
+            {
+                return Exceptions.RaiseTypeError(self.type.DeletedMessage);
+            }
 
             long[] dimensions = new long[Runtime.PyTuple_Size(args)];
             if (dimensions.Length == 0)
+            if (Runtime.PyTuple_Size(args) != 1)
             {
```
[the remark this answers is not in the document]
BadSingleton: Woops. Totally is.

src/runtime/classmanager.cs
BadSingleton: I think it is safe. If more than one thread execute this method at the same time, the list will be filled anyway. Duplicate members are not an issue as the code that clears the members deals with `KeyError`s.

src/domain_tests/Python.DomainReloadTests.15.csproj
```diff
@@ -0,0 +1,66 @@
...
+    <SolutionDir Condition="$(SolutionDir) == '' Or $(SolutionDir) == '*Undefined*'">..\..\</SolutionDir>
+    <PythonBuildDir Condition="'$(TargetFramework)'=='net40' AND '$(PythonBuildDir)' == ''">$(SolutionDir)\bin\</PythonBuildDir>
+    <PublishDir Condition="'$(TargetFramework)'!='net40'">$(OutputPath)\$(TargetFramework)_publish</PublishDir>
+    <LangVersion>7.3</LangVersion>
+    <ErrorReport>prompt</ErrorReport>
+    <BaseDefineConstants>XPLAT</BaseDefineConstants>
+    <DefineConstants>$(DefineConstants);$(BaseDefineConstants);</DefineConstants>
+    <DefineConstants Condition="'$(TargetFramework)'=='netcoreapp3.1'">$(DefineConstants);NETCOREAPP</DefineConstants>
+    <DefineConstants Condition="'$(TargetFramework)'=='netstandard2.0'">$(DefineConstants);NETSTANDARD</DefineConstants>
+    <DefineConstants Condition="'$(BuildingInsideVisualStudio)' == 'true' AND $(Configuration.Contains('Debug'))">$(DefineConstants);TRACE;DEBUG</DefineConstants>
```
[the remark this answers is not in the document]
BadSingleton: Only NETCOREAPP is.

src/runtime/StateSerialization/MaybeMethodBase.cs
```diff
@@ -0,0 +1,206 @@
...
+        {
+            MethodBase mb = null;
+            try
+            {
+                mb = tp.GetConstructor(ClassManager.BindingFlags, binder:null, types:parameters, modifiers:null);
+            }
+            catch (AmbiguousMatchException)
+            {
+                // The static constructor and non-static constructor may 
+                // have the same signature. Use ToString to disambiguate.
```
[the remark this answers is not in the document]
BadSingleton: I thought it could, but my attempts are proving me wrong.

src/runtime/StateSerialization/MaybeMethodBase.cs
```diff
@@ -0,0 +1,206 @@
...
+                    }
+                }
+            }
+
+            return mb;
+        }
+
+        public void GetObjectData(SerializationInfo serializationInfo, StreamingContext context)
+        {
+            serializationInfo.AddValue("s", name);
```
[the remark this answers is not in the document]
BadSingleton: Constants would be better. Full names would just make the serialization stream bigger.

lostmsu:
@BadSingleton can you please hit "re-request review", when you think it is ready for the next iteration?

BadSingleton:
@lostmsu travis build still failing, even though I referenced the assembly. Can you give me some pointers?

https://travis-ci.org/github/pythonnet/pythonnet/jobs/747280150

lostmsu:
@BadSingleton can you rebase this on top of https://github.com/pythonnet/pythonnet/pull/1209 (unmerged yet). Benedikt is removing the old project files, and is simplifying the new ones. We should not need to repeat that for the domain tests.

src/domain_tests/TestRunner.cs
```diff
@@ -0,0 +1,1103 @@
...
+    /// Each test case:
+    /// * Compiles some code, loads it into a domain, runs python that refers to it.
+    /// * Unload the domain.
+    /// * Compile a new piece of code, load it into a domain, run a new piece of python that accesses the code.
+    /// * Unload the domain. Reload the domain, run the same python again.
+    /// This class gets built into an executable which takes one argument:
+    /// which test case to run. That's because pytest assumes we'll run
+    /// everything in one process, but we really want a clean process on each
+    /// test case to test the init/reload/teardown parts of the domain reload
+    /// code.
```
benoithudson, line 33: The header comment here looks like a merge error?

src/domain_tests/TestRunner.cs
```diff
@@ -0,0 +1,1103 @@
...
+    assert called is True
+    called = False
+    Cls.Call()
+    assert called is False
+    #try:
+    #    assert 2 == Cls.Before
+    #except TypeError:
+    #    print('Caught expected exception')
+    #else:
+    #    raise AssertionError('Failed to throw exception')
```
benoithudson, line 355: Maybe remove this comment?

src/domain_tests/TestRunner.cs
```diff
@@ -0,0 +1,1103 @@
...
+import TestNamespace
+
+def before_reload():
+    sys.my_cls = TestNamespace.Cls
+    sys.my_inst = TestNamespace.Cls(1)
+
+def after_reload():
+     try:
+        TestNamespace.Cls(2)
+        sys.my_cls.Member()
```
benoithudson, line 401: It'd be easier to isolate the regression with two separate try/except/else here.

src/domain_tests/TestRunner.cs
```diff
@@ -0,0 +1,1103 @@
...
+        print('caught expected exception')
+    else:
+        raise AssertionError('failed to raise')
+
+    foo = TestNamespace.Data()
+    bar = TestNamespace.Cls.MyFn(foo)
+    # foo should have changed
+    assert foo.num == 7
+    assert bar.num == 7
+    # Pythonnet also returns a new object with `ref`-quialified parameters
```
benoithudson, line 848: *qualified

src/domain_tests/TestRunner.cs
```diff
@@ -0,0 +1,1103 @@
...
+        static string CreateCaseRunnerAssembly(string verb, string shutdownMode = "ShutdownMode.Reload")
+        {
+            var code = string.Format(CaseRunnerTemplate, shutdownMode, verb);
+            var name = "TestCaseRunner.exe";
+
+            return CreateAssembly(name, code, exe: true);
+        }
+        static string CreateAssembly(string name, string code, bool exe = false)
+        {
+            // Console.WriteLine(code);
```
benoithudson, line 1032: remove this printf

src/runtime/StateSerialization/MaybeMemberInfo.cs
```diff
@@ -0,0 +1,118 @@
...
+        internal MaybeMemberInfo(SerializationInfo serializationInfo, StreamingContext context)
+        {
+            // Assumption: name is always stored in "s"
+            name = serializationInfo.GetString(SerializationName);
+            info = null;
+            deserializationException = null;
+            try
+            {
+                var tp = Type.GetType(serializationInfo.GetString(SerializationType));
+                if (tp != null)
```
benoithudson, line 69: Is it reasonable to expect a null type? Otherwise, maybe just assume it's non-null and let the exception get thrown.

src/runtime/StateSerialization/MaybeMemberInfo.cs
benoithudson: You could also leave this abstract and have `class MaybePropertyInfo : MaybeMemberInfo<PropertyInfo>` and so on, enforcing that if the type of member changed then it doesn't count as a match.

benoithudson: But maybe that's not critical.

src/runtime/StateSerialization/MaybeMemberInfo.cs
```diff
@@ -0,0 +1,118 @@
...
+
+        public override string ToString()
+        {
+            return (info != null ? info.ToString() : $"missing type: {name}");
+        }
+
+        public MaybeMemberInfo(T fi)
+        {
+            info = fi;
+            name = info?.ToString();
```
benoithudson, line 56: Name isn't really required now. You only need it if info is null.

You need to always serialize the name; if info is set, then serialize info.ToString else serialize name.
You only need to deserialize the name if info failed to deserialize for whatever reason.

benoithudson: Maybe it doesn't matter, since the memory savings aren't huge, and debugging is easier if you always have the name.

src/runtime/StateSerialization/MaybeMethodBase.cs
```diff
@@ -0,0 +1,199 @@
...
+        }
+
+        internal MaybeMethodBase(SerializationInfo serializationInfo, StreamingContext context)
+        {
+            name = serializationInfo.GetString(SerializationName);
+            info = null;
+            deserializationException = null;
+            try
+            {
+                // Retrive the reflected type of the method;
```
benoithudson, line 108: *retrieve

src/runtime/StateSerialization/MaybeMethodBase.cs
```diff
@@ -0,0 +1,199 @@
...
+            }
+        }
+
+        MethodBase CheckRefTypes(MethodBase mb, ParameterHelper[] ph)
+        {
+            // One more step: Changing:
+            // void MyFn (ref int a)
+            // to:
+            // void MyFn (out int a)
+            // will still find the fucntion correctly as, `in`, `out` and `ref`
```
benoithudson, line 168: *function

src/runtime/StateSerialization/MaybeMethodBase.cs
```diff
@@ -0,0 +1,199 @@
...
+                    mb = tp.GetConstructor(ClassManager.BindingFlags&(~BindingFlags.Static), binder:null, types:types, modifiers:null);
+                }
+                else
+                {
+                    mb = tp.GetMethod(field_name, ClassManager.BindingFlags, binder:null, types:types, modifiers:null);
+                }
+                
+                if (mb != null && hasRefType)
+                {
+                    CheckRefTypes(mb, param);
```
benoithudson, line 147: is this supposed to be `mb = CheckRefTypes(...)` ?

src/runtime/classbase.cs
```diff
@@ -18,18 +18,21 @@ namespace Python.Runtime
...
         {
+            dotNetMembers = new List<string>();
             indexer = null;
             type = tp;
         }
 
         internal virtual bool CanSubclass()
         {
-            return !type.IsEnum;
+            return !type.Value.IsEnum;
```
benoithudson, line 35: Should this return false or throw on an invalid type?

src/runtime/classmanager.cs
benoithudson: The thread safety issue is no worse than before; the entry is in GetClass(Type). Two threads that hit `cache.TryGetValue(type, out cb)` simultaneously will each then create a `ClassBase` entry and `Add` it to the cache -- Dictionary isn't thread-safe.

src/runtime/constructorbinding.cs
```diff
@@ -122,7 +126,7 @@ public static IntPtr tp_repr(IntPtr ob)
                 return self.repr;
             }
             MethodBase[] methods = self.ctorBinder.GetMethods();
-            string name = self.type.FullName;
+            string name = self.type.Value.FullName;
```
benoithudson, line 129: If it's invalid then probably print something like `(missing {self.type.Name})` -- makes it possible to debug.

src/runtime/converter.cs
```diff
@@ -344,7 +344,7 @@ internal static IntPtr ToPythonImplicit(object value)
                 }
                 if (mt is ClassBase)
                 {
-                    result = ((ClassBase)mt).type;
+                    result = ((ClassBase)mt).type.Value;
```
benoithudson, line 347: probably should RaiseTypeError if invalid here

src/runtime/metatype.cs
benoithudson: to avoid the exception you could add ClassBase.IsValid and test it explicitly

benoithudson: or check cb.type.Valid like you do below

src/runtime/methodobject.cs
```diff
@@ -13,40 +17,46 @@ namespace Python.Runtime
...
                 binder.AddMethod(item);
                 if (item.IsStatic)
                 {
                     this.is_static = true;
                 }
             }
+            binder.allow_threads = allow_threads;
+        }
+
+        internal MethodInfo[] info
```
benoithudson, line 50: Does this need to be an array or can you switch to using a list?

benoithudson: Oh, I see. This is basically an attempt not to implement custom serialize/deserialize.

src/runtime/moduleobject.cs
```diff
@@ -341,6 +341,20 @@ protected override void OnSave(InterDomainContext context)
             // Decref twice in tp_clear, equilibrate them.
             Runtime.XIncref(dict);
             Runtime.XIncref(dict);
+            // destroy the cache(s)
+            foreach (var pair in cache)
+            {
+                Runtime.PyDict_DelItemString(dict, pair.Key);
+                pair.Value.DecrRefCount();
+                 if (Exceptions.ExceptionMatches(Exceptions.KeyError))
```
benoithudson, line 349: weird indent

benoithudson: also do you need to PyErr_Clear() before starting the loop? Could someone have set the error for unrelated reasons before entering here?

src/runtime/propertyobject.cs
```diff
@@ -51,8 +56,8 @@ public static IntPtr tp_descr_get(IntPtr ds, IntPtr ob, IntPtr tp)
 
                 try
                 {
-                    result = self.info.GetValue(null, null);
-                    return Converter.ToPython(result, self.info.PropertyType);
+                    result = self.info.Value.GetValue(null, null);
+                    return Converter.ToPython(result, self.info.Value.PropertyType);
```
benoithudson, line 60: I'm seeing this idiom a lot: at the top of the function, check if valid; at the end, get the Value without checking for exceptions.

I think that we should get the value immediately after checking if it's valid, so we don't somehow refactor and end up getting Value without checking for validity at all.

src/runtime/propertyobject.cs
```diff
@@ -31,7 +32,11 @@ public PropertyObject(PropertyInfo md)
         public static IntPtr tp_descr_get(IntPtr ds, IntPtr ob, IntPtr tp)
         {
             var self = (PropertyObject)GetManagedObject(ds);
-            MethodInfo getter = self.getter;
+            if (!self.info.Valid)
+            {
+                return Exceptions.RaiseTypeError(self.info.DeletedMessage);
+            }
+            MethodInfo getter = self.getter.Value;
```
benoithudson, line 39: Should we check self.getter.Valid?

src/runtime/propertyobject.cs
```diff
@@ -91,7 +96,13 @@ public static IntPtr tp_descr_get(IntPtr ds, IntPtr ob, IntPtr tp)
...
         {
             var self = (PropertyObject)GetManagedObject(ds);
-            MethodInfo setter = self.setter;
+            if (!self.info.Valid)
+            {
+                Exceptions.RaiseTypeError(self.info.DeletedMessage);
+                return -1;
+            }
+
+            MethodInfo setter = self.setter.Value;
```
benoithudson, line 105: should we check self.setter.Valid?

src/runtime/runtime_data.cs
```diff
@@ -120,8 +120,8 @@ private static void RestoreRuntimeDataImpl()
 
             var objs = RestoreRuntimeDataObjects(storage.GetStorage("objs"));
             RestoreRuntimeDataModules(storage.GetStorage("modules"));
-            var clsObjs = ClassManager.RestoreRuntimeData(storage.GetStorage("classes"));
             TypeManager.RestoreRuntimeData(storage.GetStorage("types"));
+            var clsObjs = ClassManager.RestoreRuntimeData(storage.GetStorage("classes"));
```
benoithudson, line 124: Does the order matter or can this change be reverted?

benoithudson:
FYI @BadSingleton confirmed that there's no performance implications based on the performance test suite. (We didn't really expect any impact)

Also that the performance test suite is pretty simplistic. Is there something better to use?

src/runtime/StateSerialization/MaybeMethodBase.cs
BadSingleton: Yes

src/runtime/runtime_data.cs
BadSingleton: The order matters. Because we're re-initializing the `ClassObject`s we need the types to be present or they'll be re-created then clobbered

src/domain_tests/TestRunner.cs
BadSingleton: Git history says it isn't. It does need some editing though

src/runtime/StateSerialization/MaybeMemberInfo.cs
BadSingleton: It is. If the type is not found, the method returns null. I added it because it doesn't really makes sense to have a NullReferenceException as the exception for this.

src/runtime/StateSerialization/MaybeMemberInfo.cs
BadSingleton: > debugging is easier if you always have the name.

That's why I always have it

src/runtime/classbase.cs
BadSingleton: I think it should throw. It would be false to say that something that doesn't exist can't subclass; we can't know since it doesn't exist.
Also, the code that calls `CanSubclass` throws another kind of exception when it's false `Exceptions.RaiseTypeError("delegates, enums and array types cannot be subclassed");`

src/runtime/constructorbinding.cs
BadSingleton: This one was missing it's check.

src/runtime/converter.cs
BadSingleton: I think it's not possible to have an invalid value here as the above call to GetManagedObject would return null. Still added it as a paranoia check

src/runtime/moduleobject.cs
BadSingleton: I think it's a non-problem. Either:
* Someone an exception other than KeyError and 
** we never set one and we don't clear or clobber it 
** we clobber it by setting a KeyError
* Someone set a KeyError before us and we clear in the loop (instead of clearing before the loop).

src/runtime/propertyobject.cs
BadSingleton: No, but we should use UnsafeValue (doesn't throw), as null is a valid value in this context.

src/runtime/propertyobject.cs
BadSingleton: No, but we should use UnsafeValue (doesn't throw), as null is a valid value in this context.

src/runtime/propertyobject.cs
BadSingleton: No, but we should use UnsafeValue (doesn't throw), as null is a valid value in this context.

src/runtime/propertyobject.cs
BadSingleton: No, but we should use UnsafeValue (doesn't throw), as null is a valid value in this context.

BadSingleton:
@lostmsu I'd rather wait for #1209 to be merged than to rebase.

lostmsu:
@BadSingleton it is in (though some of the old tests are randomly crashing).

src/domain_tests/TestRunner.cs
benoithudson: Probably a merge error between two parts of my brain then.

src/domain_tests/Python.DomainReloadTests.15.csproj
```diff
@@ -0,0 +1,54 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="15.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
```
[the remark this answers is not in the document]
filmor: Assemblies built with the dotnet cli build just fine with the `ReferenceAssemblies`, just look at the embedding tests for an example.

src/runtime/moduleobject.cs
lostmsu: You should only make the check when `PyDict_DelItemString` actually returns `-1`

src/runtime/StateSerialization/MaybeMethodBase.cs
```diff
@@ -0,0 +1,199 @@
...
+                if (info == null)
+                {
+                    throw new SerializationException(DeletedMessage, innerException: deserializationException);
+                }
+                return (T)info;
+            }
+        }
+
+        public T UnsafeValue { get { return (T)info; } }
+        public string Name {get{return name;}}
```
lostmsu, line 86: NIT: `Name => name;`

src/runtime/StateSerialization/MaybeMethodBase.cs
```diff
@@ -0,0 +1,199 @@
...
+        {
+            get
+            {
+                if (info == null)
+                {
+                    throw new SerializationException(DeletedMessage, innerException: deserializationException);
+                }
+                return (T)info;
+            }
+        }
```
lostmsu, line 83: Couldn't you extract this as a base class? I thought you had one in previous iterations...

src/runtime/StateSerialization/MaybeMethodBase.cs
```diff
@@ -0,0 +1,199 @@
...
+                MethodBase mb = null;
+                if (serializationInfo.GetBoolean(SerializationIsCtor))
+                {
+                    // We never want the static constructor.
+                    mb = tp.GetConstructor(ClassManager.BindingFlags&(~BindingFlags.Static), binder:null, types:types, modifiers:null);
+                }
+                else
+                {
+                    mb = tp.GetMethod(field_name, ClassManager.BindingFlags, binder:null, types:types, modifiers:null);
+                }
```
lostmsu, line 143: I think you might need to pass `modifiers` to `GetMethod` and `GetConstructor`.

Might fail a test case from `out int arg` to `out int arg`

src/runtime/StateSerialization/MaybeMethodBase.cs
```diff
@@ -0,0 +1,199 @@
...
+
+                MethodBase mb = null;
+                if (serializationInfo.GetBoolean(SerializationIsCtor))
+                {
+                    // We never want the static constructor.
+                    mb = tp.GetConstructor(ClassManager.BindingFlags&(~BindingFlags.Static), binder:null, types:types, modifiers:null);
+                }
+                else
+                {
+                    mb = tp.GetMethod(field_name, ClassManager.BindingFlags, binder:null, types:types, modifiers:null);
```
lostmsu, line 142: Considering you only get one method, how would it distinguish between `M(out int x)` and `M(ref int x)` or `M(in int x)`?

src/runtime/constructorbinder.cs
```diff
@@ -51,10 +51,15 @@ internal object InvokeRaw(IntPtr inst, IntPtr args, IntPtr kw)
...
+            {
+                return Exceptions.RaiseTypeError(_containingType.DeletedMessage);
+            }
             object result;
+            Type tp = _containingType.Value;
 
-            if (_containingType.IsValueType && !_containingType.IsPrimitive &&
-                !_containingType.IsEnum && _containingType != typeof(decimal) &&
+            if (tp.IsValueType && !tp.IsPrimitive &&
+                !tp.IsEnum && tp != typeof(decimal) &&
```
lostmsu, line 62: NIT: I wonder if we could reduce diff significantly by using the following pattern:

1. Rename fields, that were changed to `Maybe*`
2. Start methods, that use that member with `if (!field.TryGetValue(out var originalFieldName)) return Exceptions.Raise...;`

E.g.

```csharp
private MaybeType parentType; // <- creative renaming here

object InvokeRaw(...)
{
  if (!parentType.TryGet(out var _containingType)) return Exceptions.Raise(...);
  if (_containingType.IsValueType && !_containingType.IsPrimitive && // <- this and below are now untouched
```

src/runtime/metatype.cs
lostmsu: Good to have it rewritten with `.Valid`, but not required.

src/runtime/methodbinder.cs
```diff
@@ -180,6 +185,11 @@ internal MethodBase[] GetMethods()
         /// </remarks>
         internal static int GetPrecedence(MethodBase mi)
         {
+            if (mi == null)
+            {
+                return -1;
+            }
+
```
lostmsu, line 192: I think `int.Max` would be more appropriate. Higher the number - lower the priority.

UPD. can this even be null?

src/runtime/methodobject.cs
```diff
@@ -13,40 +17,46 @@ namespace Python.Runtime
...
-        }
-
-        public MethodObject(Type type, string name, MethodInfo[] info, bool allow_threads)
-        {
-            _MethodObject(type, name, info);
-            binder.allow_threads = allow_threads;
-        }
-
-        private void _MethodObject(Type type, string name, MethodInfo[] info)
+        //  `allow_threads = true`: True being the default value of MethodBinder.allow_threads
```
lostmsu, line 31: Better use a constant.

src/runtime/moduleobject.cs
```diff
@@ -341,6 +341,20 @@ protected override void OnSave(InterDomainContext context)
...
             Runtime.XIncref(dict);
             Runtime.XIncref(dict);
+            // destroy the cache(s)
+            foreach (var pair in cache)
+            {
+                if ((Runtime.PyDict_DelItemString(dict, pair.Key) == -1) &&
+                    (Exceptions.ExceptionMatches(Exceptions.KeyError)))
+                {
+                    // Trying to remove a key that's not in the dictionary 
+                    // raises an error. We don't care about it.
```
lostmsu, line 351: All other errors should actually be raised as C# exceptions.

src/runtime/runtime.cs
```diff
@@ -664,7 +664,8 @@ internal static Type[] PythonArgsToTypeArray(IntPtr arg, bool mangleObjects)
 
                 if (mt is ClassBase)
                 {
-                    t = ((ClassBase)mt).type;
+                    MaybeType _type = ((ClassBase)mt).type;
+                    t = _type.Valid ?  _type.Value : null;
```
lostmsu, line 668: NIT: couldn't this be `.type.UnsafeValue`?

lostmsu (changes requested):

BadSingleton:
I'll be on holidays for two weeks, I'll fix that when back in office

src/domain_tests/TestRunner.cs
```diff
@@ -0,0 +1,1098 @@
...
+            }
+            Console.WriteLine($"-- Running domain reload test case: {testCase.Name}");
+
+            var tempFolderPython = Path.Combine(Path.GetTempPath(), "Python.Runtime.dll");
+            if (File.Exists(tempFolderPython))
+            {
+                File.Delete(tempFolderPython);
+            }
+
+            File.Copy(PythonDllLocation, tempFolderPython);
```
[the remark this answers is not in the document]
BadSingleton: I totally agree with this

src/runtime/StateSerialization/MaybeMethodBase.cs
BadSingleton: They are structs, to minimize memory allocations. I could extract some of it to another struct

src/runtime/StateSerialization/MaybeMethodBase.cs
BadSingleton: `GetMethod` and `GetConstructor` can't. That's why there's a `CheckRefTypes` method in this struct. The `out_to_ref_param` tests this kind of scenario.

src/runtime/StateSerialization/MaybeMethodBase.cs
BadSingleton: We'd need to create a binder for that(the `modifiers` parameter is ignored when using the default binder). That scenario is handled though; see the other comment.

src/runtime/StateSerialization/MaybeMethodBase.cs
```diff
@@ -0,0 +1,199 @@
...
+                Modifier = TypeModifier.None;
+
+                if (tp.IsIn)
+                {
+                    Modifier = TypeModifier.In;
+                }
+                else if (tp.IsOut)
+                {
+                    Modifier = TypeModifier.Out;
+                }
```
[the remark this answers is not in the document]
BadSingleton: I wasn't aware of those Attributes. And yes, we'd need to check for ByRef as a parameter with the attribute would be `int`, not `int&`. I'll do that in that another PR (along with more tests).

src/runtime/runtime.cs
BadSingleton: `MaybeType` doesn't have an UnsafeValue, and it's the only place where this pattern is used for `MaybeType`.

src/runtime/constructorbinder.cs
BadSingleton: I don't think we can. Most are static methods using a class instance as a parameter to operate on like `self.type.<member or method>`. 

We could implement those members or methods on the `Maybe*`, but every time a method/member that wasn't "exposed" is used, we'd have to add it. And it might make people to forget to check validity before use.

src/runtime/StateSerialization/MaybeMethodBase.cs
lostmsu: But in this code you either receive 1 constructor or method, or none at all (an exception?). So if both `ref` and `out` are present, which one will it be?

src/domain_tests/TestRunner.cs
```diff
@@ -0,0 +1,1115 @@
...
+            // Don't delete unconditionally. It's sometimes useful to leave the
+            // folder behind to debug failing tests.
+            TeardownTestFolder();
+
+            return 0;
+        }
+
+        static void SetupTestFolder(string testCaseName)
+        {
+            TestPath = Path.Combine(Path.GetTempPath(), $"Python.TestRunner.{testCaseName}");
```
lostmsu, line 998: I think some random number should also be added to indicate the overall run, so that different runs would not clash.

src/runtime/StateSerialization/MaybeMethodBase.cs
BadSingleton: It throws an `AmbigousMatchException` if it resolves to more than one method/constructor. It's not possible to overload only with `in`/`ref`/`out` parameters. However, if in domain `1` a method has a `out` parameter and in domain `2`, same method, same parameter is now a `ref` parameter, it'll still bind.

src/domain_tests/TestRunner.cs
BadSingleton: They wouldn't, if the folder exists when setting up the test, it gets deleted first. Also, the executable can only run one test per invocation.

lostmsu: I still see a problem with running tests for multiple Python versions on the same machine in parallel.

BadSingleton: Ah, I hadn't thought about that one. I'll add the python version and architecture to the folder name.

BadSingleton: I'll add the PID instead

lostmsu (approved):

PR #1292, opened 2020-11-24
TypeOffset class no longer depends on target Python version

lostmsu:
This is a stage of removing the need to prepare a separate build of `Python.Runtime` for each Python version.

This removes the need to pick `TypeOffset` at compile time. Instead, for each supported Python version a separate class is generated (e.g. `TypeOffset36`). Then the `Runtime` (via new class `ABI`) picks the correct class based on actual Python version using reflection, and copies only the necessary `TypeOffset` members over from it.

`ManagedDataOffsets.Magic` is now also read at runtime from `tp_basicsize` of the `builtins.type` class.

### Any other comments?

N/A

### Checklist

Check all those that are applicable and complete. **N/A**


tools/geninterop/geninterop.py
```diff
@@ -244,42 +244,39 @@ def preprocess_python_headers():
...
-        "PYTHON{0}{1}".format(PY_MAJOR, PY_MINOR)
-    ]
-
     if hasattr(sys, "abiflags"):
         if "d" in sys.abiflags:
             defines.append("PYTHON_WITH_PYDEBUG")
+            raise NotImplementedError
         if "m" in sys.abiflags:
             defines.append("PYTHON_WITH_PYMALLOC")
+            raise NotImplementedError
```
filmor, line 253: This one has the exact same slots and is pretty widely used.

tools/geninterop/geninterop.py
```diff
@@ -351,6 +350,10 @@ def main():
     gen_interop_head(writer)
 
     gen_heap_type_members(ast_parser, writer)
+
+    ver_define = "PYTHON{0}{1}".format(PY_MAJOR, PY_MINOR)
+    writer.append(code="#if " + ver_define)
```
filmor, line 355: Why are these still behind a compile-time flag? Do you plan to change that?

tools/geninterop/geninterop.py
lostmsu: The code in this method actually never passes any parameters to clang. It only used to affect the generated file by adding an extra compile-time condition.

My understanding is our current tools never actually run geninterop, and the generated files we have are all generated with none of the extra flags set.

Do we actually plan to support users regenerating interop files on their own? If yes, where is the manual (I could update it for all flags then along with this code)? If we do not, I can add a string to `NotImplementedError` with minimal details.

filmor: `geninterop` is always run when in the `setup.py` we see that the respective file doesn't exist, yet. The `PYTHON_WITH_PYMALLOC` flag is quite common, in particular Anaconda uses it.

tools/geninterop/geninterop.py
lostmsu: Yeah, I am approaching this gradually.

tools/geninterop/geninterop.py
lostmsu: Not sure I understand the doc correctly, but it looks like the set of supported flags went to 0 with 3.8

lostmsu:
Hm, Travis build on Linux with Python 3.9 regularly fails with this change. Could I have screwed up something here? @filmor, you've recently added support for 3.9. Was there anything specific to Linux?

filmor:
Nope, the only reason why support for 3.9 took so long was that I had to adjust the interop generation as it wasn't working with the new header-files anymore.

lostmsu:
@filmor , oh, nvm. Looks like Travis just used older commit for 3.9 o-O

tools/geninterop/geninterop.py
lostmsu: @filmor this and the other comment should be addressed now.

.gitignore
```diff
@@ -1,3 +1,8 @@
+/src/runtime/interopNative.cs
+
+# VS Code
+.vscode/
```
filmor, line 4: Just do this in your global gitignore.

src/runtime/native/ABI.cs
```diff
@@ -0,0 +1,32 @@
...
+    static class ABI
+    {
+        internal static void Initialize(Version version, BorrowedReference pyType)
+        {
+            string offsetsClassSuffix = string.Format(CultureInfo.InvariantCulture,
+                                                      "{0}{1}", version.Major, version.Minor);
+
+            var thisAssembly = Assembly.GetExecutingAssembly();
+
+            const string nativeTypeOffsetClassName = "Python.Runtime.NativeTypeOffset";
```
filmor, line 18: Use `nameof` here.

src/runtime/runtime.cs
```diff
@@ -113,11 +114,13 @@ internal static Version PyVersion
         {
             get
             {
-                var versionTuple = new PyTuple(PySys_GetObject("version_info"));
-                var major = versionTuple[0].As<int>();
-                var minor = versionTuple[1].As<int>();
-                var micro = versionTuple[2].As<int>();
-                return new Version(major, minor, micro);
+                using (var versionTuple = new PyTuple(PySys_GetObject("version_info")))
```
filmor, line 117: :+1:

tools/geninterop/geninterop.py
```diff
@@ -244,66 +244,59 @@ def preprocess_python_headers():
...
-        if "d" in sys.abiflags:
-            defines.append("PYTHON_WITH_PYDEBUG")
-        if "m" in sys.abiflags:
-            defines.append("PYTHON_WITH_PYMALLOC")
-        if "u" in sys.abiflags:
-            defines.append("PYTHON_WITH_WIDE_UNICODE")
-
     filename = os.path.basename(__file__)
-    defines_str = " && ".join(defines)
+    abi_flags = sys.abiflags if hasattr(sys, "abiflags") else ''
```
filmor, line 248: `getattr(sys, "abiflags", "")`

filmor (commented):
I think this is fine for now, just minor things that I think could be fixed.

src/runtime/native/ABI.cs
lostmsu: This class does not actually exist, if built without setup.py (which should be supported, once Modernization project is completed).

filmor: Good point.

tools/geninterop/geninterop.py
lostmsu: I checked this out, and the rest of the conditional stuff came from 1b466df0. Since it was only used to determine if methods in Python.Runtime itself implement slots (which we can do on our own), I took a liberty to mostly revert it in this PR. Please, review the last commit "revert slot-related parts of 1b466df to avoid conditional compilation"

PR #1298, opened 2020-11-30
Detect the size of wchar_t (aka Runtime.UCS) at runtime using PyUnicode_GetMax

lostmsu:

This determines the value for `Runtime.UCS` at runtime instead of using compile-time defines, removing the need for separate builds of `Python.Runtime`.

N/A

### Any other comments?

`PyUnicode_GetMax` is deprecated, but preserved for compatibility.

### Checklist

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -34,6 +34,10 @@ csharp_new_line_before_finally = true
 [*.sln]
 indent_style = tab
 
+[*.csproj]
+charset = utf-8
+insert_final_newline = true
+
 # bumpversion reformats itself after every bump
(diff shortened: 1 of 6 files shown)


markuspalme:
This could possibly solve #114.

filmor (approved):

filmor:
Great, I wasn't aware of `PyUnicode_GetMax`. We should probably find a more sustainable solution at some point (as the function is explicitly deprecated), but this will do nicely for now :)

sekkit:
ctypes.sizeof(ctypes.c_wchar) would do the trick

PR #1304, opened 2020-12-03
Ensure that param-array matching works correctly

filmor:

...

...

### Any other comments?

...

### Checklist


lostmsu:
Can you add a test with a mix of keyword and params array arguments?

filmor:
I added some more tests, but three of them are failing. I commented them out for now and will try to fix them today.

src/tests/test_method.py
```diff
@@ -1188,3 +1188,37 @@ def test_keyword_arg_method_resolution():
...
+    assert res == "with params-array"
+
+    res = MethodTest.ParamsArrayOverloaded(1, i=1)
+    assert res == "with params-array"
+
+    # res = MethodTest.ParamsArrayOverloaded(1, 2, i=1)
+    # assert res == "with params-array"
+
+    # res = MethodTest.ParamsArrayOverloaded(1, 2, 3, i=1)
+    # assert res == "with params-array"
```
lostmsu, line 1223: These appear to duplicate ones above.

lostmsu (approved):
Please, remove the commented code first.

src/tests/test_method.py
filmor: Only one of them is a duplicate, I'll drop it. The other ones are currently failing tests.

src/runtime/methodbinder.cs
```diff
@@ -652,7 +652,7 @@ static Type TryComputeClrArgumentType(Type parameterType, IntPtr argument, bool
             {
                 match = true;
             }
-            else if (positionalArgumentCount < parameters.Length)
+            else if (positionalArgumentCount < parameters.Length && (!paramsArray || positionalArgumentCount == parameters.Length - 1))
```
christabella, line 655: Maybe it would be useful to comment explicitly on what kind of cases would fall into this condition, or simplify the branching statements somehow (refactor the default arg logic in a separate method)? Just by reading this I'm guessing that this should take care of only the following 2 cases?
1. when it's not a `params` array but there are some default values we need to fill in
2. when it's a `params` array and exactly one parameter is missing 

Maybe there are more?

christabella: Nevermind, actually I get it now! I also added an extra comment and renamed the variables for readability and consistency with the others parts of methodbinder in my PR: https://github.com/pythonnet/pythonnet/pull/1324/files#diff-237dc718ad5424f5476d2ebbbc4e5164c95901bd2a4c2daaf8ca6fc9f9d3a5fbR703-R704

PR #1307, opened 2020-12-08
Added github actions workflow to replace travis

lostmsu:
Subj

This drops coverage for `setup.py`


lostmsu:
This is not exactly ready yet, as it uses the latest version of Mono instead of 5.20, that we had before.

lostmsu:
Either we:
1. make our own docker images to test on (not sure how this works)
2. have the build take 6+ minutes instead of 1 (remove Mono 6.12 and reinstall 5.20 instead)
3. or switch Mono to 6.12 for testing

Considering we are doing a major version bump, we can pick option 3 for simplicity. @filmor thoughts?

filmor:
I'm in favour of just using the latest version.

PR #1308, opened 2020-12-08
[Out] parameters no longer added to return tuple

lostmsu:
Parameters marked with `ParameterAttributes.Out` (aka `[Out]`) are no longer returned in addition to the regular method return value (unless they are passed with `ref` or `out` keyword).

As it is now possible to pass raw .NET objects to .NET methods, there's no need to marshal `[Out]` parameters back to Python manually. Python users can simply access modified object's data directly.

### Any other comments?

It was confusing to see `read, _ = stream.Read(buff, 0, buff.Length)` in the test code, as `Read` method does not really have any `out` or `ref` parameters. It also failed on some .NET implementations, as the corresponding parameter does not always have an `[Out]` attribute.

### Related issues

This should unblock https://github.com/pythonnet/pythonnet/pull/1307 (tests there are failing because of this difference).

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,11 +18,13 @@ This document follows the conventions laid out in [Keep a CHANGELOG][].
 details about the cause of the failure
 -   `clr.AddReference` no longer adds ".dll" implicitly
 -   `PyIter(PyObject)` constructor replaced with static `PyIter.GetIter(PyObject)` method
--    Return values from .NET methods that return an interface are now automatically
+-   BREAKING: Return values from .NET methods that return an interface are now automatically
      wrapped in that interface. This is a breaking change for users that rely on being
      able to access members that are part of the implementation class, but not the
      interface.  Use the new __implementation__ or __raw_implementation__ properties to
      if you need to "downcast" to the implementation class.
+-   BREAKING: Parameters marked with `ParameterAttributes.Out` are no longer returned in addition
+     to the regular method return value (unless they are passed with `ref` or `out` keyword).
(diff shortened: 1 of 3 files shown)


filmor:
I think this was included for compatibility with IronPython. How would you handle out parameters that still have automatic conversions like `int` and `str`?

lostmsu:
@filmor this is the confusing part. This change does not affect `GetInt(out int val)` functions. It affects `GetInt([Out] int val)` functions, where `[Out]` parameter does nothing. The only real use case I see is that in `Read([Out] byte[] buf)` `[Out]` indicates, that the data will be put into the `buf`. The data can not be put into `string` or `int` without `ref int`/`out int`, and these are handled separately.

lostmsu:
For an instance of API difference between implementations in regards to `[Out]` attribute:

In .NET Core `MemoryStream.Read` has none: https://github.com/dotnet/runtime/blob/5432a66018c2499b88a54276f468ea903ed2b8de/src/libraries/System.Private.CoreLib/src/System/IO/MemoryStream.cs#L333

In Mono it is present: https://github.com/mono/mono/blob/c5b88ec4f323f2bdb7c7d0a595ece28dae66579c/mcs/class/referencesource/mscorlib/system/io/memorystream.cs#L339

Without this change the same Python code would not work with both Mono and .NET Core.

filmor:
Thanks for the explanation. I ran into this problem while testing .NET Core and couldn't really make sense of it, wasn't aware of `[Out]`.

PR #1313, opened 2020-12-10
Fixed CollectBasicObject test

lostmsu:
This contains two fixes, that must go along for the tests to pass:

1. Fixed the `CollectBasicObject` and `CollectOnShutdown` tests for Mono by fixing `MakeAGarbage` function to not keep the garbage on the stack during a conservative roots scan (looks like Mono scans some variables, that are already gone).
2. Fixed the crash during runtime shutdown by further simplifying `Finalizer` to only store raw pointers to python objects instead of instances of `IPyDisposable` and thus replacing Dispose with reliable and simple `DecRef`.

This removes some functions from `Finalizer`, which previously were public, but should not have been.

### DIFF
--- a/src/embed_tests/TestFinalizer.cs
+++ b/src/embed_tests/TestFinalizer.cs
@@ -52,11 +52,7 @@ namespace Python.EmbeddingTest
             Assert.IsFalse(called, "The event handler was called before it was installed");
             Finalizer.Instance.CollectOnce += handler;
 
-            WeakReference shortWeak;
-            WeakReference longWeak;
-            {
-                MakeAGarbage(out shortWeak, out longWeak);
-            }
+            IntPtr pyObj = MakeAGarbage(out var shortWeak, out var longWeak);
             FullGCCollect();
             // The object has been resurrected
             Warn.If(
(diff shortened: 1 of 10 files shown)


src/embed_tests/TestFinalizer.cs
```diff
@@ -104,33 +88,45 @@ public void CollectBasicObject()
         }
 
         [Test]
-        [Ignore("Ignore temporarily")]
+        [Obsolete("GC tests are not guaranteed")]
```
filmor, line 91: Is NUnit aware of this attribute or does this just generate yet another compile-time warning? Does it help to retry the test using https://docs.nunit.org/articles/nunit/writing-tests/attributes/retry.html?

filmor (commented):
Can't you use one of the `*Reference` types instead of `IntPtr`? It would be good to see at least in the code that we are talking about Python objects here.

src/embed_tests/TestFinalizer.cs
lostmsu: NUnit is not aware. This actually removes warning from the usage of other `[Obsolete]` code.

lostmsu:
@filmor not for the finalizer. `*Reference` types can not be stored.

PR #1318, opened 2020-12-12
Fix or disable warnings in Python.Runtime

filmor:

...

...

### Any other comments?

...

### Checklist

### DIFF
--- a/src/runtime/classobject.cs
+++ b/src/runtime/classobject.cs
@@ -1,3 +1,4 @@
+using System.Linq;
 using System;
 using System.Reflection;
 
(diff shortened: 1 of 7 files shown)


src/runtime/classobject.cs
```diff
@@ -143,7 +144,7 @@ public override IntPtr type_subscript(IntPtr idx)
                 return Exceptions.RaiseTypeError("type(s) expected");
             }
 
-            Type gtype = AssemblyManager.LookupType($"{type.FullName}`{types.Length}");
+            Type gtype = AssemblyManager.LookupTypes($"{type.FullName}`{types.Length}").FirstOrDefault();
```
lostmsu, line 147: Picking the first type of many randomly is not a good solution.

lostmsu (changes requested):

src/runtime/classobject.cs
filmor: That's true, but this is what we are doing right now as well. I can add a follow-up ticket, but I also want to get to a warning-free build at some point s.t. we can make them fatal.

lostmsu: That would do.

filmor: #1320.

Issue #984, opened 2019-11-08
.NET Core Support


filmor:
There are two major modes of using Python.NET:

1. Embedding Python into .NET
2. Embedding .NET into Python

Each of them has a distinct "support" status for .NET Core.

The first case should already be possible, as Python.NET can be compiled as a .NET Standard DLL that can be referenced within .NET Core.

The second case is actively being worked at in #857, I'll open a corresponding PR this weekend. There is a previous attempt in #612 that uses the older (in fact second) `coreclr` hosting API and significantly complicates the build-process, which is already annoyingly complicated.

Furthermore, we have to simplify the way how `libpython` is loaded in the embedding case, which @lostmsu has already tried out and we'll look into further once the new loading mechanism has landed.

heartacker:
Come here and see every few days, hoping to use it as soon

7934345:
Any further news?
.net core 2.2
.net core 3.0
.net core 3.1

ericjohannsen:
When I create a .NET Standard 2.0 library project in VS2019 and add a NuGet reference to pythonnet_35_dotnet, I get this error:

Error NU1202
Package pythonnet_py35_dotnet 2.3.0 is not compatible with netstandard2.0 (.NETStandard,Version=v2.0). Package pythonnet_py35_dotnet 2.3.0 does not support any target frameworks.

I tried changing the project's .NET Standard version to both 1.6 and 1.0 and received a similar error message.

What am I missing?

lostmsu:
@ericjohannsen .NET Standard is not currently supported. If you are fine with a temporary unsupported solution, try https://www.nuget.org/packages/Python.Runtime.UnofficialNetStandard/

ericjohannsen:
Thank you for that pointer. Why, though, does the original comment state, "Python.NET can be compiled as a .NET Standard DLL"?

lostmsu:
@ericjohannsen you can do that right now if you build from source. But the NuGet package is very outdated. Current release is 2.4.1, and the nuget is for 2.3.0. We are working on releasing a new NuGet in a month or two.

mikofski:
Sorry for asking, but can I import a dotnet core library using pythonnet `clr` in python?
```
import clr
assy = clr.AddReference('C:/path/to/dotnet-core-lib.dll')

from dotnet-core-lib import MyClass
```
I have been trying this all day and it doesn't seem to work, but I have done it with other .NET libraries, what's different with dotnet core?

[...]

Issue #857, opened 2019-04-29
Implement CLR loading in pure Python


filmor:
To simplify the build process (immensely), I'd like to move the CLR loading for Mono and CoreCLR to Python. I have a working prototype at https://gist.github.com/filmor/cf8a02aefff58be09a57ad2eca3cc246 for CoreCLR using CFFI and will work on an equivalent Mono implementation for this as well. For .NET Framework, I don't know of any public API to embed it into the currently running process, so we'll keep the existing loader for this one.

This implements the first section of the Plan(tm) (https://github.com/pythonnet/pythonnet/wiki/Plan#clr-loading).

spinwards:
I tried the loader from your gist in #857 and it caused a hard crash on windows nano server w/ .net core 2.2 and miniconda.

_Originally posted by @spinwards in https://github.com/pythonnet/pythonnet/pull/612#issuecomment-490568293_

spinwards:
I've tested it using docker with the mcr.microsoft.com/dotnet/core/sdk:2.2 image with both miniconda and a vanilla python install and in both cases when I call func(), python immediately crashes with a generic "Windows fatal exception: access violation".

filmor:
I don't get this error, it works for me on both Linux and Windows up to the point that it complains about `System.Reflection.Emit` from Python.NET's `CodeGenerator` class. I'll update the gist to the latest state.

spinwards:
Sure, with the net40 target platform, but I am running this on nanoserver with .net core2. 

I need the ReflectionPolyfill class to adapt the DefineDynamicAssembly method for .net standard. If I remove the net40 target framework and build the resulting assembly gives me an access violation.

filmor:
Work is ongoing in https://github.com/filmor/clr-loader.

I'm able to load up .NET Core (>=3 due to hostfxr changes) and get a callable object in Python for something that has the signature `int Func(IntPtr data, int size)`. I'll implement the same interface for Mono now and incorporate the `clrmodule` hack from Python.NET.

filmor:
We are getting there, Mono is implemented as well now, I'll try to add .NET framework support tonight.

filmor:
.NET framework support as well as a simple interface are also implemented now. I will create a PR that drops `clrmodule` in favour of a pure-python `clr.py` that uses `clr_loader` and calls a function with the correct signature (`int32 (void*, int32)`) in Python.NET to run the initialisation.

kpoman:
[...]

Issue #987, opened 2019-11-13
RFC: Use AdvancedDLSupport instead of P/Invoke or GetFunctionPointerForDelegate


filmor:
While investigating how to proceed on binding `libpython` I stumbled over this post: https://medium.com/@jarl.gullberg/an-introduction-to-adl-or-how-to-double-your-native-net-interop-performance-c008e4da54db

It showcases https://github.com/Firwood-Software/AdvancedDLSupport and also benchmarks the individual options, which look quite devastating for the combination `GetFunctionPointerForDelegate`+Mono and at least "concerning" for the other frameworks.

The library looks really slick and solves exactly our problem while (seemingly) delivering slightly better performance. It has however the catch of being LGPL. While this is perfectly fine for me, this means there might be a problem for applications trying to embed `Python.Runtime` ahead-of-time compiled or publishing to a closed app-store that prevents you from exchanging the LGPLed library locally.

Comments?

lostmsu:
I am a bit concerned about 

> Additionally, .NET Core lacks a way to set the unmanaged calling convention at runtime, resulting in unreliable results when running as a 32-bit process on Windows. This is, fortunately, a very uncommon configuration.

I don't know how much is this applicable to our users. We should discuss at the next call.

Additionally, I wonder if the calling convention is the reason for crash in https://github.com/pythonnet/pythonnet/pull/980 @SnGmng

filmor:
This is a pure .NET Core problem AFAIK (and the article is from 2018, it could be that this is fixed on .NET Core 3). Also ADL allows you to choose the style and fall back to delegates for unsupported cases.

SnGmng:
> It has however the catch of being LGPL

Well they mentioned if you dont want LGPL you can still ask them for a custom license:
"For companies, established open-source projects, or individuals wanting a custom license, please chuck us an email."


> Additionally, reading this I wonder if the calling convention is the reason for crash in #980 @SnGmng

This could be the issue, maybe. Should i try to use LoadLibrary and GetProcAdress instead?

lostmsu:
@SnGmng you can try fiddling with [DllImportAttribute.CallingConvention](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.dllimportattribute.callingconvention?view=netframework-4.8)

lostmsu:
I started prototyping Python C API interop layer: https://github.com/losttech/Python.Runtime.Native/tree/InterfaceBasedApiSets

[...]

PR #1322, opened 2020-12-13
Loader

filmor:

Integrates clr-loader.

Continues #857, last stepping stone for #984.

### Any other comments?

...

### Checklist


filmor:
@lostmsu This approach (`Python.Loader.dll` explicitly modifies and adjusts `Python.Runtime.dll`) works nicely for embedding into Python, but I can't get it to work in the other direction right no (embedding Python in C#), as the `Python.Runtime` seems to be eagerly loaded. Do you still have your branch were you stubbed out the Python functions into an interface? We can discuss this a bit more on Thursday.

Recap: The main reason we have to do something like this is that `__Internal` as the DLL name for P/Invoke only works on Mono (allows us to preload the DLL using the respective dynamic linkers, for which we already have a working implementation), and it's possible to do something similar for .NET Core, but definitely not for .NET Framework.

filmor:
#987 is relevant.

lostmsu:
@filmor I am actually working on #987. It is currently blocked on the release of https://github.com/dotnet/roslyn/issues/49760 . From the looks of it, will be released with the next VS 2019 16.9.x preview hopefully in a couple of weeks (maybe early Jan because of holidays).

lostmsu:
Oh, forgot to mention. I expect the default value for `Runtime.PythonDLL` to be `null`, and behave identical to `__Internal`, so loader should not need to change it.

filmor:
On all runtimes?

lostmsu:
Yes. However, I think we need to have TypeOffset code on Python side. E.g. clr module should pass a dict describing TypeOffset to Python.Runtime during initialization. That way we can have single Python.Runtime handle any version of Python.

filmor:
I'll have a look into this. Worst case, we just run our pycparser once on first start and generate and cache the data in an easily readable format.

Issue #1312, opened 2020-12-10
operator overloading


caxelrud:
I am using pythonnet with the .Net library Infer.net (https://dotnet.github.io/infer/ ). 
This library has some operator overloading as the '<'.
The following line fail because of the `tomorrowsTime < 18.0` part:
`probTripTakesLessThan18Minutes = engine.Infer[Bernoulli](tomorrowsTime < 18.0).GetProbTrue()
`
How to deal with operator overloading in Pythonnet?
Thanks!

caxelrud:
I got it working using the following:
`
threshold = Variable.New[Double].Named("threshold")
threshold.ObservedValue = 18.0

probTripTakesLessThan18Minutes =engine.Infer[Bernoulli(tomorrowsTime.op_GreaterThan(threshold,tomorrowsTime)).GetProbTrue()
`

Issue #1311, opened 2020-12-10
Follow up on remaining paramArray errors


filmor:
The following test cases created in PR #1304 do not work right now:

```python
    res = MethodTest.ParamsArrayOverloaded(1, 2, i=1)
    assert res == "with params-array"

    res = MethodTest.ParamsArrayOverloaded(paramsArray=[], i=1)
    assert res == "with params-array"
``` 

The respective calls are not matched to the C# method.

PR #1324, opened 2020-12-16
Operator overloads support

christabella:
Adapts @amos402's PR https://github.com/pythonnet/pythonnet/pull/907/ for the latest Pythonnet version. I can also modify Amos's PR directly if that is more convenient for @amos402 @lostmsu who have already made comments on it.

- Tests are kept the same, as well as the links in classmanager, typemanager, and runtime.
- However, the methodbinder logic in Amos's PR is quite different from the current methodbinder and cannot be copied directly.
  - Renamed 2 variables `pynargs` and `clrnargs` for readability and consistency with the rest of methodbinder.
- Added [specialized slot types](https://docs.python.org/3/c-api/typeobj.html#slots) (e.g. nb_subtract, nb_and) from the PyNumberMethods struct. Because a recent PR https://github.com/pythonnet/pythonnet/pull/1292/files defined the TypeOffset and ITypeOffset classes separately such that they no longer depends on interop3x.cs of a specific Python version.
- Put bit shift operators (`<<` and `>>`) a separate test because they are quite different from the rest of the operators (`+/-/&` etc.)---the C# language specification is such that that shift operators always need an int for the second arg: https://stackoverflow.com/questions/7586887/c-sharp-shift-operator-overload. Instead of doing `c = a >> b`, we have to do `c = a >> b.Num`  or get an ArgumentException.
- Added other operators including reverse operators (e.g. `__radd__`).

Although Amos's PR had some params array tests, it looks like params array handling is currently being looked into by @filmor  (https://github.com/pythonnet/pythonnet/pull/1304 and https://github.com/pythonnet/pythonnet/issues/1311), so I removed those for now, since they are not related to operator overloading.

Fixes https://github.com/pythonnet/pythonnet/issues/906 and https://github.com/pythonnet/pythonnet/issues/1312.

### Suggestions/future work
- Unary operators

### Checklist

  - Can add docs once reverse operators are handled.


christabella:
2 failing tests for Python 3.9 on windows, but it seems to be an unrelated test `test_subclass.py:test_events()`

 image

src/runtime/operatormethod.cs
```diff
@@ -0,0 +1,124 @@
...
+using System.Diagnostics;
+using System.Reflection;
+using System.Runtime.InteropServices;
+using System.Text;
+
+namespace Python.Runtime
+{
+    internal static class OperatorMethod
+    {
+        public static Dictionary<string, Tuple<string, int>> OpMethodMap { get; private set; }
```
tminka, line 12: Please address the numerous comments that were made on this file in PR #907

tminka: For example, there is too little documentation in this file.

src/runtime/typemanager.cs
```diff
@@ -280,6 +280,7 @@ internal static IntPtr CreateType(ManagedType impl, Type clrType)
             {
                 throw new PythonException();
             }
+            OperatorMethod.FixupSlots(type, clrType);
```
tminka, line 283: Please address the comment that was made on this file in PR #907

christabella: I have moved it up before pytype_ready

src/runtime/operatormethod.cs
```diff
@@ -0,0 +1,128 @@
...
+            OpMethodMap = new Dictionary<string, Tuple<string, int>>
+            {
+                ["op_Addition"] = Tuple.Create("__add__", TypeOffset.nb_add),
+                ["op_Subtraction"] = Tuple.Create("__sub__", TypeOffset.nb_subtract),
+                ["op_Multiply"] = Tuple.Create("__mul__", TypeOffset.nb_multiply),
+#if PYTHON2
+                ["op_Division"] = Tuple.Create("__div__", TypeOffset.nb_divide),
+#else
+                ["op_Division"] = Tuple.Create("__truediv__", TypeOffset.nb_true_divide),
+#endif
```
lostmsu, line 31: PYTHON2 is gone now

lostmsu (changes requested):
Overall looks good as a starting point. We need to keep in mind, that his does not give us full operator support.

1. In C# and .NET only one argument of a binary operator must match the type, that defines it, and it does not have to be the first argument. This needs to be fixed in this PR: when adding operator methods only the methods where the left argument is of the parent type must be considered (and maybe also inherited ones). I think without this attempts to call the following overloaded operator on two instances of `OperableObject` would fail from Python:
```csharp
class OperableObject {
  ... OperableObject operator+(int left, OperableObject right) => ...;
  ... OperableObject operator+(OperableObject left, OperableObject right) => ...;
}
```

2. Binary operators, that have the first argument of different type need to be handled correctly. I think this should be done in a separate PR, because I expect the amount of work to be comparable or even greater: loading a new type might add overloads to existing type's `__add__` for example.

3. A few operators are not covered here. I don't think all have to be done in the same PR. The full list in C# is: https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/operator-overloading . Notably missing comparison, bitwise, and obscure stuff like conversion to `true`/`false`.

src/runtime/methodbinder.cs
```diff
@@ -677,14 +723,21 @@ static Type TryComputeClrArgumentType(Type parameterType, IntPtr argument, bool
...
+                        // this is separate above because an operator method cannot have
+                        // keyword args, default args, or params arrays (exclusive cases)
                         match = false;
                     }
                 }
+                if (isOperator && defaultArgList.Count == 0)
+                {
+                    // If no default arguments were provided for an operable object.
+                    defaultArgList = null;
+                }
```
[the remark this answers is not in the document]
christabella: Operator methods can enter the branch this is in (`pynargs < clrnargs && (!paramsArray || pynargs == clrnargs - 1))`) because there will be 2 clr args (`a` and `b`) but only one py arg (`b`), since the method is bound to `a` (`a.__add__(b)`). So we don't want to set `match = false` since it's still considered a match.

christabella:
Thank you for the review. Handling operators with asymmetrical types is a really good point. I have added the "forward" case where the arguments are OperableObject and int; I will add the "reverse" case where the arguments are int and OperableObject.

lostmsu:
@christabella I think the reverse case can be done in a separate PR. Until it is in, do you want me to review this one?

christabella:
@lostmsu Yes please, would appreciate your review once again.

amos402 (commented):
The original PR #907 reduced some iterations, redundant checking, and heap allocation in order to improve the performance for the binding behavior, also it included the new style params matching, so it was quite different from the master branch.
If you don't need those enhance you don't need to port them all.

src/runtime/methodbinder.cs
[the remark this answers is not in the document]
christabella: Thanks for the catch

src/embed_tests/TestOperator.cs
```diff
@@ -0,0 +1,209 @@
...
+            }
+
+            public static OperableObject operator +(OperableObject a, OperableObject b)
+            {
+                return new OperableObject(a.Num + b.Num);
+            }
+            public static OperableObject operator +(OperableObject a, int b)
+            {
+                return new OperableObject(a.Num + b);
+            }
```
lostmsu, line 47: Please, add an overload `OperableObject operator+ (int a, OperableObject b)` before these two to ensure it does not break adding two instances of `OperableObject`.

src/runtime/methodbinder.cs
```diff
@@ -633,32 +666,52 @@ static Type TryComputeClrArgumentType(Type parameterType, IntPtr argument, bool
...
             out bool paramsArray,
             out ArrayList defaultArgList,
             out int kwargsMatched,
             out int defaultsNeeded)
         {
             defaultArgList = null;
             var match = false;
-            paramsArray = parameters.Length > 0 ? Attribute.IsDefined(parameters[parameters.Length - 1], typeof(ParamArrayAttribute)) : false;
-            var kwargCount = kwargDict.Count;
+            int clrnargs = parameters.Length;
```
lostmsu, line 691: Please, don't introduce shorthands.
Same for `pynargs`.

lostmsu (changes requested):

src/runtime/methodbinder.cs
christabella: would `numClrArgs` and `numPyArgs` respectively be okay? I agree that the names are not intuitive, I got the idea from another part of the code so I thought maybe it was a naming convention (maybe I could change that too?): https://github.com/pythonnet/pythonnet/blob/2972f49cee96eec743d99d229886245485614d7b/src/runtime/indexer.cs#L71

src/embed_tests/TestOperator.cs
christabella: ok, then I will add the equivalent for all the other operators as well. 

actually I found out that handling the reverse operators `+(int a, OperableObject b)` is just a matter of 1 line of code, should I just include that in this PR too along with the tests?

lostmsu: Sure, if you think so.

src/runtime/methodbinder.cs
lostmsu: Just should not rename without necessity, as it expands lines affected by change.

lostmsu: E.g. you should revert the part where you rename local and replace `parameters.Length` with `clrnargs`.

src/embed_tests/TestOperator.cs
christabella: Actually nevermind the reverse operators was more involved, will leave for next PR

src/runtime/methodbinder.cs
lostmsu: @christabella please, rename `pynargs` back too.

lostmsu (changes requested):
@christabella please, revert rename to `pynargs` too.

src/embed_tests/TestOperator.cs
```diff
@@ -0,0 +1,238 @@
...
+            PythonEngine.Initialize();
+        }
+
+        [OneTimeTearDown]
+        public void Dispose()
+        {
+            PythonEngine.Shutdown();
+        }
+
+        public class SampleClass
```
tminka, line 24: This is not used anymore.

src/runtime/operatormethod.cs
```diff
@@ -0,0 +1,128 @@
...
+using System.Text;
+
+namespace Python.Runtime
+{
+    internal static class OperatorMethod
+    {
+        // Maps the compiled method name in .NET CIL (e.g. op_Addition) to
+        // the equivalent Python operator (e.g. __add__) as well as the offset
+        // that identifies that operator's slot (e.g. nb_add) in heap space.
+        public static Dictionary<string, SlotDefinition> OpMethodMap { get; private set; }
```
tminka, line 15: Use an XML comment

src/runtime/methodbinder.cs
```diff
@@ -342,20 +342,40 @@ internal Binding Bind(IntPtr inst, IntPtr args, IntPtr kw, MethodBase info, Meth
                 bool paramsArray;
                 int kwargsMatched;
                 int defaultsNeeded;
-
-                if (!MatchesArgumentCount(pynargs, pi, kwargDict, out paramsArray, out defaultArgList, out kwargsMatched, out defaultsNeeded))
+                bool isOperator = OperatorMethod.IsOperatorMethod(mi);  // e.g. op_Addition is defined for OperableObject
+                if (isOperator )
+                {
+                    var a = 0;
+                    a++;
```
tminka, line 349: This code seems pointless.

src/runtime/methodbinder.cs
```diff
@@ -519,6 +541,13 @@ static IntPtr HandleParamsArray(IntPtr args, int arrayStart, int pyArgCount, out
                         op = Runtime.PyTuple_GetItem(args, paramIndex);
                     }
                 }
+                if (isOperator && paramIndex == 0)
+                {
+                    // After we've obtained the first argument from Python, we need to skip the first argument of the CLR
+                    // because operator method is a bound method in Python
+                    paramIndex++; // Leave the first .NET param as null (margs).
+                    parameter = pi[paramIndex];
+                }
```
tminka, line 550: It makes more sense to put this block before the assignment to parameter, so that it doesn't need to be done again.

tminka: The cleanest approach would be:
```
int startingParamIndex = isOperator ? 1 : 0;
for(int paramIndex = startingParamIndex; ...)
```

src/runtime/operatormethod.cs
```diff
@@ -0,0 +1,128 @@
...
+            }
+            public string MethodName { get; }
+            public int TypeOffset { get; }
+        }
+        private static PyObject _opType;
+
+        static OperatorMethod()
+        {
+            // TODO: Rich compare, inplace operator support
+            OpMethodMap = new Dictionary<string, SlotDefinition>
```
tminka, line 31: Add a helpful comment:
```
            // .NET operator method names are documented at:
            // https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/operator-overloads
            // Python operator methods and slots are documented at:
            // https://docs.python.org/3/c-api/typeobj.html
```

src/runtime/methodbinder.cs
christabella: Actually we would still need to do `op = Runtime.PyTuple_GetItem(args, paramIndex);` in a previous line so we can't completely skip `paramIndex=0` https://github.com/pythonnet/pythonnet/blob/1f40564cbff0f109b6042973527b032be194b7e8/src/runtime/methodbinder.cs#L519

I tried to convey that in the comment here, "After we've obtained the first argument from Python"

tminka: I see, so we want op to refer to args[0], while parameter refers to pi[1].  If there were a third argument, we would want op to refer to args[1], while parameter refers to pi[2].  So why does the 'if' condition include 'paramIndex == 0'?

christabella: Since we're looking at binary operators here there would be 1 python arg (just `b` from `a.__add__(b)`), while the CLR operator e.g. op_Addition has 2 args. so we need to take the first arg `b` from Python and translate it into the second arg `b` of the CLR operator

I wrote it in this comment in another function called before this one, but maybe it's worth writing the above here too?
https://github.com/pythonnet/pythonnet/pull/1324/files#diff-237dc718ad5424f5476d2ebbbc4e5164c95901bd2a4c2daaf8ca6fc9f9d3a5fbR701-R702

tminka: Yes, and the paramIndex==0 condition should be omitted if it is not needed.

src/runtime/classmanager.cs
```diff
@@ -470,6 +470,10 @@ private static ClassInfo GetClassInfo(Type type)
 
                 ob = new MethodObject(type, name, mlist);
                 ci.members[name] = ob;
+                if (OperatorMethod.IsOperatorMethod(mlist[0]))
```
lostmsu, line 473: NIT: This should really be `mlist.Any(OperatorMethod.IsOperatorMethod)`

src/runtime/methodbinder.cs
```diff
@@ -342,20 +342,35 @@ internal Binding Bind(IntPtr inst, IntPtr args, IntPtr kw, MethodBase info, Meth
                 bool paramsArray;
                 int kwargsMatched;
                 int defaultsNeeded;
-
-                if (!MatchesArgumentCount(pynargs, pi, kwargDict, out paramsArray, out defaultArgList, out kwargsMatched, out defaultsNeeded))
+                bool isOperator = OperatorMethod.IsOperatorMethod(mi);  // e.g. op_Addition is defined for OperableObject
```
lostmsu, line 345: NIT: the comment is not necessary

src/runtime/methodbinder.cs
```diff
@@ -519,6 +536,13 @@ static IntPtr HandleParamsArray(IntPtr args, int arrayStart, int pyArgCount, out
                         op = Runtime.PyTuple_GetItem(args, paramIndex);
                     }
                 }
+                if (isOperator && paramIndex == 0)
+                {
+                    // After we've obtained the first argument from Python, we need to skip the first argument of the CLR
+                    // because operator method is a bound method in Python
+                    paramIndex++; // Leave the first .NET param as null (margs).
+                    parameter = pi[paramIndex];
+                }
```
lostmsu, line 545: Seeing this discussion and `bool isOperator` doing control flow actually led me to think, that the design around operators would be much better if instead an `OperatorBinder` class would be created to derive from `MethodBinder`, and the control flow around branches, that use `isOperator` be refactored so that `OperatorBinder` could do its thing.

Or even better it could just preprocess arguments for or postprocess results of `TryConvertArguments`.

@christabella @tminka do you think you would have time to do that? It would be rather involved.

lostmsu (commented):
Approved, unless you have time and desire to refactor with `OperatorBinder`

src/runtime/methodbinder.cs
tminka: I agree that it is better to remove `IsOperator` from `TryConvertArguments` and use preprocessing and postprocessing instead.  Especially since there is already some preprocessing and postprocessing going on.

christabella: Thanks for the suggestion. When handling reverse operators in the next PR, perhaps we could do the necessary pre- and post-processing of `TryConvertArguments` and remove `isOperator` from `TryConvertArguments`, as @tminka and I just discussed.

src/runtime/methodbinder.cs
christabella: This will be removed entirely anyway in the next PR since `isOperator` is no longer passed into `TryConvertArguments` as discussed https://github.com/pythonnet/pythonnet/pull/1324/files

src/runtime/methodbinder.cs
christabella: The next PR after this one is ready on my fork https://github.com/christabella/pythonnet/pull/1/files where we do preprocessing and postprocessing instead of passing `isOperator` into `TryConvertArguments`, so this part of the code is removed completely in that one.

lostmsu:
@christabella this looks good to me. But considering the upcoming forward operator support and the `OperatorBinder` are going to rewrite parts of this PR, it will make sense to just continue updating this PR.

christabella:
> @christabella this looks good to me. But considering the upcoming forward operator support and the `OperatorBinder` are going to rewrite parts of this PR, it will make sense to just continue updating this PR.

@lostmsu sure I can do that, if you think it makes sense having seen the upcoming https://github.com/christabella/pythonnet/pull/1

lostmsu:
@christabella actually, I suggested to create an `OperatorBinder` class and possibly `OperatorBinding` class similar to existing `ConstructorBinder` and `ConstructorBinding`, which also do special argument handling, but for constructors.

christabella:
> @christabella actually, I suggested to create an `OperatorBinder` class and possibly `OperatorBinding` class similar to existing `ConstructorBinder` and `ConstructorBinding`, which also do special argument handling, but for constructors.
> 
> You could probably also move some/all of `operatormethod.cs` into it.
> 
> But! It is not a requirement, you can also just leave code as-is.

@lostmsu thanks for clarifying, I understand now. So that would be quite a refactor, for now I would try to keep this PR as-is.

src/runtime/methodbinder.cs
```diff
@@ -342,20 +342,53 @@ internal Binding Bind(IntPtr inst, IntPtr args, IntPtr kw, MethodBase info, Meth
...
                 }
                 var outs = 0;
+                int clrnargs = pi.Length;
+                isOperator = isOperator && pynargs == clrnargs - 1;  // Handle mismatched arg numbers due to Python operator being bound.
+                // Preprocessing pi to remove either the first or second argument.
+                bool isReverse = isOperator && OperatorMethod.IsReverse((MethodInfo)mi);  // Only cast if isOperator.
+                if (isOperator && !isReverse) {
+                    // The first Python arg is the right operand, while the bound instance is the left.
+                    // We need to skip the first (left operand) CLR argument.
+                    pi = pi.Skip(1).Take(1).ToArray();
```
tminka, line 358: Omit Take(1) to support unary operators.

src/runtime/classmanager.cs
```diff
@@ -470,6 +470,17 @@ private static ClassInfo GetClassInfo(Type type)
...
+                if (mlist.Any(OperatorMethod.IsOperatorMethod))
+                {
+                    string pyName = OperatorMethod.GetPyMethodName(name);
+                    string pyNameReverse = pyName.Insert(2, "r");
+                    MethodInfo[] forwardMethods, reverseMethods;
+                    OperatorMethod.FilterMethods(mlist, out forwardMethods, out reverseMethods);
+                    // Only methods where the left operand is the declaring type.
+                    ci.members[pyName] = new MethodObject(type, name, forwardMethods);
+                    // Only methods where only the right operand is the declaring type.
+                    ci.members[pyNameReverse] = new MethodObject(type, name, reverseMethods);
```
tminka, line 482: What if there are no reverseMethods?

src/runtime/classmanager.cs
```diff
@@ -470,6 +470,17 @@ private static ClassInfo GetClassInfo(Type type)
 
                 ob = new MethodObject(type, name, mlist);
                 ci.members[name] = ob;
+                if (mlist.Any(OperatorMethod.IsOperatorMethod))
+                {
+                    string pyName = OperatorMethod.GetPyMethodName(name);
+                    string pyNameReverse = pyName.Insert(2, "r");
```
tminka, line 476: Use ReversePyMethodName

src/runtime/methodbinder.cs
```diff
@@ -644,19 +698,26 @@ static Type TryComputeClrArgumentType(Type parameterType, IntPtr argument, bool
...
                 match = true;
             }
             else if (positionalArgumentCount < parameters.Length && (!paramsArray || positionalArgumentCount == parameters.Length - 1))
             {
-                // every parameter past 'positionalArgumentCount' must have either
-                // a corresponding keyword argument or a default parameter
                 match = true;
+                // operator methods will have 2 CLR args but only one Python arg,
+                // since Python operator methods are bound
+                if (isOperator)
```
tminka, line 712: `if (isOperator && positionalArgumentCount == parameters.Length - 1)`

src/embed_tests/TestOperator.cs
```diff
@@ -0,0 +1,263 @@
...
+assert c.Num == a.Num & b.Num
+
+c = a | b
+assert c.Num == a.Num | b.Num
+
+c = a ^ b
+assert c.Num == a.Num ^ b.Num
+");
+        }
+        [Test]
```
tminka, line 170: Please add this:
```csharp
        [Test]
        public void OperatorOverloadMissingArgument()
        {
            string name = string.Format("{0}.{1}",
                typeof(OperableObject).DeclaringType.Name,
                typeof(OperableObject).Name);
            string module = MethodBase.GetCurrentMethod().DeclaringType.Namespace;

            Assert.Throws<PythonException>(() =>
            PythonEngine.Exec($@"
from {module} import *
cls = {name}
a = cls(2)
b = cls(10)
a.op_Addition()
"));
        }
```

src/embed_tests/TestOperator.cs
```diff
@@ -0,0 +1,263 @@
...
+                typeof(OperableObject).DeclaringType.Name,
+                typeof(OperableObject).Name);
+            string module = MethodBase.GetCurrentMethod().DeclaringType.Namespace;
+
+            PythonEngine.Exec($@"
+from {module} import *
+cls = {name}
+a = cls(2)
+b = cls(10)
+c = a + b
```
tminka, line 148: At this point, both of these calls will succeed (and do the same thing):
```
a.op_Addition(b)
a.op_Addition(a,b)
```
Is that desirable behavior?

src/runtime/methodbinder.cs
```diff
@@ -342,20 +342,53 @@ internal Binding Bind(IntPtr inst, IntPtr args, IntPtr kw, MethodBase info, Meth
...
-
-                if (!MatchesArgumentCount(pynargs, pi, kwargDict, out paramsArray, out defaultArgList, out kwargsMatched, out defaultsNeeded))
+                bool isOperator = OperatorMethod.IsOperatorMethod(mi);
+                if (!MatchesArgumentCount(pynargs, pi, kwargDict, isOperator, out paramsArray, out defaultArgList, out kwargsMatched, out defaultsNeeded))
                 {
                     continue;
                 }
                 var outs = 0;
+                int clrnargs = pi.Length;
+                isOperator = isOperator && pynargs == clrnargs - 1;  // Handle mismatched arg numbers due to Python operator being bound.
```
tminka, line 352: If pynargs == clrnargs - 1, then we know we have the correct number of arguments.  So I think the logic can be simplified by including (pynargs == clrnargs - 1) in the original definition of isOperator, not passing isOperator to MatchesArgumentCount, and using 
```csharp
if (!MatchesArgumentCount(...) && !isOperator)
```

src/runtime/classmanager.cs
christabella: Currently it will just create a method that's never used; I can fix that by adding a hasReverse field in the SlotDefinition struct.

src/embed_tests/TestOperator.cs
christabella: So the first statement would have previously failed and now succeeds. I guess this is a question for @lostmsu if we want to let something that's wrongly written, not fail

src/runtime/classmanager.cs
christabella: So if C# reverse operator isn't defined, so the Python reverse name will have an empty method list in MethodObject

So if a user calls `b.__radd__(a)` by writing `a + b` where there is no forward +() defined on `a` so Python resorts to __radd__, AND no reverse +() defined on `b`, what happens?

Another unrelated issue is that of doing unnecessary work, if the Python slot doesn't support reverse e.g. `__truediv__`

src/embed_tests/TestOperator.cs
lostmsu: `a.op_Addition(a, b)` must be failing for operators, defined in C#.

src/runtime/classmanager.cs
lostmsu: It is also possible to have no "forward" methods.

If no reverse methods are defined, the corresponding slot should not be set. Same for forward.

> Python slot doesn't support reverse e.g. `__truediv__`

Not sure what you mean. There's `__rtruediv__`

src/embed_tests/TestOperator.cs
```diff
@@ -0,0 +1,299 @@
...
+
+c = a + b
+assert c.Num == a + b.Num
+
+c = a - b
+assert c.Num == a - b.Num
+
+c = a * b
+assert c.Num == a * b.Num
+
```
tminka, line 265:
```
c = a / b
assert c.Num == a // b.Num
```

src/embed_tests/TestOperator.cs
tminka: The code I wrote is Python code.

lostmsu: Breaking should not be a big deal, since we are making major version bump with (hopefully) a good overhaul.
But I don't think `a.op_Addition(a, b)` should be working now.

src/runtime/methodbinder.cs
christabella: Made this change but reverted it since it breaks other test cases involving default args, see: https://github.com/pythonnet/pythonnet/pull/1324/checks?sha=8cce61dd5def72485e599a071e9fd3bfc12346c6

src/runtime/classmanager.cs
christabella: That's true, I've just put a check then to not set the slot method if the method list is empty

src/runtime/methodbinder.cs
```diff
@@ -711,12 +711,8 @@ static Type TryComputeClrArgumentType(Type parameterType, IntPtr argument, bool
             else if (positionalArgumentCount < parameters.Length && (!paramsArray || positionalArgumentCount == parameters.Length - 1))
             {
                 match = true;
-                // operator methods will have 2 CLR args but only one Python arg,
-                // since Python operator methods are bound
-                if (isOperator && positionalArgumentCount == parameters.Length - 1)
+                if (positionalArgumentCount == parameters.Length - 1)
```
tminka: Since isOperator will always be false, the correct change is to remove this branch entirely.

src/runtime/methodbinder.cs
tminka: You didn't implement the change correctly.  See below.

src/runtime/methodbinder.cs
christabella: Thanks for catching this, I completely forgot that the premature exit was new...

src/runtime/operatormethod.cs
```diff
@@ -0,0 +1,172 @@
...
+                ["op_Subtraction"] = new SlotDefinition("__sub__", TypeOffset.nb_subtract),
+                ["op_Multiply"] = new SlotDefinition("__mul__", TypeOffset.nb_multiply),
+                ["op_Division"] = new SlotDefinition("__truediv__", TypeOffset.nb_true_divide),
+                ["op_BitwiseAnd"] = new SlotDefinition("__and__", TypeOffset.nb_and),
+                ["op_BitwiseOr"] = new SlotDefinition("__or__", TypeOffset.nb_or),
+                ["op_ExclusiveOr"] = new SlotDefinition("__xor__", TypeOffset.nb_xor),
+                ["op_LeftShift"] = new SlotDefinition("__lshift__", TypeOffset.nb_lshift),
+                ["op_RightShift"] = new SlotDefinition("__rshift__", TypeOffset.nb_rshift),
+                ["op_Modulus"] = new SlotDefinition("__mod__", TypeOffset.nb_remainder),
+                ["op_OneComplement"] = new SlotDefinition("__invert__", TypeOffset.nb_invert),
```
tminka, line 49: Are these two cases tested?

christabella: Thanks, I've added them

CHANGELOG.md
```diff
@@ -44,6 +44,7 @@ details about the cause of the failure
 -    Made it possible to call `ToString`, `GetHashCode`, and `GetType` on inteface objects
 -    Fixed objects returned by enumerating `PyObject` being disposed too soon
 -    Incorrectly using a non-generic type with type parameters now produces a helpful Python error instead of throwing NullReferenceException
+-    Python operator method will call C# operator method for supported binary and unary operators ([#1324][p1324]).
```
lostmsu, line 47: NIT: This should be in the "Added" section.

src/runtime/classmanager.cs
```diff
@@ -470,6 +470,19 @@ private static ClassInfo GetClassInfo(Type type)
 
                 ob = new MethodObject(type, name, mlist);
                 ci.members[name] = ob;
+                if (mlist.Any(OperatorMethod.IsOperatorMethod))
+                {
+                    string pyName = OperatorMethod.GetPyMethodName(name);
+                    string pyNameReverse = OperatorMethod.ReversePyMethodName(pyName);
+                    MethodInfo[] forwardMethods, reverseMethods;
+                    OperatorMethod.FilterMethods(mlist, out forwardMethods, out reverseMethods);
```
lostmsu, line 478: NIT: you can do `out var forwardMethods`. No need to declare the variables separately in the above line

lostmsu (approved):

src/runtime/classmanager.cs
christabella: Thanks for the suggestions, I included this small change in a related PR: https://github.com/pythonnet/pythonnet/pull/1347/

Issue #1323, opened 2020-12-15
illegal use of GetDelegateFromFunctionPointer causes problems on mono 6.12


benoithudson:
### Environment

-   Pythonnet version: master
-   Python version: 3.7 (probably any)
-   Operating System: mono on any platform, not dotnet on any platform

### Details

-   Describe what you were trying to get done.

Working on the reload improvements, we find that GetDelegateFromFunctionPointer is sometimes returning weird types that can't be casted to the proper delegate type. This is apparently a known issue in pythonnet on dotnet as well though we haven't hit that problem on that platform:
https://github.com/dotnet/runtime/issues/9074

https://docs.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.marshal.getdelegateforfunctionpointer?view=net-5.0
"You can use this method only for pure unmanaged function pointers."

We'd need to change users of NativeCall to make sure they are only calling non-managed functions, or else undefined behaviour can happen.

-   What commands did you run to trigger this issue? If you can provide a

Unit tests for the reload work that @BadSingleton is working on (#1287)

benoithudson:
Fixed by #1328

PR #1328, opened 2020-12-17
Fix illegal delegate usage

BadSingleton:

Cache the delegates to native code we've created so that when we need to
call that delegate, we don't ask for a delegate from a native pointer
(that is a delegate to managed code.)

This fixes issue #1323 .

### Any other comments?

This is needed in order for #1287 to pass the domain reload tests on Mono.

### Checklist

### DIFF
--- a/src/runtime/nativecall.cs
+++ b/src/runtime/nativecall.cs
@@ -45,8 +45,8 @@ namespace Python.Runtime
             Delegate d = null;
             if (!Interop.allocatedThunks.TryGetValue(fp, out d))
             {
-                // Use Marshal.GetDelegateForFunctionPointer<> directly after upgrade the framework
-                d = Marshal.GetDelegateForFunctionPointer(fp, typeof(T));
+                // We don't cache this delegate because this is a pure delegate ot unmanaged.
+                d = Marshal.GetDelegateForFunctionPointer<T>(fp);
             }
             return (T)d;
         }
--- a/src/runtime/nativecall.cs
+++ b/src/runtime/nativecall.cs
@@ -40,10 +40,15 @@ namespace Python.Runtime
             return d(a1, a2, a3);
         }
 
-        private static T GetDelegate<T>(IntPtr fp) where T: Delegate
+        internal static T GetDelegate<T>(IntPtr fp) where T: Delegate
         {
-            // Use Marshal.GetDelegateForFunctionPointer<> directly after upgrade the framework
-            return (T)Marshal.GetDelegateForFunctionPointer(fp, typeof(T));
+            Delegate d = null;
+            if (!Interop.allocatedThunks.TryGetValue(fp, out d))
+            {
+                // Use Marshal.GetDelegateForFunctionPointer<> directly after upgrade the framework
+                d = Marshal.GetDelegateForFunctionPointer(fp, typeof(T));
+            }
+            return (T)d;
         }
     }
 }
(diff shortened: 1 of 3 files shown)


BadSingleton:
Test failures are the same as the ones in the current master: https://github.com/pythonnet/pythonnet/runs/1571830014

src/runtime/nativecall.cs
```diff
@@ -40,10 +40,15 @@ public static int Int_Call_3(IntPtr fp, IntPtr a1, IntPtr a2, IntPtr a3)
...
-        private static T GetDelegate<T>(IntPtr fp) where T: Delegate
+        internal static T GetDelegate<T>(IntPtr fp) where T: Delegate
         {
-            // Use Marshal.GetDelegateForFunctionPointer<> directly after upgrade the framework
-            return (T)Marshal.GetDelegateForFunctionPointer(fp, typeof(T));
+            Delegate d = null;
+            if (!Interop.allocatedThunks.TryGetValue(fp, out d))
+            {
+                // Use Marshal.GetDelegateForFunctionPointer<> directly after upgrade the framework
+                d = Marshal.GetDelegateForFunctionPointer(fp, typeof(T));
```
filmor, line 49: Can you also apply the comment here? The generic function was introduced in .NET 4.5.1.

filmor: Actually, maybe just remove the comment. You are not writing this into `allocatedThunks` right now, btw.

BadSingleton: Yes, I don't cache these because they truly point to unmanaged code. We should only cache the function pointers to managed code.

src/runtime/nativecall.cs
```diff
@@ -40,10 +40,15 @@ public static int Int_Call_3(IntPtr fp, IntPtr a1, IntPtr a2, IntPtr a3)
...
         {
-            // Use Marshal.GetDelegateForFunctionPointer<> directly after upgrade the framework
-            return (T)Marshal.GetDelegateForFunctionPointer(fp, typeof(T));
+            Delegate d = null;
+            if (!Interop.allocatedThunks.TryGetValue(fp, out d))
+            {
+                // We don't cache this delegate because this is a pure delegate ot unmanaged.
+                d = Marshal.GetDelegateForFunctionPointer<T>(fp);
+            }
+            return (T)d;
```
benoithudson, line 51: You'd save a couple instructions with:
```
Delegate d; // no init
if (trygetvalue(out d)) {
  return (T)d;
} else {
  return Marshal.GetDelegate...(); // no cast
}
```

src/runtime/nativecall.cs
```diff
@@ -40,10 +40,15 @@ public static int Int_Call_3(IntPtr fp, IntPtr a1, IntPtr a2, IntPtr a3)
             return d(a1, a2, a3);
         }
 
-        private static T GetDelegate<T>(IntPtr fp) where T: Delegate
+        internal static T GetDelegate<T>(IntPtr fp) where T: Delegate
         {
-            // Use Marshal.GetDelegateForFunctionPointer<> directly after upgrade the framework
-            return (T)Marshal.GetDelegateForFunctionPointer(fp, typeof(T));
+            Delegate d = null;
+            if (!Interop.allocatedThunks.TryGetValue(fp, out d))
```
lostmsu, line 46: NIT: you don't need to assign an initial value to `d`

lostmsu (approved):

PR #1331, opened 2020-12-17
PyIter: do not force dispose previous object upon moving to the next one

lostmsu:

Prior to this change trying to use object returned by iterating over `PyObject` would fail after iterator is closed, because iterator would dispose previously returned objects on close/moving to the next one.

This also adds error handling for `PyIter_Next` calls, which can have other reasons to fail beyond reaching the end of collection.

### Checklist

### DIFF
--- a/src/runtime/runtime.cs
+++ b/src/runtime/runtime.cs
@@ -1854,6 +1854,8 @@ namespace Python.Runtime
 
         [DllImport(_PythonDll, CallingConvention = CallingConvention.Cdecl)]
         internal static extern IntPtr PyIter_Next(IntPtr pointer);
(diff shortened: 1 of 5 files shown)


src/runtime/runtime.cs
```diff
@@ -1855,6 +1855,8 @@ internal static bool PyIter_Check(IntPtr pointer)
 
         [DllImport(_PythonDll, CallingConvention = CallingConvention.Cdecl)]
         internal static extern IntPtr PyIter_Next(IntPtr pointer);
```
filmor, line 1857: Do we still need this?

lostmsu: Yes, there are still a few uses.

filmor (approved):

PR #1336, opened 2020-12-21
TestPythonEngineProperties.SetPythonPath avoids corrupting the module search path for later tests

### LINKED ISSUE #1335: TestPythonEngineProperties.SetPythonPath spoils later tests that load modules
### Environment

-   Pythonnet version: [Latest master](https://github.com/pythonnet/pythonnet/commit/d9e15a75a1f2583d2986355ca685140257145c80)
-   Python version: 3.7.8
-   .NET Runtime: Core 3.1

### Details

Add an embedded test that imports pytest.

Add the following test to TestTypeManager.cs (so that it runs last):
```csharp
        [Test]
        public void PathTest()
        {
            Py.Import("pytest");
        }
```
Then run `dotnet test --runtime any-x64 src/embed_tests`

```
  Failed PathTest [20 ms]
  Error Message:
   Python.Runtime.PythonException : ImportError : No module named pytest
```

The reason for the failure is that an earlier test named TestPythonEngineProperties.SetPythonPath corrupts the Python import mechanism.  Removing that test causes PathTest to pass.  You can see the causal effect most clearly by adding  `Py.Import("pytest");` to the end of the SetPythonPath test, where it will fail.

As explained in the documentation for [Py_SetPath](https://docs.python.org/3/c-api/init.html#c.Py_SetPath), setting the Python path has the side effect of changing sys.executable, sys.prefix, and sys.exec_prefix.  The import mechanism uses those variables to locate modules.  You can confirm that these variables are being altered by making the following change to SetPythonPath:

```csharp
        public void SetPythonPath()
        {
            PythonEngine.Initialize();

            Console.WriteLine(sys.executable);
            Console.WriteLine(sys.prefix);
            Console.WriteLine(sys.exec_prefix);

            string path = PythonEngine.PythonPath;
            PythonEngine.Shutdown();

            PythonEngine.PythonPath = path;
            PythonEngine.Initialize();

            sys = Py.Import("sys");
            Console.WriteLine(sys.executable);
            Console.WriteLine(sys.prefix);
            Console.WriteLine(sys.exec_prefix);

            Assert.AreEqual(path, PythonEngine.PythonPath);
            PythonEngine.Shutdown();
        }
```
The SetPythonPath test should restore the original values of those variables.

tminka:

Improves the test suite of Python.NET

Fixes #1335 

### Any other comments?

This change is not visible to the end user so it doesn't make sense to add it to the CHANGELOG.

### Checklist

### DIFF
--- a/src/embed_tests/TestPythonEngineProperties.cs
+++ b/src/embed_tests/TestPythonEngineProperties.cs
@@ -181,7 +181,22 @@ namespace Python.EmbeddingTest
         public void SetPythonPath()
         {
             PythonEngine.Initialize();
-            string path = PythonEngine.PythonPath;
+
+            dynamic sys = Py.Import("sys");
+            var locals = new PyDict();
+            locals.SetItem("sys", sys);
+            var result = PythonEngine.Eval("';'.join(sys.path)", null, locals.Handle);
+            // This does not work:
+            //var result = PythonEngine.Eval("import sys\n';'.join(sys.path)");
+            string path = (string)result.AsManagedObject(typeof(string));
+
+            // path should not be set to PythonEngine.PythonPath here.
+            // PythonEngine.PythonPath gets the default module search path, not the full search path.
+            // The list sys.path is initialized with this value on interpreter startup;
+            // it can be (and usually is) modified later to change the search path for loading modules.
+            // See https://docs.python.org/3/c-api/init.html#c.Py_GetPath
+            // After PythonPath is set, then PythonEngine.PythonPath will correctly return the full search path. 
+
             PythonEngine.Shutdown();
 
             PythonEngine.PythonPath = path;
--- a/src/embed_tests/TestPythonEngineProperties.cs
+++ b/src/embed_tests/TestPythonEngineProperties.cs
@@ -194,13 +194,8 @@ namespace Python.EmbeddingTest
                 importShouldSucceed = false;
             }
 
-            dynamic sys = Py.Import("sys");
-            var locals = new PyDict();
-            locals.SetItem("sys", sys);
-            var result = PythonEngine.Eval($"'{System.IO.Path.PathSeparator}'.join(sys.path)", null, locals.Handle);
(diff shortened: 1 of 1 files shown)


src/embed_tests/TestPythonEngineProperties.cs
```diff
@@ -181,13 +181,42 @@ public void SetProgramName()
...
+                importShouldSucceed = false;
+            }
+
+            dynamic sys = Py.Import("sys");
+            var locals = new PyDict();
+            locals.SetItem("sys", sys);
+            var result = PythonEngine.Eval($"'{System.IO.Path.PathSeparator}'.join(sys.path)", null, locals.Handle);
+            // This does not work:
+            //var result = PythonEngine.Eval("import sys\n';'.join(sys.path)");
+            string path = (string)result.AsManagedObject(typeof(string));
```
lostmsu, line 203: NIT: you can use `.As<string>()`

src/embed_tests/TestPythonEngineProperties.cs
```diff
@@ -181,13 +181,42 @@ public void SetProgramName()
...
+            }
+            catch
+            {
+                importShouldSucceed = false;
+            }
+
+            dynamic sys = Py.Import("sys");
+            var locals = new PyDict();
+            locals.SetItem("sys", sys);
+            var result = PythonEngine.Eval($"'{System.IO.Path.PathSeparator}'.join(sys.path)", null, locals.Handle);
```
lostmsu, line 200: NIT: you don't need `locals`. You could do something like
`Py.Import("sys").GetAttr("path").As<string[]>()`. Then use .NET `string.Join`

lostmsu (commented):
This seems fine. However, if you wanted this change to add a C# runner for Python tests, I suggest you turn the runner into a separate project instead.

I'd also suggest to use something like https://stackoverflow.com/questions/866816/dynamically-create-tests-in-nunit to have the test display as individual tests.

Issue #766, opened 2018-11-06
Interrupting running Python code from another thread


jduncalf:
Hi, does anyone know the best way to stop a currently running PyScope.Exec() call. My C# app is multi-threaded and allows users to run scripts, however where there's an infinite loop, new Python calls can be executed but I can't stop the original infinite loop.

Thanks.

filmor:
See here https://stackoverflow.com/questions/1420957/stopping-embedded-python.

We currently do not expose `Py_AddPendingCall`, though, but assuming we did (there is PR #692 that uses it), you could probably translate the given example like this:


    class Interrupter
    {
        public Interrupter()
        {
            _interruptAction = DoInterrupt;
        }

        int DoInterrupt(IntPtr arg)
        {
            Exceptions.SetError(Exceptions.KeyboardInterrupt, "interrupted");
        }

        [UnmanagedFunctionPointer(CallingConvention.Cdecl)]
        delegate int PendingCall(IntPtr arg);
        readonly PendingCall _interruptAction;

        void Interrupt()
        {
            IntPtr func = Marshal.GetFunctionPointerForDelegate(_interruptAction);
            Runtime.Py_AddPendingCall(func, IntPtr.Zero);
        }
    }

All `Runtime` functions are `internal`, so you will have to work within `pythonnet`'s codebase itself. A PR adding an `Interrupt()` function on `PythonEngine` and `PyScope` would be highly appreciated :)

jduncalf:
Many thanks, i'll do that.

jduncalf:
Hi, unfortunately I can't get this to work. I basically used the Interrupter class you posted above, but the DoInterrupt() callback never gets invoked.

Any ideas?

jduncalf:
Hi filmor, apologies, got it working. Had to change my C# host apps threading model. As a COM/DCOM MTA server the .exe primary thread was not the thread which eventually called `PyEval_InitThreads`, which is required for the `Py_AddPendingCall` supplied callback to work.

Thanks again.

filmor:
Nice, I'll see whether I can prepare a PR that includes this.

amos402:
@jduncalf I think this extension should be help https://github.com/amos402/py-xtimeout

denfromufa:
anyone would like to submit a PR?

amos402:
I think this is not a problem that relate to pythonnet, the extension I supplied above can handle this case.
Or you can reference here to make a trace callback when you try to interrupt the work thread.
[...]

PR #1333, opened 2020-12-19
Add very simple example editor

gpetrou:
### What does this implement/fix? Explain your changes.

I am doing my first steps in Python .NET and I created a small editor project in order to test the library. Looking for information on how to redirect output to C# and interrupt execution, I saw that other people had the same questions. So, I thought of making this small contribution to get other people started with this library instead of looking for the same things as I did.

### Checklist

Check all those that are applicable and complete.

-   [ ] Make sure to include one or more tests for your change
-   [ ] If an enhancement PR, please create docs and at best an example
-   [x] Add yourself to [`AUTHORS`](../blob/master/AUTHORS.md)
-   [x] Updated the [`CHANGELOG`](../blob/master/CHANGELOG.md)

PR #1337, opened 2020-12-26
Add GetPythonThreadID and Interrupt methods in PythonEngine

gpetrou:

Fixes https://github.com/pythonnet/pythonnet/issues/766 by adding an Interrupt method in PythonEngine class.

### Checklist


src/embed_tests/TestInterrupt.cs
```diff
@@ -0,0 +1,83 @@
...
+
+        [DllImport("Kernel32", EntryPoint = "GetCurrentThreadId", ExactSpelling = true)]
+        private static extern uint GetCurrentThreadId();
+
+        [DllImport("libc", EntryPoint = "pthread_self")]
+        private static extern IntPtr pthread_selfLinux();
+
+        [DllImport("pthread", EntryPoint = "pthread_self", CallingConvention = CallingConvention.Cdecl)]
+        private static extern ulong pthread_selfOSX();
+
```
filmor, line 25: Python 3.8 has `threading.get_native_id()` for this, which we should probably expose in the library with a fallback to these.

filmor:
Copying my comment from #1333:

1. This requires a test (done)
2. Strictly, the thread-id attribute is unsigned only for Python >= 3.7 and we still support 3.6
3. Also, since the original type is long, the "correct" type is (U)IntPtr on everything but Windows (Not quite sure how far we have to go there compatibility-wise, the current implementation could be fine for the usual thread IDs)
4. Maybe we should just expose the API function directly (i.e. RaiseInPythonThread on Exception objects or so) and make Interrupt use that

Additional comments on the current state: The helper API to get the current thread id should be in the library and should use `threading.get_native_id()` on Python >=3.8.

filmor:
It would also be good if you'd just use normal commits now without force-pushing a single commit over and over, otherwise it becomes very difficult to follow your changes. We can (and will) squash this in the end.

src/embed_tests/TestInterrupt.cs
gpetrou: I am not sure how to implement this. I have added some changes, what do you think?

gpetrou:
Please note that I am very unfamiliar with the codebase, so I will need more explanations from you. With regards to:
2. I added a version check. Is this what you mean?
3. I added different calls per OS and version. Is this what you mean?
4. I don't understand what that means. Can you elaborate? Is this still needed?

src/runtime/pythonengine.cs
```diff
@@ -567,6 +576,60 @@ public static void Exec(string code, IntPtr? globals = null, IntPtr? locals = nu
...
+            {
+                return (ulong)pthread_selfLinux();
+            }
+
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.OSX))
+            {
+                return pthread_selfOSX();
+            }
+
+            return 0;
```
filmor, line 606: Should raise an exception if we can't find a suitable implementation.

src/runtime/pythonengine.cs
```diff
@@ -567,6 +576,60 @@ public static void Exec(string code, IntPtr? globals = null, IntPtr? locals = nu
...
+        /// <summary>
+        /// Gets the native thread ID.
+        /// </summary>
+        /// <returns>The native thread ID.</returns>
+        public static ulong GetNativeThreadID()
+        {
+            if (Runtime.PyVersion >= new Version(3, 8))
+            {
+                dynamic threading = Py.Import("threading");
+                return threading.get_native_id();
```
filmor, line 588: Should get the `GIL`, I think.

filmor:
This goes in the right direction :)

Regarding 2.: Yes, that's what I meant.
Regarding 3.: This looks better, but the Windows type is `uint` and `int`. .NET's `int` is always 32bit, and on Windows `long` is also 32bit on both x86 and x64.

src/runtime/pythonengine.cs
gpetrou: Isn't the user supposed to do that when calling the method, as I did in the test?

gpetrou: This does not seem to be working on Ubuntu. Are you sure that this should work? Looking in Cpython tests I see that get_ident is used.

gpetrou:
@filmor can this be merged now?

src/runtime/runtime.cs
```diff
@@ -2143,6 +2143,18 @@ internal static void Py_CLEAR(ref IntPtr ob)
...
+
+        [DllImport(_PythonDll, EntryPoint = "PyThreadState_SetAsyncExc", CallingConvention = CallingConvention.Cdecl)]
+        internal static extern int PyThreadState_SetAsyncExc36Windows(int id, IntPtr exc);
+
+        [DllImport(_PythonDll, EntryPoint = "PyThreadState_SetAsyncExc", CallingConvention = CallingConvention.Cdecl)]
+        internal static extern int PyThreadState_SetAsyncExc37NonWindows(ulong id, IntPtr exc);
+
+        [DllImport(_PythonDll, EntryPoint = "PyThreadState_SetAsyncExc", CallingConvention = CallingConvention.Cdecl)]
+        internal static extern int PyThreadState_SetAsyncExc36NonWindows(long id, IntPtr exc);
+
```
lostmsu, line 2157: It is better to name these according to [64 bit data models](https://en.wikipedia.org/wiki/LP64#64-bit_data_models) as opposed to `Windows` and `NonWindows`.

src/runtime/runtime.cs
```diff
@@ -2143,6 +2143,18 @@ internal static void Py_CLEAR(ref IntPtr ob)
         [DllImport(_PythonDll, CallingConvention = CallingConvention.Cdecl)]
         internal static extern int Py_AddPendingCall(IntPtr func, IntPtr arg);
 
+        [DllImport(_PythonDll, EntryPoint = "PyThreadState_SetAsyncExc", CallingConvention = CallingConvention.Cdecl)]
+        internal static extern int PyThreadState_SetAsyncExc37Windows(uint id, IntPtr exc);
```
lostmsu, line 2147: Why do you need overloads with `uint` and `int`? They are basically equivalent at bit level.

src/runtime/pythonengine.cs
```diff
@@ -567,6 +576,54 @@ public static void Exec(string code, IntPtr? globals = null, IntPtr? locals = nu
...
+                return pthread_selfLinux();
+            }
+
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.OSX))
+            {
+                return pthread_selfOSX();
+            }
+
+            throw new InvalidOperationException("Could not retrieve native thread ID.");
+        }
```
lostmsu, line 601: I don't think this belongs to `PythonEngine`. From what I read, for the test you might be able to use `threading.get_ident()` from Python `thread` module.

src/runtime/pythonengine.cs
```diff
@@ -567,6 +576,54 @@ public static void Exec(string code, IntPtr? globals = null, IntPtr? locals = nu
...
+
+        /// <summary>
+        /// Interrupts the execution of a thread.
+        /// </summary>
+        /// <param name="nativeThreadID">The native thread ID.</param>
+        /// <returns>The number of thread states modified; this is normally one, but will be zero if the thread id isn’t found.</returns>
+        public static int Interrupt(ulong nativeThreadID)
```
lostmsu, line 608: In C# it is better to use `long` type for public method parameters, as (at least older) VB.NET and some other .NET languages do not support `ulong`.

lostmsu (changes requested):

src/runtime/pythonengine.cs
filmor: `get_ident` is a separate thing, the ODs are not compatible with the "remote raise".

src/runtime/runtime.cs
gpetrou: I added them because of "Strictly, the thread-id attribute is unsigned only for Python >= 3.7 and we still support 3.6" comment above. Do you want me to remove them?

lostmsu: @filmor on the binary side there's no difference. The caller, if necessary, can bitcast. Any thoughts?

src/runtime/pythonengine.cs
lostmsu: @filmor are you sure? Documentation is very bad on this, but here's what I found:

`get_ident` [appears to be](https://github.com/python/cpython/blob/6104013838e181e3c698cb07316f449a0c31ea96/Modules/_threadmodule.c#L1208) returning `PyThread_get_thread_ident`
`tstate->thread_id` [is also](https://github.com/python/cpython/blob/41010184880151d6ae02a226dbacc796e5c90d11/Python/pystate.c#L625) `PyThread_get_thread_ident()`

`PyThreadState_SetAsyncExc` [finds thread by its `tstate->thread_id`](https://github.com/python/cpython/blob/41010184880151d6ae02a226dbacc796e5c90d11/Python/pystate.c#L1129)

gpetrou: Using get_ident seems to work. What would you like me to do? Modify GetNativeThreadID method or remove it completely?

lostmsu: Remove it would be better.

lostmsu (changes requested):

src/runtime/pythonengine.cs
filmor: @lostmsu You are right. But then we should still expose `threading.get_ident` or `PyThread_get_thread_ident`, no?

lostmsu: Why? The user can do what tests do now - e.g. call `Py.Import("threading").Invoke("get_ident")`.

@gpetrou It might make sense to mention `threading.get_ident()` in the summary of XML doc for the `Interrupt` method.

gpetrou: I am getting mixed signals between the two of you :) Who is the BDFL for Python.NET? I added the get_ident usage in GetNativeThreadID for now.

src/runtime/runtime.cs
gpetrou: I have removed the int and long declarations for now.

src/runtime/pythonengine.cs
lostmsu: @gpetrou @filmor  well, I am voting for removal, as it is one less thing to support with potential multiple meanings, and having nothing to do with Python embedding on itself.

lostmsu: @filmor ?

src/runtime/pythonengine.cs
```diff
@@ -582,28 +573,8 @@ public static void Exec(string code, IntPtr? globals = null, IntPtr? locals = nu
         /// <returns>The native thread ID.</returns>
         public static ulong GetNativeThreadID()
```
filmor, line 574: If we keep this, it should be renamed to `GetPythonThreadId` and doesn't this require the GIL in general?

gpetrou: Can you please explain why we need GIL here and not in the Interrupt method? GIL is used for both in the tests already.

filmor: And renaming would still be nice, sorry for the back and forth, I misread the docs.

PR #1340, opened 2020-12-30
ABI.Initialize gives a helpful message when the TypeOffset interop class is not configured correctly

tminka:

When changing Python versions, it is easy to put Python.NET into a misconfigured state.  This PR provides a helpful error message in that case.

No

### Any other comments?

This change can be tested by manually editing `configured.props` to have an incorrect `PythonInteropFile`

### Checklist

### DIFF
--- a/src/runtime/Python.Runtime.csproj
+++ b/src/runtime/Python.Runtime.csproj
@@ -19,6 +19,14 @@
     <DefineConstants>$(DefineConstants);$(ConfiguredConstants)</DefineConstants>
   </PropertyGroup>
 
+  <ItemGroup Condition=" '$(PythonInteropFile)' != '' ">
+    <!-- Remove any files that might conflict with PythonInteropFile -->
+    <Compile Remove="interop*.cs" />
+    <Compile Include="interop.cs" />
+    <!-- Explicitly include PythonInteropFile because it might not be in the src/runtime folder -->
+    <Compile Include="$(PythonInteropFile)" />
+  </ItemGroup>
+
   <ItemGroup>
     <None Remove="resources\clr.py" />
     <EmbeddedResource Include="resources\clr.py">
--- a/src/runtime/Python.Runtime.csproj
+++ b/src/runtime/Python.Runtime.csproj
@@ -19,14 +19,6 @@
     <DefineConstants>$(DefineConstants);$(ConfiguredConstants)</DefineConstants>
   </PropertyGroup>
 
-  <ItemGroup Condition=" '$(PythonInteropFile)' != '' ">
-    <!-- Remove any files that might conflict with PythonInteropFile -->
-    <Compile Remove="interop*.cs" />
-    <Compile Include="interop.cs" />
-    <!-- Explicitly include PythonInteropFile because it might not be in the src/runtime folder -->
-    <Compile Include="$(PythonInteropFile)" />
-  </ItemGroup>
-
   <ItemGroup>
     <None Remove="resources\clr.py" />
     <EmbeddedResource Include="resources\clr.py">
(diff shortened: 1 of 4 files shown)


lostmsu:
I actually think the problem is this line should have been removed: https://github.com/pythonnet/pythonnet/blob/1f40564cbff0f109b6042973527b032be194b7e8/src/runtime/Python.Runtime.csproj#L23

filmor:
@lostmsu Well, that one doesn't do much, I don't see what that would bring. Adding version and ABI flags to the interop classes and comparing against those should be more stable.

@tminka The advice should be to run `python setup.py configure`.

tminka:
@filmor I have changed the advice as directed.

@lostmsu Your suggestion works.  If Python.Runtime.csproj includes all interopXX.cs files, then an incorrect value of PythonInteropFile does not break the build.  This seems like a useful robustness to have, so I added it.

src/runtime/Python.Runtime.csproj
```diff
@@ -19,12 +19,6 @@
     <DefineConstants>$(DefineConstants);$(ConfiguredConstants)</DefineConstants>
   </PropertyGroup>
 
-  <ItemGroup Condition=" '$(PythonInteropFile)' != '' ">
-    <Compile Remove="interop*.cs" />
-    <Compile Include="interop.cs" />
-    <Compile Include="$(PythonInteropFile)" />
```
lostmsu, line 25: I am unsure removing explicit `Include` for `$(PythonInteropFile)` does not break the generation and inclusion `NativeTypeOffset`. Can you verify before we merge?

tminka: How should I verify that?  Isn't this already checked by CI?

lostmsu: In the absence of `NativeTypeOffset` the tests might still pass, but we must be sure it is actually generated on install. One option is to simply add a test, that checks presence of `NativeTypeOffset` (maybe except Windows).

tminka: An even simpler option is to change ABI.Initialize to require NativeTypeOffset (except Windows), if that is the intended behavior.

lostmsu: That is actually unclear. For the time being this just reproduces existing Python.NET behavior (e.g. 2.X) in regards to offsets. But in the upcoming release we plan to make platform-independent wheels, and generating `NativeTypeOffset` should be just an option for peculiar platforms.

tminka: So then what are you recommending for this PR?  A test that fails when `NativeTypeOffset` is missing would have the same problem.

lostmsu: I think it makes sense to have a test that fails when `NativeTypeOffset` is missing. It must be passing in CI (again I am not sure if that's true for Windows). Outside CI we could have it simply ignored. See [GitHub support](https://github.com/actions/runner/issues/368)

tminka:
The test failures imply that NativeTypeOffset is in fact not generated on some non-Windows platforms.  I'm not sure what to do about that.

lostmsu:
It seems to be a temporary fluke, the test failures are identical between this and your other PR, which are unrelated.

lostmsu:
Oh, you meant this failure: https://github.com/pythonnet/pythonnet/pull/1340/commits/f59e5d0e392dbba4e319c5c1a44ca33cb08f43fa

I think this just because you must keep

```xml
  <ItemGroup Condition=" '$(PythonInteropFile)' != '' ">	
    <Compile Include="$(PythonInteropFile)" />	
  </ItemGroup>
```

I believe the file is generated outside `src\runtime` folder.

tminka:
I have restored the lines that you mentioned, and the same tests fail.  So I don't think that was the problem.

lostmsu (approved):

Issue #1325, opened 2020-12-17
NullReferenceException when non-generic type is given type parameters


tminka:
### Environment

-   Pythonnet version: [latest git commit](https://github.com/pythonnet/pythonnet/commit/bd264cf003ca1f9727d7eaeb90ddf913e81fdad2)
-   Python version: 3.7
-   Operating System: Windows 10 64-bit

### Details

The following incorrect Python code produces a confusing NullReferenceException, instead of a more helpful error message:
```python
from System.Collections import IList
IList[bool]
```

-   If there was a crash, please include the traceback here.

```
Unhandled Exception: System.NullReferenceException: Object reference not set to an instance of an object.
   at Python.Runtime.GenericUtil.GenericByName(String ns, String name, Int32 paramCount)
   at Python.Runtime.ClassBase.type_subscript(IntPtr idx)
```

A more helpful message would be: `System.Collections.IList is not a generic type`

PR #1341, opened 2020-12-31
Python tests can now be debugged by running them as embedded tests within NUnit

tminka:

When debugging a problem such as issue #1325, it is immensely useful to have the Visual Studio debugger stop at the location of the error.  This is tricky to do with pytest, even when running pytest from Visual Studio.  Running the test via Python.Exec makes debugging simple and easy.  You don't need to re-install Python.NET every time you make a change.  You don't need to edit project options, and you don't need to figure out how to run pytest from Visual Studio.

With this change, you can just type in the name of the python test you want to debug, build the solution, and debug the new NUnit test that appears.

No

### Any other comments?

No

### Checklist

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -53,9 +53,9 @@ jobs:
         run: dotnet test --runtime any-${{ matrix.platform }} src/embed_tests/
         if: ${{ matrix.os != 'macos' }} # Not working right now, doesn't find libpython
 
-      - name: Python tests runner
+      - name: Python tests run from .NET
         run: dotnet test --runtime any-${{ matrix.platform }} src/python_tests_runner/
-        if: ${{ matrix.os != 'macos' }} # Not working right now, doesn't find libpython
+        if: ${{ matrix.os == 'windows' }} # Not working for others right now
 
       # TODO: Run perf tests
       # TODO: Run mono tests on Windows?
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -49,8 +49,12 @@ jobs:
       - name: Python Tests
         run: pytest
 
-      - name: Run Embedding tests
+      - name: Embedding tests
         run: dotnet test --runtime any-${{ matrix.platform }} src/embed_tests/
+        if: ${{ matrix.os != 'macos' }} # Not working right now, doesn't find libpython
+
+      - name: Python tests runner
+        run: dotnet test --runtime any-${{ matrix.platform }} src/python_tests_runner/
         if: ${{ matrix.os != 'macos' }} # Not working right now, doesn't find libpython
 
       # TODO: Run perf tests
--- a/src/python_tests_runner/PythonTestRunner.cs
+++ b/src/python_tests_runner/PythonTestRunner.cs
@@ -1,0 +1,82 @@
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Reflection;
+using System.Text;
+
+using NUnit.Framework;
+
+using Python.Runtime;
+
+namespace Python.PythonTestsRunner
+{
+    public class PythonTestRunner
+    {
+        [OneTimeSetUp]
+        public void SetUp()
+        {
+            PythonEngine.Initialize();
+        }
+
+        [OneTimeTearDown]
+        public void Dispose()
+        {
+            PythonEngine.Shutdown();
+        }
+
+        /// <summary>
+        /// Selects the Python tests to be run as embedded tests.
+        /// </summary>
+        /// <returns></returns>
+        static IEnumerable<string[]> PythonTestCases()
+        {
+            // Add the test that you want to debug here.
+            yield return new[] { "test_enum", "test_enum_standard_attrs" };
+            yield return new[] { "test_generic", "test_missing_generic_type" };
+        }
+
+        /// <summary>
+        /// Runs a test in src/tests/*.py as an embedded test.  This facilitates debugging.
+        /// </summary>
+        /// <param name="testFile">The file name without extension</param>
+        /// <param name="testName">The name of the test method</param>
+        [TestCaseSource(nameof(PythonTestCases))]
+        public void RunPythonTest(string testFile, string testName)
+        {
+            // Find the tests directory
+            string folder = typeof(PythonTestRunner).Assembly.Location;
+            while (Path.GetFileName(folder) != "src")
+            {
+                folder = Path.GetDirectoryName(folder);
+            }
+            folder = Path.Combine(folder, "tests");
+            string path = Path.Combine(folder, testFile + ".py");
+            if (!File.Exists(path)) throw new FileNotFoundException("Cannot find test file", path);
+
+            // We could use 'import' below, but importlib gives more helpful error messages than 'import'
+            // https://docs.python.org/3/library/importlib.html#importing-a-source-file-directly
+            // Because the Python tests sometimes have relative imports, the module name must be inside the tests package
+            PythonEngine.Exec($@"
+import sys
+import os
+sys.path.append(os.path.dirname(r'{folder}'))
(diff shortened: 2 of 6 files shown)


lostmsu:
These must not be run during CI. As I suggested, this should be a separate project file.

tminka:
Why can't they run during CI?  They do test the embedding system.

lostmsu:
That makes sense. But still I would put them into a separate project file.

appveyor.yml
```diff
@@ -51,6 +51,7 @@ test_script:
   #- ps: .\ci\appveyor_run_tests.ps1
   - pytest
   - dotnet test src/embed_tests/
+  - dotnet test src/embed_python_tests/
```
lostmsu, line 55: Let's skip AppVeyor. We are planning to deprecate it soon.

src/embed_python_tests/packages.config
```diff
@@ -0,0 +1,6 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<packages>
+  <package id="NUnit" version="3.12.0" targetFramework="net40" />
+  <package id="NUnit.ConsoleRunner" version="3.11.1" targetFramework="net40" />
+  <package id="System.ValueTuple" version="4.5.0" targetFramework="net40" />
+</packages>
```
lostmsu, line 6: I don't think this is actually needed.

src/embed_python_tests/Python.EmbeddingPythonTest.csproj
```diff
@@ -0,0 +1,25 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
```
lostmsu, line 3: How about naming the project file simply PythonTests(Runner)?

lostmsu (changes requested):

appveyor.yml
lostmsu: GitHub actions are in .github\workflows\main.yml

tminka:
It appears that `Py.Import("pytest")` is failing on some platforms, for reasons unrelated to this PR.  Any ideas?

.github/workflows/main.yml
```diff
@@ -49,9 +49,13 @@ jobs:
       - name: Python Tests
         run: pytest
 
-      - name: Run Embedding tests
+      - name: Embedding tests
         run: dotnet test --runtime any-${{ matrix.platform }} src/embed_tests/
         if: ${{ matrix.os != 'macos' }} # Not working right now, doesn't find libpython
 
+      - name: Python tests runner
```
lostmsu, line 56: NIT: Python tests run from .NET

src/python_tests_runner/PythonTestRunner.cs
```diff
@@ -0,0 +1,82 @@
...
+        /// <summary>
+        /// Selects the Python tests to be run as embedded tests.
+        /// </summary>
+        /// <returns></returns>
+        static IEnumerable<string[]> PythonTestCases()
+        {
+            // Add the test that you want to debug here.
+            yield return new[] { "test_enum", "test_enum_standard_attrs" };
+            yield return new[] { "test_generic", "test_missing_generic_type" };
+        }
```
lostmsu, line 37: I thought you would discover all tests. Are they not displayed in Test Explorer after a run?

.github/workflows/main.yml
```diff
@@ -49,9 +49,13 @@ jobs:
       - name: Python Tests
         run: pytest
 
-      - name: Run Embedding tests
+      - name: Embedding tests
```
lostmsu, line 52: Happy New Year!

src/python_tests_runner/PythonTestRunner.cs
tminka: There is no need to discover all tests.  That would only slow down the normal testing process.  The lists in PythonTestCases are indeed displayed in Test Explorer after a run.

lostmsu:
@tminka I think the cause might be that the OS image running actions in CI has Python 3.8 installed as system Python, and when embed tests are running, it is being picked up instead of the one in custom environment.

lostmsu (approved):

src/embed_tests/TestPythonEngineProperties.cs
```diff
@@ -183,15 +183,13 @@ public void SetPythonPath()
...
             try
             {
                 Py.Import(moduleName);
-                importShouldSucceed = true;
             }
-            catch
+            catch (PythonException ex)
             {
-                importShouldSucceed = false;
+                throw new Exception($"Py.Import(\"{moduleName}\") failed before setting PythonEngine.PythonPath", ex);
```
lostmsu, line 192: You can't raise an exception here, because that would prevent `PythonEngine.Shutdown();` a few lines below from running, which brings it into inconsistent state.

lostmsu (changes requested):
Sorry, just noticed the issue with skipped `Shutdown`

src/embed_tests/TestPythonEngineProperties.cs
tminka: Technically, I am not raising a new exception, I am just appending to the message of an exception that was already raised.  There are many tests in that file that do not call Shutdown in case an exception is raised, e.g. GetPythonPathDefault, GetProgramNameDefault, GetPythonHomeDefault, SetPythonHomeTwice, SetProgramName.  Are you saying these are all wrong?

tminka: I notice that a few tests in pyinitialize.cs also have this problem.  LoadDefaultArgs uses `using (new PythonEngine())` to handle Initialize/Shutdown.  Perhaps that is how all these other tests should be written?

tminka:
@lostmsu I don't think it is picking up the wrong Python installation because an [earlier commit](https://github.com/pythonnet/pythonnet/pull/1341/checks?check_run_id=1631718489) showed the value of sys.path in the embedded tests: 
```
Python.Runtime.PythonException : ImportError : No module named pytest when sys.path=/opt/hostedtoolcache/Python/3.8.6/x64/lib/python38.zip:/opt/hostedtoolcache/Python/3.8.6/x64/lib/python3.8:/opt/hostedtoolcache/Python/3.8.6/x64/lib/python3.8/lib-dynload:/opt/hostedtoolcache/Python/3.8.6/x64/lib/python3.8/site-packages:/usr/share/dotnet/shared/Microsoft.NETCore.App/3.1.10/:/home/runner/work/pythonnet/pythonnet/src:/home/runner/work/pythonnet/pythonnet/src/tests/fixtures:/home/runner/work/pythonnet/pythonnet/src:/home/runner/work/pythonnet/pythonnet/src/tests/fixtures
```
The "Install dependencies" step installed pytest into `/opt/hostedtoolcache/Python/3.8.6/x64/lib/python3.8/site-packages` which is on that path.

lostmsu:
@tminka since you moved python test running from .NET to a separate project and CI step, you no longer need the changes to `TestPythonEngineProperties.cs` for the purposes this PR is trying to achieve. Moving it to a separate PR should let us merge the runner.

tminka:
I agree, but I think the change to importhook is important enough (and harmless enough) to keep in this PR.

src/runtime/importhook.cs
```diff
@@ -342,7 +345,7 @@ public static IntPtr __import__(IntPtr self, IntPtr args, IntPtr kw)
                 ManagedType mt = tail.GetAttribute(name, true);
                 if (!(mt is ModuleObject))
                 {
-                    Exceptions.SetError(Exceptions.ImportError, $"No module named {name}");
+                    Exceptions.SetError(Exceptions.ImportError, originalExceptionMessage);
```
lostmsu, line 348: Use `PythonException.Restore()`

lostmsu (changes requested):

lostmsu (approved):

Issue #1342, opened 2020-12-31
PyFloats should not be implicitly rounded to integers


tminka:
### Environment

-   Pythonnet version: [Latest master](https://github.com/pythonnet/pythonnet/commit/1f40564cbff0f109b6042973527b032be194b7e8)
-   Python version: 3.7.8
-   Operating System: Windows 10 64-bit
-   .NET Runtime: Core 3.1

### Details

Incorrectly passing a float value to a method accepting an integer argument causes the float value to be rounded.  

```python
from Python.Test import MethodTest
MethodTest.TestOverloadedNoObject(5.5)
```

Surprisingly, the above code invokes the method with the integer 5.  I expect this to raise an error instead, even if the value passed was `5.0`.  I expect a method with an integer argument to behave like list indexing in Python, which refuses any float value:

```python
a = [1,2,3]
a[1.0]
TypeError: list indices must be integers or slices, not float
```

tminka:
There is a similar problem with array indexing.  The below test cases (extending `test_array.py`) fail by not raising an error:
```python
def test_typed_array():
    ...
    with pytest.raises(TypeError):
        ob = Test.TypedArrayTest()
        _ = ob.items[0.5]

    with pytest.raises(TypeError):
        ob = Test.TypedArrayTest()
        ob.items[0.5] = Spam("0")

def test_multi_dimensional_array():
    ...
    with pytest.raises(TypeError):
        ob = Test.MultiDimensionalArrayTest()
        _ = ob.items[0.5, 0]

    with pytest.raises(TypeError):
        ob = Test.MultiDimensionalArrayTest()
        ob.items[0.5, 0] = 0
```

PR #1343, opened 2020-12-31
Disable implicit conversion from PyFloat to .NET integer types

lostmsu:

Previously, when a floating point value was passed to .NET parameter of integer type, it would be silently truncated. After this change it would no longer be permitted to pass a floating point value to integer parameter.

#1342

### Any other comments?

I removed poorly-behaved `PyLong_As*` functions, as they would always implicitly convert argument to integer type, which in most cases is undesired behavior. Instead, to avoid roundtrips, `PyLong_AsSize_t` is used in most cases. The only exception is `Int64` which on 32 bit platforms is larger than `size_t`, so `PyLong_AsLongLong` function is used after explicit check for `PyLong`.

Draft because it depends on C# 9, which is not yet ready.

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,8 @@ details about the cause of the failure
      to the regular method return value (unless they are passed with `ref` or `out` keyword).
 -   BREAKING: Drop support for the long-deprecated CLR.* prefix.
 -   `PyObject` now implements `IEnumerable<PyObject>` in addition to `IEnumerable`
+-   floating point values passed from Python are no longer silently truncated
+when .NET expects an integer [#1342][i1342]
 
 ### Fixed
 
@@ -807,3 +809,4 @@ This version improves performance on benchmarks significantly compared to 2.3.
 [i755]: https://github.com/pythonnet/pythonnet/pull/755
 [p534]: https://github.com/pythonnet/pythonnet/pull/534
 [i449]: https://github.com/pythonnet/pythonnet/issues/449
+[i1342]: https://github.com/pythonnet/pythonnet/issues/1342
--- a/src/runtime/Python.Runtime.csproj
+++ b/src/runtime/Python.Runtime.csproj
@@ -4,6 +4,7 @@
     <Platforms>AnyCPU</Platforms>
     <RootNamespace>Python.Runtime</RootNamespace>
     <AssemblyName>Python.Runtime</AssemblyName>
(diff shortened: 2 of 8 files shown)


tminka (commented):
test_uint64_conversion should similarly raise a TypeError [here](https://github.com/pythonnet/pythonnet/blob/efe84c2bef3771fa82f946e760b6100201982de1/src/tests/test_conversion.py#L385)

christabella:
In what sense is C# 9 not ready? It would be great if this fix was merged, since that would unblock other issues

lostmsu:
@christabella @tminka if we bump version now, users of many older Mono builds will be unable to install pythonnet `master` due to lack of C# 9 support. This will no longer be a problem after merging https://github.com/losttech/pythonnet/commits/features/VersionIndependent , which will allow building single `Python.Runtime.dll` for all Python versions (so they don't have to build it).

The problem is this branch relies on a new C# feature - unmanaged function pointers, which is [broken in the latest .NET Core SDK C# compiler](https://github.com/dotnet/roslyn/issues/49760). The fix is up in master, but has not been released yet.

filmor:
@lostmsu The only C# 9 feature you are using here is `nint`, isn't it? As far as I understood this, the feature doesn't require help from the CLR, so it should work with older Mono versions as well. @BadSingleton @benoithudson Could you verify that this branch still works with the Mono version you are using?

lostmsu:
I expect older Mono versions to have old C# compiler, which simply won't recognize `nint`.

benoithudson:
I don't know what this is about exactly. For Unity, if we can target NET Standard 2.0 we're good. We don't use the Unity mcs to compile anyway, we ship this code as DLLs.

Helpfully, Microsoft now tells you which version of Unity supports which standard.
https://docs.microsoft.com/en-us/dotnet/standard/net-standard

filmor:
We don't compile this project with the Mono C# compiler either, we expect a .NET Core SDK to be installed nowadays. This just bumps the required SDK version to 5.0, I think.

filmor:
@lostmsu Since this passes tests, I'd vote we merge it but keep the bumped requirement in mind.

lostmsu:
@filmor heads up, MS just released VS 16.9 Preview 3 along with a new .NET SDK Preview, and the blocking issue I mentioned above appears to be fixed there.

PR #1347, opened 2021-01-06
Support comparison operators

christabella:

Continuation of https://github.com/pythonnet/pythonnet/pull/1324/files but specifically for comparison operators e.g. >, >=, <, == etc. based on a discussion with @tminka --- in a nutshell, we want to check each C# class for any comparison `operator` methods when calling `ClassBase.tp_richcompare`, before proceeding with the usual logic (see https://github.com/pythonnet/pythonnet/pull/294) which handles C# classes that implement an `IComparable` interface.

Would appreciate a review from @lostmsu who helped with the previous PR that this builds on. Thanks!

Closes https://github.com/pythonnet/pythonnet/issues/1312
More concrete examples can also be found in other Infer.NET tutorials e.g. having to use `op_GreaterThan` https://github.com/dotnet/infer/blob/67b4f80d97018460bcb817f76ec874d0f33f1651/test/TestPython/test_tutorials.py#L31
### Any other comments?

Some remaining tasks could be:
- Add tests for derived objects, and for other edge cases (e.g. if truly no operator methods defined, asserting good error messages)
- When C# class defines `operator ==()`, the class should also define .Equals and .GetHashCode() https://stackoverflow.com/questions/10790370/whats-wrong-with-defining-operator-but-not-defining-equals-or-gethashcode

### Checklist

The tests pass in net472 but not on netcoreapp3.1

### DIFF
--- a/src/embed_tests/TestOperator.cs
+++ b/src/embed_tests/TestOperator.cs
@@ -149,6 +149,58 @@ namespace Python.EmbeddingTest
                 return new OperableObject(a.Num ^ b);
             }
 
+            public static bool operator <=(int a, OperableObject b)
+            {
+                return (a <= b.Num);
+            }
+            public static bool operator <=(OperableObject a, OperableObject b)
+            {
+                return (a.Num <= b.Num);
+            }
+            public static bool operator <=(OperableObject a, int b)
+            {
+                return (a.Num <= b);
+            }
+
+            public static bool operator >=(int a, OperableObject b)
+            {
+                return (a >= b.Num);
+            }
+            public static bool operator >=(OperableObject a, OperableObject b)
+            {
+                return (a.Num >= b.Num);
+            }
+            public static bool operator >=(OperableObject a, int b)
+            {
+                return (a.Num >= b);
+            }
+
+            public static bool operator<(int a, OperableObject b)
+            {
+                return (a < b.Num);
+            }
+            public static bool operator<(OperableObject a, OperableObject b)
+            {
+                return (a.Num < b.Num);
+            }
+            public static bool operator<(OperableObject a, int b)
+            {
+                return (a.Num < b);
+            }
+
+            public static bool operator>(int a, OperableObject b)
+            {
+                return (a > b.Num);
+            }
+            public static bool operator>(OperableObject a, OperableObject b)
+            {
+                return (a.Num > b.Num);
+            }
+            public static bool operator >(OperableObject a, int b)
+            {
+                return (a.Num > b);
+            }
+
             public static OperableObject operator <<(OperableObject a, int offset)
             {
                 return new OperableObject(a.Num << offset);
@@ -161,7 +213,7 @@ namespace Python.EmbeddingTest
         }
 
         [Test]
-        public void OperatorOverloads()
+        public void SymmetricalOperatorOverloads()
         {
             string name = string.Format("{0}.{1}",
                 typeof(OperableObject).DeclaringType.Name,
@@ -206,6 +258,18 @@ assert c.Num == a.Num | b.Num
 
 c = a ^ b
 assert c.Num == a.Num ^ b.Num
+
+c = a <= b
+assert c == (a.Num <= b.Num)
+
+c = a >= b
+assert c == (a.Num >= b.Num)
+
+c = a < b
+assert c == (a.Num < b.Num)
+
+c = a > b
+assert c == (a.Num > b.Num)
 ");
         }
 
@@ -263,6 +327,18 @@ assert c.Num == a.Num | b
 
 c = a ^ b
 assert c.Num == a.Num ^ b
+
+c = a <= b
+assert c == (a.Num <= b)
+
+c = a >= b
+assert c == (a.Num >= b)
+
+c = a < b
+assert c == (a.Num < b)
+
+c = a > b
+assert c == (a.Num > b)
 ");
         }
 
@@ -304,6 +380,18 @@ assert c.Num == a | b.Num
 
 c = a ^ b
 assert c.Num == a ^ b.Num
+
+c = a <= b
+assert c == (a <= b.Num)
+
+c = a >= b
+assert c == (a >= b.Num)
+
+c = a < b
+assert c == (a < b.Num)
+
+c = a > b
+assert c == (a > b.Num)
 ");
 
         }
--- a/src/embed_tests/TestOperator.cs
+++ b/src/embed_tests/TestOperator.cs
@@ -212,6 +212,25 @@ namespace Python.EmbeddingTest
                 return (a.Num >= b);
             }
 
+            public static bool operator >=(OperableObject a, PyObject b)
+            {
+                using (Py.GIL())
+                {
+                    // Assuming b is a tuple, take the first element.
+                    int bNum = (int)b.ToArray()[0].AsManagedObject(typeof(int));
+                    return a.Num >= bNum;
+                }
+            }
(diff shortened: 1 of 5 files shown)


src/runtime/operatormethod.cs
```diff
@@ -49,6 +49,13 @@ static OperatorMethod()
                 ["op_OnesComplement"] = new SlotDefinition("__invert__", TypeOffset.nb_invert),
                 ["op_UnaryNegation"] = new SlotDefinition("__neg__", TypeOffset.nb_negative),
                 ["op_UnaryPlus"] = new SlotDefinition("__pos__", TypeOffset.nb_positive),
+                ["op_OneComplement"] = new SlotDefinition("__invert__", TypeOffset.nb_invert),
+                ["op_Equality"] = new SlotDefinition("__eq__", TypeOffset.tp_richcompare),
+                ["op_Inequality"] = new SlotDefinition("__ne__", TypeOffset.tp_richcompare),
+                ["op_LessThanOrEqual"] = new SlotDefinition("__le__", TypeOffset.tp_richcompare),
+                ["op_GreaterThanOrEqual"] = new SlotDefinition("__ge__", TypeOffset.tp_richcompare),
+                ["op_LessThan"] = new SlotDefinition("__lt__", TypeOffset.tp_richcompare),
+                ["op_GreaterThan"] = new SlotDefinition("__gt__", TypeOffset.tp_richcompare),
```
lostmsu, line 58: I think this is error-prone. `op_LessThan` does not really implement `tp_richcompare`. It would be better to have a separate `ComparisonOpMap`, `string -> string`, and fix `IsOperatorMethod` correspondingly.

It would also remove the need in the change in `FixupSlots` below.

src/runtime/operatormethod.cs
```diff
@@ -86,7 +99,7 @@ public static void FixupSlots(IntPtr pyType, Type clrType)
             Debug.Assert(_opType != null);
             foreach (var method in clrType.GetMethods(flags))
             {
-                if (!IsOperatorMethod(method))
+                if (!IsOperatorMethod(method) || IsComparisonOp(method))  // We don't want to override ClassBase.tp_richcompare.
```
lostmsu, line 102: Direct comment to the reason why we don't want to override, otherwise it does not explain anything and simply says what code in this method does (which is generally not very useful).

Something like "comparison operators are handled by `ClassBase.tp_richcompare`"

lostmsu (changes requested):

src/embed_tests/TestOperator.cs
```diff
@@ -25,6 +25,17 @@ public class OperableObject
         {
             public int Num { get; set; }
 
+            public override int GetHashCode()
+            {
+                return 159832395 + Num.GetHashCode();
```
tminka, line 30: `unchecked(159832395 + Num.GetHashCode())`

src/runtime/classbase.cs
```diff
@@ -72,6 +82,42 @@ public static IntPtr tp_richcompare(IntPtr ob, IntPtr other, int op)
...
+            // C# operator methods take precedence over IComparable.
+            // We first check if there's a comparison operator by looking up the richcompare table,
+            // otherwise fallback to checking if an IComparable interface is handled.
+            if (PyToCilOpMap.ContainsKey(op)) {
+                string CilOp = PyToCilOpMap[op];
+                if (cls.richcompare.Contains(CilOp)) {
+                    var methodObject = (MethodObject)cls.richcompare[CilOp];
+                    IntPtr args = other;
+                    var free = false;
+                    if (!Runtime.PyTuple_Check(other))
```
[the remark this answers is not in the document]
christabella: Indeed if `b` were a tuple this `if` would be skipped. Once removing the `if`, the operator method does receive  `b`  as `PyObject`,  but I get an `AccessViolationException : Attempted to read or write protected memory.` so I think `b` is wrongly parsed?

image

src/runtime/operatormethod.cs
christabella: Makes sense, thanks! Should that `ComparisonOpMap` also be merged with the other map for `op_LessThan` -> `Py_LT`?

christabella: I added this `OperatorMethod.ComparisonOpMap` in the latest commit, without merging it with the `ClassBase.CilToPyOpMap`, although they both have the same keys so it would make sense to put them together in `OperatorMethod`.

src/runtime/classbase.cs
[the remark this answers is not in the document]
christabella: I think that error message might have been a flaky bug, because now the variable states are the same but the error is:
```
Message: 
    Python.Runtime.PythonException : TypeError : '>=' not supported between instances of 'int' and 'tuple'
```

Still it means the `PyObject` was not recognized as a tuple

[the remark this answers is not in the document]
lostmsu: This is because to work with `PyObject` instances in C# callbacks you must hold GIL. E.g. the method body must be inside `using (Py.GIL()) { ...` block.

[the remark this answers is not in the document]
christabella: Thanks, I didn't know about that. https://github.com/pythonnet/pythonnet/blob/d6c0081f87b67707b59598efe877378b0c36d822/src/runtime/methodbinder.cs#L847 

image


I also put the `operator >(Obj a, PyObject b)` body in a `Py.GIL` block but that doesn't do anything.

src/embed_tests/TestOperator.cs
```diff
@@ -149,6 +160,103 @@ public OperableObject(int num)
...
+            public static bool operator >=(OperableObject a, int b)
+            {
+                return (a.Num >= b);
+            }
+
+            public static bool operator >=(OperableObject a, PyObject b)
+            {
+                using (Py.GIL())
+                {
+                    // Assuming b is a tuple, take the first element.
```
christabella, line 219: @lostmsu is this the right way to implement the rest of the comparison operators on tuples, as a test?

src/runtime/classbase.cs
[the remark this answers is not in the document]
lostmsu: I don't think you need to add a test with a tuple for each comparison operator. One is enough.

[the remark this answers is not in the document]
lostmsu: The failure on the screenshot is only happening because you have debugger attached. It tries to call `ToString` on some `PyObject` to show in watches, but it does not work without `Py.GIL`

src/embed_tests/TestOperator.cs
lostmsu: I think you should be able to do `b[0].As<int>()`

src/runtime/operatormethod.cs
```diff
@@ -86,7 +104,9 @@ public static void FixupSlots(IntPtr pyType, Type clrType)
             Debug.Assert(_opType != null);
             foreach (var method in clrType.GetMethods(flags))
             {
-                if (!IsOperatorMethod(method))
+                // We don't want to override slots for either non-operators or
+                // comparison operators, which are handled by ClassBase.tp_richcompare.
+                if (!IsOperatorMethod(method) || IsComparisonOp(method))
```
lostmsu, line 109: NIT:  `!OpMethodMap .ContainsKey(method)`

lostmsu (changes requested):

src/runtime/operatormethod.cs
christabella: Of course, thanks!

lostmsu (approved):

PR #1360, opened 2021-01-21
Fix exception string

### LINKED ISSUE #1359: Ugly traceback in PythonException
### Environment

-   Pythonnet version: 2.5.0 and earlier
-   Python version: 3.7 and 2.7
-   Operating System: any
-   .NET Runtime: any

### Details

Within Unity, when a PythonException is thrown, we usually print the exception. It comes out looking like:
```
PythonException: TypeError : cannot create weak reference to 'PySideTestUI' object
['  File "<string>", line 3, in <module>/n', '  File "/Users/bhudson/projects/com.unity.scripting.python/com.unity.scripting.python/Samples/PySideExample/PySideExample.py", line 168, in create_or_reinitialize/n    _PYSIDE_UI = PySideTestUI()/n', '  File "/Users/bhudson/projects/com.unity.scripting.python/com.unity.scripting.python/Samples/PySideExample/PySideExample.py", line 54, in __init__/n    self._dialog = self.load_ui_widget(ui_path.replace("////", "/"))/n', '  File "/Users/bhudson/projects/com.unity.scripting.python/com.unity.scripting.python/Samples/PySideExample/PySideExample.py", line 130, in load_ui_widget/n    ui.useCameraButton.clicked.connect(self.use_camera)/n']
```
All on one line. It's very hard to read.

The pythonnet issue is that we have a list of strings that got str() called on it rather than joining them. Then a separate bug is that something is flipping backslashes to slashes but I don't think that's pythonnet's responsibility.

The reason is here: https://github.com/pythonnet/pythonnet/blob/master/src/runtime/pythonexception.cs#L53

filmor:

Fixes #1359 by joining the traceback lines and separates the .NET stacktrace by an additional line.

#1359 

### Any other comments?

...

### Checklist

### DIFF
--- a/src/runtime/pythonexception.cs
+++ b/src/runtime/pythonexception.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Linq;
 using System.Runtime.CompilerServices;
 using System.Text;
 
@@ -44,6 +45,7 @@ namespace Python.Runtime
                 }
                 _message = type + " : " + message;
             }
+
             if (_pyTB != IntPtr.Zero)
             {
                 using PyObject tb_module = PythonEngine.ImportModule("traceback");
@@ -54,7 +56,9 @@ namespace Python.Runtime
                 using var tbList = tb_module.InvokeMethod("format_tb", pyTB);
 
                 var sb = new StringBuilder();
-                foreach (var line in tbList) {
+                // Reverse Python's traceback list to match the order used in C#
+                // stacktraces
+                foreach (var line in tbList.Reverse()) {
                     sb.Append(line.ToString());
                 }
                 _tb = sb.ToString();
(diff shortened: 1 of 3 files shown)


src/runtime/pythonexception.cs
```diff
@@ -45,14 +45,18 @@ public PythonException()
...
+                Runtime.XIncref(_pyTB);
+                using var pyTB = new PyObject(_pyTB);
+
+                using var tbList = tb_module.InvokeMethod("format_tb", pyTB);
+
+                var sb = new StringBuilder();
+                foreach (var line in tbList) {
+                    sb.Append(line.ToString());
                 }
+                _tb = sb.ToString();
```
lostmsu, line 59: I vaguely remember you need to reverse the python stack trace lines int order for them to match the order in `base.StackTrace`. Can you please doublecheck.

filmor: You are right. Python tracebacks are sorted from outer to inner functions, C# are sorted from inner to outer. Since we retain the respective formatting, I wonder whether it's really a good idea to turn own of them (likely the Python one) around as that may mess with expectations of the debugging user...

lostmsu: I did so in my fork (e.g. reversed `tbList`), because it was really confusing during debugging.

benoithudson:
That was quick, thanks!

PR #1361, opened 2021-01-21
Better error messages for method argument mismatch and others

tminka:

This change produces more specific error messages by passing setError=true to Converter.ToManaged whenever possible, such as when there is only one method in methodbinder.  To support this:
- Converter.ToManaged is fixed to obey setError consistently
- Converter.ToArray catches exceptions
- RaiseTypeError appends any existing error message
- Fixed missing information on 'No method matches given arguments' by adding the method name.

No, but it provides more information to the user when they come across issues like #1099 .  Instead of "No method matches" or "No match found" it now says `TypeError : 'int' value cannot be converted to System.Enum`

### Any other comments?

I added some failing test cases to test_array.py that should be addressed by a future PR.

### Checklist

### DIFF
--- a/src/runtime/converter.cs
+++ b/src/runtime/converter.cs
@@ -303,6 +303,11 @@ namespace Python.Runtime
         /// Return a managed object for the given Python object, taking funny
         /// byref types into account.
         /// </summary>
+        /// <param name="value">A Python object</param>
+        /// <param name="type">The desired managed type</param>
+        /// <param name="result">Receives the managed object</param>
+        /// <param name="setError">If true, call <c>Exceptions.SetError</c> with the reason for failure.</param>
+        /// <returns>True on success</returns>
         internal static bool ToManaged(IntPtr value, Type type,
             out object result, bool setError)
         {
@@ -341,7 +346,10 @@ namespace Python.Runtime
                         result = tmp;
                         return true;
                     }
-                    Exceptions.SetError(Exceptions.TypeError, $"value cannot be converted to {obType}");
+                    if (setError)
+                    {
+                        Exceptions.SetError(Exceptions.TypeError, $"value cannot be converted to {obType}");
+                    }
                     return false;
                 }
                 if (mt is ClassBase)
@@ -374,6 +382,15 @@ namespace Python.Runtime
                 }
                 // Set type to underlying type
                 obType = obType.GetGenericArguments()[0];
+            }
+
+            if (obType.ContainsGenericParameters)
+            {
+                if (setError)
+                {
+                    Exceptions.SetError(Exceptions.TypeError, $"Cannot create an instance of the open generic type {obType}");
+                }
+                return false;
             }
 
             if (obType.IsArray)
@@ -777,7 +794,7 @@ namespace Python.Runtime
             IntPtr ob = Runtime.PyObject_Repr(value);
             string src = Runtime.GetManagedString(ob);
             Runtime.XDecref(ob);
-            Exceptions.SetError(Exceptions.TypeError, $"Cannot convert {src} to {target}");
+            Exceptions.RaiseTypeError($"Cannot convert {src} to {target}");
         }
 
 
@@ -791,32 +808,58 @@ namespace Python.Runtime
             Type elementType = obType.GetElementType();
             result = null;
 
-            bool IsSeqObj = Runtime.PySequence_Check(value);
-            var len = IsSeqObj ? Runtime.PySequence_Size(value) : -1;
-
             IntPtr IterObject = Runtime.PyObject_GetIter(value);
-
-            if(IterObject==IntPtr.Zero) {
+            if (IterObject == IntPtr.Zero)
+            {
                 if (setError)
                 {
+                    SetConversionError(value, obType);
+                }
+                else
+                {
+                    // PyObject_GetIter will have set an error
+                    Exceptions.Clear();
+                }
+                return false;
+            }
+
+            IList list;
+            try
+            {
+                // MakeGenericType can throw because elementType may not be a valid generic argument even though elementType[] is a valid array type.
+                // For example, if elementType is a pointer type.
+                // See https://docs.microsoft.com/en-us/dotnet/api/system.type.makegenerictype#System_Type_MakeGenericType_System_Type
+                var constructedListType = typeof(List<>).MakeGenericType(elementType);
+                bool IsSeqObj = Runtime.PySequence_Check(value);
+                if (IsSeqObj)
+                {
+                    var len = Runtime.PySequence_Size(value);
+                    list = (IList)Activator.CreateInstance(constructedListType, new Object[] { (int)len });
+                }
+                else
+                {
+                    // CreateInstance can throw even if MakeGenericType succeeded.
+                    // See https://docs.microsoft.com/en-us/dotnet/api/system.activator.createinstance#System_Activator_CreateInstance_System_Type_
+                    list = (IList)Activator.CreateInstance(constructedListType);
+                }
+            }
+            catch (Exception e)
+            {
+                if (setError)
+                {
+                    Exceptions.SetError(e);
                     SetConversionError(value, obType);
                 }
                 return false;
             }
 
-            Array items;
-
-            var listType = typeof(List<>);
-            var constructedListType = listType.MakeGenericType(elementType);
-            IList list = IsSeqObj ? (IList) Activator.CreateInstance(constructedListType, new Object[] {(int) len}) :
-                                        (IList) Activator.CreateInstance(constructedListType);
             IntPtr item;
 
             while ((item = Runtime.PyIter_Next(IterObject)) != IntPtr.Zero)
             {
-                object obj = null;
+                object obj;
 
-                if (!Converter.ToManaged(item, elementType, out obj, true))
+                if (!Converter.ToManaged(item, elementType, out obj, setError))
                 {
                     Runtime.XDecref(item);
                     return false;
@@ -827,7 +870,7 @@ namespace Python.Runtime
             }
             Runtime.XDecref(IterObject);
 
-            items = Array.CreateInstance(elementType, list.Count);
+            Array items = Array.CreateInstance(elementType, list.Count);
             list.CopyTo(items, 0);
 
             result = items;
--- a/src/runtime/eventobject.cs
+++ b/src/runtime/eventobject.cs
@@ -86,7 +86,7 @@ namespace Python.Runtime
             }
 
             nint hash = Runtime.PyObject_Hash(handler);
-            if (hash == -1 || Exceptions.ErrorOccurred())
+            if (hash == -1 && Exceptions.ErrorOccurred())
             {
                 return false;
             }
--- a/src/runtime/eventobject.cs
+++ b/src/runtime/eventobject.cs
@@ -72,6 +72,12 @@ namespace Python.Runtime
         /// </summary>
         internal bool RemoveEventHandler(IntPtr target, IntPtr handler)
         {
+            if (reg == null)
+            {
+                Exceptions.SetError(Exceptions.ValueError, "unknown event handler");
+                return false;
+            }
+
             object obj = null;
             if (target != IntPtr.Zero)
             {
@@ -79,10 +85,9 @@ namespace Python.Runtime
                 obj = co.inst;
             }
 
-            IntPtr hash = Runtime.PyObject_Hash(handler);
-            if (Exceptions.ErrorOccurred() || reg == null)
+            nint hash = Runtime.PyObject_Hash(handler);
+            if (hash == -1 || Exceptions.ErrorOccurred())
             {
-                Exceptions.SetError(Exceptions.ValueError, "unknown event handler");
                 return false;
             }
 
--- a/src/runtime/exceptions.cs
+++ b/src/runtime/exceptions.cs
@@ -72,31 +72,18 @@ namespace Python.Runtime
                 return Exceptions.RaiseTypeError("invalid object");
             }
 
-            string message = string.Empty;
-            if (e.Message != string.Empty)
+            string message = e.ToString();
+            string fullTypeName = e.GetType().FullName;
+            string prefix = fullTypeName + ": ";
+            if (message.StartsWith(prefix))
             {
-                message = e.Message;
+                message = message.Substring(prefix.Length);
             }
-            if (e is AggregateException ae)
+            else if (message.StartsWith(fullTypeName))
             {
-                message += GetAggregateExceptionString(ae);
-            }
-            if (!string.IsNullOrEmpty(e.StackTrace))
-            {
-                message = message + "\n" + e.StackTrace;
+                message = message.Substring(fullTypeName.Length);
             }
             return Runtime.PyUnicode_FromString(message);
-        }
-
-        private static string GetAggregateExceptionString(AggregateException aggregateException)
-        {
-            StringBuilder stringBuilder = new StringBuilder();
-            foreach(var innerException in aggregateException.InnerExceptions)
-            {
-                stringBuilder.AppendLine();
-                stringBuilder.Append(innerException.Message);
-            }
-            return stringBuilder.ToString();
         }
     }
 
--- a/src/runtime/exceptions.cs
+++ b/src/runtime/exceptions.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Reflection;
 using System.Runtime.InteropServices;
+using System.Text;
 
 namespace Python.Runtime
 {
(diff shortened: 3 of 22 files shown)


src/runtime/converter.cs
```diff
@@ -800,32 +808,52 @@ private static bool ToArray(IntPtr value, Type obType, out object result, bool s
...
+                    var len = Runtime.PySequence_Size(value);
+                    list = (IList)Activator.CreateInstance(constructedListType, new Object[] { (int)len });
+                }
+                else
+                {
+                    list = (IList)Activator.CreateInstance(constructedListType);
+                }
+            }
+            catch
+            {
```
lostmsu, line 845: Why would this fail? We should only catch specific exceptions.

Also, the exception information should be included into the resulting error unless we fully process it.

src/runtime/eventobject.cs
```diff
@@ -80,7 +80,11 @@ internal bool RemoveEventHandler(IntPtr target, IntPtr handler)
             }
 
             IntPtr hash = Runtime.PyObject_Hash(handler);
-            if (Exceptions.ErrorOccurred() || reg == null)
+            if (Exceptions.ErrorOccurred())
```
lostmsu, line 83: Would make sense to check if `hash == -1`

src/runtime/eventobject.cs
```diff
@@ -80,7 +80,11 @@ internal bool RemoveEventHandler(IntPtr target, IntPtr handler)
...
+            {
+                return false;
+            }
+            if (reg == null)
             {
                 Exceptions.SetError(Exceptions.ValueError, "unknown event handler");
                 return false;
```
lostmsu, line 90: I'd move this up, as this check is independent from Python

src/runtime/exceptions.cs
```diff
@@ -370,6 +370,10 @@ public static void deprecation(string message)
 
         internal static IntPtr RaiseTypeError(string message)
         {
+            if (Exceptions.ErrorOccurred())
+            {
+                message += Environment.NewLine + "  ----> " + new PythonException().Message;
+            }
```
lostmsu, line 376: I think this should only be done explicitly. It is theoretically possible Python calls our code with the error flag already set, then it is not really a direct cause for the TypeError you are about to report.

Additionally, it is better to keep message fixed, and set `__cause__` (which, admittedly, I have not mastered yet).

src/runtime/methodbinder.cs
```diff
@@ -378,6 +378,7 @@ internal Binding Bind(IntPtr inst, IntPtr args, IntPtr kw, MethodBase info, Meth
                 var outs = 0;
                 var margs = TryConvertArguments(pi, paramsArray, args, pynargs, kwargDict, defaultArgList,
                     needsResolution: _methods.Length > 1,  // If there's more than one possible match.
+                    setError: _methods.Length == 1 && !isGeneric,   // True when there are no alternatives.
```
lostmsu, line 381: This is problematic in the following scenario:
```python
try:
  UInt64Method(some_val)
except OverflowError:
  print('value is too large')
```

When somebody adds `UInt64Method(string)`, the code above will stop working despite the new overload not getting matched.

lostmsu, line 381: Since you set an error here, it must be cleared unless propagated (e.g. when next overload is tried).

lostmsu (changes requested):

src/runtime/converter.cs
tminka: See [Activator.CreateInstance](https://docs.microsoft.com/en-us/dotnet/api/system.activator.createinstance) for a list of the cases where it can throw.  test_generic_list_array_conversion tests one such case.  I will change the code to include the error.

lostmsu: Most of them are not applicable, since the type is known to be `List<T>`

src/runtime/methodbinder.cs
lostmsu: An option is to set `__cause__` of the "unable to find overload" `TypeError`  to an instance of `AggregateException`

src/runtime/eventobject.cs
tminka: An IntPtr cannot be -1.  Perhaps IntPtr is not the correct return type for PyObject_Hash?

lostmsu: You can still compare it to `new IntPtr(-1)`.
Ideally we should have C types mirrored for clarity though.

src/runtime/exceptions.cs
tminka: Is that really a problem though?  Extra information is always good for tracking down a bug.

src/runtime/methodbinder.cs
tminka: Yes, there will be an inconsistency in error messages depending on whether a method is overloaded.  I think this is the best that can be done in the current bad situation wrt overloading.

tminka: The condition here is meant to exclude all cases where the error might need to be cleared, such as another overload.

lostmsu: The situation is mostly bad due to unexpected default conversions, which will be gone in 3.0.0. We should not introduce an inconsistent behavior to patch it temporarily, when it is possible to improve it without inconsistencies.

tminka: I think the solution you want is to preserve "No method matches" as the top-level error, but include the OverflowError as an inner error.  I will change that.

src/runtime/eventobject.cs
tminka: Is `nint` the correct type?

src/runtime/methodbinder.cs
lostmsu: Right, I could have stated it clearly.

Consider also collecting errors from all overloads.

src/runtime/eventobject.cs
lostmsu: It is definitely a better replacement in this case (hooray C# 9!). Judging by https://docs.microsoft.com/en-us/cpp/build/common-visual-cpp-64-bit-migration-issues?redirectedfrom=MSDN&view=msvc-160 , it should be of the right size.

src/runtime/converter.cs
tminka: Yes but some still are applicable.  In fact, MakeGenericType can also fail in some cases, so it should also be included in the try/catch.

tminka: Actually, since `elementType` is the element type of an array, MakeGenericType cannot fail.  So the only possible exception seems to be ArgumentException.

tminka: I have added the new tests test_method_with_pointer_array_argument, test_invalid_generic_type_parameter, test_invalid_generic_method_type_parameter to emphasize the importance of try/catch in these cases.

src/runtime/eventobject.cs
tminka: I have changed PyObject_Hash to use nint.

src/runtime/exceptions.cs
tminka: I have changed it to use __cause__.  If you don't want the existing error to be attached as the cause, then use SetError instead of RaiseTypeError.

src/runtime/eventobject.cs
```diff
@@ -72,17 +72,22 @@ internal bool AddEventHandler(IntPtr target, IntPtr handler)
...
             if (target != IntPtr.Zero)
             {
                 var co = (CLRObject)GetManagedObject(target);
                 obj = co.inst;
             }
 
-            IntPtr hash = Runtime.PyObject_Hash(handler);
-            if (Exceptions.ErrorOccurred() || reg == null)
+            nint hash = Runtime.PyObject_Hash(handler);
+            if (hash == -1 || Exceptions.ErrorOccurred())
```
lostmsu, line 89: It must be `&&`, not `||`. `PyObject_Hash` on error returns `-1` (but also when no error and hash is -1), so if `hash` is not `-1`, there's no error.

src/runtime/methodbinder.cs
```diff
@@ -102,7 +102,18 @@ internal static MethodInfo MatchParameters(MethodInfo[] mi, Type[] tp)
...
+                {
+                    // MakeGenericMethod can throw ArgumentException if the type parameters do not obey the constraints.
+                    MethodInfo method = t.MakeGenericMethod(tp);
+                    Exceptions.Clear();
+                    return method;
+                }
+                catch (ArgumentException e)
+                {
+                    Exceptions.SetError(e);
+                    // The error will remain set until cleared by a successful match.
```
lostmsu, line 115: This needs to mentioned in the XmlDoc of the function.

src/runtime/methodbinder.cs
lostmsu: Since you can set the `__cause__`, I see no reason to ever pass `setError: false`. If there were > 1 overload, and one matched, you can just clear the error.

src/runtime/methodbinder.cs
```diff
@@ -600,10 +614,11 @@ static IntPtr HandleParamsArray(IntPtr args, int arrayStart, int pyArgCount, out
         /// <param name="op">Pointer to the object at a particular parameter.</param>
         /// <param name="parameterType">That parameter's managed type.</param>
         /// <param name="needsResolution">There are multiple overloading methods that need resolution.</param>
+        /// <param name="setError">If true, call <c>Exceptions.SetError</c> with the reason for failure.</param>
         /// <param name="arg">Converted argument.</param>
         /// <param name="isOut">Whether the CLR type is passed by reference.</param>
         /// <returns></returns>
-        static bool TryConvertArgument(IntPtr op, Type parameterType, bool needsResolution,
+        static bool TryConvertArgument(IntPtr op, Type parameterType, bool needsResolution, bool setError,
```
lostmsu, line 621: Same here: is there a reason to ever pass `setError: false`?

src/runtime/pythonexception.cs
```diff
@@ -148,6 +149,18 @@ public string PythonTypeName
             get { return _pythonTypeName; }
         }
 
+        /// <summary>
+        /// Replaces PyValue with an instance of PyType, if PyValue is not already an instance of PyType.
+        /// Often PyValue is a string and this method will replace it with a proper exception object.
+        /// Must not be called when an error is set.
```
lostmsu, line 155: Hm. Maybe add a check then?

Also, I can't find that requirement in the docs. Where is it coming from?

src/testing/Python.Test.csproj
```diff
@@ -2,6 +2,9 @@
   <PropertyGroup>
     <TargetFrameworks>netstandard2.0</TargetFrameworks>
   </PropertyGroup>
+  <PropertyGroup Condition="'$(Configuration)|$(TargetFramework)|$(Platform)'=='Debug|netstandard2.0|AnyCPU'">
+    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
+  </PropertyGroup>
```
lostmsu, line 7: Seems like this should be set unconditionally

lostmsu (changes requested):

src/runtime/methodbinder.cs
tminka: I can see two reasons:
1. If there were > 1 overload, and none matched, you would only get the last error.
2. If there were 10 overloads, and the last matched, you would waste time constructing error messages.
If you are happy with these consequences then I will make setError always true.

src/runtime/pythonexception.cs
tminka: I found that requirement by experimenting.  I will add a code comment explaining what happens.

src/runtime/methodbinder.cs
lostmsu: 1. Can you make in 1. so that all errors are kept as `AggregateException`?
2. I think that is fine. We might need a guide suggesting to choose overload if they need performance anyway. Because time is always wasted trying to convert parameters.

src/runtime/pythonexception.cs
lostmsu: This is weird. In the Python source code there's no indication, that this function has anything to do with the current error. I guess it is fine to leave the comment though.

src/runtime/exceptions.cs
```diff
@@ -181,6 +197,7 @@ internal static void SetArgsAndCause(IntPtr ob)
 
             if (e.InnerException != null)
             {
+                // Note: For an AggregateException, InnerException is only the first of the InnerExceptions.
```
lostmsu, line 200: The note is probably not needed, as an instance of `AggregateException` will itself have a property `InnerExceptions`. E.g. in this case `ob.InnerExceptions`.

src/runtime/methodbinder.cs
```diff
@@ -446,6 +463,13 @@ internal Binding Bind(IntPtr inst, IntPtr args, IntPtr kw, MethodBase info, Meth
                 {
                     // Best effort for determining method to match on gives multiple possible
                     // matches and we need at least one default argument - bail from this point
+                    StringBuilder stringBuilder = new StringBuilder("Not enough arguments provided to disambiguate the method.  Found:");
```
lostmsu, line 466: 🚀

Just want to double check that the message is correct, and adding arguments is the only way to disambiguate further.

lostmsu (changes requested):

src/runtime/exceptions.cs
tminka: I think the note is needed because this behaviour was surprising to me (and I consider it a bug in Python.NET that it defines the cause to be the first InnerException only).

src/runtime/methodbinder.cs
tminka: How else would you disambiguate?

src/runtime/methodbinder.cs
```diff
@@ -475,26 +499,44 @@ internal Binding Bind(IntPtr inst, IntPtr args, IntPtr kw, MethodBase info, Meth
...
+            }
+            if (mismatchedMethods.Count == 1)
+            {
+                mismatchedMethods[0].Exception.Restore();
+            }
+            else if (mismatchedMethods.Count > 1)
+            {
+                var aggregateException = GetAggregateException(mismatchedMethods);
+                Exceptions.SetError(aggregateException);
             }
```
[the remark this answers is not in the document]
tminka: I'm not changing the kind of top-level error.  This part is only configuring the cause.

src/runtime/exceptions.cs
```diff
@@ -76,12 +77,27 @@ public new static IntPtr tp_str(IntPtr ob)
             {
                 message = e.Message;
             }
+            if (e is AggregateException ae)
+            {
+                message += GetAggregateExceptionString(ae);
+            }
```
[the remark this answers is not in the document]
tminka: It is necessary because most users will not want or care to unpack an AggregateException.  They just want a meaningful error message by default.

src/runtime/methodbinder.cs
lostmsu: Perhaps by specifying arguments of different types?
But I am not sure which scenario this handles.

src/runtime/methodbinder.cs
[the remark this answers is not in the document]
lostmsu: I still think you should not be unwrapping `AggregateException` for the caller. Now callers have to handle two cases: when overloads exist, and when they don't.

src/runtime/methodbinder.cs
tminka: See test_default_params_overloads_ambiguous_call

src/runtime/exceptions.cs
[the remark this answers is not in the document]
lostmsu: What I mean is that this should not be done at `SetError` level. The caller of `SetError` can handle it on their own. This violates single responsibility.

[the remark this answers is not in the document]
tminka: The underlying problem is that this method doesn't call ToString on the exception.  It assumes that the exception message has all relevant information, which is not true for AggregateException.  I can change it to call ToString instead.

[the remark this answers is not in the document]
tminka: BTW ClassBase.tp_str does call ToString.  For some reason, ExceptionClassObject changes this behavior for exceptions.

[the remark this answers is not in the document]
lostmsu: Actually, that makes sense. It should just call `ToString`. I think custom `tp_str` was made to avoid "System.Exception: " prefix in the message.

Perhaps we could strip the prefix if the result of `ToString()` starts with it.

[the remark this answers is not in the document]
tminka: For all exceptions, or just AggregateException?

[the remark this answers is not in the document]
lostmsu: For all.

src/runtime/methodbinder.cs
```diff
@@ -475,26 +499,44 @@ internal Binding Bind(IntPtr inst, IntPtr args, IntPtr kw, MethodBase info, Meth
...
             return null;
         }
 
+        static AggregateException GetAggregateException(IEnumerable<MismatchedMethod> mismatchedMethods)
+        {
+            return new AggregateException(mismatchedMethods.Select(m => new ArgumentException($"{m.Exception.Message} in method {m.Method}")));
```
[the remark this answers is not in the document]
tminka: Unfortunately, this causes test_method_parameters_change to fail (see the build results)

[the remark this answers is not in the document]
lostmsu: I think you should disable the test for now. I dug around a bit, and it seems like domain reload would fail when a .NET object pointed to by Python has in turn references to Python objects (e.g. has either `PythonException` which causes `SerializationException` like in this case or `PyObject`, which would cause a refcount disbalance).

@BadSingleton and/or @benoithudson can later fix this issue and reenable the test case.

lostmsu (approved):

PR #1362, opened 2021-01-21
Disable implicit conversion from PyFloat to uint64

tminka:

Fixes an oversight in #1343 that allows PyFloat to continue being converted to uint64.

Fixes the test failures in #1361

### Any other comments?

...

### Checklist

### DIFF
--- a/src/runtime/converter.cs
+++ b/src/runtime/converter.cs
@@ -701,16 +701,12 @@ namespace Python.Runtime
                         op = value;
                         if (Runtime.PyObject_TYPE(value) != Runtime.PyLongType)
                         {
-                            op = Runtime.PyNumber_Long(value);
-                            if (op == IntPtr.Zero)
-                            {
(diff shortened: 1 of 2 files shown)


src/runtime/converter.cs
```diff
@@ -701,16 +701,12 @@ private static bool ToPrimitive(IntPtr value, Type obType, out object result, bo
                         op = value;
                         if (Runtime.PyObject_TYPE(value) != Runtime.PyLongType)
                         {
-                            op = Runtime.PyNumber_Long(value);
-                            if (op == IntPtr.Zero)
-                            {
-                                goto convert_error;
-                            }
+                            goto type_error;
```
lostmsu, line 704: I think this can be skipped alltogether. https://docs.python.org/3.8/c-api/long.html#c.PyLong_AsUnsignedLongLong already checks the type.

PR #1363, opened 2021-01-21
Disable implicit conversion from float to array index

tminka:

A float value as an array index will raise an error.  Fixes an oversight in #1343

The comment in #1342 

### Any other comments?

...

### Checklist

### DIFF
--- a/src/runtime/arrayobject.cs
+++ b/src/runtime/arrayobject.cs
@@ -159,6 +159,10 @@ namespace Python.Runtime
 
             if (rank == 1)
             {
+                if (!Runtime.PyInt_Check(idx))
+                {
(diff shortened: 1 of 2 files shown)


src/runtime/arrayobject.cs
```diff
@@ -320,6 +338,11 @@ public static new int mp_ass_subscript(IntPtr ob, IntPtr idx, IntPtr v)
             return 0;
         }
 
+        private static IntPtr RaiseTypeError(IntPtr idx)
```
lostmsu, line 341: NIT: `RaiseIndexMustBeIntegerError`

lostmsu (approved):

Issue #1355, opened 2021-01-14
Crash on Event invocation with ref parameter


noambonnieclear:
### Environment

-   Pythonnet version: 2.5.1
-   Python version: 3.6.5
-   Operating System: Windows 10
-   .NET Runtime: 4.8.4250.0

### Details

-   When I try to invoke a .NET event that takes a `ref` parameter, the python process crashes. I provided the code to reproduce the issue below.

***Note:*** I cannot change the code to not use `ref` as the `ref` delegate is coming from a 3rd party library. I can wrap around it with a lightweight C# piece of code but would rather avoid it if possible.
***Note:*** This may be related to the following issue: https://github.com/pythonnet/pythonnet/issues/965

#### Windows Event Viewer Error 1: Source - .NET Runtime
```
Application: python.exe
Framework Version: v4.0.30319
Description: The process was terminated due to an internal error in the .NET Runtime at IP 00007FF81CEA3D2E (00007FF81CEA0000) with exit code 80131506.
```

#### Windows Event Viewer Error 2: Source - Application Error
```
Faulting application name: python.exe, version: 3.6.5150.1013, time stamp: 0x5abbcb08
Faulting module name: clr.dll, version: 4.8.4250.0, time stamp: 0x5f2a059c
Exception code: 0xc0000005
Fault offset: 0x0000000000003d2e
Faulting process id: 0x4348
Faulting application start time: 0x01d6eaa4a034a863
Faulting application path: C:\Users\noam.bonnie\dev\hp-bluetooth-service\.tox\py36\Scripts\python.exe
Faulting module path: C:\Windows\Microsoft.NET\Framework64\v4.0.30319\clr.dll
Report Id: c2031029-3b2f-4e61-aa84-5e5502e904d2
Faulting package full name: 
Faulting package-relative application ID: 
```

#### The C# class with `ref` parameter for the delegate. Removing the `ref` resolves the problem
```C#
using System;

namespace TestEvent
{
    public delegate void MyEvent(ref string data); 

    public class MyTest
    {
        public event MyEvent myEvent;

        public virtual void triggerEvent()
        {
            Console.WriteLine("Invoking Event!");
            string myStr = "This is the event data";
            myEvent?.Invoke(ref myStr);
        }
    }
}    
```
#### Python program to invoke the event. Crashes. 
```python
import clr
import os
import sys


def handleEventData(data):
    print("=== event invoked. data: {} ===".format(data))


try:
[...]

PR #1364, opened 2021-01-21
Support ByRef arguments in event handlers

tminka:

When an event handler has a ByRef argument, extra logic is needed in DelegateManager.GetDispatcher or else the program will access invalid memory and crash.

#1355 

### Any other comments?

...

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ This document follows the conventions laid out in [Keep a CHANGELOG][].
 -   Ability to instantiate new .NET arrays using `Array[T](dim1, dim2, ...)` syntax
 -   Python operator method will call C# operator method for supported binary and unary operators ([#1324][p1324]).
 -   Add GetPythonThreadID and Interrupt methods in PythonEngine
+-   Ability to implement delegates with `ref` and `out` parameters in Python, by returning the modified parameter values in a tuple. ([#1355][i1355])
 
 ### Changed
 -   Drop support for Python 2, 3.4, and 3.5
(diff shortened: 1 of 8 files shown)


lostmsu:
This still does not allow python to change the value under the memory location passed by reference?

Ideally, a special Python `ByRef` type should be constructed to hold the value until function exits, then the caller should check if Python updated the value in the `ByRef` instance, and update the value pointed by reference accordingly.

tminka:
Perhaps, but your scenario is not the one in #1355 .  This PR at least prevents crashes.

lostmsu:
Since you can't override Python assignment operator, this fix is fundamentally incompatible with the proper solution with `ByRef`, so will have to be replaced by 3.0.0.

Even if we don't implement updating the value right now, for compatibility the value has to be wrapped into some Python object instance.

E.g. in the linked issue the handler code should be
```python
def handleEventData(data):
    print("=== event invoked. data: {} ===".format(data.Value)) # NOTE .Value
```

tminka:
When Python calls a .NET method with ref arguments, it returns a tuple of out/ref parameters, and if there's a return value it is the first in the tuple.  See [methodbinder.cs](https://github.com/pythonnet/pythonnet/blob/92932fd2a615e7ddcf16c4e4441a4840050b6bc9/src/runtime/methodbinder.cs#L962) and [stackoverflow](https://stackoverflow.com/questions/2857287/writing-iron-python-method-with-ref-or-out-parameter). Therefore the "proper solution" is for `handleEventData` to return a tuple of out/ref parameter values.  No special `ByRef` type is needed.

lostmsu:
Good point. Still for forward compatibility we must at least check Python returns a tuple of the correct size, or better yet actually implement the marshaling of data back to the event.

lostmsu (approved):

noambonnieclear:
So is this supposed to be released with 2.5.2?

PR #1365, opened 2021-01-22
Build single Python.Runtime.dll for all platforms

lostmsu:

This allows building the same Python.Runtime.dll on all platforms and supported Python versions.

To invoke Python code from .NET, one needs to set `Python.Runtime.Runtime.PythonDLL` or manually preload DLL before invoking any Python.NET functions or other properties.

### Implementation details

PInvoke signatures have been replaced with unmanaged functions pointers, that are loaded using `ILibraryLoader` on the first attempt to call any Python C API functions.

Marshaling attributes have been replaced with explicit marshaling calls.

`PythonDLL` gets its default value in the following order from:
1. `PYTHONNET_PYDLL` environment variable
2. Is set to `null` when launched from Python (so that DLL is already loaded in the process).
3. Set to a platform-dependent value, based on `PYTHONNET_PYVER` environment variable (e.g. '3.6' -> 'libpython3.6.so')

### Checklist


lostmsu:
A few issues still in the works:
1. Presence of 'm' suffix in the python C API DLL name. Used to be determined by `setup.py` and embedded into `Python.Runtime.dll`. We need some reliable way to get it from CI environment.
2. On Mac build fails due to some internal .NET SDK/Roslyn issue.

lostmsu:
Ideally, we should vend `Python.Runtime.dll` separately from the Python embedding module, and build it separately too.

filmor:
>     1. Presence of 'm' suffix in the python C API DLL name. Used to be determined by `setup.py` and embedded into `Python.Runtime.dll`. We need some reliable way to get it from CI environment.
On Linux, `pkg-config` is the way to go. I also think that we should be able to pass the libpython directly, not just as version + abiflags.

> Ideally, we should vend `Python.Runtime.dll` separately from the Python embedding module, and build it separately too.
That is the plan, with `clr-loader` we just put `Python.Runtime.dll` somewhere and load it via pure Python. I have this working against the P/Invoke implementation.

lostmsu:
Mac build appears to be blocked on https://github.com/dotnet/roslyn/issues/46772

lostmsu:
One check still failing, that's the same one everywhere else.

src/runtime/runtime.cs
```diff
@@ -722,6 +715,15 @@ internal static unsafe void XIncref(IntPtr op)
 #endif
         }
 
+        internal static NewReference NewRef(BorrowedReference reference)
```
filmor, line 718: Shouldn't this rather be on `BorrowedReference` directly?

src/runtime/runtime.cs
```diff
@@ -1026,44 +1030,69 @@ internal static bool PyObject_IsIterable(IntPtr pointer)
...
+        {
+            using var namePtr = new StrPtr(name, Encoding.UTF8);
+            return Delegates.PyObject_HasAttrString(pointer, namePtr);
+        }
 
-        [DllImport(_PythonDll, CallingConvention = CallingConvention.Cdecl)]
-        internal static extern IntPtr PyObject_GetAttrString(IntPtr pointer, [MarshalAs(UnmanagedType.CustomMarshaler, MarshalTypeRef = typeof(Utf8Marshaler))] string name);
+        internal static IntPtr PyObject_GetAttrString(IntPtr pointer, string name)
+        {
+            IntPtr nameMem = Marshal.StringToHGlobalAnsi(name);
```
filmor, line 1042: This will not work for unicode strings, is it just an oversight that you don't use your `StrPtr` here as for other functions?

src/runtime/runtime.cs
```diff
@@ -1026,44 +1030,69 @@ internal static bool PyObject_IsIterable(IntPtr pointer)
...
+            finally
+            {
+                Marshal.FreeHGlobal(nameMem);
+            }
+        }
 
-        [DllImport(_PythonDll, CallingConvention = CallingConvention.Cdecl)]
-        internal static extern int PyObject_SetAttrString(IntPtr pointer, [MarshalAs(UnmanagedType.CustomMarshaler, MarshalTypeRef = typeof(Utf8Marshaler))] string name, IntPtr value);
+        
+        internal static IntPtr PyObject_GetAttrString(IntPtr pointer, IntPtr name) => Delegates.PyObject_GetAttrString(pointer, name);
```
filmor, line 1054: Is this used somewhere? If not, I'd argue we should remove it from here, the string handling is not trivial.

src/runtime/runtime.cs
```diff
@@ -1026,44 +1030,69 @@ internal static bool PyObject_IsIterable(IntPtr pointer)
...
-        internal static extern IntPtr PyObject_GetAttr(IntPtr pointer, IntPtr name);
+        internal static int PyObject_SetAttrString(IntPtr pointer, string name, IntPtr value)
+        {
+            using var namePtr = new StrPtr(name, Encoding.UTF8);
+            return Delegates.PyObject_SetAttrString(pointer, namePtr, value);
+        }
 
-        [DllImport(_PythonDll, CallingConvention = CallingConvention.Cdecl)]
-        internal static extern int PyObject_SetAttr(IntPtr pointer, IntPtr name, IntPtr value);
+        internal static int PyObject_HasAttr(IntPtr pointer, IntPtr name) => Delegates.PyObject_HasAttr(pointer, name);
```
filmor, line 1063: Same as above, if not required, `Runtime` should not contain any delegates that have an `IntPtr` parameter for strings.

src/runtime/runtime.cs
```diff
@@ -1026,44 +1030,69 @@ internal static bool PyObject_IsIterable(IntPtr pointer)
...
+        internal static IntPtr PyObject_GetAttrString(IntPtr pointer, IntPtr name) => Delegates.PyObject_GetAttrString(pointer, name);
 
-        [DllImport(_PythonDll, CallingConvention = CallingConvention.Cdecl)]
-        internal static extern int PyObject_HasAttr(IntPtr pointer, IntPtr name);
 
-        [DllImport(_PythonDll, CallingConvention = CallingConvention.Cdecl)]
-        internal static extern IntPtr PyObject_GetAttr(IntPtr pointer, IntPtr name);
+        internal static int PyObject_SetAttrString(IntPtr pointer, string name, IntPtr value)
+        {
+            using var namePtr = new StrPtr(name, Encoding.UTF8);
```
filmor, line 1059: Since this pattern is used in a lot of places, we should have a function for this (`GetUtf8String` or something like that). That could open up potential optimisation paths with caching or preallocated buffers per thread.

src/runtime/runtime.cs
```diff
@@ -1287,199 +1332,193 @@ internal static bool PyFloat_Check(IntPtr ob)
...
 
-        [DllImport(_PythonDll, CallingConvention = CallingConvention.Cdecl)]
-        internal static extern IntPtr PyFloat_FromDouble(double value);
+        
+        internal static IntPtr PyFloat_FromDouble(double value) => Delegates.PyFloat_FromDouble(value);
 
-        [DllImport(_PythonDll, CallingConvention = CallingConvention.Cdecl)]
-        internal static extern IntPtr PyFloat_FromString(IntPtr value, IntPtr junk);
+        
+        internal static IntPtr PyFloat_FromString(IntPtr value, IntPtr junk) => Delegates.PyFloat_FromString(value, junk);
```
filmor, line 1347: Is this in use? The API docs don't say anything about this ominous `junk` parameter.

src/runtime/runtime.cs
```diff
@@ -1852,87 +1916,111 @@ internal static bool PyIter_Check(IntPtr pointer)
...
+        {
+            using var namePtr = new StrPtr(name, Encoding.UTF8);
+            return Delegates.PySys_SetObject(namePtr, ob);
+        }
 
 
         //====================================================================
         // Python type object API
         //====================================================================
+        static readonly delegate* unmanaged[Cdecl]<IntPtr, bool> pyType_Check;
```
filmor, line 2012: Apparently unused, is this a left-over from a previous attempt? Is this one also the reason `Runtime` is declared `unsafe`?

src/runtime/runtime.cs
```diff
@@ -2185,6 +2276,539 @@ internal static IntPtr GetBuiltins()
         {
             return PyImport_Import(PyIdentifier.builtins);
         }
+
+        private static class Delegates
```
filmor, line 2280: Please make `Runtime` a `partial` class and move this into a separate file.

src/runtime/runtime.cs
```diff
@@ -2185,6 +2276,539 @@ internal static IntPtr GetBuiltins()
...
+                PyBuffer_Release = (delegate* unmanaged[Cdecl]<ref Py_buffer, void>)GetFunctionByName(nameof(PyBuffer_Release), GetUnmanagedDll(_PythonDll));
+                try
+                {
+                    PyBuffer_SizeFromFormat = (delegate* unmanaged[Cdecl]<StrPtr, IntPtr>)GetFunctionByName(nameof(PyBuffer_SizeFromFormat), GetUnmanagedDll(_PythonDll));
+                }
+                catch (MissingMethodException)
+                {
+                    // only in 3.9+
+                }
+                PyBuffer_IsContiguous = (delegate* unmanaged[Cdecl]<ref Py_buffer, char, int>)GetFunctionByName(nameof(PyBuffer_IsContiguous), GetUnmanagedDll(_PythonDll));
```
filmor, line 2370: Did you generate this via a script? That would be a good addition (although I'd hope we wouldn't need too many changes of this class ;)).

src/runtime/pydict.cs
```diff
@@ -22,6 +22,7 @@ public PyDict(IntPtr ptr) : base(ptr)
         {
         }
 
+        internal PyDict(BorrowedReference reference) : base(reference) { }
```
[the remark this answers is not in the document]
lostmsu: This overload actually does incref internally (which makes sense, since to create a potentially long-lived `PyDict` object pointing to something we borrowed, we need to `IncRef`).

src/runtime/runtime.cs
lostmsu: This was a mistake, probably leftover from the first try.

src/runtime/runtime.cs
lostmsu: I changed the signature to accept `StrPtr` instead

src/runtime/runtime.cs
lostmsu: At this point it would be a premature optimization. Also, I don't see how we could cache disposable `struct` instances easily.

src/runtime/runtime.cs
lostmsu: I did not really walk through the APIs to check them apart from string usages. This looks bogus. I will change the signature.

src/runtime/importhook.cs
```diff
@@ -74,7 +75,7 @@ static void RestoreImport()
         /// <summary>
         /// Initialization performed on startup of the Python runtime.
         /// </summary>
-        internal static void Initialize()
+        internal static unsafe void Initialize()
```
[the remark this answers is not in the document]
lostmsu: Because of `_PyObject_GetDictPtr`

src/runtime/runtime.cs
lostmsu: I can do this just before pushing. Otherwise it would require me to update the roslyn package, that does automatic conversion of [DllImport] into the pattern with `Delegates`.

src/runtime/runtime.cs
lostmsu: The script converted `DllImport` declarations. It did not work on the Python source, headers, or docs. Essentially, signatures are identical to what we had except `string` parameters and a few minor exceptions where I switched to `*Reference` types.

src/runtime/native/StrPtr.cs
```diff
@@ -0,0 +1,73 @@
+#nullable enable
+using System;
+using System.Runtime.InteropServices;
+using System.Text;
+
+namespace Python.Runtime.Native
+{
+    [StructLayout(LayoutKind.Sequential)]
+    struct StrPtr : IDisposable
```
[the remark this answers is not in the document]
lostmsu: Naming things :-D

Not sure about `NativeString`. It has connotations to native string type supported on the platform, but this is not it.

It might make sense to have a `StrPtr` type per encoding for clarity. Then signatures could enforce it. But maybe that would be overengineering.

src/runtime/runtime.cs
lostmsu: `Runtime` in unsafe because any use of unmanaged function pointer requires `unsafe`. So all the methods with bodies like `=> Delegates.PyFunc_Pointer(args)`, so basically all methods.

src/runtime/native/StrPtr.cs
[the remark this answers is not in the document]
filmor: Do we use any encoding apart from UTF8? I thought all `FromString` functions nowadays used that.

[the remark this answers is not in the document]
filmor: The main thing I find "problematic" is that this object is not a pure pointer, it has ownership over the buffer it points to.

filmor (approved):

PR #1367, opened 2021-01-23
Simplify PyScope by delegating ownership to PyObject instance

lostmsu:

I noticed `PyScope` ownership implementation is so-so. One of the potential reasons for the crash could be `PyScope` being improperly deallocated.

This is an attempt to fix crash in CI on Ubuntu with Python 3.9.

No


lostmsu:
Dang :(

PR #1369, opened 2021-01-25
Modernize import hook

BadSingleton:
What does this implement/fix? Explain your changes.

This replaces the current import hook replacement with a MetaPathFinder, letting python handle non-CLR module import.
Does this close any currently open issues?

pythonnet#727 pythonnet#1245 pythonnet#547 pythonnet#111 and probably others.
Any other comments?

This fix also helps working around a mono runtime behaviour where a mono owns all threads that execute .NET code.
Checklist

Make sure to include one or more tests for your change
If an enhancement PR, please create docs and at best an example
Add yourself to AUTHORS
Updated the CHANGELOG


src/runtime/importhook.cs
```diff
@@ -94,6 +87,10 @@ internal static void Initialize()
             IntPtr dict = Runtime.PyImport_GetModuleDict();
             Runtime.PyDict_SetItemString(dict, "CLR", py_clr_module);
             Runtime.PyDict_SetItemString(dict, "clr", py_clr_module);
+
+            // Add/create the MetaPathLoader
+            SetupNamespaceTracking();
+            PythonEngine.Exec(LoaderCode);
```
filmor, line 93: I think this should rather be an actual module than just a "script" that is executed. Doesn't change much with regard to the C# code, but allows users to access the loader classes directly (e.g. to deactivate the loading for a short while and reactivate it later).

src/runtime/importhook.cs
```diff
@@ -9,13 +8,46 @@ namespace Python.Runtime
...
         private static IntPtr module_def = IntPtr.Zero;
 
+        private const string LoaderCode = @"
+import importlib.abc
+import sys
+
+class DotNetLoader(importlib.abc.Loader):
+
+    def __init__(self):
+        super(DotNetLoader, self).__init__()
```
filmor, line 23: `super().__init__()` should be enough for Python 3.

src/runtime/importhook.cs
```diff
@@ -9,13 +8,46 @@ namespace Python.Runtime
...
+
+    @classmethod
+    def exec_module(klass, mod):
+        # This method needs to exist.
+        pass
+
+    @classmethod
+    def create_module(klass, spec):
+        import clr
+        return clr._LoadClrModule(spec)
```
filmor, line 33: Maybe we should introduce another module within `clr` for these kind of interface functions. In any case, the exposed functions should (IMO) be "pythonic", in this case something like `clr.internal.load_clr_module`.

src/runtime/importhook.cs
```diff
@@ -9,13 +8,46 @@ namespace Python.Runtime
...
+
+class DotNetFinder(importlib.abc.MetaPathFinder):
+    
+    def __init__(self):
+        super(DotNetFinder, self).__init__()
+    
+    @classmethod
+    def find_spec(klass, fullname, paths=None, target=None): 
+        import clr
+        if (hasattr(clr, '_availableNamespaces') and fullname in clr._availableNamespaces):
```
filmor, line 43: Why is this written so defensively?

src/runtime/importhook.cs
```diff
@@ -134,199 +130,135 @@ internal static void SaveRuntimeData(RuntimeDataStorage storage)
 
         internal static void RestoreRuntimeData(RuntimeDataStorage storage)
         {
-            InitImport();
             storage.GetValue("py_clr_module", out py_clr_module);
             var rootHandle = storage.GetValue<IntPtr>("root");
             root = (CLRModule)ManagedType.GetManagedObject(rootHandle);
+            IntPtr dict = Runtime.PyImport_GetModuleDict();
+            Runtime.PyDict_SetItemString(dict, "CLR", py_clr_module);
```
filmor, line 137: `CLR` should be gone since a few commits.

src/runtime/importhook.cs
```diff
@@ -134,199 +130,135 @@ internal static void SaveRuntimeData(RuntimeDataStorage storage)
...
+                foreach (var ns in AssemblyManager.GetNamespaces())
                 {
-                    Runtime.XIncref(py_mod_dict);
-                    using (var mod_dict = new PyDict(py_mod_dict))
+                    var pyNs = Runtime.PyString_FromString(ns);
+                    try
                     {
-                        Runtime.XIncref(fromList.GetValueOrDefault());
-                        using (var from = new PyTuple(fromList.GetValueOrDefault()))
+                        if(Runtime.PySet_Add(newset, pyNs) != 0)
```
filmor, line 158: Nit, but please run a formatter over the changed files.

src/runtime/importhook.cs
```diff
@@ -134,199 +130,135 @@ internal static void SaveRuntimeData(RuntimeDataStorage storage)
...
-                return Exceptions.RaiseTypeError("__import__() takes at least 1 argument (0 given)");
-            }
-
-            // borrowed reference
-            IntPtr py_mod_name = Runtime.PyTuple_GetItem(args, 0);
-            if (py_mod_name == IntPtr.Zero ||
-                !Runtime.IsStringType(py_mod_name))
+            AssemblyManager.namespaceAdded -= OnNamespaceAdded;
+            // If the C# runtime isn't loaded, then there is no namespaces available
+            if ((Runtime.PyDict_DelItemString(root.dict, availableNsKey) != 0) &&
```
filmor, line 191: Set it to `None` instead.

src/runtime/importhook.cs
```diff
@@ -134,199 +130,135 @@ internal static void SaveRuntimeData(RuntimeDataStorage storage)
...
-                }
-                module = Runtime.PyDict_GetItemString(modules, names[0]);
-                Runtime.XIncref(module);
-                return module;
-            }
-            Exceptions.Clear();
+        /// <summary>
+        /// Return the clr python module (new reference)
+        /// </summary>
+        public static IntPtr GetCLRModule()
```
filmor, line 247: Use `NewReference` for the return value.

src/runtime/importhook.cs
```diff
@@ -134,199 +130,135 @@ internal static void SaveRuntimeData(RuntimeDataStorage storage)
...
+            Runtime.XIncref(py_clr_module);
+            return py_clr_module;
+        }
 
-            // Traverse the qualified module name to get the named module
-            // and place references in sys.modules as we go. Note that if
+        /// <summary>
+        /// The hook to import a CLR module into Python
+        /// </summary>
+        public static ModuleObject __import__(string modname)
```
filmor, line 258: This doesn't replace Python's `__import__` anymore so it should probably just be called `Import`

src/runtime/importhook.cs
BadSingleton: As things are right now, when the domain unloads, the attribute is removed.

src/runtime/importhook.cs
```diff
@@ -9,14 +8,45 @@ namespace Python.Runtime
...
         private static IntPtr module_def = IntPtr.Zero;
 
+        private const string LoaderCode = @"
+import importlib.abc
+import sys
+
+class DotNetLoader(importlib.abc.Loader):
+
+    def __init__(self):
+        super().__init__()
```
benoithudson, line 24: Is this init needed? It does nothing other than tell the MRO to keep looking for init, which is what it was going to do anyway if this function didn't exist.

benoithudson: (same comment in the finder below)

src/runtime/importhook.cs
```diff
@@ -99,8 +83,12 @@ internal static unsafe void Initialize()
 
             Runtime.PyDict_Update(mod_dict, clr_dict);
             BorrowedReference dict = Runtime.PyImport_GetModuleDict();
-            Runtime.PyDict_SetItemString(dict, "CLR", ClrModuleReference);
-            Runtime.PyDict_SetItemString(dict, "clr", ClrModuleReference);
+            Runtime.PyDict_SetItemString(dict.DangerousGetAddress(), "CLR", py_clr_module);
+            Runtime.PyDict_SetItemString(dict.DangerousGetAddress(), "clr", py_clr_module);
```
benoithudson, line 87: why are you switching to DangerousGetAddress here?

src/runtime/importhook.cs
```diff
@@ -141,189 +128,177 @@ internal static void SaveRuntimeData(RuntimeDataStorage storage)
...
+                throw new PythonException();
             }
-            Runtime.XIncref(py_clr_module);
-            return NewReference.DangerousFromPointer(py_clr_module);
+
+            // Finally, add the hook to the meta path
+            var finder_inst = Runtime.PyDict_GetItemString(mod_dict, "finder_inst").DangerousGetAddressOrNull();
+            Runtime.XIncref(finder);
+            var metapath = Runtime.PySys_GetObject("meta_path");
+            Runtime.PyList_Append(metapath, finder);
```
benoithudson, line 190: These four lines look fishy: you don't use finder_inst and you are doing an incref on finder for no obvious reason.

src/runtime/importhook.cs
```diff
@@ -141,189 +128,177 @@ internal static void SaveRuntimeData(RuntimeDataStorage storage)
...
-                return Exceptions.RaiseTypeError("string expected");
-            }
-
-            // Check whether the import is of the form 'from x import y'.
-            // This determines whether we return the head or tail module.
-
-            BorrowedReference fromList = default;
-            var fromlist = false;
-            if (num_args >= 4)
+            var newset = Runtime.PySet_New(new BorrowedReference(IntPtr.Zero));
```
benoithudson, line 201: You Dispose in a finally; should this be a `using var` ?

src/runtime/importhook.cs
BadSingleton: oh.. that should be `finder_inst`, not finder

src/runtime/importhook.cs
```diff
@@ -141,189 +128,177 @@ internal static void SaveRuntimeData(RuntimeDataStorage storage)
...
+        {
+            using (Py.GIL())
             {
-                if (fromlist)
+                var pyNs = Runtime.PyString_FromString(name);
+                try
                 {
-                    if (IsLoadAll(fromList))
+                    var nsSet = Runtime.PyDict_GetItemString(new BorrowedReference(root.dict), availableNsKey);
+                    if (!nsSet.IsNull || nsSet.DangerousGetAddress() != Runtime.PyNone)
```
benoithudson, line 253: Should BorrowedReference have an IsNullOrNone property? It seems likely to be a common idiom.

src/runtime/importhook.cs
lostmsu: BTW, it makes sense to add an overload for `PyList_Append` with proper reference types, then avoid manual refcounting.

Same for `PySys_GetObject` - you can rename the original one to `Dangerous_PySys_GetObject` to avoid refactoring other call sites.

This is a NIT though. Should come up with a better plan to switch everything.

src/runtime/importhook.cs
lostmsu: NIT: I'd replace `new BorrowedReference(IntPtr.Zero)` with either `BorrowedReference.Null` or simply `default`.

src/runtime/importhook.cs
lostmsu: I'd keep separate `IsNull` and `IsNone`. Thought there's already `IsNone`.

src/runtime/importhook.cs
BadSingleton: I think that was an aritfact from rebasing

src/runtime/importhook.cs
```diff
@@ -115,187 +97,154 @@ internal static void SaveRuntimeData(RuntimeDataStorage storage)
 
         internal static void RestoreRuntimeData(RuntimeDataStorage storage)
         {
-            InitImport();
             storage.GetValue("py_clr_module", out py_clr_module);
             var rootHandle = storage.GetValue<IntPtr>("root");
             root = (CLRModule)ManagedType.GetManagedObject(rootHandle);
+            BorrowedReference dict = Runtime.PyImport_GetModuleDict();
+            Runtime.PyDict_SetItemString(dict.DangerousGetAddress(), "clr", py_clr_module);
```
lostmsu, line 104: I'd add a property `ClrModuleReference` that returns a `BorrowedReference` of `py_clr_module`, and then used it in `PyDict_SetItemString` overload, that works with `BorrowedReferences` to avoid using `Dangerous*` methods.

src/runtime/importhook.cs
```diff
@@ -115,187 +97,154 @@ internal static void SaveRuntimeData(RuntimeDataStorage storage)
...
+            var exec = Runtime.PyDict_GetItemString(builtins, "exec");
+            using var args = NewReference.DangerousFromPointer(Runtime.PyTuple_New(2));
+
+            IntPtr codeStr = Runtime.PyString_FromString(LoaderCode);
+            Runtime.PyTuple_SetItem(args.DangerousGetAddress(), 0, codeStr);
+            // PyTuple_SetItem steals a reference, mod_dict is borrowed.
+            var mod_dict = Runtime.PyModule_GetDict(import_hook_module);
+            Runtime.XIncref(mod_dict.DangerousGetAddress());
+            Runtime.PyTuple_SetItem(args.DangerousGetAddress(), 1, mod_dict.DangerousGetAddress());
+            Runtime.PyObject_Call(exec.DangerousGetAddress(), args.DangerousGetAddress(), IntPtr.Zero);
```
lostmsu, line 124: It is better to make an overload with the new reference types when possible without rewriting existing code, than to call `Dangerous*` methods. See for example https://github.com/pythonnet/pythonnet/blob/7d8f754ff6c4787133427f707f58727d4ddc4981/src/runtime/runtime.cs#L996

One can also be made for `PyTuple_SetItem` using the new `StolenReference`.

src/runtime/importhook.cs
```diff
@@ -115,187 +97,154 @@ internal static void SaveRuntimeData(RuntimeDataStorage storage)
...
+
+            // Set as a sub-module of clr.
+            if(Runtime.PyModule_AddObject(ClrModuleReference, "loader", import_hook_module) != 0)
             {
-                Runtime.PyDict_Update(py_mod_dict, clr_dict);
+                Runtime.XDecref(import_hook_module.DangerousGetAddress());
```
lostmsu, line 129: Instead of calling decref on a `NewReference` here you can rewrite it as `using var import_hook_module = ...` above to dispose it automatically.

src/runtime/importhook.cs
lostmsu: This applies to a few Runtime methods across the PR. Basically anything where you can add a new overload.

src/runtime/importhook.cs
```diff
@@ -115,187 +97,154 @@ internal static void SaveRuntimeData(RuntimeDataStorage storage)
...
+
+        /// <summary>
+        /// Sets up the tracking of loaded namespaces. This makes available to 
+        /// Python, as a Python object, the loaded namespaces. The set of loaded
+        /// namespaces is used during the import to verify if we can import a 
+        /// CLR assembly as a module or not. The set is stored on the clr module.
+        /// </summary>
+        static void SetupNamespaceTracking()
+        {
+            var newset = Runtime.PySet_New(default);
```
lostmsu, line 150: `using var newset = ...` would remove the need for `try ... finally ...`

src/runtime/importhook.cs
```diff
@@ -115,187 +97,154 @@ internal static void SaveRuntimeData(RuntimeDataStorage storage)
...
+        }
 
-            string mod_name = Runtime.GetManagedString(py_mod_name);
-            // Check these BEFORE the built-in import runs; may as well
-            // do the Incref()ed return here, since we've already found
-            // the module.
-            if (mod_name == "clr")
+        public static void AddNamespace(string name)
+        {
+            using (Py.GIL())
```
lostmsu, line 192: Are there scenarios when this method will be called without GIL acquired?
If it is a module method, can it be attributed with `allowthreads=false`?

src/runtime/importhook.cs
```diff
@@ -115,187 +97,154 @@ internal static void SaveRuntimeData(RuntimeDataStorage storage)
...
-                    fromlist = true;
-                }
-            }
+        /// <summary>
+        /// Removes the set of available namespaces from the clr module.
+        /// </summary>
+        static void TeardownNameSpaceTracking()
+        {
+            // If the C# runtime isn't loaded, then there are no namespaces available
+            Runtime.PyDict_SetItemString(root.dict, availableNsKey, Runtime.PyNone);
```
lostmsu, line 187: Why not remove the key entirely? You would not need to compare it to `None` below in that case.

src/runtime/importhook.cs
```diff
@@ -115,187 +97,154 @@ internal static void SaveRuntimeData(RuntimeDataStorage storage)
...
-            // do the Incref()ed return here, since we've already found
-            // the module.
-            if (mod_name == "clr")
+        public static void AddNamespace(string name)
+        {
+            using (Py.GIL())
             {
-                NewReference clr_module = GetCLRModule(fromList);
-                if (!clr_module.IsNull())
+                var pyNs = Runtime.PyString_FromString(name);
```
lostmsu, line 194: NIT: `using var pyNs = NewReference.DangerousFromPoiner(Runtime.PyString_FromString(name))` would remove both `try ... finally ...` and a cast for `PySet_Add`

src/runtime/moduleobject.cs
```diff
@@ -181,7 +187,23 @@ public void LoadNames()
...
                 }
-                GetAttribute(name, true);
+
+                if(GetAttribute(name, true) != null)
+                {
+                    // if it's a valid attribute, add it to __all__
+                    var pyname = Runtime.PyString_FromString(name);
+                    try
+                    {
+                        if (Runtime.PyList_Append(new BorrowedReference(__all__), pyname) != 0)
```
lostmsu, line 197: Frankly, I screwed up `PyList_Append`. Its only overload schizophrenically mixes references and raw pointers. Can you change it to either have two overloads with only references and with only pointers, or just one that takes only references?

src/runtime/moduleobject.cs
```diff
@@ -320,6 +349,25 @@ protected override void Clear()
             base.Clear();
         }
 
+        /// <summary>
+        /// Override the setattr implementation.
+        /// This is needed because the import mechanics need
+        /// to set a few attributes
+        /// </summary>
+        [ForbidPythonThreads]
```
lostmsu, line 357: Why do you have to forbid threads here?

src/runtime/moduleobject.cs
```diff
@@ -526,5 +584,25 @@ public static string[] ListAssemblies(bool verbose)
...
+        public static ModuleObject _load_clr_module(PyObject spec)
+        {
+            ModuleObject mod = null;
+            using (var modname = spec.GetAttr("name"))
+            {
+                mod = ImportHook.Import(modname.ToString());
+            }
+            return mod;
+        }
```
lostmsu, line 606: NIT:
```
using var  modname = ...;
return ...;
```

src/runtime/moduleobject.cs
```diff
@@ -526,5 +584,25 @@ public static string[] ListAssemblies(bool verbose)
...
+        /// <summary>
+        /// Note: This should *not* be called directly.
+        /// The function that get/import a CLR assembly as a python module.
+        /// This function should only be called by the import machinery as seen
+        /// in importhook.cs
+        /// </summary>
+        /// <param name="spec">A ModuleSpec Python object</param>
+        /// <returns>A new reference to the imported module, as a PyObject.</returns>
+        [ModuleFunction]
+        [ForbidPythonThreads]
```
lostmsu, line 597: Why do you have to forbid python threads here?

src/runtime/runtime.cs
```diff
@@ -1944,6 +1944,11 @@ internal static string PyModule_GetFilename(IntPtr module)
 
 
         internal static IntPtr PyImport_Import(IntPtr name) => Delegates.PyImport_Import(name);
+        internal static int PyModule_AddObject(BorrowedReference module, string name, BorrowedReference stolenObject)
```
lostmsu, line 1947: Looks like `stolenObject` parameter should be of type `StolenReference` here and below.

lostmsu (changes requested):
Only reviewed in regards to usage of references and exceptions.

@filmor should still sign off on functionality.

src/runtime/importhook.cs
BadSingleton: It used to be possible. The method calling it now has the attribute

src/runtime/importhook.cs
BadSingleton: The DotNetFinder relies on that key being present on the object. If I remove it in AddNamespace, I need to add a check over there instead.

src/runtime/runtime.cs
BadSingleton: > StolenReference here and below (present in the latest master)

🎉

BadSingleton: Although I'm reviewing the semantics, and this function only steals on success, so by using `StolenReference`, we leak on failure.

src/runtime/moduleobject.cs
BadSingleton: Because functions called from `ImportHook.Import` calls CPython internally.

src/runtime/moduleobject.cs
BadSingleton: Because there's a call to `Runtime.PyDict_SetItem`. I might be missing something and this is always called from CPython and the GIL is guaranteed to be held.

src/runtime/moduleobject.cs
BadSingleton: There are only a few calls to `PyList_Append` throughout the code base, I changed all of them to use only references.

src/runtime/runtime.cs
lostmsu: Dang you are right. Why did they not make it consistent? `PyTuple_SetItem` for example does steal even when there's an error. ugh...

Maybe we should make a wrapper, that always steals?

lostmsu: I think that we should not use `BorrowedReference` for the last parameter here. It makes false suggestion, that signature is valid, and borrowed reference is always enough when it is really not. I'd keep it `IntPtr` for now, and explain the behavior in the parameter doc.

Or alternatively, make a wrapper, that takes either `BorrowedReference` or `StolenReference` and handles refcounting internally, and keep the real PInvoke private.

lostmsu (changes requested):

src/runtime/runtime.cs
BadSingleton: I'm leaning towards the `IntPtr` parameter to be as "consistant" as possible with the CPython API.

src/runtime/converter.cs
```diff
@@ -202,6 +202,14 @@ internal static IntPtr ToPython(object value, Type type)
                     return ClassDerivedObject.ToPython(pyderived);
             }
 
+            // ModuleObjects are created in a way that their wrapping them as
+            // a CLRObject fails, the ClassObject has no tpHandle. Return the
+            // pyHandle as is, do not convert.
+            if (value is ModuleObject modobj)
+            {
+                return modobj.pyHandle;
```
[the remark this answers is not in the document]
lostmsu: @BadSingleton what about this one?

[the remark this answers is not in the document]
BadSingleton: My bad missed that one.

lostmsu (approved):
@filmor I'm good with the code. Please see if you are OK with the approach.

filmor:
I'll review this today.

hiaselhans:
great news! :)

filmor (approved):

Issue #1374, opened 2021-02-04
Create an F# test project


filmor:
F# generated CIL has some unique properties (like potentially missing `ParameterInfo.Name`) that should be caught in our test-suite. For that we should have a test project like `Python.Tests` written in F# that is called from the pytest test-suite in the same fashion as the C# tests.

PR #1375, opened 2021-02-04
ParameterInfo.Name needs to be checked for null before usage

filmor:

This occured in trying to use F# code from Python. As the `.Name` property returns `null`, `ContainsKey` fails.

I don't think so.

### Any other comments?

Related documentation: https://docs.microsoft.com/en-us/dotnet/api/system.reflection.parameterinfo.name

I don't know if there is a way to create a C# method without a parameter name. I'll try to set up a few simple F# test-cases in the future in a separate PR (follow-up ticket: #1374).

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -56,6 +56,7 @@ or the DLL must be loaded in advance. This must be done before calling any other
 -   `import` may now raise errors with more detail than "No module named X"
 -   Exception stacktraces on `PythonException.StackTrace` are now properly formatted
 -   Providing an invalid type parameter to a generic type or method produces a helpful Python error
+-   Empty parameter names (as can be generated from F#) do not cause crashes
 
 ### Removed
 
--- a/src/runtime/methodbinder.cs
+++ b/src/runtime/methodbinder.cs
@@ -622,7 +622,7 @@ namespace Python.Runtime
             for (int paramIndex = 0; paramIndex < pi.Length; paramIndex++)
             {
                 var parameter = pi[paramIndex];
-                bool hasNamedParam = kwargDict.ContainsKey(parameter.Name);
+                bool hasNamedParam = parameter.Name != null ? kwargDict.ContainsKey(parameter.Name) : false;
                 bool isNewReference = false;
 
                 if (paramIndex >= pyArgCount && !(hasNamedParam || (paramsArray && paramIndex == arrayStart)))


lostmsu:
Can you add a test?

lostmsu:
I see #1374

noambonnieclear:
I tried testing this with the latest 2.5.2 release. Could it be that there was something wrong with the distribution of the package? We checked on multiple environments and it looks like when we install 2.5.2 we still get 2.5.1. Really odd but we're running out of ideas of what we might be doing wrong for a straight forward upgrade... :-)

For example:
```
PS C:\Users\user> pip install --upgrade pythonnet==2.5.2
Collecting pythonnet==2.5.2
  Downloading https://files.pythonhosted.org/packages/0c/b3/e136267c9299ac5572286ccb807f4b0a6bad9b882e7ca94691004a6acc83/pythonnet-2.5.2-cp36-cp36m-win_amd64.whl (81kB)
     |████████████████████████████████| 92kB 737kB/s
Requirement already satisfied, skipping upgrade: pycparser in c:\users\apaganini\appdata\local\programs\python\python36\lib\site-packages (from pythonnet==2.5.2) (2.20)
Installing collected packages: pythonnet
  Found existing installation: pythonnet 2.5.1
    Uninstalling pythonnet-2.5.1:
      Successfully uninstalled pythonnet-2.5.1
Successfully installed pythonnet-2.5.2
WARNING: You are using pip version 19.3.1; however, version 21.0.1 is available.
You should consider upgrading via the 'python -m pip install --upgrade pip' command.
PS C:\Users\user> py
Python 3.6.5 (v3.6.5:f59c0932b4, Mar 28 2018, 17:00:18) [MSC v.1900 64 bit (AMD64)] on win32
>>> import clr
>>> clr.__version__
'2.5.1'
```

filmor:
It's possible that I missed one version string somewhere, sorry about that. The process on the backports branch is completely manual.

noambonnieclear:
Thank you @filmor - what's the path forward? is it going to be re-released as 2.5.2 or changed to 2.5.3? i.e. - will I get notified of a new release when this is corrected?

noambonnieclear:
@filmor I apologize but I actually commented on the wrong issue. It was supposed to be on https://github.com/pythonnet/pythonnet/pull/1364 - But regardless looks like something in the distribution is wrong based on the above comment.

filmor:
I don't intend to do a 2.5.3 release just for this. 2.5 is frozen, only patches for regressions are backported. Work is happening on `master`. The issue you wanted to comment on is not merged into backports-2.5 anyway.

Issue #1379, opened 2021-02-11
Neglected Python.NET in NuGet package


Livius90:
Hi!

As i see [NuGet package for Python.NET](https://www.nuget.org/packages/pythonnet/2.3.0-py35-dotnet) is really old, last updated was in 2017. Could you update NuGet package to the latest version and can you do more maintaining in your NuGet package to be able to download the newest version in all the time?

filmor:
This is a very bold first contribution. Nuget package creation is already discussed in #165, #778, #1194, you are not the first one to notice this. With version 3.0 (and probably some dev versions inbetween) we will be able to automatically produce a unified Nuget package for all architectures and Python versions. The necessary code in the C# part is already there, you are welcome to contribute the required CI steps.

lostmsu:
BTW, I like the idea of starting previews for 3.0. I think monthly release from `master` would make sense. @filmor what do you think?

filmor:
Fine by me :)

Will you set this up? I don't have access to our nuget org.

lostmsu:
@filmor I need to give you access then. What's you NuGet email (can mail me)?

PR #1380, opened 2021-02-11
Monthly NuGet release previews

lostmsu:

This generates monthly release previews on NuGet using GitHub actions

 #1379

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -38,6 +38,9 @@ indent_style = tab
 charset = utf-8
 insert_final_newline = true
 
+[Python.Runtime.csproj]
+charset = latin1
+
 # bumpversion reformats itself after every bump
 [.bumpversion.cfg]
 trim_trailing_whitespace = false
--- a/.github/workflows/nuget-preview.yml
+++ b/.github/workflows/nuget-preview.yml
@@ -1,0 +1,67 @@
+name: GitHub Actions
+
+on: [ pull_request ] # testing only
+#on:
+#  schedule:
+#  - cron: "5 4 3 */1 *" # once a month, at 4:05 on 3rd
+
+jobs:
+  release:
+    name: Release Preview
+    runs-on: ubuntu-latest
+    environment: NuGet
+    timeout-minutes: 10
+
+    env:
+      PYTHONNET_SHUTDOWN_MODE: Normal
+
+    steps:
+      - name: Get Date
+        run: |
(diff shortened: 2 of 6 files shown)


filmor:
Does this really need to run tests? IMHO it's perfectly fine to keep this as simple as possible, i.e. to just run `dotnet pack`. I also wouldn't call it explicitly `monthly`, it might be useful to be able to trigger these manually. `preview` should do.

lostmsu:
I think it is better to keep tests to avoid non-functional releases.

.editorconfig
```diff
@@ -38,6 +38,9 @@ indent_style = tab
 charset = utf-8
 insert_final_newline = true
 
+[Python.Runtime.csproj]
+charset = latin1
```
filmor, line 42: This is an odd choice for XML, is this intentional?

.github/workflows/nuget-preview.yml
```diff
@@ -0,0 +1,69 @@
...
+        env:
+          PYTHONNET_PYDLL: libpython3.8.so
+
+      - name: Pack
+        run: dotnet pack --configuration Release --version-suffix preview${{env.DATE_VER}}
+
+      - name: Publish NuGet
+        run: dotnet nuget push --api-key ${NUGET_KEY}
+        env:
+          NUGET_KEY: ${{ secrets.NUGET_MONTHLY }}
```
filmor, line 66: I think this way you'll leak the key.

lostmsu:
Superseded by #1381

.github/workflows/nuget-preview.yml
lostmsu: Without this Visual Studio tends to add UTF-8 BOM header to solution, leading to PRs with the entire file replaced. It simply ignores `utf-8`.

.editorconfig
lostmsu: Without this Visual Studio tends to add UTF-8 BOM header to solution, leading to PRs with the entire file replaced. It simply ignores `utf-8`.

.github/workflows/nuget-preview.yml
lostmsu: I don't think so, but the reason I had to try to add this turned out to be bogus: it was not picked up, because PR was from my fork. I moved the PR branch into this repository instead, now it is getting picked up correctly. See #1381

PR #1391, opened 2021-02-19
Remove Utf8Marshaler, set PyScope base class to PyObject, added PyModule

lostmsu:
Also allow `BorrowedReference == null` comparison.

### DIFF
--- a/src/runtime/pymodule.cs
+++ b/src/runtime/pymodule.cs
@@ -1,0 +1,41 @@
+using System;
+
+namespace Python.Runtime
+{
+    public class PyModule : PyScope
+    {
+        internal PyModule(ref NewReference reference) : base(ref reference, PyScopeManager.Global) { }
+        public PyModule(PyObject o) : base(o.Reference, PyScopeManager.Global) { }
+
+        /// <summary>
+        /// Given a module or package name, import the
+        /// module and return the resulting module object as a <see cref="PyModule"/>.
+        /// </summary>
+        /// <param name="name">Fully-qualified module or package name</param>
+        public static PyModule Import(string name)
+        {
+            NewReference op = Runtime.PyImport_ImportModule(name);
+            PythonException.ThrowIfIsNull(op);
+            return new PyModule(ref op);
+        }
+
+        /// <summary>
+        /// Reloads the module, and returns the updated object
+        /// </summary>
+        public PyModule Reload()
+        {
+            NewReference op = Runtime.PyImport_ReloadModule(this.Reference);
+            PythonException.ThrowIfIsNull(op);
+            return new PyModule(ref op);
+        }
+
+        public static PyModule FromString(string name, string code)
+        {
+            using NewReference c = Runtime.Py_CompileString(code, "none", (int)RunFlagType.File);
+            PythonException.ThrowIfIsNull(c);
+            NewReference m = Runtime.PyImport_ExecCodeModule(name, c);
+            PythonException.ThrowIfIsNull(m);
+            return new PyModule(ref m);
+        }
+    }
+}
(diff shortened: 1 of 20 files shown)


src/runtime/pymodule.cs
```diff
@@ -0,0 +1,40 @@
...
+        public PyModule Reload()
+        {
+            NewReference op = Runtime.PyImport_ReloadModule(this.Reference);
+            PythonException.ThrowIfIsNull(op);
+            return new PyModule(ref op);
+        }
+
+        public static PyModule FromString(string name, string code)
+        {
+            using NewReference c = Runtime.Py_CompileString(code, "none", (int)RunFlagType.File);
```
filmor, line 33: We can do a little better using caller information: https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/attributes/caller-information

src/runtime/pythonengine.cs
```diff
@@ -469,61 +469,15 @@ public static void EndAllowThreads(IntPtr ts)
...
 
-        /// <summary>
-        /// ImportModule Method
-        /// </summary>
-        /// <remarks>
-        /// Given a fully-qualified module or package name, import the
-        /// module and return the resulting module object as a PyObject
-        /// or null if an exception is raised.
-        /// </remarks>
-        public static PyObject ImportModule(string name)
```
filmor, line 480: We should probably leave these for backward-compat with an `[Obsolete]` annotation. Since they can all be easily implemented in terms of the "new" way of doing things, this is very little additional code to maintain and we can drop it with 3.1 or so.

src/runtime/pythonengine.cs
```diff
@@ -744,7 +698,7 @@ public static KeywordArguments kw(params object[] kv)
 
         public static PyObject Import(string name)
```
filmor, line 699: I think this should also be obsoleted. Having a clear separation of concerns in our API will help us documenting it and keeping it stable for a long time.

src/runtime/tricks/NullOnly.cs
```diff
@@ -0,0 +1,7 @@
+namespace Python.Runtime
+{
+    class NullOnly
```
filmor, line 3: Maybe give this one a few lines of documentation.

src/runtime/pymodule.cs
lostmsu: If you are suggesting this for `filename` parameter, then passing C# code file path here does not seem like a good idea. Line numbers will be off, and there's no guarantee, that the source code being compiled is present at the caller, rather than somewhere higher in the callstack.

filmor: It would just allow the user to actually see where the code is coming from. The `filename` of a code object is mostly informational, as far as I can see, and if the respective file is not readily available, it will just not show up in stack traces.

lostmsu: I am not going to make this change, as I find it potentially confusing, and of limited usefulness.

src/runtime/pythonengine.cs
```diff
@@ -742,9 +707,10 @@ public static KeywordArguments kw(params object[] kv)
             return dict;
         }
 
+        [Obsolete("Use PyModule.Import")]
         public static PyObject Import(string name)
```
filmor, line 711: I wouldn't obsolete this one right now. Having `Py.` as a central entry-point has merit.

lostmsu: Perhaps we should move `Py` class into a new file
image

filmor: :D
Yep, sorry for the inconsistency. I'd like to obsolete `PythonEngine` for the most part and collect entry points in `Py`.

lostmsu: Removed this `[Obsolete]`, added a new issue to move `Py` out.

PR #1395, opened 2021-02-22
PyType class, wrapper for Python types

lostmsu:

This introduces a new `PyType` class (similar to `PyInt`, but for types). It can:
- wrap existing `PyObject`, that points to a type
- create a new heap type from spec (`TypeSpec` class)

It is WIP on https://github.com/pythonnet/pythonnet/issues/1033
It is related to https://github.com/pythonnet/pythonnet/pull/1196

### Checklist

### DIFF
--- a/src/runtime/clrobject.cs
+++ b/src/runtime/clrobject.cs
@@ -14,7 +14,7 @@ namespace Python.Runtime
             System.Diagnostics.Debug.Assert(tp != IntPtr.Zero);
             IntPtr py = Runtime.PyType_GenericAlloc(tp, 0);
 
-            long flags = Util.ReadCLong(tp, TypeOffset.tp_flags);
+            var flags = (TypeFlags)Util.ReadCLong(tp, TypeOffset.tp_flags);
             if ((flags & TypeFlags.Subclass) != 0)
             {
                 IntPtr dict = Marshal.ReadIntPtr(py, ObjectOffset.TypeDictOffset(tp));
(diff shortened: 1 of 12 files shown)


lostmsu:
Hm, on non-Windows, utf8 decoding fails inside Python with a funky name.

src/runtime/clrobject.cs
```diff
@@ -14,7 +14,7 @@ internal CLRObject(object ob, IntPtr tp)
             System.Diagnostics.Debug.Assert(tp != IntPtr.Zero);
             IntPtr py = Runtime.PyType_GenericAlloc(tp, 0);
 
-            long flags = Util.ReadCLong(tp, TypeOffset.tp_flags);
+            var flags = (TypeFlags)Marshal.ReadInt32(tp, TypeOffset.tp_flags);
```
filmor, line 17: `tp_flags` is a `long`, not an `int32`.

lostmsu:
Damn Mono incorrectly marshaled
```csharp
[StructLayout(Sequential)]
class ByReference
{
  Some Fields;
}
```

No idea why. .NET Core and .NET Framework were ok on Windows and Linux. According to [Mono docs](https://www.mono-project.com/docs/advanced/pinvoke/#class-marshaling) this should have worked.

lostmsu:
P.S. I tried to isolate commits, so best to rebase or merge.

src/runtime/clrobject.cs
lostmsu: Reverted back to using `Util.ReadCLong`.
Surprisingly in `PyType_Spec` flags are of type `unsigned int`, which implies they should never exceed `2^32`.

lostmsu: Hm. Actually, because of the above we can probably use `ReadInt32`.
`tp_flags` is the only field, that is assumed to be declared as `long`. So we could remove `ReadCLong` and `WriteCLong` helpers then.

lostmsu:
@filmor can you re-review this one? I am continuing to do cleanup, and in-progress work on `TypeManager` would greatly benefit from this class.

Issue #1389, opened 2021-02-19
Officially support Python 3.9


saroad2:
### Environment

-   Pythonnet version: latest
-   Python version: python 3.9
-   Operating System: Windows
-   .NET Runtime:

### Details

-   Describe what you were trying to get done.

My project [EddingtonGUI](https://github.com/EddLabs/eddington-gui) is dependant on *Pythonnet*. I wanted for quite some time to update it to use Python 3.9 as default, but I noticed that you don't support it yet.

When are you planning on officially supporting Python 3.9? What prevents you from currently supporting it?
I'm asking so I could update my release scheduling accordingly.

Thank you!

sschukat:
The latest release 2.5 already supports Python 3.9 see [2.5.2] (https://github.com/pythonnet/pythonnet/releases/tag/v2.5.2). Version 3.0.0 is still in development.

saroad2:
Ye, but when I took a look at [PyPi](https://pypi.org/project/pythonnet/), Python3.9 is not written as supported.
I guess it's just a release error :)

filmor:
After backporting the Python 3.9 support from master to the 2.5 branch, I noticed crashes in CI (I don't remember the details right now). That's why, even though the support is nominally there, I didn't publish the respective wheels and didn't update the trove specifier. You are welcome to give debugging this a go, development happens on master, though, we haven't done any release planning for 3.0 yet.

lostmsu:
@filmor considering our current master also has crashes on 3.9, random crashes in v2.5 on Python 3.9 could be related.

We should consider 3.9 support in v2.5 as-is, e.g. support installation, but not necessarily fix all crashes.

jrobbins-LiveData:
@lostmsu is there a recipe for reproducing the current master 3.9 crashes?

Jeff17Robbins:
@lostmsu, is there a recipe for reproducing the current master 3.9 crashes?

lostmsu:
They happen in CI occasionally in both Ubuntu and MacOS with the normal shutdown mode.
Here's a sample failure: https://github.com/pythonnet/pythonnet/runs/1944796763
The CI script is this: https://github.com/pythonnet/pythonnet/blob/master/.github/workflows/main.yml

ppena-LiveData:
[...]

PR #1397, opened 2021-02-24
Allow GIL state debugging; require GILState to be properly disposed

lostmsu:

`PyGILState_Release` must always be called from the same thread, that called `PyGILState_Acquire`. Which means it is illegal to try to dispose `GILState` from .NET finalizer thread.

See the following thread https://github.com/pythonnet/pythonnet/issues/1389#issuecomment-784716437

### DIFF
--- a/src/runtime/pythonengine.cs
+++ b/src/runtime/pythonengine.cs
@@ -4,6 +4,7 @@ using System.IO;
 using System.Linq;
 using System.Reflection;
 using System.Runtime.InteropServices;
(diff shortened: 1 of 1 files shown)


filmor:
Is there any particular reason why we shouldn't just do this unconditionally?

filmor (approved):

lostmsu:
I am concerned about performance. Once we get perf tests back online, I'd compare.

PR #1400, opened 2021-02-26
ImportHook cleanup + PyObject.Length exception

lostmsu:
#### Import
- adds validation to `PyImportTest`
- before restoring original `__import__`, validates, that `__import__` was not replaced by a third party (only in `DEBUG` build)
- drops unsupported `_PyObject_GetDictPtr` in favor of `PyObject_GenericGetDict`

#### PyObject
- `Length()` method now raises an exception when object does not support concept of length (was returning 0).


filmor:
Why do you run the validation only in `DEBUG` mode?

filmor (approved):

Issue #1371, opened 2021-01-28
Domain reload crashes when .NET object points to PythonException


lostmsu:
Please see this comment: https://github.com/pythonnet/pythonnet/pull/1361#discussion_r564786550

We disabled `test_method_parameters_change`. Domain reload needs to be fixed so that holding a reference to `PythonException` is permitted.

I also suspect holding a reference to `PyObject` in this situation would break refcount based on the source code (serialization does not IncRef => original and deserialized object will both call DecRef).

lostmsu:
@benoithudson @BadSingleton can one of you pick this up?

BadSingleton:
Yup, I'll give it a go

BadSingleton:
The issue is actually that the `PythonException` class is not explicitly marked as `[Serializable]` (also needs the deserializaiton ctor)

BadSingleton:
There is also something going on with the refcount. Not sure yet how, but the AggregateException (and InnerExceptions) lives for longer than it should.

filmor:
I ran into something like this on #1373, see https://github.com/pythonnet/pythonnet/pull/1373/commits/53e51a28df8604bd016ee83f81f41d32a5f84de3. `_pyvalue` doesn't seem to be tracked properly.

filmor:
The test is enabled nowadays, I guess this one is solved.

PR #1402, opened 2021-03-01
Fix exception refcnt

BadSingleton:

Adds a tp_dealloc function to ExceptionClassObject, so we clean the BaseException members the same way as other the derived exceptions in the cpython code does.

Fixes issue #1371.

### Any other comments?

Fixes a bug where C# Exceptions set as the Inner of another C# Exception, then converted into a Python exception object (and thus having `__cause__` set to a valid object) would leak the `Inner`/`__cause__` objects.

Simplified diagram :
```
AggregateException.Inner        -> ArgumentException(...)
      ^                                    ^
      |                                    |
PyAggregateException.__cause__  -> PyArgumentException
```
In this case, where the AggregateException would be thrown from C# and caught in Python, the AggregateException and PyAggregateException would be freed, but not the ArgumentException set as the cause.

### Checklist

### DIFF
--- a/src/runtime/clrobject.cs
+++ b/src/runtime/clrobject.cs
@@ -106,7 +106,7 @@ namespace Python.Runtime
 
         public override string ToString()
         {
-            return $"<CLRObject wrapping {inst}>";
+            return $"<CLRObject: {inst}>";
         }
     }
 }
--- a/src/runtime/clrobject.cs
+++ b/src/runtime/clrobject.cs
@@ -103,5 +103,10 @@ namespace Python.Runtime
             GCHandle gc = AllocGCHandle(TrackTypes.Wrapper);
             Marshal.WriteIntPtr(pyHandle, ObjectOffset.magic(tpHandle), (IntPtr)gc);
         }
+
+        public override string ToString()
+        {
+            return $"<CLRObject wrapping {inst}>";
+        }
     }
 }
--- a/src/runtime/exceptions.cs
+++ b/src/runtime/exceptions.cs
@@ -85,6 +85,25 @@ namespace Python.Runtime
             }
             return Runtime.PyUnicode_FromString(message);
         }
+        
+        static void ClearOffsetHelper(IntPtr ob, int offset)
+        {
+            var field = Marshal.ReadIntPtr(ob, offset);
+            Runtime.Py_CLEAR(ref field);
+        }
+
+        // As seen in exceptions.c, every derived type must also clean the base.
+        // TODO: once the call tp_traverse and tp_clear on the instance and not
+        // the type, implement those.
+        public new static void tp_dealloc(IntPtr self)
+        {
+            ClearOffsetHelper(self, ExceptionOffset.dict);
+            ClearOffsetHelper(self, ExceptionOffset.args);
+            ClearOffsetHelper(self, ExceptionOffset.traceback);
+            ClearOffsetHelper(self, ExceptionOffset.context);
+            ClearOffsetHelper(self, ExceptionOffset.cause);
+            ClassBase.tp_dealloc(self);
+        }
     }
 
     /// <summary>
@@ -181,9 +200,19 @@ namespace Python.Runtime
 
             if (e.InnerException != null)
             {
-                // Note: For an AggregateException, InnerException is only the first of the InnerExceptions.
-                IntPtr cause = CLRObject.GetInstHandle(e.InnerException);
-                Marshal.WriteIntPtr(ob, ExceptionOffset.cause, cause);
+                if(e.InnerException is PythonException)
+                {
+                    // If the error is a Python exception, write the real one
+                    var pyerr =  (e.InnerException as PythonException);
+                    Runtime.XIncref(pyerr.PyValue);
+                    Runtime.PyException_SetCause(ob, pyerr.PyValue);
+                }
+                else
+                {
+                    // Note: For an AggregateException, InnerException is only the first of the InnerExceptions.
+                    IntPtr cause = CLRObject.GetInstHandle(e.InnerException);
+                    Runtime.PyException_SetCause(ob, cause);
+                }
             }
         }
 
@@ -279,10 +308,7 @@ namespace Python.Runtime
             var pe = e as PythonException;
             if (pe != null)
             {
-                Runtime.XIncref(pe.PyType);
-                Runtime.XIncref(pe.PyValue);
-                Runtime.XIncref(pe.PyTB);
-                Runtime.PyErr_Restore(pe.PyType, pe.PyValue, pe.PyTB);
+                pe.Restore();
                 return;
             }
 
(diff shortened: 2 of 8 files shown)


tests/test_exceptions.py
```diff
@@ -9,6 +9,53 @@
 import pickle
 
 
+# begin code from https://utcc.utoronto.ca/~cks/space/blog/python/GetAllObjects
```
filmor, line 12: It's not entirely clear to me how this is license-wise, did you ping the owner?

src/runtime/exceptions.cs
```diff
@@ -85,6 +85,25 @@ public new static IntPtr tp_str(IntPtr ob)
             }
             return Runtime.PyUnicode_FromString(message);
         }
+        
+        static void ClearOffsetHelper(IntPtr ob, int offset)
+        {
+            var field = Marshal.ReadIntPtr(ob, offset);
+            Runtime.Py_CLEAR(ref field);
+        }
```
lostmsu, line 93: In C# 9 it should be possible to do a `ref` return. E.g. we could have `ref T Util.Ref<T>(IntPtr ob, int offest) where T: unmanaged`, then pass the reference to `Py_CLEAR`

lostmsu: BTW, the way you are doing right now, you are not nulling the field in the original object.

src/runtime/clrobject.cs
```diff
@@ -103,5 +103,10 @@ protected override void OnLoad(InterDomainContext context)
             GCHandle gc = AllocGCHandle(TrackTypes.Wrapper);
             Marshal.WriteIntPtr(pyHandle, ObjectOffset.magic(tpHandle), (IntPtr)gc);
         }
+
+        public override string ToString()
+        {
+            return $"<CLRObject wrapping {inst}>";
```
lostmsu, line 109: NIT: "CLRObject: {inst}"

src/runtime/exceptions.cs
```diff
@@ -181,9 +200,19 @@ internal static void SetArgsAndCause(IntPtr ob)
 
             if (e.InnerException != null)
             {
-                // Note: For an AggregateException, InnerException is only the first of the InnerExceptions.
-                IntPtr cause = CLRObject.GetInstHandle(e.InnerException);
-                Marshal.WriteIntPtr(ob, ExceptionOffset.cause, cause);
+                if(e.InnerException is PythonException)
```
lostmsu, line 203: NIT: formatting. Missing space after `if`

lostmsu: NIT: you can also do `if (e.InnerException is PythonException pyerr) ...`

src/runtime/exceptions.cs
```diff
@@ -181,9 +200,19 @@ internal static void SetArgsAndCause(IntPtr ob)
...
+                {
+                    // If the error is a Python exception, write the real one
+                    var pyerr =  (e.InnerException as PythonException);
+                    Runtime.XIncref(pyerr.PyValue);
```
lostmsu, line 207: You might need to call `PythonException.Normalize`. Some python exceptions don't have `Value`

lostmsu (changes requested):
Mostly NITs, but Clear and Normalize must be rechecked.

tests/test_exceptions.py
BadSingleton: I stumbled upon this code on stack overflow and (wrongly) assumed that the poster was the author. I'll reach out to the original author.

src/runtime/exceptions.cs
filmor: As it stands, I think this is just equivalent to `Runtime.XDecref(field)`, only the local variable `field` is set to `null`.

BadSingleton: Oh right. That's *not* the cpython macro Py_CLEAR

src/runtime/exceptions.cs
BadSingleton: I'll revert that part of the code, the call to Normalize makes the python (but not the C# side) exception leak. Probably a variation of the same problem @filmor found with the pyValue of PythonException.

src/runtime/exceptions.cs
```diff
@@ -89,7 +89,8 @@ public new static IntPtr tp_str(IntPtr ob)
         static void ClearOffsetHelper(IntPtr ob, int offset)
         {
             var field = Marshal.ReadIntPtr(ob, offset);
-            Runtime.Py_CLEAR(ref field);
+            Runtime.XDecref(field);
+            Marshal.WriteIntPtr(ob, offset, IntPtr.Zero);
```
lostmsu, line 93: Actually, you must null the field before calling `DecRef`, otherwise decref might eventually call code, that will access the field and read value, that is already dead.

This is the main purpose of `Py_CLEAR` macro in the CPython codebase.

lostmsu (changes requested):

tests/test_exceptions.py
BadSingleton: I haven't found a way to contact the author, but he has a github account, so I'll try to ping him here. 

Hello, @siebenmann I've stumbled on your code through stack overflow, and I wanted to use it in this PR. What is the license to this code, if there is any?
Thanks

lostmsu:
@BadSingleton can you investigate timeout in Ubuntu Py 3.6 tests?

BadSingleton:
Huh.. that's weird. I'll take a look into it.

tests/test_exceptions.py
siebenmann: I consider the code to be in the public domain, so you can use it freely. However, it was written in 2005 and Python has changed a lot since then, so I would test it carefully before you trust it. (You may already have done this.)

BadSingleton: Thanks!
I made sure it worked for my use case at least.

BadSingleton:
I'm blocked due to #1412, this seems like an issue with garbage collection. This does raise the question of "How much do we care about Python 3.6?" There's less than a year before it's EOL: https://www.python.org/dev/peps/pep-0494/#lifespan .

mjmvisser:
@BadSingleton is this still blocked?

lostmsu:
@mjmvisser @BadSingleton AFAIK you should now be able to use a debug build of Python.

BadSingleton:
Apologies for the late reply, was on vacations. Yes this needs a rebase. I hadn't had the time to take a look at it in a while, but we're planning on resuming work on this Soon ™️

lostmsu:
@BadSingleton can you please see if the issue is still present in `master`, and if it is not, I would suggest to open a new PR with regression tests (you seem to have added a few here).

BadSingleton:
Looks like cb4bb9acf157a9122e149e41d0bfa4fe90ee955d supersedes this PR which re-enabled the test `test_method_parameters_change`. I'll see if the other test case is still applicable

BadSingleton:
Closing. Superseded by #1679

Issue #1403, opened 2021-03-01
Domain Reload test suite isn't run by default


BadSingleton:
### Environment

-   Pythonnet version: master@ f2dc8f19c1dcf3a6735d5d9bdabd83ab69121b88

### Details

-   Running `pytest` form the root of the repo doesn't run the "Domain reload test suite" (test_domain_reload). Probably broken by moving the test files in #1373.

-   What commands did you run to trigger this issue? If you can provide a
    [Minimal, Complete, and Verifiable example](http://stackoverflow.com/help/mcve)
    this will help us understand the issue.

```shell
git checkout f2dc8f19c1dcf3a6735d5d9bdabd83ab69121b88
python setup.py develop
pytest
```
Workaround: run pytest with the directory where the tests are located: `pytest src/domain_tests`

PR #1404, opened 2021-03-01
Re-enable the domain reload tests

BadSingleton:
Fixes issue #1403

### Checklist


filmor:
Sorry about that. Could you please add these as separate actions, though, instead of adding them to the normal testing? They are only relevant on Mono and .NET Framework anyhow.

BadSingleton:
I think adding them as a separate action is not optimal, as devs will have to run pytest specifying explicitly the domain tests. I added pytest filtering instead

BadSingleton:
I think the timeouts are false positive. The domain tests takes about 1:15m to execute.

PR #1413, opened 2021-03-10
Removed unnecessary dictionary initialization in CLRObject

lostmsu:
I don't see any reason for it to be present. The code does not have any associated comments, and I was not able to hit it running our tests. Any proper access to `ob_dict` should be going through `PyObject_GenericGetDict`, which initializes `ob_dict` if necessary.


lostmsu:
Superseded by https://github.com/pythonnet/pythonnet/pull/1441

Issue #1412, opened 2021-03-09
Unable to run with Python debug library


BadSingleton:
### Environment

-   Pythonnet version: master@ f2dc8f1
-   Python version: 3.6 (probably others)
-   Operating System: Ubuntu 20.04 (probably others)
-   .NET Runtime: netcore 3.1, mono

### Details

-   While debugging a build failure in #1402 , I tried to use a debug build of python,

note: Python 3.6 was installed from the deadsnakes PPA.

Repro:
-  Install Python debug packages
- Create a and use a virtual environment using the Python debug executable
- Apply/cherry-pick https://github.com/filmor/pythonnet/commit/ee3503b25b0c475fc9dee74398107fc61ca411ab to run with a static libpython
- change the `dotnet_config` to `"debug"` (line ~43 in setup.py)
- run `python -m pythonnet.find_libpython --export` and export the result in the environment, make sure it points to the debug library
- `python setup.py develop` then run `pytest` (with either `--runtime mono` or `--runtime netcore`)
- It will error out with:
```
Failed to initialize pythonnet: System.TypeInitializationException: The type initializer for 'Delegates' threw an exception. ---> System.MissingMethodException: Failed to load symbol PyModule_Create2: /home/user/dev/36-dbg/bin/python: undefined symbol: PyModule_Create2
  at Python.Runtime.Platform.LinuxLoader.GetFunction (System.IntPtr dllHandle, System.String name) [0x00050] in <12881a58dd5544c59504de515c25928b>:0
  at Python.Runtime.Runtime+Delegates.GetFunctionByName (System.String functionName, System.IntPtr libraryHandle) [0x00000] in <12881a58dd5544c59504de515c25928b>:0
  at Python.Runtime.Runtime+Delegates..cctor () [0x01793] in <12881a58dd5544c59504de515c25928b>:0
   --- End of inner exception stack trace ---
  at Python.Runtime.Runtime.PyGILState_Ensure () [0x00000] in <12881a58dd5544c59504de515c25928b>:0
  at Python.Runtime.Loader.Initialize (System.IntPtr data, System.Int32 size) [0x0003c] in <12881a58dd5544c59504de515c25928b>:0
  at Python.Runtime.Runtime.PyGILState_Ensure () [0x00000] in <12881a58dd5544c59504de515c25928b>:0
[...]

Issue #1388, opened 2021-02-18
pythonnet.load crashes if the Python interpreter is called from PATH in Ubuntu 20.04


tibortakacs:
### Environment

-   Pythonnet version: 3.0.0-dev (latest master on 18/2/2021)
-   Python version: 3.8.5
-   Operating System: Ubuntu 20.04
-   .NET Runtime: .NET 5.0.2

### Details

-   Describe what you were trying to get done.
I am trying to use the latest `pythonnet` in a virtual environment. If I call my simple demo script using `python` command in PATH, pythonnet.load() crashes. It is the same with any other PATH-based `pythonX` command. Using a full path works fine. See the details and the analysis below.

Preparation
```
# Create virtual env.
python3 -m virtualenv venv
. venv/bin/activate
# Install the latest pythonnet
git clone https://github.com/pythonnet/pythonnet.git
pip install ./pythonnet/
# Get .NET 5 runtime
wget https://download.visualstudio.microsoft.com/download/pr/824175f2-5577-46ec-a675-95d2f652575f/07dafc2398e669afa7d25f2d47398c36/dotnet-runtime-5.0.2-linux-x64.tar.gz
mkdir dotnet_5.0.2
tar xfzv dotnet-runtime-5.0.2-linux-x64.tar.gz -C dotnet_5.0.2/
```

Runtime config (`cat runtimeconfig.json`)
``` 
{
  "runtimeOptions": {
    "tfm": "net5.0",
    "framework": {
      "name": "Microsoft.NETCore.App",
      "version": "5.0.2"
    }
  }
}
```

Python script (`cat bug.py`):
```python
import os

import clr_loader
import pythonnet

runtime_config = os.path.join(os.path.dirname(__file__), "runtimeconfig.json")
dotnet_root = os.path.join(os.path.dirname(__file__), "dotnet_5.0.2")

runtime = clr_loader.get_coreclr(runtime_config=runtime_config, dotnet_root=dotnet_root)

pythonnet.set_runtime(runtime)
pythonnet.load()

import System
from System import String

s = String("Hello World from Python.NET for .NET 5!")
System.Console.WriteLine(s)
```

-   What commands did you run to trigger this issue? If you can provide a
    [Minimal, Complete, and Verifiable example](http://stackoverflow.com/help/mcve)
    this will help us understand the issue.

```
> which python
/home/takacs/src/pythonnet/venv/bin/python

>`which python` bug.py 
Hello World from Python.NET for .NET 5!
[...]

PR #1440, opened 2021-04-11
Do not call `find_libpython` during `import clr` as Python.Runtime can find it on its own

lostmsu:

`find_libpython` spectacularly fails to find libpython for debug build of Python on latest Ubuntu (e.g. `apt install python3-dbg; python3-dbg`). It is also not necessary at runtime, since `Python.Runtime` can now find C API functions, since they are already loaded into the process.

Probably these:
https://github.com/pythonnet/pythonnet/issues/1388
https://github.com/pythonnet/pythonnet/issues/1412 (partially)

### DIFF
--- a/pythonnet/__init__.py
+++ b/pythonnet/__init__.py
@@ -29,7 +29,6 @@ def load():
     if _LOADED:
         return
 
-    from .find_libpython import linked_libpython
     from os.path import join, dirname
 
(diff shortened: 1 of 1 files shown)


filmor:
Yeah, let's try it like this. I'll create a follow-up PR to make the CI use the (now adjusted) upstream `find_libpython`.

PR #1442, opened 2021-04-13
Fixed `__cause__` on overload bind failure and array conversion

lostmsu:

Calls to `PyObject_Str` and `PyObject_Repr` should not be made with error set, as they may clear it. When using debug build of Python, this causes an assertion at runtime, preventing use of Python.NET.

Partial fix for https://github.com/pythonnet/pythonnet/issues/1412

### DIFF
--- a/src/runtime/methodbinder.cs
+++ b/src/runtime/methodbinder.cs
@@ -876,7 +876,7 @@ namespace Python.Runtime
                     {
                         try
                         {
-                            var description = Runtime.PyObject_Unicode(type);
+                            var description = Runtime.PyObject_Str(type);
                             if (description != IntPtr.Zero)
                             {
                                 to.Append(Runtime.GetManagedString(description));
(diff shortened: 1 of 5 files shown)


filmor:
As there are probably a few more places like this, shouldn't we add a convenience `IDisposable` that does the fetching in startup and restores on `Dispose`?

lostmsu:
@filmor I do not think it is justified at this moment: with just these two places I don't get any more related assertion failures of that nature with debug builds of Python.

src/runtime/methodbinder.cs
```diff
@@ -926,7 +926,9 @@ internal virtual IntPtr Invoke(IntPtr inst, IntPtr args, IntPtr kw, MethodBase i
                 }
 
                 value.Append(": ");
+                Runtime.PyErr_Fetch(out var errType, out var errVal, out var errTrace);
```
BadSingleton, line 929: This might also be a fix for #1371

BadSingleton (approved):
LGTM. However this is something easy to miss. It'd be great if we had pythonnet in debug (without Py_Debug) be able to detect known cases.

src/runtime/methodbinder.cs
lostmsu: It might let us reenable the test case, but the issue that caused it will still be there.

PR #1447, opened 2021-04-21
[WIP] Add ability to create module from C#

slide:

This allows doing something like this, which I haven't found any other way to do:

```csharp
var mod = PythonEngine.AddModule("mymod");
var myobj = new object();
mod.SetAttr("myprop".ToPython(), myobj.ToPython());
```

```python
import mymod
print(mymod.myprop)
```

This is very useful for providing objects to an embedded Python engine.

I could not find any specific issues that would be related.

### Any other comments?

This is an initial PR to get feedback if this would be allowed, I have a build of the library locally with this change and use this in my code, but would like to have it upstream if possible. I will fill out the remaining details and add tests if folks think this would be ok.

### Checklist


lostmsu:
We now have https://github.com/pythonnet/pythonnet/blob/master/src/runtime/pymodule.cs
I think it would make sense to give it a static factory method `Create` instead of placing code into `PythonEngine`.

Also, you should probably use https://github.com/pythonnet/pythonnet/blob/16f04e9c0281cd1e4c266399b35ee89069eb36fe/src/runtime/runtime.cs#L1895

slide:
Cool, I didn't know about pymodule.cs, I'll definitely migrate there and use the PyModule_New

src/embed_tests/TestPyModule.cs
```diff
@@ -0,0 +1,52 @@
+
+using System;
+
+using NUnit.Framework;
+
+using Python.Runtime;
+
+namespace Python.EmbeddingTest
+{
+    class TestPyModule
```
lostmsu, line 10: I think test class must be `public`.

src/runtime/pymodule.cs
```diff
@@ -37,5 +40,40 @@ public static PyModule FromString(string name, string code)
             PythonException.ThrowIfIsNull(m);
             return new PyModule(ref m);
         }
+
+        public static PyModule Create(string name, string filename = "none")
```
lostmsu, line 44: This method violates single responsibility principle.

First, `Create` should never return an existing module. Instead, the caller should be able to check if module already exists by name by calling a separate function.

Second, I an uncertain about setting up `__builtins__` by default. It would be great to know why Python does not do it before discarding that potentially good rationale. Users can always manually do `newModule.Dict.Set("__builtins__", GetBuiltIns())`. If this is a common scenario in your product, just make a factory method combining the two.

Third, I don't think adding new module to `sys.modules` right away is a good idea. Especially, overwriting the existing value. This might cause hard to debug surprise issues.

src/runtime/pymodule.cs
```diff
@@ -37,5 +40,40 @@ public static PyModule FromString(string name, string code)
...
+            PythonException.ThrowIfIsNull(op);
+
+            // setup the module basics (__builtins__ and __file__)
+            BorrowedReference globals = Runtime.PyModule_GetDict(new BorrowedReference(op.DangerousGetAddress()));
+            PythonException.ThrowIfIsNull(globals);
+            BorrowedReference __builtins__ = Runtime.PyEval_GetBuiltins();
+            PythonException.ThrowIfIsNull(__builtins__);
+            int rc = Runtime.PyDict_SetItemString(globals, "__builtins__", __builtins__);
+            PythonException.ThrowIfIsNotZero(rc);
+            rc = Runtime.PyDict_SetItemString(globals, "__file__", new BorrowedReference(filename.ToPython().Handle));
```
lostmsu, line 69: The default value for `filename` should be `null` and when it is null `__file__` should simply not be set.

lostmsu (changes requested):

LimpingNinja:
This is a good add feature, is anyone working on it? I am migrating from C where I create a handful of accessible modules from C using Py_InitModule3() and passing in the methods (C function pointers) and vars. Is there an alternative method here or is this something that would pend this change being actioned?

slide:
I need to get back to it and make the changes that were requested. I just have a local compiled version with my changes that I have been using, so I haven't had a huge push to get me back to fixing this up. I'll try and work on it soon.

src/runtime/pymodule.cs
slide: > Second, I am uncertain about setting up `__builtins__` by default. It would be great to know why Python does not do it before discarding that potentially good rationale. Users can always manually do `newModule.Dict.Set("__builtins__", GetBuiltIns())`. If this is a common scenario in your product, just make a factory method combining the two.
> 

I didn't see a public method that would allow access to the __builtins__. There is an internal method, on `Runtime`, but not a public method that could be used outside.

slide:
@lostmsu I think I addressed the feedback you provided, please let me know if there are more changes you would like me to make. If not, I will make this a normal PR and remove the WIP.

src/runtime/pymodule.cs
```diff
@@ -29,6 +32,13 @@ public PyModule Reload()
             return new PyModule(ref op);
         }
 
+        public void AddToSysModules()
+        {
+            BorrowedReference modules = Runtime.PyImport_GetModuleDict();
+            int rc = Runtime.PyDict_SetItemString(modules, this.Name, this.Reference);
+            PythonException.ThrowIfIsNotZero(rc);
+        }
+
```
lostmsu, line 41: Instead of adding potentially even more methods, that have to do with `sys.modules` in `PyModule`, expose a static member `PyModule.SysModules`, and let users do checking and adding on their own.

It should replace `AddToSysModules` and `IsInSysModules`.

src/runtime/pymodule.cs
```diff
@@ -37,5 +47,39 @@ public static PyModule FromString(string name, string code)
...
+        public static PyModule Create(string name, string filename=null)
+        {
+            NewReference op = Runtime.PyModule_New(name);
+            PythonException.ThrowIfIsNull(op);
+
+            if (filename != null)
+            {
+                BorrowedReference globals = Runtime.PyModule_GetDict(new BorrowedReference(op.DangerousGetAddress()));
```
lostmsu, line 66: should be just `PyModule_GetDict(op)`

src/runtime/pymodule.cs
```diff
@@ -37,5 +47,39 @@ public static PyModule FromString(string name, string code)
...
+                PythonException.ThrowIfIsNull(globals);
+                int rc = Runtime.PyDict_SetItemString(globals, "__file__", new BorrowedReference(filename.ToPython().Handle));
```
lostmsu, line 68: should be just `filename.ToPython().Reference`

src/runtime/pymodule.cs
```diff
@@ -37,5 +47,39 @@ public static PyModule FromString(string name, string code)
...
+                PythonException.ThrowIfIsNotZero(rc);
+            }
+
+            return new PyModule(ref op);
+        }
+
+        public void AddBuiltins()
```
lostmsu, line 75: Can you split this into `void SetBuiltins(PyModule builtins)` and `PyModule.Builtins` (or maybe `Runtime.Builtins`)

src/runtime/pymodule.cs
```diff
@@ -37,5 +47,39 @@ public static PyModule FromString(string name, string code)
...
+        {
+            // first check if there is an existing module
+            BorrowedReference modules = Runtime.PyImport_GetModuleDict();
+            BorrowedReference m = Runtime.PyDict_GetItemString(modules, name);
+            return !m.IsNull && Runtime.PyObject_TypeCheck(m, new BorrowedReference(Runtime.PyModuleType));
+        }
+
+        public static PyModule Create(string name, string filename=null)
+        {
+            NewReference op = Runtime.PyModule_New(name);
```
lostmsu, line 61: Add a null check for the `name` parameter.

lostmsu (changes requested):

src/runtime/pymodule.cs
```diff
@@ -37,5 +47,39 @@ public static PyModule FromString(string name, string code)
...
+            if (filename != null)
+            {
+                BorrowedReference globals = Runtime.PyModule_GetDict(new BorrowedReference(op.DangerousGetAddress()));
+                PythonException.ThrowIfIsNull(globals);
+                int rc = Runtime.PyDict_SetItemString(globals, "__file__", new BorrowedReference(filename.ToPython().Handle));
+                PythonException.ThrowIfIsNotZero(rc);
+            }
+
+            return new PyModule(ref op);
+        }
```
lostmsu, line 73: I suggest to make `Create` method private, and add a constructor instead. E.g.
`public PyModule(string name, string filename = null): this(Create(name, filename)) { }`

src/runtime/pymodule.cs
slide: BorrowedReference is an internal struct, what is the best thing to return then since this would be a public API?

lostmsu: For `SysModules`? `PyDict` - it is a dictionary.

slide: Sorry, missed the PyDict in your original comment.

slide:
Feedback hopefully addressed. Thanks for being patient. I still don't know the codebase very well.

lostmsu (approved):

PR #1458, opened 2021-05-21
Handle ProcessExit event

lostmsu:
This avoids segfault when CLR (in particular Mono) is unloaded before Python stops.

1. Added a handler to `AppDomain.ProcessExit`, that fires in the situations like this instead of `AppDomain.DomainUnload`. This allows Python.NET to remove all slots, that are implemented in C#.
2. When initialized as extension (e.g.  Python hosts CLR), release GIL acquired during shutdown so that Python finalization can resume.

This is (at least one of) the reason of CI test crashes in https://github.com/pythonnet/pythonnet/pull/1134

### DIFF
--- a/src/runtime/runtime.cs
+++ b/src/runtime/runtime.cs
@@ -393,6 +393,10 @@ namespace Python.Runtime
                 {
                     Py_Finalize();
                 }
+                else
+                {
+                    PyGILState_Release(state);
+                }
             }
         }
 
--- a/src/runtime/pythonengine.cs
+++ b/src/runtime/pythonengine.cs
@@ -192,6 +192,7 @@ namespace Python.Runtime
 
             // Make sure we clean up properly on app domain unload.
             AppDomain.CurrentDomain.DomainUnload += OnDomainUnload;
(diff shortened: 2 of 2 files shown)


src/runtime/runtime.cs
```diff
@@ -393,6 +393,10 @@ internal static void Shutdown(ShutdownMode mode)
                 {
                     Py_Finalize();
                 }
+                else
+                {
+                    PyGILState_Release(state);
```
filmor, line 398: This change seems unrelated, why did you add this?

lostmsu: Without this `Py_Finalize` called when Python finally shuts down hangs because it can't acquire GIL.

lostmsu:
I debugged the test failures, and they appear to be caused by gaps in domain reload support: when .NET objects that Python sees get severed from CLR, subsequent calls to their `tp_dealloc` fail because `GetManagedObject` returns `null`.

lostmsu:
Found a way to solve the same problem and more without handling `ProcessExit`

lostmsu:
Or maybe not. Seems to still be essential for https://github.com/pythonnet/pythonnet/pull/1134

PR #1474, opened 2021-06-21
Use Borrowed/NewReference in a few more places and get rid of some warnings

filmor:

...

...

### Any other comments?

...

### Checklist

### DIFF
--- a/src/runtime/extensiontype.cs
+++ b/src/runtime/extensiontype.cs
@@ -76,10 +76,10 @@ namespace Python.Runtime
         /// <summary>
         /// Type __setattr__ implementation.
         /// </summary>
-        public static int tp_setattro(IntPtr ob, IntPtr key, IntPtr val)
+        public static int tp_setattro(BorrowedReference ob, BorrowedReference key, BorrowedReference val)
         {
             var message = "type does not support setting attributes";
-            if (val == IntPtr.Zero)
+            if (val.IsNull)
             {
                 message = "readonly attribute";
             }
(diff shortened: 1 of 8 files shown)


src/runtime/extensiontype.cs
```diff
@@ -76,10 +76,10 @@ protected virtual void Clear()
         /// <summary>
         /// Type __setattr__ implementation.
         /// </summary>
-        public static int tp_setattro(IntPtr ob, IntPtr key, IntPtr val)
+        public static int tp_setattro(BorrowedReference ob, BorrowedReference key, BorrowedReference val)
```
lostmsu, line 79: Neat, I did not know this would work

src/runtime/managedtype.cs
```diff
@@ -102,7 +102,7 @@ internal void FreeGCHandle()
                 if ((flags & TypeFlags.HasClrInstance) != 0)
                 {
                     var gc = TryGetGCHandle(new BorrowedReference(ob));
-                    return (ManagedType)gc?.Target;
+                    return gc?.Target as ManagedType;
```
lostmsu, line 105: I think this is wrong. If gc handle has `Target`, it must be a `ManagedType` instance here. If it is not, this would violate a strong assumption.

src/runtime/extensiontype.cs
lostmsu: Dang. Judging by failures it did not. You'd need to construct a delegate type with the right signature at runtime using [Expression.GetFuncType](https://docs.microsoft.com/en-us/dotnet/api/system.linq.expressions.expression.getfunctype?redirectedfrom=MSDN&view=net-5.0#System_Linq_Expressions_Expression_GetFuncType_System_Type___)

BTW, in the future, we might make this faster by using [`[UnmanagedCallersOnly]`](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.unmanagedcallersonlyattribute) (.NET 5+)

lostmsu (changes requested):
Minor tweaks needed.

Love PRs that reduce total line count.

PR #1509, opened 2021-08-05
Names of .NET types changed to better support generic types

lostmsu:

Previously, Python.NET would produce confusing type names in certain scenarios (generic and/or nesting types).

### Checklist

### DIFF
--- a/src/runtime/typemanager.cs
+++ b/src/runtime/typemanager.cs
@@ -211,18 +211,7 @@ namespace Python.Runtime
 
         static PyType CreateClass(Type clrType)
         {
-            // Cleanup the type name to get rid of funny nested type names.
-            string name = $"clr.{clrType.FullName}";
-            int i = name.LastIndexOf('+');
-            if (i > -1)
-            {
-                name = name.Substring(i + 1);
-            }
-            i = name.LastIndexOf('.');
-            if (i > -1)
-            {
-                name = name.Substring(i + 1);
-            }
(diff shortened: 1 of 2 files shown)


src/runtime/typemanager.cs
```diff
@@ -251,6 +240,65 @@ static PyType CreateClass(Type clrType)
...
+            }
+
+            string name = CleanupFullName(clrType.FullName);
+            target.Append(name);
+        }
+
+        static string CleanupFullName(string fullTypeName)
+        {
+            // Cleanup the type name to get rid of funny nested type names.
+            string name = "CLR." + fullTypeName;
```
filmor, line 286: Maybe `clr.`?

src/runtime/typemanager.cs
```diff
@@ -251,6 +240,65 @@ static PyType CreateClass(Type clrType)
...
+                string fullName = clrType.GetGenericTypeDefinition().FullName;
+                int argCountIndex = fullName.LastIndexOf('`');
+                if (argCountIndex >= 0)
+                {
+                    string nonGenericFullName = fullName.Substring(0, argCountIndex);
+                    string nonGenericName = CleanupFullName(nonGenericFullName);
+                    target.Append(nonGenericName);
+
+                    var arguments = clrType.GetGenericArguments();
+                    target.Append('<');
```
filmor, line 263: Could it make sense to use the "Python syntax" instead, i.e. square brackets?

PR #1524, opened 2021-08-21
Fix #1523

### LINKED ISSUE #1523: Parameter resolution for integer types doesn't work in some cases
### Environment

-   Pythonnet version: master
-   Python version: 3.7
-   .NET Runtime: .NET 4.8

### Details

  Calling .NET methods from Python

-   What commands did you run to trigger this issue? 

Given a class like the following, any value you pass for address to the first MethodA will cause the exception underneath, it looks like because when needsResolution is true, the Converter will see the the value as an Int32 in GetTypeByAlias since in 3.7 there is no distinction between int and long anymore.

```csharp
public class MethodResolutionInt
    {
        public IEnumerable<byte> MethodA(ulong address, int size)
        {
            return new byte[10];
        }

        public int MethodA(string dummy, ulong address, int size)
        {
            return 0;
        }
    }
```

```python
mri = MethodResolutionInt()
data = list(mri.MethodA(0x1000, 10))
assert len(data) == 10
assert data[0] == 0

data = list(mri.MethodA(0x100000000, 10))
assert len(data) == 10
assert data[0] == 0
```

```
Python.Runtime.PythonException: Expected UInt64, got Int32

The above exception was the direct cause of the following exception:
Python.Runtime.PythonException: Expected UInt64, got Int32

The above exception was the direct cause of the following exception:

System.ArgumentException: Expected UInt64, got Int32 in method System.Collections.Generic.IEnumerable`1[System.Byte] MethodA(UInt64, Int32) ---> Python.Runtime.PythonException: Expected UInt64, got Int32
   --- End of inner exception stack trace ---

The above exception was the direct cause of the following exception:
Python.Runtime.PythonException: Expected UInt64, got Int32

The above exception was the direct cause of the following exception:

System.ArgumentException: Expected UInt64, got Int32 in method System.Collections.Generic.IEnumerable`1[System.Byte] MethodA(UInt64, Int32) ---> Python.Runtime.PythonException: Expected UInt64, got Int32
   --- End of inner exception stack trace ---

The above exception was the direct cause of the following exception:

System.AggregateException: One or more errors occurred. ---> System.ArgumentException: Expected UInt64, got Int32 in method System.Collections.Generic.IEnumerable`1[System.Byte] MethodA(UInt64, Int32) ---> Python.Runtime.PythonException: Expected UInt64, got Int32
   --- End of inner exception stack trace ---
   --- End of inner exception stack trace ---
---> (Inner Exception #0) System.ArgumentException: Expected UInt64, got Int32 in method System.Collections.Generic.IEnumerable`1[System.Byte] MethodA(UInt64, Int32) ---> Python.Runtime.PythonException: Expected UInt64, got Int32
   --- End of inner exception stack trace ---<---
```

I have a PR almost ready with A solution to the issue, but it may not be the correct solution.

slide:

Adds a preferred type parameter to GetTypeByAlias and then the preferred type gets passed in in methodbinder's TryComputeClrArgumentType. I am not sure if this is the best fix for the issue, I know that the test fails before the fix and passes after.

#1523

### Checklist

### DIFF
--- a/src/runtime/converter.cs
+++ b/src/runtime/converter.cs
@@ -29,6 +29,8 @@ namespace Python.Runtime
         private static Type int64Type;
         private static Type boolType;
         private static Type typeType;
+        private static Type uint64Type;
+        private static Type uint32Type;
 
         static Converter()
         {
@@ -43,13 +45,15 @@ namespace Python.Runtime
             decimalType = typeof(Decimal);
             boolType = typeof(Boolean);
             typeType = typeof(Type);
+            uint64Type = typeof(UInt64);
+            uint32Type = typeof(UInt32);
         }
 
 
         /// <summary>
         /// Given a builtin Python type, return the corresponding CLR type.
         /// </summary>
-        internal static Type GetTypeByAlias(IntPtr op)
+        internal static Type GetTypeByAlias(IntPtr op, Type preferred=null)
         {
             if (op == Runtime.PyStringType)
                 return stringType;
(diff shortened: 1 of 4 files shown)


lostmsu:
This fails some tests.
Have you explored what happens when you remove `needsResolution` altogether?

slide:
It looks like master is also failing the tests? I may be wrong since I only looked on my phone. I have not tried removing needsResolution, I can try that.

lostmsu:
@slide master fails pretty rarely, and only in 1 configuration, probably due to heap corruption. The failure here is directly related to the change.

slide:
Ok, I'll definitely take a look. I need to stop using my phone for looking at github

slide:
Ok, I can see the failures much easier on my computer. It looks like a different overload is being chosen for the <= operator. I'll track down the issue and see why it is happening.

slide:
I don't know if the solution I proposed is a good one for all circumstances. Another way of fixing the tests would be to change the overload that takes a PyObject to not assume that it is a tuple.

Issue #1476, opened 2021-06-23
Memory leak when throwing exceptions in .NET and catching in Python


Rookfighter:
### Environment

-   Pythonnet version: 2.5.2
-   Python version: issue verified on 3.7.2, 3.7.7 and 3.8.6
-   Operating System: Windows 10 64bit
-   .NET Runtime: .NET Framework 4.7.2

### Details

Hello everybody and thanks for this great piece of Software!

The basic issue is:
* if I throw a .NET exception from within C# code
* and the exception moves into the python space
  * i.e. the C# method was called from python
* then the memory of the exception does not get released anymore

This becomes apparent when you throw quite a lot of exceptions and your applications runs 24/7.

#### Reproducability

The issue is fairly easy to reproduce. I have also setup a sample repository, which reproduces the issue.

https://github.com/Rookfighter/pythonnet-memleak-exceptions

The basis instruction are:

* setup a Visual Studio Project of the type "Class Library .NET Framework 4.7.2"
* create a `ExceptionThrower` class

```cs
using System;

namespace LeakingExceptions
{
    public class ExceptionThrower
    {
        public void ThrowException()
        {
            throw new Exception("a simple exception");
        }
    }
}
```

* create a python script which imports the assembly
* create a `ExceptionThrower` object and call `ThrowException` within a looped `try except` block

```python
import os
import __main__

import clr

# bin dir must be adjusted to your relative path!
bin_dir = os.path.abspath(os.path.join(os.path.dirname(__main__.__file__), '../Source/LeakingExceptions/bin/Debug'))
clr.AddReference(os.path.join(bin_dir, 'LeakingExceptions'))

from LeakingExceptions import ExceptionThrower

thrower = ExceptionThrower()

while True:
    try:
        thrower.ThrowException()
    except:
        pass
```

* in the task manager you can then watch the application to accumulate memory and never release it

![memleak_taskmgr](https://user-images.githubusercontent.com/2543949/123053735-7dc5ba00-d404-11eb-95a2-9d9fc8ea1c63.png)

PR #1541, opened 2021-09-05
Call tp_clear of base unmanaged type

lostmsu:

Calls `tp_clear` (if any) of base unmanaged type when a reflected CLR object instance is cleared.

https://github.com/pythonnet/pythonnet/issues/1476

### Any other comments?

C# exceptions inherit from Python `BaseException` class, which has `tp_clear`. Prior to this change it was not called when Python deallocated them leaving dangling references to `BaseException.args` and `BaseException.__cause__`.

### Checklist

### DIFF
--- a/src/embed_tests/Inheritance.cs
+++ b/src/embed_tests/Inheritance.cs
@@ -1,7 +1,5 @@
 using System;
 using System.Collections.Generic;
-using System.Diagnostics;
-using System.Runtime.InteropServices;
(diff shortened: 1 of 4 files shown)


lostmsu:
@amos402 can you review this one?

The 3.7 MacOS CI failure is preexisting: https://github.com/pythonnet/pythonnet/runs/3267749344

PR #1548, opened 2021-09-07
Allow to specificy full path in AddReference

MarioBeretta:
### What does this implement/fix? Explain your changes.
If you specify a full path to an assembly in the *clr.AddReference* it fails with a FileLoadException
...

### Does this close any currently open issues?

...

### Any other comments?

...

### Checklist

Check all those that are applicable and complete.

-   [ ] Make sure to include one or more tests for your change
-   [ ] If an enhancement PR, please create docs and at best an example
-   [ ] Add yourself to [`AUTHORS`](../blob/master/AUTHORS.md)
-   [ ] Updated the [`CHANGELOG`](../blob/master/CHANGELOG.md)

PR #1549, opened 2021-09-08
Allow to specify a full path in clr.AddReference

MarioBeretta:

Allow to specify a full path in clr.AddReference.

...

### Any other comments?

...

### Checklist

### DIFF
--- a/src/runtime/assemblymanager.cs
+++ b/src/runtime/assemblymanager.cs
@@ -206,6 +206,10 @@ namespace Python.Runtime
             {
                 return null;
             }
+            catch(FileLoadException)
+            {
+                return null;
+            }
         }
 
 
--- a/src/runtime/assemblymanager.cs
+++ b/src/runtime/assemblymanager.cs
@@ -206,10 +206,8 @@ namespace Python.Runtime
             {
                 return null;
             }
-            catch(FileLoadException)
-            {
-                return null;
-            }
+            
+
         }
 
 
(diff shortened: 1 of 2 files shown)


filmor:
This is essentially the same thing as #1548. This doesn't allow you to load anything as far as I can see, it just hides one particular error.

MarioBeretta:
Given that the AddReference already have a retry mechanism based on the value of the variable assembly==null,
catching the FileLoadException will ensure that we try to load the Assembly using the LoadFrom (called inside LoadAssemblyFullPath), that doesn't have the same problem of the Assembly.load 


```
public static Assembly AddReference(string name)
        {
            AssemblyManager.UpdatePath();
            var origNs = AssemblyManager.GetNamespaces();
            Assembly assembly = null;
            assembly = AssemblyManager.FindLoadedAssembly(name);
            if (assembly == null)
            {
                assembly = AssemblyManager.LoadAssemblyPath(name);
            }
            if (assembly == null)
            {
                try
                {
                    assembly = AssemblyManager.LoadAssembly(name);
                }
                // The Assembly.Load fail with FileLoadException in case name contains a path
                catch(FileLoadException)
                {
                    assembly=null;
                }
            }
            if (assembly == null)
            {
                assembly = AssemblyManager.LoadAssemblyFullPath(name);
            }
            if (assembly == null)
            {
                throw new FileNotFoundException($"Unable to find assembly '{name}'.");
            }

```

lostmsu:
Same comment as for the previous PR: you need to distinguish cases when file exists and fails to load for some reason (e.g. text file was specified instead of a .NET DLL) and when there's no file in the first place, in which case the original exception has to be preserved.

MarioBeretta:
Thanks I got you point, I will push a new proposed solution

PR #1557, opened 2021-09-15
Cleaning up public API

lostmsu:
This handles https://github.com/pythonnet/pythonnet/issues/1539

Deleted:
- `PyLong`
- `PyAnsiString`
- a few internal overloads to convert between Python integer type and .NET int types

### Checklist


src/embed_tests/TestRuntime.cs
```diff
@@ -91,6 +97,8 @@ public static void PyCheck_Iter_PyObject_IsIterable_ThreadingLock_Test()
         {
             Runtime.Runtime.Py_Initialize();
 
+            Runtime.Native.ABI.Initialize(Runtime.Runtime.PyVersion);
```
filmor, line 100: What is this for?

lostmsu: `PyIter_Check` down in the function is not something we import from Python DLL. In Python it is a macro, that we reimplemented in C#. That implementation depends on `TypeOffset` to be initialized, which is handled by `ABI.Initialize`

lostmsu:
@filmor any other comments?

src/runtime/delegatemanager.cs
```diff
@@ -76,7 +76,7 @@ private Type GetDispatcher(Type dtype)
             var cc = CallingConventions.Standard;
             Type[] args = { ptrtype, typetype };
             ConstructorBuilder cb = tb.DefineConstructor(ma, cc, args);
-            ConstructorInfo ci = basetype.GetConstructor(args);
+            ConstructorInfo ci = basetype.GetConstructor(BindingFlags.Instance | BindingFlags.NonPublic, null, args, null);
```
filmor, line 79: Does this change behaviour?

lostmsu: Users can no longer construct their own instances of `Dispatcher` manually (which is the goal of the change anyway).

I doubt anyone did that anyway, it is very obscure.

filmor (approved):

PR #1568, opened 2021-09-23
Disable implicit conversions that might lose information

lostmsu:

This removes the following conversions:
`PyInt` -> `int32` (when trying to convert to `object` only)
.NET arrays and collections -> Python `list` (due to mutability)

Additionally, `PyObject` no longer implements `IEnumerable<PyObject>`, as not all python objects are iterable. Instead, a new class `PyIterable` is added.

### Checklist

### DIFF
--- a/src/runtime/converter.cs
+++ b/src/runtime/converter.cs
@@ -1,8 +1,9 @@
+#nullable enable
 using System;
 using System.Collections;
 using System.Collections.Generic;
-using System.ComponentModel;
 using System.Globalization;
+using System.Reflection;
 using System.Runtime.InteropServices;
 using System.Security;
 
@@ -18,12 +19,10 @@ namespace Python.Runtime
         {
         }
 
-        private static NumberFormatInfo nfi;
         private static Type objectType;
         private static Type stringType;
         private static Type singleType;
         private static Type doubleType;
-        private static Type decimalType;
         private static Type int16Type;
         private static Type int32Type;
         private static Type int64Type;
@@ -32,7 +31,6 @@ namespace Python.Runtime
 
         static Converter()
         {
-            nfi = NumberFormatInfo.InvariantInfo;
             objectType = typeof(Object);
             stringType = typeof(String);
             int16Type = typeof(Int16);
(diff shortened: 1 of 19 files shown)


tests/test_conversion.py
```diff
@@ -413,16 +413,10 @@ def test_single_conversion():
     assert ob.SingleField == 0.0
 
     ob.SingleField = 3.402823e38
-    assert ob.SingleField == 3.402823e38
+    assert ob.SingleField == System.Single(3.402823e38)
```
filmor, line 416: When comparing to the `float` literal instead, does this fail with an exception or just return `False`?

filmor (approved):

tests/test_conversion.py
lostmsu: There's no exception. The reason you have to explicitly cast the literal to `float32` is that `(double)3.402823e38 != (float)3.402823e38`. This was previously hacked around in `Converter` via gross conversion to string representation and back, and I removed that hack.

src/runtime/converter.cs
```diff
@@ -541,6 +495,30 @@ internal static IntPtr ToPythonImplicit(object value)
...
+                {
+                    Exceptions.SetError(ex.InnerException);
+                }
+                return null;
+            }
+            catch (SecurityException ex)
+            {
+                if (setError)
+                {
+                    Exceptions.SetError(ex.InnerException);
```
lostmsu, line 516: Should be just `SetError(ex)`

src/runtime/converter.cs
```diff
@@ -390,19 +352,15 @@ internal static IntPtr ToPythonImplicit(object value)
...
                     {
                         Exceptions.SetError(Exceptions.TypeError, cb.type.DeletedMessage);
                         return false;
                     }
                     result = cb.type.Value;
                     return true;
-                }
-                // shouldn't happen
-                return false;
             }
```
lostmsu, line 364: Fallback branch should be added.

lostmsu (commented):
A few minor isssues

Issue #1390, opened 2021-02-19
Proposal: remove PyScopeManager


lostmsu:
Purpose of `PyScopeManager` within `Python.Runtime` is unclear. It is not used internally in the library, and users can re-implement it if necessary.

lostmsu:
`PyScopeManager` has been removed

PR #1569, opened 2021-09-23
`PyScope`/`PyModule` API cleanup

lostmsu:
- removed `PyScopeManager`
- merged `PyScope` into `PyModule` (every `PyScope` was a Python module anyway)
- minor behavioral changes

This is part of [3.0 API cleanup](https://github.com/pythonnet/pythonnet/issues/1539), targeting https://github.com/pythonnet/pythonnet/issues/1390

### Checklist


lostmsu:
@filmor can you take a look at this one at least in principle?

src/runtime/module.cs
```diff
@@ -101,26 +151,15 @@ public PyScope NewScope()
         /// Import a scope or a module of given name,
         /// scope will be looked up first.
```
filmor, line 152: This part of the comment is not true anymore.

filmor:
I'm okay with this one, though for a slight backwards-compatibility I'd propose we add `[Obsolete]public class PyScope : PyModule` with both public constructors copied. Is there any particular reason why you renamed the file from `pymodule` to `module`? Matching the filename makes sense to me.

filmor (approved):

lostmsu:
@filmor re: filename. Without renaming the file, diff is larger (that's because the file has more code from `PyScope` than from `PyModule`)

lostmsu:
@filmor re: `PyScope`. The original `PyScope` did not have any public constructors.

filmor:
It had public static functions, though (and that's also how I always used it). So you can probably ignore the constructors, but deriving still makes sense.

lostmsu:
@filmor I don't see any public static functions on old `PyScope`. There was a single factory method `Py.Scope()` and I preserved that one.

Issue #533, opened 2017-08-30
assemblymanager.cs: ResolveHandler finds assemblies based on StartsWith


rmadsen-ks:
### Environment

-   Pythonnet version: All
-   Python version: All
-   Operating System: All

### Details

-   Describe what you were trying to get done.
Loading an assembly called MyAssembly. Assembly did not exist, should have resulted in an error, but instead it loaded MyAssemblyTest.
This can be done using Assembly.Load("AssemblyName").

Effectively this changes the way .NET normally loads assemblies and it ignores things such as strong naming. This could cause all kinds of unexpected effects in larger systems. 

```cs
 static Assembly ResolveHandler(Object ob, ResolveEventArgs args){
            string name = args.Name.ToLower();
            for (int i = 0; i < assemblies.Count; i++) {
                Assembly a = (Assembly)assemblies[i];
                string full = a.FullName.ToLower();
                if (full.StartsWith(name)) { // this is a problem!
                    return a;
                }
            }
            return LoadAssemblyPath(args.Name);
        }
```

The FindAssembly method has obvious issues as well, but it only searches inside the python path, so the effect there is limited. The ResolveHandler searches through all currently loaded assemblies, which is a bigger problem.

lostmsu:
This was fixed

PR #1570, opened 2021-09-24
Simplify assembly `ResolveHandler`, and use official assembly name parsing

lostmsu:

https://github.com/pythonnet/pythonnet/issues/533


src/runtime/importhook.cs
```diff
@@ -37,6 +36,11 @@ class DotNetFinder(importlib.abc.MetaPathFinder):
         if 'clr' not in sys.modules:
             return None
         clr = sys.modules['clr']
+
+        _add_pending = clr._add_pending_namespaces
+        if _add_pending:
```
filmor, line 41: Why are you checking it instead of calling it directly?

filmor (approved):

src/runtime/importhook.cs
lostmsu: I wasn't sure if there are no scenarios where clr module won't have its members loaded yet, but would already be present in `sys.modules`. The uncertainty was based on the lines that immediately follow this one. I guess I could skip the check, and reintroduce it if some scenario is reported.

PR #1603, opened 2021-10-19
Use new references

lostmsu:
TBA


lostmsu:
Some work is also required on the Roslyn analyzer, that ensures that we don't copy `NewReference` instances, because it currently reports this scenario as error:

```csharp
var obj = new NewReference(...);
PyObject_Str(obj.Borrow());
return obj;
```

where `obj` is never copied, instead its value is "moved" to the caller.

In this PR the issue is bypassed using a `.AnalyzerWorkaround()` method in about 5 places.

filmor:
Should we just fork the analyzer into this codebase?

lostmsu:
@filmor I have a fork already under lostmsu

PR #1615, opened 2021-11-13
Include NonCopyableAnalyzer

filmor:
- Include NonCopyableAnalyzer
- Fix analyzer for array creation handling and deactivate return handling

...

...

### Any other comments?

...

### Checklist


lostmsu:
I don't think including analyzer in pythonnet is a good idea.

Also, simply disabling return analysis is not correct. It needs to be counted as a "move".

filmor:
Well, we have to do something :)

1. Somehow your fork's nuget is not visible on Nuget.org, I couldn't find the sources yesterday either
2. The last version published (0.7.0-m04?) doesn't have the fix for array creation, so it throws warnings without end for me
3. Do you intend to upsteam your changes? Do you use the analyzer somewhere else?

lostmsu:
Yeah, it is in my personal repository, dev branch: https://github.com/lostmsu/NonCopyableAnalyzer/commits/dev

I guess upstreaming is an option. It just took extra time to upstream.

filmor:
The reason why I'd like to include it here is that I don't actually want to maintain it more than necessary to keep it working within this project (as in: I (myself) don't want to publish a nuget package and react on other people using it). If you want to do that, that's fine by me, but please publish an updated version then (with the array creation fix).

lostmsu:
@filmor I was actually planning to upstream all the changes in that package. We are only using a fork because I did not get to doing it last time. My plan is (once I get to it):

1. make the changes in my fork
2. publish an updated personal version of NuGet
3. consume it from this repository
4. upstream changes to the original `NonCopyableAnalyzer` repo
5. once upstream publishes new version, switch this repo to it

PR #1620, opened 2021-11-23
Use .NET 6.0 LTS and C# 10

lostmsu:
.NET 5 that we use expires in May 2022. For the new version we should use LTS if possible.

Also, new goodies in C#!

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -1,15 +1,15 @@
 <Project>
   <PropertyGroup>
     <VersionPrefix>3.0.0</VersionPrefix>
-    <AssemblyCopyright>Copyright (c) 2006-2020 The Contributors of the Python.NET Project</AssemblyCopyright>
+    <AssemblyCopyright>Copyright (c) 2006-2021 The Contributors of the Python.NET Project</AssemblyCopyright>
     <AssemblyCompany>pythonnet</AssemblyCompany>
     <AssemblyProduct>Python.NET</AssemblyProduct>
-    <LangVersion>9.0</LangVersion>
+    <LangVersion>10.0</LangVersion>
     <IsPackable>false</IsPackable>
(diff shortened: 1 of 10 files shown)


Directory.Build.props
```diff
@@ -1,15 +1,15 @@
 <Project>
   <PropertyGroup>
     <VersionPrefix>3.0.0</VersionPrefix>
-    <AssemblyCopyright>Copyright (c) 2006-2020 The Contributors of the Python.NET Project</AssemblyCopyright>
+    <AssemblyCopyright>Copyright (c) 2006-2022 The Contributors of the Python.NET Project</AssemblyCopyright>
```
filmor, line 4: Bold move ;)

pythonnet.sln
```diff
@@ -34,6 +34,13 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Solution Items", "Solution
 		setup.py = setup.py
 	EndProjectSection
 EndProject
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "conda.recipe", "conda.recipe", "{7FD2404D-0CE8-4645-8DFB-766470E2150E}"
```
filmor, line 37: We don't maintain a conda recipe anymore.

Directory.Build.props
lostmsu: No worries, I committed this using my DeLorean. :-D

pythonnet.sln
lostmsu: Removed. Looks like some VS glitch with solution file on branch change.

Directory.Build.props
filmor: Just fix this one, please and you can go ahead and merge :)

PR #1621, opened 2021-11-23
Fix enum codec

filmor:
A boxed enum value can't be casted directly to an integer, but using `System.Convert`
functions instead works fine.

...

...

### Any other comments?

...

### Checklist


lostmsu:
Is there a test for `UInt64.MaxValue`? Removal of `try ... catch ...` looks suspicious.

filmor:
Apparently there were no tests at all for this :)

I'll look into creating some.

filmor:
After finally understanding what you meant, it works now ;)

The old codec didn't handle `ulong` either, that's why I didn't consider it.

Issue #1582, opened 2021-10-02
Random crash on engine shutdown: 'clr' has no attribute '_add_pending_namespaces'


lostmsu:
### Environment

-   Pythonnet version: dee41b8
-   Python version: Any
-   Operating System: Windows x86
-   .NET Runtime: .NET Framework

### Details

  Error Message:
   Python.Runtime.PythonException : partially initialized module 'clr' has no attribute '_add_pending_namespaces' (most likely due to a circular import)

PR #1623, opened 2021-11-29
Try to detect whether the clr module is partially initialised

filmor:
This should fix issue #1582. While the respective behaviour (setting the member value `_initializing`) is not documented, it has been that way since at least 3.4. As we control `clr`, it can also not happen, that the value is accidentally set.

...

...

### Any other comments?

...

### Checklist


lostmsu:
Do you know how to reproduce the original issue? Can you confirm you tried with and without the fix, and it at least seems not to repro with the fix?

filmor:
Nope. I just read the Python source code, and it's clear how exactly this can happen. There is however no public interface to check for this situation, so apart from this hacky version, we are out of luck anyhow.

lostmsu:
I don't see any code that would set that `_initializing` property. Is it set implicitly somehow?

filmor:
These are the relevant parts of CPython:

https://github.com/python/cpython/blob/f4c03484da59049eb62a9bf7777b963e2267d187/Lib/importlib/_bootstrap.py#L678-L707

https://github.com/python/cpython/blob/f4c03484da59049eb62a9bf7777b963e2267d187/Objects/moduleobject.c#L722-L741

The latter is exactly the part that's `False` in the error case that you have observed.

lostmsu:
@filmor do you have a good theory why this could happen during shutdown then? The original bug speaks about shutdown scenario.

filmor:
It must be that the module is gone and some code (the loader?) tries to reinitialise it, races with the initialisation and hits this bump.

filmor:
@lostmsu What do you think about this PR?

lostmsu:
@filmor I am not convinced unless we have a repro.

PR #1638, opened 2021-12-20
Removed `ShutdownMode`. Now always behaves like original `Reload`

lostmsu:

This removes all shutdown modes, except `Reload` (which you don't need to specify, because it is the only one left).

This means Python C runtime is never actually shut down when .NET calls `PythonEngine.Shutdown()`. Instead, anything from .NET exposed to Python before `Shutdown` becomes unavailable until `PythonEngine.Initialize()` is called again (which can be done from a different `AppDomain`).

### Any other comments?

Also in this change:
- dropped Python 3.6 support
- fixed Python derived types not being decrefed when an instance is deallocated
- reduced time and amount of storage needed for runtime reload
- removed circular reference loop between Type <-> `ConstructorBinding`(s)
+ exposed `Runtime.TryCollectingGarbage`

A review from @amos402 would be welcome

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -14,10 +14,6 @@ jobs:
         os: [windows, ubuntu, macos]
         python: ["3.6", "3.7", "3.8", "3.9", "3.10"]
         platform: [x64]
-        shutdown_mode: [Normal, Soft]
-
-    env:
-      PYTHONNET_SHUTDOWN_MODE: ${{ matrix.SHUTDOWN_MODE }}
 
     steps:
       - name: Set Environment on macOS
--- a/src/runtime/pythonengine.cs
+++ b/src/runtime/pythonengine.cs
@@ -17,14 +17,6 @@ namespace Python.Runtime
     /// </summary>
     public class PythonEngine : IDisposable
     {
-        public static ShutdownMode ShutdownMode
-        {
-            get => Runtime.ShutdownMode;
-            set => Runtime.ShutdownMode = value;
-        }
-
-        public static ShutdownMode DefaultShutdownMode => Runtime.GetDefaultShutdownMode();
-
         private static DelegateManager? delegateManager;
         private static bool initialized;
         private static IntPtr _pythonHome = IntPtr.Zero;
@@ -182,9 +174,9 @@ namespace Python.Runtime
             Initialize(setSysArgv: true);
         }
 
-        public static void Initialize(bool setSysArgv = true, bool initSigs = false, ShutdownMode mode = ShutdownMode.Default)
+        public static void Initialize(bool setSysArgv = true, bool initSigs = false)
         {
-            Initialize(Enumerable.Empty<string>(), setSysArgv: setSysArgv, initSigs: initSigs, mode);
+            Initialize(Enumerable.Empty<string>(), setSysArgv: setSysArgv, initSigs: initSigs);
         }
 
         /// <summary>
@@ -197,7 +189,7 @@ namespace Python.Runtime
         /// interpreter lock (GIL) to call this method.
         /// initSigs can be set to 1 to do default python signal configuration. This will override the way signals are handled by the application.
         /// </remarks>
-        public static void Initialize(IEnumerable<string> args, bool setSysArgv = true, bool initSigs = false, ShutdownMode mode = ShutdownMode.Default)
+        public static void Initialize(IEnumerable<string> args, bool setSysArgv = true, bool initSigs = false)
         {
             if (initialized)
             {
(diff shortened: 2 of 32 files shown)


.github/workflows/main.yml
```diff
@@ -12,12 +12,8 @@ jobs:
       fail-fast: false
       matrix:
         os: [windows, ubuntu, macos]
-        python: ["3.6", "3.7", "3.8", "3.9", "3.10"]
+        python: ["3.7", "3.8", "3.9", "3.10"]
```
filmor, line 15: Is there any particular reason to do this now?

lostmsu: It [EOLs in 3 days](https://endoflife.date/python). Does not have to be a part of this PR, just thought I'd clean up the matrix. Considering EOL, IMHO we should stop supporting it in 3.0.

lostmsu:
@filmor looks like I got tests to pass, so this is ready.

.github/workflows/main.yml
filmor: I don't see a compelling reason to remove it now, though. If we see it acting up, we can remove support, but that means (to me) that we also adjust the package metadata and remove the interop file. Please revert this.

lostmsu: @filmor OK, moved into https://github.com/pythonnet/pythonnet/issues/1640

src/runtime/pythonengine.cs
```diff
@@ -318,7 +310,16 @@ public static IntPtr InitExt()
         {
             try
             {
-                Initialize(setSysArgv: false, mode: ShutdownMode.Extension);
+                if (Runtime.IsInitialized)
+                {
+                    using var builtins = Runtime.PyImport_ImportModule("builtins");
+                    using var xx = Runtime.PyObject_GetAttrString(builtins.Borrow(), "RuntimeError");
+                    Exceptions.SetError(xx.Borrow(), ".NET Runtime is already initialized");
```
filmor, line 317: 1. The error should be something like `"Python.NET runtime is already initialized"`
2. We should probably use `PyEval_GetBuiltins` everywhere

src/runtime/pytype.cs
```diff
@@ -111,7 +111,7 @@ internal IntPtr GetSlot(TypeSlotID slot)
         internal static TypeFlags GetFlags(BorrowedReference type)
         {
             Debug.Assert(TypeOffset.tp_flags > 0);
-            return (TypeFlags)Util.ReadCLong(type, TypeOffset.tp_flags);
+            return (TypeFlags)Util.ReadInt32(type, TypeOffset.tp_flags);
```
filmor, line 114: This is not right. It's documented to be an `unsigned long`: https://docs.python.org/3/c-api/typeobj.html#c.PyTypeObject.tp_flags

src/runtime/runtime.cs
```diff
@@ -351,26 +305,23 @@ internal static void Shutdown(ShutdownMode mode)
...
 
+                ExtensionType.loadedExtensions.Clear();
+                CLRObject.reflectedObjects.Clear();
             }
             else
             {
                 ResetPyMembers();
-                if (mode != ShutdownMode.Extension)
-                {
-                    Py_Finalize();
```
filmor, line 360: This is my only real gripe with this: Embedders would not be able to finalise the Python runtime at all anymore. Maybe we should provide this (`Shutdown()` + `Py_Finalize()`) as a possible footgun with a long name?

src/runtime/runtime.cs
```diff
@@ -1985,7 +1923,7 @@ static Delegates()
                 }
                 catch (MissingMethodException e)
                 {
-                    throw new NotSupportedException(Util.MinimalPythonVersionRequired, innerException: e);
+                    throw new NotSupportedException(Util.MinimalPythonVersionRequired35, innerException: e);
```
filmor, line 1926: The minimal version we support right now is 3.6.

src/runtime/runtime.cs
lostmsu: We can always add it if there is a strong request. Instead what happened is people opening bugs about crashes after restarting runtime because NumPy does not support it. I'd rather avoid it.

Moreover, the behavior of the whole thing when either of runtimes is shut down completely is pretty much undefined.

src/runtime/runtime.cs
lostmsu: Reverted constant name, edited constant text instead

src/runtime/pythonengine.cs
lostmsu: Changed error message, switched to `PyEval_GetBuiltins` here

lostmsu:
@filmor I am waiting for an explicit approval or more comments

filmor (approved):

PR #1646, opened 2021-12-23
Fix warning regarding undefined __module__ on GC Offset Base

filmor:
Also makes the type names a bit more Python-esque. The `clr._internal`
module doesn't exist, but that is no difference to the previous setup.

...

...

### Any other comments?

...

### Checklist

### DIFF
--- a/src/runtime/typemanager.cs
+++ b/src/runtime/typemanager.cs
@@ -456,7 +456,7 @@ namespace Python.Runtime
             int size = Util.ReadInt32(Runtime.PyTypeType, TypeOffset.tp_basicsize)
                        + IntPtr.Size // tp_clr_inst_offset
             ;
-            var result = new PyType(new TypeSpec("GC Offset Base", basicSize: size,
+            var result = new PyType(new TypeSpec("clr._internal.GCOffsetBase", basicSize: size,
                 new TypeSpec.Slot[]
                 {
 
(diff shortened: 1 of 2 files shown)


src/runtime/typemanager.cs
```diff
@@ -456,7 +456,7 @@ internal static PyType CreateMetatypeWithGCHandleOffset()
             int size = Util.ReadInt32(Runtime.PyTypeType, TypeOffset.tp_basicsize)
                        + IntPtr.Size // tp_clr_inst_offset
             ;
-            var result = new PyType(new TypeSpec("GC Offset Base", basicSize: size,
+            var result = new PyType(new TypeSpec("clr._internal.GCOffsetBase", basicSize: size,
```
lostmsu, line 459: NIT: either `GcOffsetBase` or `CLRMetatype`

tests/utils.py
```diff
@@ -24,7 +24,7 @@ def is_clr_root_module(ob):
 
 
 def is_clr_class(ob):
-    return type(ob).__name__ == 'CLR Metatype'  # for now
+    return type(ob).__name__ == 'ClrMetatype' and type(ob).__module__ == 'clr._internal'  # for now
```
lostmsu, line 27: This will fail )

filmor: Gnah :)

PR #1653, opened 2021-12-30
Drop Python 3.6 support

lostmsu:

This drops official support for Python 3.6 which EOLed a week ago.

fixes https://github.com/pythonnet/pythonnet/issues/1640

### Checklist


filmor:
I personally don't want to do this for 3.0. There is no pressing need right now AFAICT, and keeping a slightly bigger overlap in the compatibility with 2.5 is a good thing.

lostmsu:
@filmor it would still work, but won't be officially supported. Our CI runs would be slightly less wasteful.

If you look at NumPy, their [drop support policy](https://numpy.org/neps/nep-0029-deprecation_policy.html#drop-schedule) for new releases is much more aggressive. In fact, they already dropped support for 3.7 in all upcoming releases.

lostmsu:
@mjmvisser what are Unity plans in regards to supported Python versions?

lostmsu:
@filmor reconsider before 3.0? What's the overlap for?
Most of the new packages I see on PyPi do not support 3.6 anymore. If people need legacy code, they can continue using 2.5.2

filmor:
I still don't see the point. It costs us nothing to keep the support around until we require a >= 3.7 feature.

lostmsu:
@filmor and once we do, it would require a major version bump to drop support for 3.6 if we follow semver

filmor:
What, why? I have never seen that logic practiced *anywhere*, see also https://semver.org/#what-should-i-do-if-i-update-my-own-dependencies-without-changing-the-public-api.

lostmsu:
@filmor this might be OK on PyPi, but on NuGet pythonnet does not explicitly depend on Python, and I would consider supported Python version to be a part of public API.

Anyway, I am fine with keeping it. I just don't like the approach, when stuff is kept around without explicit justification. It is always "easy" to "just keep it around", but it is not a good justification on its own. Neither of the 3 active participants in pythonnet use 3.6 anymore.

filmor:
Being stuck with old `setuptools` convinces me :)

Issue #1368, opened 2021-01-23
Restore performance tests


lostmsu:
They have been disabled when 3.5 support was dropped.

filmor:
Note:
- Use the pythonnet 2.5.1 wheel from Pypi to get the Python.Runtime.dll

PR #1658, opened 2022-01-05
Reenable perf tests for 3.0

lostmsu:

use pythonnet 2.5.2 built for Python 3.8 as the baseline build

fixes https://github.com/pythonnet/pythonnet/issues/1368


.gitignore
```diff
@@ -1,7 +1,6 @@
 /src/runtime/interopNative.cs
 
 # General binaries and Build results
-*.dll
```
filmor, line 4: This is not something we want.

filmor:
Would it be possible to acquire the baseline from somewhere else? It is included in all wheels and the nuget packages, could that be an option?

lostmsu:
NuGet packages only include 3.0. We can switch to that after 3.1

It would be non-trivial to find the DLL inside a wheel. I think the current solution serves the purpose and is pretty simple (which imo is good), should last us fine until we switch to 3.0 release as the baseline.

.gitignore
lostmsu: Removing this line did have any effect except making the baseline DLL visible to GIT again. Build outputs are covered by `bin/` and `obj/` patterns.

filmor: It makes it too easy to commit things by accident. If you really want to commit `Python.Runtime.dll` (which is what I am mainly concerned about since this approach bloats the repo over time), you can just exclude it explicitly here with `!.../Python.Runtime.dll`.

lostmsu:
Superseded by https://github.com/pythonnet/pythonnet/commit/7e2ec4d9cfc71a4c54de56707bb3fccc469b5c9b

PR #1665, opened 2022-01-08
Move code to subdirectories and rename or split up

filmor:

...

...

### Any other comments?

...

### Checklist


src/runtime/ManagedTypes.cd
```diff
@@ -1,196 +0,0 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
```
lostmsu, line 1: Is there something wrong with this file? It is useful when you need to wrap your head around the hierarchy of `ManagedType`.

src/runtime/TypeManager.cs
```diff
@@ -609,7 +609,6 @@ internal static PyType AllocateTypeObject(string name, PyType metatype)
         /// </summary>
         static void InheritSubstructs(IntPtr type)
         {
-            #warning dead code?
```
lostmsu, line 612: Some explanation of what this code does would be useful.
If you don't like the warning - figure it out! :)

src/runtime/Types/ExceptionClassObject.cs
```diff
@@ -0,0 +1,86 @@
...
+    /// </summary>
+    /// <remarks>
+    /// The Python wrapper for managed exceptions LIES about its inheritance
+    /// tree. Although the real System.Exception is a subclass of
+    /// System.Object the Python type for System.Exception does NOT claim that
+    /// it subclasses System.Object. Instead TypeManager.CreateType() uses
+    /// Python's exception.Exception class as base class for System.Exception.
+    /// </remarks>
+    [Serializable]
+    internal class ExceptionClassObject : ClassObject
```
lostmsu, line 17: Since this detects as new code anyway, maybe switch to file-scoped namespace too?

src/runtime/TypeManager.cs
filmor: Why do you think that it's dead? The function is being called, after all...

lostmsu: @filmor it might be doing something that is no longer needed and is essentially no-op

I guess 'dead' might not be the best term here, but from the looks of the code I can't determine its intent.

filmor: I'll give it a try understanding it, but this warning is currently extremely verbose for something that is at least not really a problem :)

lostmsu: @filmor see you after returning from the rabbit hole! (P.S. I never tried this particular one yet)

src/runtime/ManagedTypes.cd
filmor: Nope, I recovered it now.

src/runtime/Types/ExceptionClassObject.cs
```diff
@@ -0,0 +1,85 @@
...
+/// <summary>
+/// Base class for Python types that reflect managed exceptions based on
+/// System.Exception
+/// </summary>
+/// <remarks>
+/// The Python wrapper for managed exceptions LIES about its inheritance
+/// tree. Although the real System.Exception is a subclass of
+/// System.Object the Python type for System.Exception does NOT claim that
+/// it subclasses System.Object. Instead TypeManager.CreateType() uses
+/// Python's exception.Exception class as base class for System.Exception.
```
lostmsu, line 14: NIT: this is not specific to `System.Exception` anymore, any type can mix-in any compatible Python base class. The one for exception is handled in the default base type provider. E.g. this remark can be removed.

src/runtime/TypeManager.cs
filmor: Wow, this one tracks through until the initial commit, 15 years ago :) @brianlloyd, you don't happen to know what exactly you were trying to achieve here? :)

filmor: This code is required for things like `PyNumber_Check` to work, see e.g. https://github.com/python/cpython/blob/31ff96712e8f89ac1056c2da880b44650002219f/Objects/abstract.c#L830-L837

I don't think it is necessarily required, since our types are structured exactly like `PyHeapTypeObject`, but this can be revisited if we ever restructure our generated types to be created completely using Python's methods instead of this hackery. It's definitely valid to do this explicit initialisation here.

src/runtime/TypeManager.cs
```diff
@@ -587,7 +587,7 @@ internal static PyType AllocateTypeObject(string name, PyType metatype)
             // Clr type would not use __slots__,
             // and the PyMemberDef after PyHeapTypeObject will have other uses(e.g. type handle),
             // thus set the ob_size to 0 for avoiding slots iterations.
-            Util.WriteIntPtr(type, TypeOffset.ob_size, IntPtr.Zero);
+            Util.WriteCLong(type, TypeOffset.ob_size, 0);
```
lostmsu, line 590: Sorry, confused because the other PR also includes this change. I believe it is incorrect. The field seems to have type `Py_ssize_t` in CPython code, which is equivalent to `IntPtr`, not C's `long`.

src/runtime/TypeManager.cs
lostmsu: It seems wrong to have it present for types that do not implement numbers protocol. From the code you linked it looks like our types would pass `PyNumber_Check`.

filmor: `sq_length` etc. are the first fields of the respective abstract types. If your type /could/ implement one of them, they have to be set. And our types (being modelled in memory after heap types) always have the respective sections, so it's correct to set them. The checks pass as soon as one of the slots is filled.

src/runtime/TypeManager.cs
filmor: I think you are right, will revert.

lostmsu: LGTM if this commit is skipped.

PR #1666, opened 2022-01-08
GIL-protect overridden methods

filmor:
- Move code to subdirectories and rename or split up
- Protect all overridden calls from System classes by GIL

...

...

### Any other comments?

...

### Checklist

### DIFF
--- a/src/runtime/Py.cs
+++ b/src/runtime/Py.cs
@@ -32,14 +32,13 @@ public static class Py
 
         internal GILState()
         {
-            state = PythonEngine.AcquireLock();
+            state = Runtime.PyGILState_Ensure();
         }
 
         public virtual void Dispose()
         {
             if (this.isDisposed) return;
-
-            PythonEngine.ReleaseLock(state);
+            Runtime.PyGILState_Release(state);
             GC.SuppressFinalize(this);
             this.isDisposed = true;
         }
--- a/src/runtime/Py.cs
+++ b/src/runtime/Py.cs
@@ -27,18 +27,29 @@ public static class Py
 
     public class GILState : IDisposable
     {
+        private static readonly ThreadLocal<int> internalGILCount = new();
         private readonly PyGILState state;
         private bool isDisposed;
 
         internal GILState()
         {
-            state = Runtime.PyGILState_Ensure();
+            if (internalGILCount.Value == 0)
+            {
+                state = Runtime.PyGILState_Ensure();
+            }
+
+            internalGILCount.Value += 1;
         }
 
         public virtual void Dispose()
         {
             if (this.isDisposed) return;
-            Runtime.PyGILState_Release(state);
+
+            internalGILCount.Value -= 1;
+            if (internalGILCount.Value == 0)
+            {
+                Runtime.PyGILState_Release(state);
+            }
             GC.SuppressFinalize(this);
             this.isDisposed = true;
         }
--- a/src/runtime/Types/ClassDerived.cs
+++ b/src/runtime/Types/ClassDerived.cs
@@ -719,7 +719,7 @@ namespace Python.Runtime
             if (null != self.Ref)
             {
                 var disposeList = new List<PyObject>();
-                PyGILState gs = Runtime.PyGILState_Ensure();
+                using var _ = new Py.GILState();
                 try
                 {
                     using var pyself = new PyObject(self.CheckRun());
@@ -751,7 +751,6 @@ namespace Python.Runtime
                     {
                         x?.Dispose();
                     }
-                    Runtime.PyGILState_Release(gs);
                 }
             }
 
@@ -774,7 +773,7 @@ namespace Python.Runtime
             if (null != self.Ref)
             {
                 var disposeList = new List<PyObject>();
-                PyGILState gs = Runtime.PyGILState_Ensure();
+                using var _ = new Py.GILState();
                 try
                 {
                     using var pyself = new PyObject(self.CheckRun());
@@ -804,7 +803,6 @@ namespace Python.Runtime
                     {
                         x?.Dispose();
                     }
-                    Runtime.PyGILState_Release(gs);
                 }
             }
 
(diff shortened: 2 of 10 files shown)


lostmsu:
Maybe post this PR after the rename one?

lostmsu:
As I mentioned, we can't just always acquire GIL, as this can lead to unavoidable deadlocks.

filmor:
Well, our whole test suite (Python as well as .NET embedding) runs through without failures. Since this just "ensures" the GIL, it will also not break existing (valid) code. What deadlocks do you envision here?

lostmsu:
@filmor well, you are right. With the current version I don't see any possibility of deadlocks. My original concern was about the nuclear solution, that would make all .NET methods behave as if they are marked with `[ForbidPythonThreads]`. The risk of deadlock arises if any code we force wrap in GIL call back to user code.

However, it is inconsistent at the moment, as there are some library methods, that require GIL, and some acquire it automatically, and the pattern is at least very much not obvious.

I also have a concern about performance. The few perf tests we have have no coverage for this change. What is the impact of acquiring GIL 5000 times to `GetAttr` 5000 times instead of just acquiring it once?

filmor:
The pattern I am going for is to GIL protect all methods, that implement a C# builtin interface or override a method. So `GetAttr` etc are left as they are, so there should be no performance problem in existing working code. "Reacquiring"  the GIL if your thread already has it is cheap.

lostmsu:
Re: performance, we should have a test though.

Having auto GIL on `TryInvokeMembber` is equivalent to having it on `InvokeMethod` in

```csharp
using var _ = Py.GIL();

dynamic npCounter  = numpy.int64(0);

for (int i = 0; i < 5000; i++)
  npCounter += 1;
```

lostmsu:
```csharp
using BenchmarkDotNet.Attributes;

using Python.Runtime;
using Python.PerformanceTests;
using BenchmarkDotNet.Running;

var summary = BenchmarkRunner.Run<GilOnEveryAdd>();

[SimpleJob]
public class GilOnEveryAdd
{
    const int loop = 50000;

    [GlobalSetup]
    public void Setup()
    {
        Runtime.PythonDLL = @"python38.dll";
    }

    [Benchmark]
    public void OneGil()
    {
        using var _ = Py.GIL();

        dynamic counter = Py.Import("numpy").InvokeMethod("int64");

        for (int i = 0; i < loop; i++)
            counter+=1;
    }

    [Benchmark]
    public void GilEveryIteration()
    {
        using var _ = Py.GIL();

        dynamic counter = Py.Import("numpy").InvokeMethod("int64");

        for (int i = 0; i < loop; i++)
        {
            using var _2 = Py.GIL();
            counter+=1;
        }
    }
}
```

  DefaultJob : .NET 6.0.1 (6.0.121.56705), X64 RyuJIT


|            Method |     Mean |    Error |   StdDev |
|------------------ |---------:|---------:|---------:|
|            OneGil | 23.82 ms | 0.467 ms | 0.640 ms |
| GilEveryIteration | 28.80 ms | 0.567 ms | 0.557 ms |

filmor:
I'm not sure how you'd want to test this, but I'll have a look at the current perf tests and adjust them accordingly. If you're going for performance, using DLR expressions is probably not a good idea anyhow.

lostmsu:
The example above uses ends up using `TryBinaryOperation` and I am not even sure if there's a relatively fast alternative to it, as that function directly invokes `PyNumber_InPlaceAdd`.

filmor:
Well, we could do a performance optimisation for when `GILState` is used by staying within C#. Then we'll get the same safety and (apart from an additional boolean check) same performance. I'll try that.

src/runtime/TypeManager.cs
```diff
@@ -587,7 +587,7 @@ internal static PyType AllocateTypeObject(string name, PyType metatype)
             // Clr type would not use __slots__,
             // and the PyMemberDef after PyHeapTypeObject will have other uses(e.g. type handle),
             // thus set the ob_size to 0 for avoiding slots iterations.
-            Util.WriteIntPtr(type, TypeOffset.ob_size, IntPtr.Zero);
+            Util.WriteCLong(type, TypeOffset.ob_size, 0);
```
lostmsu, line 590: This looks wrong to me. ob_size is `Py_ssize_t` which AFAIK is explicitly defined to match pointer sizes.

src/runtime/Types/ClassDerived.cs
```diff
@@ -867,18 +865,11 @@ public class PythonDerivedType
...
-                    return pyvalue.As<T>();
-                }
-            }
-            finally
+            using var _ = new Py.GILState();
+            using var pyself = new PyObject(self.CheckRun());
+            using (PyObject pyvalue = pyself.GetAttr(propertyName))
             {
-                Runtime.PyGILState_Release(gs);
+                return pyvalue.As<T>();
```
lostmsu, line 872: Not coming from your change, but disposing `PyObject`, that was used in `As<T>` operation is semantically incorrect since `As<T>` can store that `PyObject` instance elsewhere.

src/runtime/Py.cs
```diff
@@ -27,18 +27,29 @@ public static PyModule CreateScope(string name)
 
     public class GILState : IDisposable
     {
+        private static readonly ThreadLocal<int> internalGILCount = new();
```
lostmsu, line 30: You can check if this resolves the performance issue with the script I published above.

This also leaves the scenario, where upon the initialization of runtime `BeginAllowThreads` was never called. In that case you don't need to explicitly acquire GIL unless you reenter .NET code from Python. E.g. in the sample benchmark with `dynamic counter` I posted above you don't actually need to call `Py.GIL` unless you called `PythonEngine.BeginAllowThreads`.

lostmsu:
BTW, when you will do perf testing note that the current master already has ~30% regression from 2.5.2, which I hope to narrow down at least partially.

src/runtime/Types/ClassDerived.cs
lostmsu: NVM, that was wrong. Currently `As<T>` takes this as a `BorrowedReference`.

PR #1672, opened 2022-01-13
Temporary fix method binder for out parameters

eirannejad:

Current method binder code needs cleanup. There is an active issue with pythonnet failing to find the appropriate overload when the method signature includes `out` parameters e.g. `SomeMethod(int a, out int b)`. Current workaround is to call the method by providing bogus value for the `out` parameter e.g. `SomeMethod(1, 0)` instead of `SomeMethod(1)`

This PR pushes a temporary fix for this problem until the binder code is cleaned up. The change is intentionally minimal.

Not sure. Can not find anything related to this by looking at open issue titles

### Any other comments?

No

### Checklist

### DIFF
--- a/tests/test_method.py
+++ b/tests/test_method.py
@@ -267,7 +267,13 @@ def test_params_method_with_lists():
 
 def test_string_out_params():
     """Test use of string out-parameters."""
-    result = MethodTest.TestStringOutParams("hi")
+    result = MethodTest.TestStringOutParams("hi", "there")
+    assert isinstance(result, tuple)
+    assert len(result) == 2
+    assert result[0] is True
+    assert result[1] == "output string"
+
+    result = MethodTest.TestStringOutParams("hi", None)
     assert isinstance(result, tuple)
     assert len(result) == 2
     assert result[0] is True
@@ -291,11 +297,15 @@ def test_string_ref_params():
 
 def test_value_out_params():
     """Test use of value type out-parameters."""
-    result = MethodTest.TestValueOutParams("hi")
+    result = MethodTest.TestValueOutParams("hi", 1)
     assert isinstance(result, tuple)
     assert len(result) == 2
     assert result[0] is True
     assert result[1] == 42
+
+    # None cannot be converted to a value type like int, long, etc.
+    with pytest.raises(TypeError):
+        MethodTest.TestValueOutParams("hi", None)
 
 
 def test_value_ref_params():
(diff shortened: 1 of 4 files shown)


lostmsu:
@eirannejad, please add a test or two to cover the new code.

tests/test_method.py
```diff
@@ -267,13 +267,7 @@ def test_params_method_with_lists():
...
     """Test use of string out-parameters."""
-    result = MethodTest.TestStringOutParams("hi", "there")
-    assert isinstance(result, tuple)
-    assert len(result) == 2
-    assert result[0] is True
-    assert result[1] == "output string"
-
-    result = MethodTest.TestStringOutParams("hi", None)
+    result = MethodTest.TestStringOutParams("hi")
     assert isinstance(result, tuple)
```
lostmsu, line 271: Since your change does not break the "old way" to call functions with `out` parameters, I think you should keep existing test scenarios, and add yours as new ones. This applies to the test changes below too.

I think this is important, because the old way enabled users to choose an overload from multiple overloads with different `out` parameter types, and the new one does not seem to be able to do so. Please, correct me if I am wrong. (BTW if there's not yet a test for overload resolution over the type of `out` parameter, it would be nice to add one in this PR).

lostmsu (changes requested):
Please, keep old test scenarios unless they are actually not supported anymore.

tests/test_method.py
eirannejad: You are absolutely right. I'll revert the test modifications and will add new ones. I'd like to suggest passing types instead of values to help method matching e.g. `TestOutParams("hi", int)` or `TestOutParams("hi", System.Guid)`. Most of my experience is using IronPython and I'm not sure what is the overall design goal for handling `out` parameters in PythonNet. I'm thinking out loudly but I think we'll end up supporting 3 methods of matching:

1 - `TestOutParams("hi")`
2 - `TestOutParams("hi", "some value matching out type")`
3 - `TestOutParams("hi", System.Guid)`

lostmsu: @eirannejad I was just thinking about that and I think having something like `clr.Uninitialized[System.Int32]` or `System.Int32.__default__` would do the trick. Passing types might be confusing as methods might actually have `System.Type` parameters.

eirannejad: Uh true. I personally like the `clr.default[System.Int32]` slightly more as it is explicitly part of `clr` and easier to teach and remember. We can also maybe box the Type e.g. `clr.out[System.Int32]` to pass to the method. Both of those forces more typing for the user of course so I'd like to see if unconventional approaches like `TestOutParams("hi") # signature: string, System.Guid` would even be considerable

eirannejad: Reverted the existing test modifications and added new tests instead

eirannejad:
(I'm a bit new to Github review/request change process. I apologize in advance if I clicked too many things)

filmor:
@eirannejad We'd need you to sign the .NET Foundation CLA: https://cla.dotnetfoundation.org/pythonnet/pythonnet

@lostmsu Please have an eye on this for first-time contributors until the CLA bot is fixed.

eirannejad:
@filmor @lostmsu Signed the CLA. Thanks for merging this.

PR #1673, opened 2022-01-15
added missing PyThreadState_Swap to Runtime and its Delegates

eirannejad:

`PyThreadState_Swap` was missing in `Runtime` and `Runtime.Delegates` code

No

### Any other comments?

N/A

### Checklist

-   Make sure to include one or more tests for your change
    ⚠️ Skipping tests since there are not any tests for specific runtime p/invokes. This method is not used anywhere in original pythonnet code base as of creating this PR

### DIFF
--- a/src/runtime/Runtime.Delegates.cs
+++ b/src/runtime/Runtime.Delegates.cs
@@ -23,6 +23,7 @@ public unsafe partial class Runtime
             Py_EndInterpreter = (delegate* unmanaged[Cdecl]<PyThreadState*, void>)GetFunctionByName(nameof(Py_EndInterpreter), GetUnmanagedDll(_PythonDll));
             PyThreadState_New = (delegate* unmanaged[Cdecl]<PyInterpreterState*, PyThreadState*>)GetFunctionByName(nameof(PyThreadState_New), GetUnmanagedDll(_PythonDll));
             PyThreadState_Get = (delegate* unmanaged[Cdecl]<PyThreadState*>)GetFunctionByName(nameof(PyThreadState_Get), GetUnmanagedDll(_PythonDll));
+            PyThreadState_Swap = (delegate* unmanaged[Cdecl]<PyThreadState*, PyThreadState*>)GetFunctionByName(nameof(PyThreadState_Swap), GetUnmanagedDll(_PythonDll));
             _PyThreadState_UncheckedGet = (delegate* unmanaged[Cdecl]<PyThreadState*>)GetFunctionByName(nameof(_PyThreadState_UncheckedGet), GetUnmanagedDll(_PythonDll));
             try
             {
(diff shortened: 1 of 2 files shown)


lostmsu:
This looks good but without usages there's not much point merging it.

eirannejad:
@lostmsu I'm using this in a fork that I have working with sub-interpreters. Whichever you decide. I thought it's better to PR to master

lostmsu:
Are you planning to upstream subinterpreters? Or is it not yet ready?

eirannejad:
I definitely am. To be honest I'm a little in the dark as to what is the overall scope and design goal for pythonnet. I'd appreciate your help and info on this. It'd help me push more generic code upstream.

**Q:** Is `PythonEngine` intended to be a facade over the runtime and wrap more complex functionality, simplifying runtime usage? or is it intended to be as close as possible to python runtime itself? `Runtime.cs` currently wraps `Py_NewInterpreter` and `Py_EndInterpreter` without them being used in pythonnet code AFAIK. I'm not sure if I should just wrap these in `PythonEngine` or provide a [context manager like this](https://github.com/mcneel/pythonnet/blob/ce219ce4be2b5b65db1dfa8eb80ba417e456c146/src/runtime/McNeel.PythonEngine.cs#L73) to simplify usage?

- The linked code is custom to my specific use case but I'd love to make it more generic and push upstream
- It's more of an overall design api goal question really.

eirannejad:
@lostmsu I'm embedding pythonnet in a dotnet application and creating a cpython IDE over it. I need to run a mini language server on the python engine as well that doesn't have anything to do with the actual user scripts being executed by the engine. Effectively this creates two completely independent usages for pythonnet under the same process. I need to create a sub-interpreter for these reasons:

- The language server imports lots of support libraries for linting and autocompletion. If this is just another `PyScope` in the main interpreter it would sprinkle `sys.modules` with those imported modules
- I need to assign independent `stdout` and `stderr` streams to this language server and therefore need a separate interpreter

lostmsu:
My understanding is that the issue requires more involved changes.

> For modules using single-phase initialization, e.g. `PyModule_Create`() (**this is what Python.NET does AFAIR**), the first time a particular extension is imported, it is initialized normally, and a (shallow) copy of its module’s dictionary is squirreled away. When the same extension is imported by another (sub-)interpreter, a new module is initialized and filled with the contents of this copy; the extension’s init function is not called. Objects in the module’s dictionary thus end up shared across (sub-)interpreters, which might cause unwanted behavior (see Bugs and caveats below).

Now if you think about it, the modules that Python.NET creates to reflect .NET namespaces if simply copied to the new subinterpreter won't get notified of newly loaded assemblies.

The following passage is also unclear to the exact effect:

> Also note that combining this functionality with PyGILState_* APIs is delicate, because these APIs assume a bijection between Python thread states and OS-level threads, an assumption broken by the presence of sub-interpreters. It is highly recommended that you don’t switch sub-interpreters between a pair of matching PyGILState_Ensure() and PyGILState_Release() calls. Furthermore, extensions (such as ctypes) using these APIs to allow calling of Python code from non-Python created threads will probably be broken when using sub-interpreters.

We can probably defer that issue to users, but the problem is that because most usages of Python.NET are higher-level, we will get a lot of people asking why it does not work like they expect it should.

BTW, @eirannejad, according to the first quote above in the example you mentioned, Python.NET will only be initialized once.

As for running a language server, why not use [Language Server Protocol](https://microsoft.github.io/language-server-protocol/) used in VS Code for all languages. Then you should be able to just reuse existing Python language server that Microsoft developed for VS Code. This does not use subinterpreters, and instead just runs language server library in a separate process.

AFAIK, Python language server in VS Code does not autocomplete .NET types, and this is something we might want to address instead.

lostmsu:
As discussed, we are not taking it in the current form. I hope the LSP is good alternative for you @eirannejad

Issue #1683, opened 2022-01-28
SegFault before Python shutdown


AndyWatterman:
### Environment

-   Pythonnet version: 3.0.0.dev1
-   Python version: 3.9.9
-   Operating System: Kali Linux
-   .NET Runtime: 
mono-complete:amd64/kali-rolling 6.8.0.105+dfsg-3.2 uptodate
mono-devel:all/kali-rolling 6.8.0.105+dfsg-3.2

### Details

This example causes segfault before python shutdown:

```
import clr
from System import IntPtr
a = clr.GetClrType(IntPtr)
```
Actually no matter what type is imported (IntPtr, Console, etc.) - crashing anyway. However, crash happen exactly after the program finished its executing and only when NO debugging.

```
gef➤  bt

#0  0x0000000040071c50 in ?? ()
#1  0x0000000000508a13 in _Py_Dealloc (op=<optimized out>) at ../Objects/object.c:2209
#2  _Py_DECREF (op=<optimized out>) at ../Include/object.h:430
#3  _Py_XDECREF (op=<optimized out>) at ../Include/object.h:497
#4  free_keys_object (keys=0x11ac070) at ../Objects/dictobject.c:598
#5  0x0000000000508824 in dictkeys_decref (dk=<optimized out>) at ../Objects/dictobject.c:333
#6  dict_dealloc (mp=0x7ffff78671c0) at ../Objects/dictobject.c:2026
#7  0x00000000005a9f60 in _Py_DECREF (op=<optimized out>) at ../Include/object.h:430
#8  _Py_XDECREF (op=<optimized out>) at ../Include/object.h:497
#9  module_dealloc (m=0x7ffff7863cc0) at ../Objects/moduleobject.c:677
#10 0x00000000004f9749 in _Py_Dealloc (op=<optimized out>) at ../Objects/object.c:2209
#11 _Py_DECREF (op=<optimized out>) at ../Include/object.h:430
#12 _Py_XDECREF (op=<optimized out>) at ../Include/object.h:497
#13 insertdict (mp=<optimized out>, key='__main__', hash=<optimized out>, value=<optimized out>) at ../Objects/dictobject.c:1123
#14 0x00000000006294a0 in _PyImport_Cleanup (tstate=0x94e6e0) at ../Python/import.c:526
#15 0x000000000062732f in Py_FinalizeEx () at ../Python/pylifecycle.c:1426
[...]

PR #1712, opened 2022-03-04
On shutdown from Python release all slot holders

lostmsu:

Before this fix slots of reflected types were not released to Python defaults leaving Python referencing managed code after managed runtime has already shut down.

In the original bug, the runtime in question is Mono after `mono_jit_cleanup`

https://github.com/pythonnet/pythonnet/issues/1683


lostmsu:
This is the old slots holder magic created by Amos. I think it is better, that skipping critical finalizers.

This should make https://github.com/pythonnet/clr-loader/issues/26 irrelevant.

lostmsu:
Looks like this does not work. I am going to investigate further.

lostmsu:
@filmor yeet!

filmor:
Nice :)

Does it still work if you pull in 39db9f4d4d9e3e7d8d3e687b3ed74cb3d8f3ddb8?

lostmsu:
@filmor looks like it

PR #1720, opened 2022-03-10
On shutdown from Python release all slot holders

filmor:

Before this fix slots of reflected types were not released to Python defaults leaving Python referencing managed code after managed runtime has already shut down.

In the original bug, the runtime in question is Mono after `mono_jit_cleanup`.

This is an extension of #1712 and actually calls shutdown from Python.

https://github.com/pythonnet/pythonnet/issues/1683


filmor:
The errors we are hitting here are the reason why we didn't do the full shutdown in the past. They only occur when running the full suite *in sequence* (I'm trying to find the minimal permutation that "works", but it's tricky).

filmor:
The minimal call I have found so far that reproduces the segfault is
```
pytest --runtime mono tests/test_class.py tests/test_conversion.py tests/test_exceptions.py tests/test_field.py tests/test_generic.py tests/test_import.py tests/test_indexer.py tests/test_method.py tests/test_subclass.py
```

filmor:
@lostmsu This runs through now, but we are not doing the `PyObject_GC_Del` anymore. Where is the corresponding object creation?

lostmsu:
@filmor you would be looking for callers of `PythonDerivedType.SetPyObj`

lostmsu:
I know it is annoying, but can you untangle the fix for the issue from the syntax changes?

Also, I think you might have accidentally changed line endings in the interop .tt file.

lostmsu:
Superseded by https://github.com/pythonnet/pythonnet/pull/1712

PR #1758, opened 2022-04-08
Clear weakref list when reflected object is destroyed

lostmsu:

According to [official documentation](https://docs.python.org/3/extending/newtypes.html#weak-reference-support) for defining extension types, if they are to support weak references, they must clear weak reference list associated with the instance in `tp_clear`.

This was missed when https://github.com/pythonnet/pythonnet/pull/1267 was superseded.

P.S. without the fix corresponding test would crash with segfault due to an attempt to dereference an object that no longer exists via a weak reference to it.

N/A

### Checklist

### DIFF
--- a/src/embed_tests/TestInstanceWrapping.cs
+++ b/src/embed_tests/TestInstanceWrapping.cs
@@ -34,6 +34,16 @@ namespace Python.EmbeddingTest
             }
         }
 
+        [Test]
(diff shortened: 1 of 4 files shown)


lostmsu:
Also @amos402 if you have time. This is based on your old PR from Oct 2020.

filmor (approved):

PR #1760, opened 2022-04-09
Miscellaneous fixes and cleanups

filmor:
- Use Py.GIL directly, now that it doesn't try to init anymore
- Modernise syntax a bit
- Fix Py.SetArgv for systems that don't support GetCommandLineArgs
- Modernise syntax in ClassDerived

...

...

### Any other comments?

...

### Checklist


src/runtime/Types/ClassBase.cs
```diff
@@ -276,42 +275,46 @@ public static nint tp_hash(BorrowedReference ob)
...
+                    {
+                        e = e.InnerException;
+                    }
+                    Exceptions.SetError(e);
+                    return default;
                 }
-                Exceptions.SetError(e);
-                return default;
             }
+            return Exceptions.RaiseTypeError("invalid object");
```
lostmsu, line 294: I think keeping input parameter validation returns before the main body is a good practice.

src/runtime/Types/EventObject.cs
```diff
@@ -27,24 +27,20 @@ public EventObject(EventInfo info)
...
+                    return new EventBinding(self.name, self.reg, new PyObject(ob)).Alloc();
+                }
 
-            if (Runtime.PyObject_IsInstance(ob, tp) < 1)
-            {
-                return Exceptions.RaiseTypeError("invalid argument");
             }
-
-            return new EventBinding(self.name, self.reg, new PyObject(ob)).Alloc();
+            return Exceptions.RaiseTypeError("invalid argument");
```
lostmsu, line 43: Same, validate and quick exit.

src/runtime/Types/PropertyObject.cs
```diff
@@ -71,26 +71,24 @@ public static NewReference tp_descr_get(BorrowedReference ds, BorrowedReference
...
+                    {
+                        e = e.InnerException;
+                    }
+                    Exceptions.SetError(e);
+                    return default;
                 }
-                Exceptions.SetError(e);
-                return default;
             }
+            return Exceptions.RaiseTypeError("invalid target");
```
lostmsu, line 91: Same

src/runtime/Types/EventObject.cs
filmor: I agree with this in general, but here this would mean that instead of a single `RaiseTypeError` with a particular string, we need two, which does not look like an advantage to me.

lostmsu: I think the old version is easier to read. Nesting makes it harder.

filmor: Is this your only gripe with this PR? I find the new version easier to read because it puts emphasis on the happy path.

lostmsu: Yes. I also think that since we are in disagreement, a change should not be made without some kind of necessity.

filmor: Meh. I'll revert these particular changes, the rest is important (introducing `readonly` wherever possible etc.).

PR #1762, opened 2022-04-10
Implement IConvertible on PyObject

### LINKED ISSUE #1755: Python int type is translated to pyInt when passed as a .net Dictionary value only in 3.x version
### Environment

-   Pythonnet version: 3.0.0a2
-   Python version: 3.7, 3.8, 3.9, 3.10
-   Operating System: Windows
-   .NET Runtime: 4.7.2 framework

### Details

After upgrading to pre release 3.0.0a2 version, the python int type is getting translated to PyInt instead of System.int32 inside the .NET DLL, and hence getting a unable to cast exception. this is working fine in 2.x version.

DLL method: 
```csharp
public void DictMethod(Dictionary<string, object> paramDict)
  {
            Console.WriteLine("Inside");
            Console.WriteLine(paramDict["index"].GetType().ToString());
            var index = Convert.ToInt32((paramDict["index"]));
            
        }
```
Python script: 

```python
   clr.AddReference("System.Collections")
from System.Collections.Generic import Dictionary
from System import String
from System import Object
clr.AddReference("PyTest")
from PyTest import PyClass

class Utils(object):

    @staticmethod
    def prepare_dict(**kwargs):

        dict_net = Dictionary[String,Object]()
        for key_value in kwargs.keys():
            dict_net.Add(key_value,kwargs[key_value])
        return dict_net

parameters = {}
parameters['index'] = 1
net_dict = Utils.prepare_dict(**parameters)
obj = PyClass()
obj.DictMethod(net_dict)
```

Output in 3.x version:
```
C:\Users\Athul\Desktop>c:\Python37\python.exe pynettest.py
Inside
Python.Runtime.PyInt
  File "pynettest.py", line 50, in <module>
    obj.DictMethod(net_dict)
System.InvalidCastException: Unable to cast object of type 'Python.Runtime.PyInt' to type 'System.IConvertible'.
   at System.Convert.ToInt32(Object value)
   at PyTest.PyClass.DictMethod(Dictionary`2 paramDict) in C:\Users\Athul\source\repos\PyTest\PyClass.cs:line 46
```

Output in 2.x version : 

```
C:\Users\Athul\Desktop>c:\Python37\python.exe pynettest.py
Inside
System.Int32

C:\Users\Athul\Desktop>
```

filmor:

Implements the [IConvertible](https://docs.microsoft.com/en-us/dotnet/api/system.iconvertible) interface for `PyObject` instances.

Should fix #1755.

### Any other comments?

...

### Checklist

### DIFF
--- a/src/runtime/PythonTypes/PyInt.cs
+++ b/src/runtime/PythonTypes/PyInt.cs
@@ -230,5 +230,7 @@ namespace Python.Runtime
             using var _ = Py.GIL();
             return ToBigInteger().ToString(format, formatProvider);
         }
+
+        public override TypeCode GetTypeCode() => TypeCode.Int64;
     }
 }
--- a/src/runtime/Converter.cs
+++ b/src/runtime/Converter.cs
@@ -550,7 +550,7 @@ namespace Python.Runtime
         /// <summary>
         /// Convert a Python value to an instance of a primitive managed type.
         /// </summary>
-        private static bool ToPrimitive(BorrowedReference value, Type obType, out object? result, bool setError)
+        internal static bool ToPrimitive(BorrowedReference value, Type obType, out object? result, bool setError)
         {
             result = null;
             if (obType.IsEnum)
--- a/src/runtime/PythonTypes/PyFloat.cs
+++ b/src/runtime/PythonTypes/PyFloat.cs
@@ -101,5 +101,7 @@ namespace Python.Runtime
             PythonException.ThrowIfIsNull(op);
             return new PyFloat(op.Steal());
         }
+
+        public override TypeCode GetTypeCode() => TypeCode.Double;
     }
 }
--- a/src/runtime/PythonTypes/PyObject.IConvertible.cs
+++ b/src/runtime/PythonTypes/PyObject.IConvertible.cs
@@ -1,0 +1,53 @@
+using System;
+
+namespace Python.Runtime;
+
+public partial class PyObject : IConvertible
+{
+    public virtual TypeCode GetTypeCode() => TypeCode.Object;
+
+    private T DoConvert<T>()
+    {
+        using var _ = Py.GIL();
+        if (Converter.ToPrimitive(Reference, typeof(T), out object? result, setError: false))
+        {
+            return (T)result!;
+        }
+        else
+        {
+            throw new InvalidCastException();
+        }
+    }
+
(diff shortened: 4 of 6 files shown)


lostmsu:
.NET team now avoids `IConvertible`: https://github.com/dotnet/runtime/issues/936#issuecomment-663743656

lostmsu:
See also https://github.com/dotnet/runtime/issues/62022

filmor:
It doesn't look like there is an alternative yet, though, right? And I also don't think that they are actually going to drop the interface, they are just not extending it for new primitive types, which have no representation in Python anyway.

lostmsu:
@filmor the new `Half` type does not implement `IConvertible` though it is nearly identical to `Single` and `Double`. So I guess it is kinda deprecated (not that I completely agree with everything .NET team does).

src/runtime/PythonTypes/PyInt.cs
```diff
@@ -230,5 +230,7 @@ public string ToString(string format, IFormatProvider formatProvider)
             using var _ = Py.GIL();
             return ToBigInteger().ToString(format, formatProvider);
         }
+
+        public override TypeCode GetTypeCode() => TypeCode.Int64;
```
lostmsu, line 234: Generally, this is wrong.

filmor: The docs are a bit unclear on this function. I used this as an "should almost always work" value, the `InvalidCastException` might still be thrown if the conversion fails. We could drop the `GetTypeCode` overrides entirely and just return `TypeCode.Object` if you'd prefer that.

lostmsu: I can't say for others, just for this one. Honestly, not sure if anyone ever uses this.

filmor: I'm not sure I follow. `GetTypeCode` is supposed to be a hint that allows you to write a `switch` like
```csharp
switch (convertible.GetTypeCode()) {
    case TypeCode.Int64:
        return Convert.ToInt64(convertible);
}
```
It doesn't mean that the object *is* a `System.Int64` already.

lostmsu: You seem to be correct on that. Still not sure `Int64` is valid for this specific case, but float and string probably are. I am fine either way though. I don't think we will get people complaining one way or another any time soon.

PR #1763, opened 2022-04-10
Ensure that codec tests are run

filmor:
- Move test_codec.py to the right directory
- Make PyObjectConversions.Reset public to allow calling it from Python

...

...

### Any other comments?

...

### Checklist


src/runtime/Codecs/PyObjectConversions.cs
```diff
@@ -125,7 +125,7 @@ bool TryDecode(BorrowedReference pyHandle, out object? result)
 
         #endregion
 
-        internal static void Reset()
+        public static void Reset()
```
lostmsu, line 128: This method should not be a part of public API

Issue #1729, opened 2022-03-14
Access Violation when using 32bit version of python


SillieWous:
### Environment

-   Pythonnet version: 3.0.0-preview2021-12-25
-   Python version: 3.9.10
-   Operating System: Windows 10
-   .NET Runtime: 4.7.2

### Details

-   Describe what you were trying to get done.

Embed python into C# using 32bit python.

-   What commands did you run to trigger this issue? If you can provide a
    [Minimal, Complete, and Verifiable example](http://stackoverflow.com/help/mcve)
    this will help us understand the issue.

Example assumes [python3.9.10 32bit embed](https://www.python.org/ftp/python/3.9.10/python-3.9.10-embed-win32.zip) to be extracted to `./python`

```C#
Python.Runtime.Runtime.PythonDLL = "./python/python39.dll";

Python.Runtime.PythonEngine.PythonHome = ./python/python39.dll;
Python.Runtime.PythonEngine.PythonPath = $"{PythonEngine.PythonPath}{Path.PathSeparator}.";

Python.Runtime.PythonEngine.Initialize();
```

Last line causes an access violation after preview2021-12-25 (preview2021-12-03 works).
[python3.9.10 64bit embed](https://www.python.org/ftp/python/3.9.10/python-3.9.10-embed-amd64.zip) version of python does not cause the same issue.

-   If there was a crash, please include the traceback here.

```C#
   System.AccessViolationException: Attempted to read or write protected memory. This is often an indication that other memory is corrupt.
   at Python.Runtime.Runtime.PyType_FromSpecWithBases(NativeTypeSpec& spec, BorrowedReference bases)
   at Python.Runtime.PyType.FromSpec(TypeSpec spec, PyTuple bases)
   at Python.Runtime.TypeManager.CreateMetatypeWithGCHandleOffset()
   at Python.Runtime.TypeManager.CreateMetaType(Type impl, SlotsHolder& slotsHolder)
   at Python.Runtime.MetaType.Initialize()
   at Python.Runtime.Runtime.Initialize(Boolean initSigs)
   at Python.Runtime.PythonEngine.Initialize(IEnumerable`1 args, Boolean setSysArgv, Boolean initSigs)
   at Python.Runtime.PythonEngine.Initialize()
```

jborbely:
I am also interested in hearing about whether there are plans to continue to support 32-bit Python (e.g., issue #1210).

[...]

PR #1765, opened 2022-04-10
Fix layout of `NativeTypeSpec` on 32 bit platforms

lostmsu:

#1729 

### Checklist


src/runtime/Native/NativeTypeSpec.cs
```diff
@@ -10,7 +10,7 @@ struct NativeTypeSpec : IDisposable
         public readonly StrPtr Name;
         public readonly int BasicSize;
         public readonly int ItemSize;
-        public readonly TypeFlags Flags;
+        public readonly IntPtr Flags;
```
filmor, line 13: Hmm, `Flags` is clearly defined as a C `int`, which should "always" be 32bit, right? (https://docs.python.org/3/c-api/type.html#c.PyType_Spec)

Issue #1766, opened 2022-04-11
Embedding tests randomly hang in MacOS


lostmsu:
Example: https://github.com/pythonnet/pythonnet/runs/5978191426?check_suite_focus=true

Appears to have started with https://github.com/pythonnet/pythonnet/pull/1712

filmor:
Seems to always hang after the `TestEval` or `TestExec` tests.

lostmsu:
@filmor do you have a MacOS machine to investigate this?

PR #1779, opened 2022-05-02
Work around potential Mono bug, that hangs the runtime when new threads start

lostmsu:
I suspect the issue is actually with Mono GC and the relatively new mode of suspending threads, which has a few (closed) bugs against it, complaining about hangs. In many of those bugs (you can find them [here](https://github.com/mono/mono/issues?q=is%3Aissue+MONO_THREADS_SUSPEND+hangs)) switching to the old `preemptive` mode works around the issue. I tried it for our tests, and it seems to solve it too.

The issue also vaguely mentioned in https://www.mono-project.com/docs/advanced/runtime/docs/coop-suspend/#thread-startfinish-still-bad

So the fix is to simply set `MONO_THREADS_SUSPEND` to `preemptive` in CI before running tests in Mono.

Fixes https://github.com/pythonnet/pythonnet/issues/1766

This workaround seems to solve the issue on my local machine as well as in CI. I did 5+ CI full matrix reruns to confirm. Before the fix a few MacOS legs would time out in every run.

Below is some history of finding this workaround.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -66,6 +66,8 @@ jobs:
 
       - name: Embedding tests
         run: dotnet test --runtime any-${{ matrix.platform }} --logger "console;verbosity=detailed" src/embed_tests/
+        env:
+          MONO_THREADS_SUSPEND: preemptive # https://github.com/mono/mono/issues/21466
 
       - name: Python Tests (Mono)
         if: ${{ matrix.os != 'windows' }}


lostmsu:
:/ it does not fail when the blame mode is on in `dotnet test` :/

lostmsu:
Looks like the issue may be with the `TestThread` function. Perhaps we are missing some synchronization there. I previously suspected that it needs memory barriers, because AFAIK acquiring GIL does not insert them.

filmor:
This is probably also something that we should include in our docs.

PR #1780, opened 2022-05-02
Disallow runtime shutdown when the Python error indicator is set

lostmsu:

Shutting down runtime when the Python error indicator is set has unpredictable behavior.

### Checklist


filmor:
Couldn't we just clear it instead?

lostmsu:
@filmor that would be silent information loss.

filmor:
This way will probably lead to a lot more unwanted behaviour in that a program that ends with `Shutdown` will not have its resources closed if an error is still set (as it will just crash on this exception), or am I misunderstanding something?

lostmsu:
@filmor if they are at this point, they already have unhandled exception. Generally speaking it is better to crash, that to try to close resources as you might persist a corrupted state and not get notified about it at all.

filmor:
Hmm, I'll tentatively agree with this, let's see whether someone complains :)

Issue #1785, opened 2022-05-06
Python handling of C# collection


Martin-Molinero:
### Environment

-   Pythonnet version: master/3.0
-   Python version: *
-   Operating System: *
-   .NET Runtime: *

### Details

-   Describe what you were trying to get done.

Seems that since `CollectionMixinsProvider` was added this case started failing (`if key in collection.Keys:`), not sure why `CollectionMixinsProvider` is required, if this is a bug or not.. 
Test passes commenting out https://github.com/pythonnet/pythonnet/blob/master/src/runtime/InteropConfiguration.cs#L24

-   What commands did you run to trigger this issue? If you can provide a
    [Minimal, Complete, and Verifiable example](http://stackoverflow.com/help/mcve)
    this will help us understand the issue.

```csharp
    class ATests
    {
        private static dynamic containsTest;
        private static string testModule = @"
from clr import AddReference
AddReference(""System"")
AddReference(""Python.EmbeddingTest"")

def ContainsTest(key, collection):
    if key in collection.Keys:
        return True
    return False
";

        [OneTimeSetUp]
        public void Setup()
        {
            PythonEngine.Initialize();
            var pyModule = PyModule.FromString("module", testModule);
            containsTest = pyModule.GetAttr("ContainsTest");
        }

        [TestCase("AAPL", false)]
        [TestCase("SPY", true)]
        public void ContainsTest(string key, bool expected)
        {
            var dic = new Dictionary<string, object> { { "SPY", new object() } };
            Assert.AreEqual(expected, (bool)containsTest(key, dic));
        }
    }
```

PR #1786, opened 2022-05-07
Multiple fixes related to Dictionary.Keys bug

lostmsu:

This has individual commits for:
1. bad name generation for nested classes which caused the type of `Dictionary.Keys` to be confused with the `Dictionary` itself
2. quality of life change, that allows cast `Generic.IEnumerable(obj)`, which previously required the generic type argument to be specified. E.g. `Generic.IEnumerable[String](obj)` was required
3. collection mixins now cast objects to corresponding interfaces when the interfaces are implemented privately

fixes https://github.com/pythonnet/pythonnet/issues/1785

### Checklist

@Martin-Molinero


filmor:
Could you split part 2 out into a separate PR for discussion and add a few tests?

lostmsu:
Do you have any specific concerns? The reason I made quality of life change is to use it in `IteratorMixin`. E.g. it has a minimal test in that it is actually used internally. I don't think GitHubs allows PRs basing off other PRs.

PR #1793, opened 2022-05-22
Move to modern setuptools with pyproject.toml

filmor:
Also moves the version to a common ̀`version.txt` that is read by both
the .NET and Python builds.

...

...

### Any other comments?

...

### Checklist


lostmsu:
> Python 3.6
>  No matching distribution found for setuptools>=61

How about we drop 3.6 [as suggested](https://github.com/pythonnet/pythonnet/pull/1653)? 8-)

Directory.Build.props
```diff
@@ -1,11 +1,12 @@
+<?xml version="1.0"?>
 <Project>
   <PropertyGroup>
-    <VersionPrefix>3.0.0</VersionPrefix>
```
lostmsu, line 3: `VersionPrefix` is necessary to create preview packages for NuGet

lostmsu (changes requested):

Directory.Build.props
filmor: I'd suggest that we just write the generated preview version to the `version.txt` with a small script.

Directory.Build.props
```diff
@@ -6,7 +6,9 @@
     <AssemblyProduct>Python.NET</AssemblyProduct>
     <LangVersion>10.0</LangVersion>
     <IsPackable>false</IsPackable>
-    <Version>$([System.IO.File]::ReadAllText("version.txt"))</Version>
+    <FullVersion>$([System.IO.File]::ReadAllText("version.txt"))</FullVersion>
+    <VersionPrefix>$(FullVersion.Split('-', 2)[0])</VersionPrefix>
```
lostmsu, line 10: Can you fix the monthly NuGet script right away? Should be easy to try locally

lostmsu: Currently the version looks like this: `3.0.0-preview2022-04-11` 3.0.0 being the prefix and the rest is the suffix.

filmor: I'm not sure I understand, with this change I get exactly the result you would expect. When specifying the suffix on the command-line, it overrides whatever suffix (if one was set) from the `version.txt` file.

lostmsu (approved):

PR #1798, opened 2022-05-25
`Min`/`MaxSupportedVersion` and `IsSupportedVersion` on `PythonEngine`

lostmsu:

fixes https://github.com/pythonnet/pythonnet/discussions/1724

### Checklist

### DIFF
--- a/src/runtime/PythonEngine.cs
+++ b/src/runtime/PythonEngine.cs
@@ -127,6 +127,10 @@ namespace Python.Runtime
             }
         }
 
+        public static Version MinSupportedVersion => new(3, 7);
+        public static Version MaxSupportedVersion => new(3, 10, int.MaxValue, int.MaxValue);
+        public static bool IsSupportedVersion(Version version) => version >= MinSupportedVersion && version <= MaxSupportedVersion;
+
         public static string Version
         {
             get { return Marshal.PtrToStringAnsi(Runtime.Py_GetVersion()); }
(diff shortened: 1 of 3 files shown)


lostmsu:
:/ I am puzzled. What does this change have to do with segfaults? Only on *nix? Ugh...

src/runtime/PythonEngine.cs
```diff
@@ -127,6 +127,10 @@ public static string PythonPath
             }
         }
 
+        public static Version MinSupportedVersion => new(3, 7);
```
filmor, line 130: `readonly`?

src/runtime/PythonEngine.cs
```diff
@@ -127,6 +127,10 @@ public static string PythonPath
...
+        public static Version MaxSupportedVersion => new(3, 10, int.MaxValue, int.MaxValue);
```
filmor, line 131: Maybe something more "readable", like 999 instead of maxint?

lostmsu: I thought of that, or maybe do `MaxSupportedVersionExclusive`. Alternatively, we can only leave the `IsSupportedVersion` check and provide an error string in a separate property.

filmor:
> :/ I am puzzled. What does this change have to do with segfaults? Only on *nix? Ugh...

Just tested it as well, it only happens on Mono...

filmor:
Further observations:

- The culprit is the call, the comparisons in the tests seem fine
- Only the `enum` conversions with max int/long/ulong fail, if I remove these, the tests run though

Issue #1810, opened 2022-06-08
cannot get anymore int of enumclass (alpha2)


maxcapodi78:
### Environment

-   Pythonnet version: 3.0.0dev
-   Python version: 3.9
-   Operating System: Windows
-   .NET Runtime:

### Details

-   Describe what you were trying to get done.

In v2.5.2 I'm able to get a C# enum and get the int value. In alpha 2 it seems that the full enum class is returned but the int command is not working. In 3.0.0dev it makes crashing the Debugger.

-   What commands did you run to trigger this issue? If you can provide a
    [Minimal, Complete, and Verifiable example](http://stackoverflow.com/help/mcve)
    this will help us understand the issue.
![image](https://user-images.githubusercontent.com/77293250/172635930-3f5d51aa-3989-410f-9189-12570843135c.png)

-   If there was a crash, please include the traceback here.

```python
    Out[13]: <bound method 'int'>

Unhandled Exception: System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation. ---> System.AccessViolationException: Attempted to read or write protected memory. This is often an indication that other memory is corrupt.
at Python.Runtime.Runtime.PyLong_FromLongLong(Int64 value)
at Python.Runtime.PyInt..ctor(Int64 value)
at Python.Runtime.EnumOps`1.int(T value)
--- End of inner exception stack trace ---
at System.RuntimeMethodHandle.InvokeMethod(Object target, Object[] arguments, Signature sig, Boolean constructor)
at System.Reflection.RuntimeMethodInfo.UnsafeInvokeInternal(Object obj, Object[] parameters, Object[] arguments)
at System.Reflection.RuntimeMethodInfo.Invoke(Object obj, BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture)
at Python.Runtime.MethodBinder.Invoke(BorrowedReference inst, BorrowedReference args, BorrowedReference kw, MethodBase info, MethodBase[] methodinfo)
at Python.Runtime.MethodObject.Invoke(BorrowedReference target, BorrowedReference args, BorrowedReference kw, MethodBase info)
at Python.Runtime.MethodBinding.tp_call(BorrowedReference ob, BorrowedReference args, BorrowedReference kw)

Process finished with exit code -532462766
```

filmor:
I can't reproduce this, you have given almost no code here. Does the code run outside of the debugger? Please provide a Minimal, Complete and Verifiable example.

[...]

PR #1811, opened 2022-06-09
Add test for enum int conversion

filmor:

#1810 

#1810

### Any other comments?

...

### Checklist

### DIFF
--- a/tests/test_enum.py
+++ b/tests/test_enum.py
@@ -87,6 +87,11 @@ def test_ulong_enum():
     assert Test.ULongEnum.Two == Test.ULongEnum(2)
 
 
+def test_simple_enum_to_int():
+    from System import DayOfWeek
+    assert int(DayOfWeek.Sunday) == 0
+
+
 def test_long_enum_to_int():
     assert int(Test.LongEnum.Max) == 9223372036854775807
     assert int(Test.LongEnum.Min) == -9223372036854775808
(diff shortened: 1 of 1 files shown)


src/runtime/Util/OpsHelper.cs
```diff
@@ -82,10 +82,20 @@ internal static class EnumOps<T> where T : Enum
     {
         [ForbidPythonThreads]
 #pragma warning disable IDE1006 // Naming Styles - must match Python
-        public static PyInt __int__(T value)
+        public static PyInt __int__(T value) {
```
lostmsu, line 85: The `[ForbidPythonThreads]` attribute should be ensuring that when this method is called, the calling thread holds GIL.

filmor: It's the other way round, right? If this attribute is not given, the GIL is explicitly released. If for some reason it has been released before, it will not be taken here.

lostmsu: The point is: who and how releases GIL when this is called from Python?

filmor: That's a great question, but in the end, the code here is wrong as `ForbidPythonThreads` does not take the GIL, right? After this one is merged, I'd declare the current master rc1 and start off a release-3.0 branch, ok?

lostmsu: @filmor you should take this fix instead. I actually addressed ignoring `ForbidPythonThreads`: [025cc5cdcf17d33f619b8582970b8fe8b28c113e (from https://github.com/losttech/pythonnet.git branch ForbidThreadsFix)](https://github.com/losttech/pythonnet/commit/025cc5cdcf17d33f619b8582970b8fe8b28c113e)

Issue #1831, opened 2022-06-23
Crash on shutdown when embedding Python in C#


BadSingleton:
### Environment

-   Pythonnet version: a80c685d197d47e959d65f82941068286c481276
-   Python version: 3.9.10
-   Operating System: Windows (probably others too)
-   .NET Runtime: mono/Unity

### Details

Using the Python for Unity package with at a commit hash later or equal to the one above, pythonnet causes a crash on Shutdown. I've traced the issue to this PR: https://github.com/pythonnet/pythonnet/pull/1780 which is cause by checking a Python internal without holding the GIL; which the shutdown process isn't doing (or at least not at this point).

filmor:
Is the `Exceptions.ErrorOccurred()` the culprit? Does it work if that whole section is put in a `using (Py.GIL)` scope?

Martin-Molinero:
Just adding to this: we are seeing a similar segmentation fault while shutting down, only while debugging for same reason. Manually shutting down seems to avoid the seg fault https://github.com/QuantConnect/Lean/pull/6439

PR #1832, opened 2022-06-23
Grab the GIL on shutdown when checking for exceptions

BadSingleton:

The GIL isn't taken when being called from the DomainUnload handler. Take it to avoid a crash when calling `Exceptions.ErrorOccurred()`

#1831 

### Checklist

### DIFF
--- a/src/runtime/PythonEngine.cs
+++ b/src/runtime/PythonEngine.cs
@@ -355,11 +355,25 @@ namespace Python.Runtime
             {
                 return;
             }
+
+            // If we're embedded, we need to take the GIL, if it wasn't already 
+            // taken, as we may get called from the DomainUnload handler.
+            var pyGILState = PyGILState.PyGILState_UNLOCKED;
+            if (!Runtime.HostedInPython)
+            {
+                pyGILState = Runtime.PyGILState_Ensure();
+            }
+
             if (Exceptions.ErrorOccurred())
             {
                 throw new InvalidOperationException(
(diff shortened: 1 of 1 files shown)


src/runtime/PythonEngine.cs
```diff
@@ -355,12 +355,26 @@ public static void Shutdown()
...
             }
+
+            // If we're embedded, we need to take the GIL, if it wasn't already 
+            // taken, as we may get called from the DomainUnload handler.
+            var pyGILState = PyGILState.PyGILState_UNLOCKED;
+            if (!Runtime.HostedInPython)
+            {
+                pyGILState = Runtime.PyGILState_Ensure();
+            }
+
```
lostmsu, line 366: Since `Py.GIL()` no longer tries to force runtime initialization, I think the code would be easier to read with a simple `using (Py.GIL()) { ... }` around the `if`. I don't think it is necessary to check if we are hosted in Python

lostmsu (changes requested):

src/runtime/PythonEngine.cs
lostmsu: Another issue is that the code below throws when Python error is set. And that `throw` prevents the GIL release code to be reached making the GIL unreleasable.

filmor: I'll adjust it accordingly s.t. we can tag rc1.

PR #1861, opened 2022-07-08
Fix (U)IntPtr construction

filmor:

Allows creating an `IntPtr` from Python. This is currently impossible, because we rely on automatic conversion (as `IntPtr` is a primitive type) while we don't actually implement one for this type (rightly so!).

The relevant code section is:
https://github.com/pythonnet/pythonnet/blob/87d4db92f6da00b4be27b842ac2ba9875569e04d/src/runtime/Types/ClassObject.cs#L73-L89

#1116 

### Any other comments?

...

### Checklist

### DIFF
--- a/src/runtime/Types/ClassObject.cs
+++ b/src/runtime/Types/ClassObject.cs
@@ -173,9 +173,14 @@ namespace Python.Runtime
                 else if (Runtime.PyInt_Check(op))
                 {
                     long? num = Runtime.PyLong_AsLongLong(op);
-                    if (num is long n)
+                    if (num is long n && n >= Runtime.IntPtrMinValue && n <= Runtime.IntPtrMaxValue)
                     {
                         result = new IntPtr(n);
+                    }
+                    else
+                    {
+                        Exceptions.SetError(Exceptions.OverflowError, "value not in range for IntPtr");
+                        return default;
                     }
                 }
             }
@@ -200,9 +205,14 @@ namespace Python.Runtime
                 else if (Runtime.PyInt_Check(op))
                 {
                     ulong? num = Runtime.PyLong_AsUnsignedLongLong(op);
-                    if (num is ulong n)
+                    if (num is ulong n && n <= Runtime.UIntPtrMaxValue)
                     {
                         result = new UIntPtr(n);
+                    }
+                    else
+                    {
+                        Exceptions.SetError(Exceptions.OverflowError, "value not in range for UIntPtr");
+                        return default;
                     }
                 }
             }
--- a/src/runtime/Types/ClassObject.cs
+++ b/src/runtime/Types/ClassObject.cs
@@ -129,7 +129,7 @@ namespace Python.Runtime
 
         /// <summary>
         /// Create a new Python object for a primitive type
-        /// 
+        ///
         /// The primitive types are Boolean, Byte, SByte, Int16, UInt16,
         /// Int32, UInt32, Int64, UInt64, IntPtr, UIntPtr, Char, Double,
         /// and Single.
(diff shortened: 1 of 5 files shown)


filmor:
I'm not entirely happy with this, yet. I think we might want to do the special-casing in `ClassObject` instead, like it's done for enums and bigints. Thoughts?

lostmsu:
Re: IntPtr

Why can't we just treat IntPtr as non-primitive type? Then its constructor will be invoked by normal means.

lostmsu:
Maybe we should have an explicit set of types for which this constructor masking behavior should be used, instead of relying on `IsPrimitive` which was created for a totally different purpose.

filmor:
> Re: IntPtr
> 
> Why can't we just treat IntPtr as non-primitive type? Then its constructor will be invoked by normal means.

Because then it's always default constructed, apparently calling a cobstructor on a primitive type after it has been created "uninitialised" is a no-op. I can check again, though I'm quite certain that that was among the first things I've tried.

lostmsu:
> apparently calling a cobstructor on a primitive type after it has been created "uninitialised" is a no-op

@filmor this would be highly problematic depending on what types this behavior apply to. If it is just `IsPrimitive` types, then we just need to adjust handling explicitly. However, if it is the case for any value types, we are in trouble.

tests/test_conversion.py
```diff
@@ -679,3 +679,20 @@ def test_iconvertible_conversion():
...
+    assert ob.UIntPtrField == UIntPtr.Zero
+
+    ob.IntPtrField = IntPtr(Int64(-1))
+    assert ob.IntPtrField.ToInt64() == -1
+
+    ob.IntPtrField = IntPtr(Int64(1024))
+    assert ob.IntPtrField.ToInt64() == 1024
+
+    ob.UIntPtrField = UIntPtr(UInt64(1024))
+    assert ob.UIntPtrField.ToUInt64() == 1024
```
lostmsu, line 698: It should also be possible to construct `UIntPtr` and `IntPtr` from Python `int` type with overflow exception.

lostmsu:
@filmor this PR was rushed. There are few artifacts of development left in the final code, the latest commits never reviewed, and the earliest not reviewed completely.

src/runtime/Types/ClassObject.cs
```diff
@@ -112,6 +104,127 @@ static NewReference tp_new_impl(BorrowedReference tp, BorrowedReference args, Bo
...
+                    }
+                    else
+                    {
+                        Exceptions.SetError(Exceptions.OverflowError, "value not in range for UIntPtr");
+                        return default;
+                    }
+                }
+            }
+
+            if (result == null && !Converter.ToManaged(op, type, out result, true))
```
lostmsu, line 220: I think we need a test for `IntPtr` constructor call with this path (e.g. parameter is neither CLR object nor Python `int`

src/runtime/Types/ClassObject.cs
```diff
@@ -112,6 +104,127 @@ static NewReference tp_new_impl(BorrowedReference tp, BorrowedReference args, Bo
...
+        /// All numeric types and Boolean can be handled by a simple
+        /// conversion, (U)IntPtr has to be handled separately as we
+        /// do not want to convert them automically to/from integers.
+        /// </summary>
+        /// <param name="type">.NET type to construct</param>
+        /// <param name="tp">Corresponding Python type</param>
+        /// <param name="args">Constructor arguments</param>
+        private static NewReference NewPrimitive(BorrowedReference tp, BorrowedReference args, Type type)
+        {
+            // TODO: Handle IntPtr
```
lostmsu, line 146: dead TODO?

filmor: True, I'll remove it.

filmor:
> @filmor this PR was rushed. There are few artifacts of development left in the final code, the latest commits never reviewed, and the earliest not reviewed completely. This would be OK if there was no active response for a few days, but simply bypassing review on new code is not a good idea.

It might have been a bit rushed, but that was mostly because I wanted to tackle the other remaining issues (regressions from the constructor refactor) and get a new rc out soon for people to test with. This one would have actually bitten myself for a change if we had released this as stable. :)

Issue #1900, opened 2022-08-08
Cannot convert to managed type


Tiehong:
### Environment

-   Pythonnet version: 3.0.0-rc4
-   Python version: 3.8
-   Operating System: Windows11
-   .NET Runtime: .NET Framework 4.8

### Details

With the following code in C#, when a Python class has __getitem__() defined, it will be converted to an array (System.Object[]) in managed code:

List<object> lst = new List<object>();
                using (Py.GIL())
                {
                    using (var scope = Py.CreateScope())
                    {
                        scope.Set("lst", lst);
                        string text = @"
class DBRow:
    def __init__(self, row, col_names):
        self.row = row
        self.col_names = col_names    
        
    def __getitem__(self, key):   
        if isinstance(key, int):
            return self.row[key]

        if key in self.col_names:
            index = self.col_names.index(key)
            return self.row[index]
        else:
            return None 
            
col_names=['col1', 'col2', 'col3']
row=[1, '2', 3]

r = DBRow(row, col_names)
lst.Add(r)
";

                        var code = PythonEngine.Compile(text);
                        scope.Execute(code);
                    }
                }
                string str = lst[0].GetType().ToString();

lostmsu:
Can you clarify. The bug is that `str` variable ends up being `System.Object[]`?

Tiehong:
Yes, @lostmsu. I expect it to be PyObject. Then I can view the class members/properties in the C# world. System.Object[] does not allow me to do that.

lostmsu:
I believe this should be fixed now.

PR #1901, opened 2022-08-08
Disable implicit seq to array conversion when target type is `object`

lostmsu:

Disabled implicit conversion from Python objects implementing sequence protocol to .NET arrays when the target .NET type is `System.Object`. The conversion is still attempted when the target type is a `System.Array`.

Many Python types implement sequence protocol. In many cases such conversion looses too much important data.

fixes https://github.com/pythonnet/pythonnet/issues/1900

### Checklist


lostmsu:
Wrong branch

PR #1904, opened 2022-08-10
Explicit float and int conversion for builtin types

### LINKED ISSUE #1896: Numerical accuracy when converting from Single to float
### Environment

-   Pythonnet version: 3.0.0rc4
-   Python version: 3.10.5
-   Operating System: Windows 10 20H2
-   .NET Runtime: 6.0.6

### Details

    I'm sorry it's me again with some weird low-level issue. I use a DLL where a function returns a `System.Single`. Its return value is `0.01`. Before pythonnet version 3.0 this got converted to a Python `float` also of value `0.01`. With Pythonnet 3.0 I get a Python `float` `0.009999999776482582` instead. I can reproduce this issue by creating `System.Single` objects.

    ```python
    import clr
    import System

    # Parse a string to avoid conversion from Python float to dotnet
    print(System.Single.Parse('1e-2'))
    # Create a pure dotnet object without conversion back to Python (I think .ToString() gets called here?)
    print(System.Single(0.01))
    # Create a pure dotnet object without conversion back to Python and convert back to Python
    print(System.Single(0.01).MemberwiseClone())
    ```
    ```
    0.009999999776482582
    0.01
    0.009999999776482582
    ```

    With pythonnet 2.5.2 this will return
    ```
    0.01
    0.01
    0.01
    ```

filmor:
Fixes #1896

### DIFF
--- a/src/runtime/Types/ClassBase.cs
+++ b/src/runtime/Types/ClassBase.cs
@@ -529,6 +529,13 @@ namespace Python.Runtime
             return callBinder.Invoke(ob, args, kw);
         }
 
+        static NewReference DoConvert(BorrowedReference ob)
+        {
+            var self = (CLRObject)GetManagedObject(ob)!;
+            using var python = self.inst.ToPython();
+            return python.NewReferenceOrNull();
+        }
+
         static IEnumerable<MethodInfo> GetCallImplementations(Type type)
             => type.GetMethods(BindingFlags.Public | BindingFlags.Instance)
                 .Where(m => m.Name == "__call__");
@@ -563,6 +570,24 @@ namespace Python.Runtime
             if (MpLengthSlot.CanAssign(type.Value))
             {
                 TypeManager.InitializeSlotIfEmpty(pyType, TypeOffset.mp_length, new Interop.B_P(MpLengthSlot.impl), slotsHolder);
+            }
+
+            switch (Type.GetTypeCode(type.Value))
+            {
+                case TypeCode.Byte:
+                case TypeCode.SByte:
+                case TypeCode.UInt16:
+                case TypeCode.UInt32:
+                case TypeCode.UInt64:
+                case TypeCode.Int16:
+                case TypeCode.Int32:
+                case TypeCode.Int64:
+                    TypeManager.InitializeSlotIfEmpty(pyType, TypeOffset.nb_int, new Interop.B_N(DoConvert), slotsHolder);
+                    break;
+                case TypeCode.Double:
+                case TypeCode.Single:
+                    TypeManager.InitializeSlotIfEmpty(pyType, TypeOffset.nb_float, new Interop.B_N(DoConvert), slotsHolder);
+                    break;
             }
         }
 
--- a/src/runtime/Types/ClassBase.cs
+++ b/src/runtime/Types/ClassBase.cs
@@ -536,6 +536,20 @@ namespace Python.Runtime
             return python.NewReferenceOrNull();
         }
 
+        static NewReference DoConvertBooleanInt(BorrowedReference ob)
+        {
+            var self = (CLRObject)GetManagedObject(ob)!;
+            using var python = ((bool)self.inst ? 1 : 0).ToPython();
+            return python.NewReferenceOrNull();
+        }
+
+        static NewReference DoConvertIntFloat(BorrowedReference ob)
+        {
+            var self = (CLRObject)GetManagedObject(ob)!;
+            using var python = (Convert.ToDouble(self.inst)).ToPython();
+            return python.NewReferenceOrNull();
+        }
+
         static IEnumerable<MethodInfo> GetCallImplementations(Type type)
             => type.GetMethods(BindingFlags.Public | BindingFlags.Instance)
                 .Where(m => m.Name == "__call__");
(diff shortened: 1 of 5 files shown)


lostmsu:
I suggest we also add: `int(System.Boolean)`, `float(System.Boolean)`, and `float(System.*Int*)`
Anything else you can think of?

filmor:
There are probably more valid conversions (`Boolean` to `bool` works already due to `__int__`), but I wouldn't want to actually change behaviour now. The implemented conversions are sensible and fix a "regression".

src/runtime/Types/ClassBase.cs
```diff
@@ -529,6 +529,27 @@ static NewReference tp_call_impl(BorrowedReference ob, BorrowedReference args, B
             return callBinder.Invoke(ob, args, kw);
         }
 
+        static NewReference DoConvert(BorrowedReference ob)
+        {
+            var self = (CLRObject)GetManagedObject(ob)!;
+            using var python = self.inst.ToPython();
+            return python.NewReferenceOrNull();
```
lostmsu, line 536: Avoid creating `PyObject` by directly calling `Runtime.PyInt_FromInt64` and/or `Runtime.PyLong_FromUnsignedLongLong`

src/runtime/Types/ClassBase.cs
```diff
@@ -529,6 +529,27 @@ static NewReference tp_call_impl(BorrowedReference ob, BorrowedReference args, B
...
+        }
+
+        static NewReference DoConvertBooleanInt(BorrowedReference ob)
+        {
+            var self = (CLRObject)GetManagedObject(ob)!;
+            using var python = ((bool)self.inst ? 1 : 0).ToPython();
```
lostmsu, line 542: `PyInt_FromInt32`

src/runtime/Types/ClassBase.cs
```diff
@@ -529,6 +529,27 @@ static NewReference tp_call_impl(BorrowedReference ob, BorrowedReference args, B
...
+            return python.NewReferenceOrNull();
+        }
+
+        static NewReference DoConvertIntFloat(BorrowedReference ob)
+        {
+            var self = (CLRObject)GetManagedObject(ob)!;
+            using var python = (Convert.ToDouble(self.inst)).ToPython();
```
lostmsu, line 549: `Runtime.PyFloat_FromDouble`

tests/test_conversion.py
```diff
@@ -720,3 +720,17 @@ def test_intptr_construction():
...
+    from System import (
+        Int64, UInt64, Int32, UInt32, Int16, UInt16, Byte, SByte, Boolean
+    )
+    from System import Double, Single
+
+    assert int(Boolean(True)) == 1
+
+    for t in [Int64, UInt64, Int32, UInt32, Int16, UInt16, Byte, SByte]:
+        assert int(t(127)) == 127
+        assert float(t(127)) == 127.0
```
lostmsu, line 733: Try edge cases (`UInt64.MaxValue`, `Int64.MinValue`) in addition to regular values

lostmsu (changes requested):

src/runtime/Types/ClassBase.cs
filmor: The code is much simpler this way, we can optimise it later if it's necessary.

lostmsu: The problem with this approach is that it will generate .NET exceptions instead of Python errors (which I don't expect to happen here though), so Python will have problems unwinding the stack.

filmor: Okay, got it. The `Converter.To<something>` functions would throw as well (crashing the process), but I think it's guaranteed in the way that they are called that this can't happen.

tests/test_conversion.py
```diff
@@ -720,3 +720,28 @@ def test_intptr_construction():
...
+
+    for t in [Int64, Int32, Int16, SByte]:
+        assert int(t(127)) == 127
+        assert int(t(-127)) == -127
+        assert float(t(127)) == 127.0
+        assert float(t(-127)) == -127.0
+
+    assert int(Int64.MaxValue) == 2**63 - 1
+    assert int(Int64.MinValue) == -2**63
+    assert int(UInt64.MaxValue) == 2**64 - 1
```
lostmsu, line 744: I believe this does not invoke the new code. Only explicit construction.

lostmsu (approved):
See comment about new tests

tests/test_conversion.py
filmor: No, explicit construction is the other dierction. This does call the new code.

lostmsu: Have you actually run it? Accessing `Int64.MaxValue` is reading a field, and upon reading any fields or properties I believe conversions happen (like `Int32` -> `int` or `String` -> `str`), so `Int64.MaxValue` is already returning `int` type to Python. And then `int(int value)` does nothing.

filmor: Oooh, yes, sorry. I have a follow-up PR (#1907), I'll add this. Maybe we should just do away with the implicit conversions altogether :/

filmor: Fixed in https://github.com/pythonnet/pythonnet/pull/1907/commits/c7068a522a8c84584515a0bfc2c97c2bfc5e88b8 (I verified it in IPython this time ;))

lostmsu: That would involve quite a bit of work, and may be slightly unexpected. Most Python code expects `int` and even slight deviation from its behavior will cause unexpected issues.

For example, I am not sure what behavior `int` + `System.Byte` has. E.g. does it convert `Byte` to `int` and performs full addition, or converts `int` to one of .NET integer types, and subsequently wraps on overflow.

Some code will even have explicit `isinstance` checks.

PR #1907, opened 2022-08-12
Fill nb_index slot for integer types

filmor:

Fills slot for `__index__` as a follow up to #1904. In Python >= 3.10, use of `__int__` for quite a few conversion cases has been removed.

-/-

### Any other comments?

...

### Checklist

### DIFF
--- a/src/runtime/Types/OperatorMethod.cs
+++ b/src/runtime/Types/OperatorMethod.cs
@@ -54,6 +54,7 @@ namespace Python.Runtime
 
                 ["__int__"] = new SlotDefinition("__int__", TypeOffset.nb_int),
                 ["__float__"] = new SlotDefinition("__float__", TypeOffset.nb_float),
+                ["__index__"] = new SlotDefinition("__float__", TypeOffset.nb_index),
             };
             ComparisonOpMap = new Dictionary<string, string>
             {
--- a/src/runtime/Types/OperatorMethod.cs
+++ b/src/runtime/Types/OperatorMethod.cs
@@ -54,7 +54,7 @@ namespace Python.Runtime
 
                 ["__int__"] = new SlotDefinition("__int__", TypeOffset.nb_int),
                 ["__float__"] = new SlotDefinition("__float__", TypeOffset.nb_float),
-                ["__index__"] = new SlotDefinition("__float__", TypeOffset.nb_index),
+                ["__index__"] = new SlotDefinition("__index__", TypeOffset.nb_index),
             };
             ComparisonOpMap = new Dictionary<string, string>
             {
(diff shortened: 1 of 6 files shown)


src/runtime/Types/OperatorMethod.cs
```diff
@@ -54,6 +54,7 @@ static OperatorMethod()
 
                 ["__int__"] = new SlotDefinition("__int__", TypeOffset.nb_int),
                 ["__float__"] = new SlotDefinition("__float__", TypeOffset.nb_float),
+                ["__index__"] = new SlotDefinition("__float__", TypeOffset.nb_index),
```
lostmsu, line 57: `SlotDefinition("__float__", ...)` is this a mistake?

src/runtime/Util/OpsHelper.cs
```diff
@@ -82,7 +82,7 @@ internal static class EnumOps<T> where T : Enum
     {
         [ForbidPythonThreads]
 #pragma warning disable IDE1006 // Naming Styles - must match Python
-        public static PyInt __int__(T value)
+        public static PyInt __index__(T value)
```
lostmsu, line 85: Shouldn't `__index__` supplement `__int__` and not completely replace it?

lostmsu (changes requested):

src/runtime/Util/OpsHelper.cs
filmor: According to the docs, `__index__` should be used for lossless conversions, `int(...)` will use it if available.

src/runtime/Types/OperatorMethod.cs
filmor: It is, thanks!

src/runtime/Util/OpsHelper.cs
lostmsu: Are you sure there are no common scenarios where only `__int__` will be looked up?

filmor: No valid scenarios. /We/ do that, but only for our own types :). The correct way to handle things is `operator.index` or `builtins.int`, depending on the expected semantics.

lostmsu: Still not sure what is the point of removing `__int__` here. Is this not a valid implementation of `__int__`? Is there some kind of instruction saying `__int__` should not be set in addition to `__index__`?

filmor: E.g. https://docs.python.org/3/c-api/long.html#c.PyLong_AsLong, functions like these will not work anymore with `__int__`. `__int__` is for lossy conversions, this one is lossless => `__index__` is correct.

filmor: Compromise: I'll add both, is that okay with you?

Issue #1833, opened 2022-06-24
Converting numpy float64 of python float to Decimal is not the same (get rounded in numpy case)


seb5g:
### Environment

-   Pythonnet version: '2.5.1'
-   Python version: 3.8.12
-   Operating System: Win10
-   .NET Runtime: 4.8.4470.0
-   numpy version: 1.21.4

### Details

-   Describe what you were trying to get done.

   I had to convert floats to Decimal but somehow the current Double I got out of Decimal when looking at it using ToDouble would give me only the round part of the double. I figured that my floats were not native python but numpy float64 . This issue happens only with numpy floats..

-   What commands did you run to trigger this issue? If you can provide a
    [Minimal, Complete, and Verifiable example](http://stackoverflow.com/help/mcve)
    this will help us understand the issue.

```python
    import clr
    from System import Decimal
    import numpy as np

    d= 45.5
    assert d == np.array([d])[0]  # should be fine
    assert Decimal(d) == Decimal(np.array([d])[0])  # assertion error 
    
```

lostmsu:
Python.NET 2.x permits all sorts of conversions in attempts to match the function arguments, some of which are surprising. Please, try 3.0 previews of Python.NET. I believe 3.0 will tell you your `Decimal(np.float64)` call has no matching overloads forcing you to convert `np.float64` to either Python `float` or to `System.Double` first.

Also, to save time please add info about investigation from your side. E.g. in the example the equality of two decimals fails. What are their values at the time of comparison? Surely you checked before opening the bug.

filmor:
This is fixed in rc6.

PR #1908, opened 2022-08-12
Implicit float conversion in function calls

filmor:

Uses `__float__` when available if a Python object is passed to a function with a `Double` or `Single` parameter.

#1833

### Any other comments?

This is not yet supported for integer types, which would use `__int__`
or `__index__` (for Python >=3.10), as the respective logic is for some
reason only implemented for `AsLongLong` and `AsLong`, not for the
unsigned counterparts and also not for the `AsSize_t` variants that we
are using.

### Checklist

### DIFF
--- a/src/runtime/Converter.cs
+++ b/src/runtime/Converter.cs
@@ -710,7 +710,7 @@ namespace Python.Runtime
                     {
                         if (Runtime.Is32Bit)
                         {
-                            if (!Runtime.PyLong_Check(value))
+                            if (!Runtime.PyInt_Check(value))
                             {
                                 goto type_error;
                             }
--- a/src/runtime/Runtime.cs
+++ b/src/runtime/Runtime.cs
@@ -1145,9 +1145,7 @@ namespace Python.Runtime
         }
 
         internal static bool PyFloat_Check(BorrowedReference ob)
-        {
-            return PyObject_TYPE(ob) == PyFloatType;
-        }
+            => PyObject_TypeCheck(ob, PyFloatType);
 
         /// <summary>
         /// Return value: New reference.
--- a/src/runtime/Runtime.cs
+++ b/src/runtime/Runtime.cs
@@ -1101,11 +1101,6 @@ namespace Python.Runtime
 
         internal static NewReference PyInt_FromInt64(long value) => PyLong_FromLongLong(value);
 
-        internal static bool PyLong_Check(BorrowedReference ob)
-        {
-            return PyObject_TYPE(ob) == PyLongType;
-        }
-
         internal static NewReference PyLong_FromLongLong(long value) => Delegates.PyLong_FromLongLong(value);
 
 
--- a/tests/test_method.py
+++ b/tests/test_method.py
@@ -1256,3 +1256,41 @@ def test_method_encoding():
 def test_method_with_pointer_array_argument():
     with pytest.raises(TypeError):
         MethodTest.PointerArray([0])
+
+def test_method_call_implicit_conversion():
+
+    class IntAnswerMixin:
+        # For Python >= 3.8
+        def __index__(self):
+            return 42
+
+        # For Python < 3.10
+        def __int__(self):
+            return 42
+
+    class Answer(int, IntAnswerMixin):
+        pass
+
+    class FloatAnswer(float, IntAnswerMixin):
+        def __float__(self):
+            return 42.0
+
+    # TODO: This should also work for integer types but due to some complexities
+    # in the C-API functions (some call __int__/__index__, some don't), it's not
+    # supported, yet.
+    for v in [Answer(), FloatAnswer()]:
+        for t in [System.Double, System.Single]:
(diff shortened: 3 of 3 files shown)


lostmsu:
Implicit casts to float were disabled intentionally. For instance, `float(torch.tensor(42))` is a `float`. However, if implicit casts are enabled, Python.NET can not choose an overload between `float` and `Tensor` if the C# side has a `Tensor` wrapper via a custom codec.

filmor:
The (expected) order here is:

1. Check if a type matches exactly
2. Check if a custom codec matches
3. Check if the parameter type is `Double` or `Single` and if that is the case and the Python parameter is "floaty" (i.e. [PyNumber_Float](https://docs.python.org/3/c-api/number.html#c.PyNumber_Float) returns something `!= NULL`), use the converted value
4. Check if the parameter type is an integer type and do the same with `PyNumber_Long` or `PyNumber_Index`

This should lead to a sensible overload resolution.

lostmsu:
> Check if a type matches exactly

This can never happen with a custom Python type.

Also, we do not really perform overload resolution prioritization like that.

filmor:
Can you please provide a test-case that you expect to fail with this PR? I'm fine with working on the overload resolution, but I think making `CsharpFunc(float_like)` fail is complicating things on the user side in a very error-prone manner.

lostmsu:
I described the test case above.

1. Create an empty class `Tensor` in C#, and a codec along the lines of `return isinstance(o, torch.tensor) ? new Tensor() : null;`
2. Get a C# function with two overloads: `Foo(float)` (or maybe `Foo(double)` and `Foo(Tensor)`
3. Call it from Python with `Foo(torch.tensor(42))`
 
See it pick the overload based on the weather. This is because we actually do not perform any overload resolution, but this behavior prevents user from forcing exact type match when the type on Python side has `__float__`.

lostmsu:
@filmor I think we still should **not** just enable `__float__` for all types that have it, because there are bogus ones that do that and should never be converted without an explicit cast. Implicit conversion should be reserved only for types, that explicitly represent identically sized numbers, e.g. `numpy.float64`, `ctypes.int32`, etc

filmor:
That effectively boils down to just using the "real" `Runtime.PyFloat_Check` instead of our botched version. I'd be completely fine with that.

filmor:
Just to recap: This will now check if before attempting a conversion whether the object identifies as a `float` using `PyFloat_Check`, which has been adjusted in the way @rzindler proposed in #1936. All other changes are just cosmetics (plus one minor bug-fix adjustment for the `Int64` conversion) and test-cases.

filmor:
@lostmsu If this is mergeable for you, I think that would solve the reported issue sufficiently that we can continue with the releases without having to touch method resolution right now (which I also really don't want to). PyTorch's `tensor` does not derive from `float64`, so this particular case should not be a problem.

src/runtime/Converter.cs
```diff
@@ -782,7 +782,7 @@ internal static bool ToPrimitive(BorrowedReference value, Type obType, out objec
                             goto type_error;
                         }
                         double num = Runtime.PyFloat_AsDouble(value);
-                        if (num == -1.0 && Exceptions.ErrorOccurred())
```
lostmsu, line 785: Why remove the `num` check? The doc says that is the value returned on failure, and checking for it saves us a PInvoke trip in `ErrorOcurred`.

lostmsu:
My problem with this approach is that it does not allow us to do:
```csharp
string Foo(float x) => "float32";
string Foo(double x) => "float64";
```
```python
Foo(numpy.float32(42.0))
```

src/runtime/Converter.cs
filmor: Good point, I just wanted to simplify it, but the previous variant is better.

filmor:
> My problem with this approach is that it does not allow us to do:
> 
> ```cs
> string Foo(float x) => "float32";
> string Foo(double x) => "float64";
> ```
> 
> ```python
> Foo(numpy.float32(42.0))
> ```

It does, you just need to be explicit. But I get your point, I just think that the reasonable default is more important than optimising this overload. We can document this particular issue.

rzindler:
Thanks guys!

lostmsu:
I think we can do better than document, and actually special case ctypes, NumPy, PyTorch, and TensorFlow types that have exact precision. Preferably also leave a room for users to special case other precise types.

It is not that hard, and if we don't do it now we'd have to wait until 4.0 to make another breaking change.

filmor:
I'm strongly against special-casing particular library types. If you want to allow proper `float32`, that's an entirely new issue. Allowing modifying primitive type conversion as well.

lostmsu:
That's exactly what I want to do in a week or two.

filmor:
> That's exactly what I want to do in a week or two.

I'll merge this in the meantime. Feel free to revert or adjust the behaviour when you are attempting your refactoring. In this state, I'd be fine to publish it.

Issue #1887, opened 2022-07-25
Cannot convert Numpy Float64 array to double[]


Tarudro:
### Environment

-   Pythonnet version: 3.0.0-rc4
-   Python version: 3.7 (also tested with 3.10)
-   Operating System: Windows 10 Enterprise (10.0.19044 Build 19044)
-   .NET Runtime: .NET 6.0

### Details
  
Hi,
We want to update our PythonEnvironment (conda env) to 3.10 (previously were using 3.7). Therefore I'm trying to update the our used pythonnet version to the latest prerelease one. (3.0.0-rc4 right now) However, when doing so, our unit tests for converting 1, 2 and 3 dimensional numpy (float) arrays to a CLR (double[] type) are breaking. I debugged it and could nail it down to be reproducible using the following code (which executes using version 3.0.0-preview2021-08-03):

```C#
// call PythonEngine.Initialize() and set PythonHome and Paths correctly
using (Py.GIL())
{
    PyModule scope = Py.CreateScope();
    scope.Exec(
        "import numpy as np \n" +
        "arr = np.array([1., 2.])"
    );
    PyObject arr = scope.Get("arr");
    Type dataType = typeof(double[]);
    var array = arr.AsManagedObject(dataType);

    // CollectionAssert.AreEqual(new double[] { 1.0, 2.0 }, (double[])result);
}
```
This throws an `InvalidCastException` in `AsManagedObject`:

> System.InvalidCastException: cannot convert object to target type
>  ---> Python.Runtime.PythonException: 'numpy.float64' value cannot be converted to System.Double

Let me know if you need something else or if I made a mistake. We were initially using pythonnet version 2.5.2 but I tried to refactor and change all the code as required for 3.0.0.
Help is highly appreciated. I couldn't find something/someone else that is having problems converting numpy arrays. Which seems a bit weird to me. 

Thanks for your efforts developing this great and helpful library, keep it up!

Cheers, 
Lukas

filmor:
We don't include non-trivial conversions by default anymore. You should use a codec like this for the conversion: https://github.com/fdieulle/pandasnet/blob/main/dotnet/PandasNet/NumpyCodec.cs

Tarudro:
Hi @filmor,
Thanks for your guidance. I've added the pandasnet code from @fdieulle to my project and I can convert numpy arrays and pandas dataframes once again. 

However, for the conversion of a numpy.float64 I keep ending up in a conversion error within pythonnet. 

[...]

PR #1926, opened 2022-08-29
Added docstrings and minor refactoring

fractus:
* Documentation and formatting Converter.cs (max line length = 100, static member in UPPER_CASE, reorder method for readability, remove many empty lines). I didn't notice any particular guideline, so I used some common rules - no auto-formatter yet.
* Some methods converted to private (main change in `ToManagedValue` now always go via the `ToManaged`) - I expect no performance degradation
* Threw a few more exception when nullable

No - I started debugging a Numpy converter (as suggested in #1887) and ended up formatting this file; I will demonstrate the issue in another PR

### Any other comments?
Some instruction to run the benchmarking would be useful (including the copy PythonRunTime.dll to the baseline folder; otherwise, build fails)

### Checklist

### DIFF
--- a/src/runtime/Converter.cs
+++ b/src/runtime/Converter.cs
@@ -9,127 +9,125 @@ using System.Security;
 namespace Python.Runtime
 {
     /// <summary>
-    /// Performs data conversions between managed types and Python types.
+    /// Performs data conversions between managed types and Python types; it provides both static
+    /// functions and static member grouped grouped at start of the file for readability.
+    /// * Mapping between Python/CLR types; <see cref="GetTypeByAlias"/> and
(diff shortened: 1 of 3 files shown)


src/runtime/Converter.cs
```diff
@@ -9,250 +9,253 @@
...
-            if (op == Runtime.PyStringType)
-                return stringType;
-
-            if (op == Runtime.PyUnicodeType)
-                return stringType;
-
-            if (op == Runtime.PyLongType)
-                return int32Type;
-
-            if (op == Runtime.PyLongType)
```
fractus, line 59: This block can never be reached so it's removed ; PyLongType maps to INT32_TYPE now

PR #1946, opened 2022-09-17
Add __version__

### LINKED ISSUE #1909: Ability to find pythonnet version at runtime / define __version__
### Details

-   I would like to be able to find out which version of pythonnet is installed in a particular runtime environment from the python side. Most packages define a `package.__version__` for that purpose. I couldn't find anything like that in pythonnet.

I tried
```python
    print(pythonnet.__version__)
    print([v for v in dir(pythonnet) if "version" in v])
```
I checked on pythonnet==3.0.0rc4.

filmor:
Add `__version__` property to `pythonnet`, trying first `importlib.metadata`, then `pkg_resources`, then `version.txt` before using `"unknown"`.

Fixes #1909.

@amueller I'm still not sure whether this is needed in modern Python versions (we only support 3.7 upwards nowadays), but since you are also a Bonn Maths alumnus ... ;)

### DIFF
--- a/pythonnet/__init__.py
+++ b/pythonnet/__init__.py
@@ -10,6 +10,21 @@ _LOADER_ASSEMBLY: Optional[clr_loader.Assembly] = None
 _LOADED: bool = False
 
 
+try:
+    import importlib.metadata
+    __version__ = importlib.metadata.version("pythonnet")
+except Exception:
+    try:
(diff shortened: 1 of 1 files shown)


amueller:
I am! How did you know? But I don't think that's a good enough reason to add unnecessary complexity to your code! Sorry I was unaware that `importlib.metadata` does that now.

Issue #1957, opened 2022-09-30
Only `float` type should automatically convert to `System.Double` when target type is `System.Object`


lostmsu:
### Environment

-   Pythonnet version: 3.0.0

### Details

```csharp
var numpy = Py.Import("numpy");
var float64 = numpy.GetAttr("float64");
var float64Value = float64.Invoke(42.ToPython());
Console.WriteLine(float64Value.As<object>().GetType());
```

**Expected:** `Python.Runtime.PyObject`
**Actual**: `System.Double`

PR #1958, opened 2022-09-30
Stricter coversion of primitive Python types to `System.Object`

lostmsu:
When trying to convert to `System.Object`, only convert `float`, `bool`, and `str` to corresponding .NET primitives types. Do **NOT** convert instances of types derived from them.

This changes `Converter` to perform exact type checks when target type is `System.Object`

fixes https://github.com/pythonnet/pythonnet/issues/1957

### Checklist

### DIFF
--- a/src/embed_tests/TestConverter.cs
+++ b/src/embed_tests/TestConverter.cs
@@ -133,6 +133,26 @@ namespace Python.EmbeddingTest
         }
 
         [Test]
+        public void DerivedFloatIsNotFloat()
+        {
+            using var scope = Py.CreateScope();
+            scope.Exec(@"class DerivedFloat(float): pass");
+            using var derivedFloat = scope.Eval("DerivedFloat(42.0)");
+            object dotnetValue = derivedFloat.As<object>();
+            Assert.IsInstanceOf<PyObject>(dotnetValue);
+        }
+
+        [Test]
+        public void DerivedIntIsNotInt()
+        {
+            using var scope = Py.CreateScope();
+            scope.Exec(@"class DerivedInt(int): pass");
+            using var derivedInt = scope.Eval("DerivedInt(42)");
+            object dotnetValue = derivedInt.As<object>();
+            Assert.IsInstanceOf<PyObject>(dotnetValue);
+        }
+
+        [Test]
         public void BigIntExplicit()
         {
             BigInteger val = 42;
--- a/src/runtime/Converter.cs
+++ b/src/runtime/Converter.cs
@@ -361,17 +361,17 @@ namespace Python.Runtime
             // conversions (Python string -> managed string).
             if (obType == objectType)
             {
-                if (Runtime.PyString_Check(value))
+                if (Runtime.PyString_CheckExact(value))
                 {
                     return ToPrimitive(value, stringType, out result, setError);
                 }
 
-                if (Runtime.PyBool_Check(value))
+                if (Runtime.PyBool_CheckExact(value))
                 {
                     return ToPrimitive(value, boolType, out result, setError);
                 }
 
-                if (Runtime.PyFloat_Check(value))
+                if (Runtime.PyFloat_CheckExact(value))
                 {
                     return ToPrimitive(value, doubleType, out result, setError);
                 }
--- a/src/runtime/Runtime.cs
+++ b/src/runtime/Runtime.cs
@@ -1094,8 +1094,13 @@ namespace Python.Runtime
         internal static bool PyInt_Check(BorrowedReference ob)
             => PyObject_TypeCheck(ob, PyLongType);
 
+        internal static bool PyInt_CheckExact(BorrowedReference ob)
+            => PyObject_TypeCheckExact(ob, PyLongType);
+
         internal static bool PyBool_Check(BorrowedReference ob)
             => PyObject_TypeCheck(ob, PyBoolType);
+        internal static bool PyBool_CheckExact(BorrowedReference ob)
+            => PyObject_TypeCheckExact(ob, PyBoolType);
 
         internal static NewReference PyInt_FromInt32(int value) => PyLong_FromLongLong(value);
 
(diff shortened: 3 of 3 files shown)


lostmsu:
This should go into 3.0.0
Did you remove the `release` branch?

lostmsu:
@filmor I don't think you got the scenario. It is not about converting `np.float64` to `System.Double` when the requested type is `System.Double`. It is about what to convert to when the requested type is `System.Object`.

Without this change codecs can't be applied because the conversion to primitives takes priority.

Converting `np.float64` to `System.Double` without `Double` being the desired type is also a loss of information, because at the receiving side you can no longer know that the original value was more specific.

lostmsu:
@filmor any comments?

filmor:
I'm on vacation right now, will check once I'm back. I find it unlikely that I will accept the PR as it stands, but I'll need to be a bit around with it.

lostmsu:
@filmor I think I came up with a good demonstration why this PR should be accepted as is. Without it the following code will work for any numpy value except `numpy.float64` because `numpy.float64` derives from `float` but other NumPy types do not, so `float64` will be converted to `System.Double`, which does not have any NumPy members on it.

```csharp
public static dynamic Centroid(dynamic value)  => value - value.mean();
```

```python
v = numpy.float32(42)
print(Centroid(v)) # OK
v = numpy.float64(42)
print(Centroid(v)) # RuntimeBinderException: System.Double does not have .mean()
```

lostmsu:
@filmor I don't think it applies to primitive types, as
1) they don't really have that many members specific to Python
2) they are present in all Python installations and have natural mapping to .NET types
3) we actually already did that to `int` for much the same reason, and integer objects in this situation are converted to `PyInt` precisely because `int` does not map directly to any of `System.Int*`

filmor:
> I don't think it applies to primitive types, as
> 
>     1. they don't really have that many members specific to Python
I'm not sure what you mean by that. How many would be "too many"? Why

>     2. they are present in all Python installations and have natural mapping to .NET types
Well, strictly speaking only `float <-> System.Double` and `str <-> System.String` are like this. These are sufficiently few to 

>     3. we actually already did that to `int` for much the same reason, and integer objects in this situation are converted to `PyInt` precisely because `int` does not map directly to any of `System.Int*`
Exactly. And I think instead of breaking working code further, we should just make this an option. There are (to me) two kinds of functions for which this is relevant:

1. Functions that pass an element through (e.g. an imaginary `Dictionary<object, object>.GetOrDefault`)
2. Functions that use Python members and thus want to use a `PyObject`

For the latter, I don't see any reason why these shouldn't just take a `PyObject` parameter. I get that `dynamic` makes things more convenient, but a `dynamic Centroid(PyObject obj) { dynamic o = obj; return obj - obj.mean(); }` is not a lot more work. For these, an additional attribute would also do nicely, e.g. `[DisablePythonnetConversions]` or `[PythonnetFunction(conversion = None, gil = true)]`.

For the former, things might be a bit more tricky, but I'd suggest (in general) to just expose more of the conversion machinery publicly and solve this case using a `pythonnet.conversion.disable(method_wrapper)` call (or a corresponding `Python.Runtime` call in C#).

lostmsu:
> I'm not sure what you mean by that. How many would be "too many"?

`float` is a primitive type and only has primitive operations implemented. Addition, subtraction, etc. There are maybe a few 10 ops in total, and the set hasn't changed for 10+ years. NumPy's `float64` though has over 40 members on top of that. And they add new ones now and then.

> For the latter, I don't see any reason why these shouldn't just take a PyObject parameter.

Then you proceed to name a reason. I'd also say that the only good reason to have a `System.Object` parameter in library code is that you either only ever plan to call `.ToString` and/or `.GetHashCode`, or your parameter is actually a `dynamic`. And since the first scenario is not broken after we added GIL to these members, the only remaining scenario for an `System.Object` parameter is made worse by this autoconversion.

filmor:
> > I'm not sure what you mean by that. How many would be "too many"?
> 
> `float` is a primitive type and only has primitive operations implemented. Addition, subtraction, etc. There are maybe a few 10 ops in total, and the set hasn't changed for 10+ years. NumPy's `float64` though has over 40 members on top of that. And they add new ones now and then.

That is not the point. `float` has `is_integer`, `conjugate`, etc. If you want to access those via `dynamic`, it would not work either. Same for all of the `str` members.
 
> > For the latter, I don't see any reason why these shouldn't just take a PyObject parameter.
> 
> Then you proceed to name a reason. I'd also say that the only good reason to have a `System.Object` parameter in library code is that you either only ever plan to call `.ToString` and/or `.GetHashCode`, or your parameter is actually a `dynamic`. And since the first scenario is not broken after we added GIL to these members, the only remaining scenario for a `System.Object` parameter is made worse by this autoconversion.

And I also give a way to make this work by being more explicit. That is the gist of a lot of the changes for Python.NET 3 (that you also introduced yourself!). There is no way to distinguish `dynamic` from `object` parameters, so the proper way to modify this behaviour without breaking existing code is to introduce attributes or runtime configuration.

filmor:
> > And I also give a way to make this work by being more explicit.
> 
> Point is why even do it this way in the first place? What is the scenario where `float` should be passed to `System.Object` parameter and converted to `System.Double`?

We agree here. I just want to not break the behaviour right now, and if we break it, remove all special-casing in one go instead of keeping it for (some) primitives.

If we want to keep things simple, we can make this a runtime configuration parameter.

lostmsu:
@filmor as far as I am concerned, this issue is a consequence of introducing last minute changes to 3.0 before release. And considering its use case rarity, it should be considered a bug in 3.0.0, so should be fixed in 3.0.1. If you are unhappy about interface breakage, 3.0.0 should be yanked.

filmor:
It is not. As said, with the change as it stands, I'm not okay and I think we have discussed it enough now. The correct thing to do would be to not convert at all for `object` parameters, which is a much bigger breakage (compared to 2.5) and can't be introduced in a patch version. I'll draft a PR to enforce this behaviour with runtime configuration.

lostmsu:
@filmor I don't think we need a global toggle for this. There was another option:

> the checks need to move below codec application

That way this could be fixed with a codec.

filmor:
Closing this in favour of #1986.

PR #1986, opened 2022-10-24
Allow decoders to decode Python types derived from primitives

lostmsu:

When a Python instance needs to be converted to `System.Object`, and the Python type is derived from one of the primitives (e.g. `numpy.float64` is derived from `float`), allow a custom decoder to override the default behavior (which would have been converting to `System.Double`).

It addresses the primary concern of https://github.com/pythonnet/pythonnet/issues/1957

### Checklist

### DIFF
--- a/src/embed_tests/Codecs.cs
+++ b/src/embed_tests/Codecs.cs
@@ -356,6 +356,19 @@ DateTimeDecoder.Setup()
         }
 
         [Test]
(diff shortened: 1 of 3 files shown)


filmor:
Looks good to me. I think this is a good stepping stone to eventually retire built-in primitive conversion in favour of using codecs for everything.

PR #2015, opened 2022-11-17
Converting Callable Python to Dotnet Delegate

eirannejad:

When a dotnet method requries a delegate as a parameter, pythonnet fails to convert a python function into a dotnet delegate

```csharp
  public delegate void DelegateHandler(int first, float second);

  public static class DelegateParameterTest
  {
    public static void DoSomething(DelegateHandler handler) => handler(1, 2);
  }
```

⚠️  This will fail:
```python
from PyNetTests import DelegateParameterTest

def Handler(first, second):
	print(first, second)

DelegateParameterTest.DoSomething(Handler)
```

Not that I know of

### Any other comments?

No

### Checklist

### DIFF
--- a/src/runtime/Converter.cs
+++ b/src/runtime/Converter.cs
@@ -282,19 +282,6 @@ namespace Python.Runtime
                 return true;
             }
 
-            if (typeof(System.Delegate).IsAssignableFrom(obType))
-            {
-                result = null;
-                if (Runtime.PyCallable_Check(value) < 1)
-                {
-                    Exceptions.SetError(Exceptions.TypeError, "object must be callable");
-                    return false;
-                }
-
-                result = PythonEngine.DelegateManager.GetDelegate(obType, new PyObject(value));
-                return true;
-            }
-
             if (obType.IsSubclassOf(typeof(PyObject))
                 && !obType.IsAbstract
                 && obType.GetConstructor(new[] { typeof(PyObject) }) is { } ctor)
@@ -337,6 +324,20 @@ namespace Python.Runtime
 
                 default:
                     throw new ArgumentException("We should never receive instances of other managed types");
+            }
+
+            // NOTE: generate a delegate to cast a python callable into a System.Delegate
+            if (typeof(System.Delegate).IsAssignableFrom(obType))
+            {
+                result = null;
+                if (Runtime.PyCallable_Check(value) < 1)
+                {
+                    Exceptions.SetError(Exceptions.TypeError, "object must be callable");
+                    return false;
+                }
+
+                result = PythonEngine.DelegateManager.GetDelegate(obType, new PyObject(value));
+                return true;
             }
 
             if (value == Runtime.PyNone && !obType.IsValueType)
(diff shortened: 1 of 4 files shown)


eirannejad:
Hello friends. I'd love to get a review on this to improve and fix. I'll then create the necessary tests and will mark this as final

eirannejad:
Self notes:
- [ ] Do I need to check the number of parameters on delegate and python callable? See[ this for getting delegate definition from its type](https://stackoverflow.com/a/429564/2350244)

eirannejad:
@filmor Sorry to tag you directly here but all tests are failing on this one commit I added to the master branch. They are related to this analyzer

```
CSC : warning AD0001: Analyzer 'NonCopyable.NonCopyableAnalyzer' threw an exception of type 'System.InvalidCastException' with message 'Unable to cast object of type 'Microsoft.CodeAnalysis.Operations.ArrayInitializerOperation' to type 'Microsoft.CodeAnalysis.Operations.IArrayCreationOperation'.'. [/home/ubuntu/actions-runner/_work/pythonnet/pythonnet/src/testing/Python.Test.csproj]
```

Not sure where the problem is

lostmsu:
@eirannejad
> ... all tests are failing ... They are related to
> CSC : warning AD0001

No they are not. That message is just a warning.

filmor:
Looks good to me. @lostmsu?

src/runtime/MethodBinder.cs
```diff
@@ -888,6 +888,30 @@ internal virtual NewReference Invoke(BorrowedReference inst, BorrowedReference a
...
+            // to both add_ and remove_ event property methods. Therefore once event handler is
+            // added, it can not be removed since the generated delegate on the remove_ method
+            // is not identical to the first.
+            // Here, we ensure given argument to add/remove special method is a Delegate
+            if (binding.info.IsSpecialName &&
+                    (binding.info.Name.StartsWith("add_") || binding.info.Name.StartsWith("remove_")))
+            {
+                if (Runtime.PyTuple_Size(args) != 1)
+                {
+                    throw new Exception("Event handler methods only accept one Delegate argument");
```
lostmsu, line 903: This must be generating `TypeError`

src/runtime/Converter.cs
```diff
@@ -326,6 +326,20 @@ internal static NewReference ToPythonImplicit(object? value)
                     throw new ArgumentException("We should never receive instances of other managed types");
             }
 
+            // NOTE: generate a delegate to cast a python callable into a System.Delegate
```
lostmsu, line 329: This should either be part of `ToPrimitive`, or at very least moved down so that codecs could override this behavior. 

Also @filmor even if this is corrected this is a breaking change as it might fail or change overload resolution.

lostmsu (changes requested):

src/runtime/Converter.cs
lostmsu: `void Foo(PyObject x)` vs `void Foo(System.Action x)` previously would unambiguously resolve to the first variant. With this change it will either fail due to non-unique resolution or attempt to prioritize, which might or might not result in the second one being selected.

Issue #1774, opened 2022-04-27
Classes without __namespace__ cannot be used for reflection


rmadsen-ks:
### Environment

-   Pythonnet version: 3.0.0-preview2022-04-11
-   Python version: 3.10
-   Operating System: MacOS
-   .NET Runtime: .net 6

### Details

My software does lots of reflection on "plugin" types. Simply inheriting from a specific type is enough in C#, but when I use python.net I also have to define `__namespace__`. Since my users generally have the namespace being <Module>.<Class>, there is no logic reason for them to have to do this.

I can see which part of the code causes this:
```cs
    // from MetaType.cs:133 (git: 7247da55c174be1b733e5f9fc4e1c356f6c42dc4)
    if (clsDict.HasKey("__assembly__") || clsDict.HasKey("__namespace__"))
    {
            return TypeManager.CreateSubType(name, base_type, clsDict);
    }
```

I am wonder if we could broaden the cases where TypeManager.CreateSubType is used. For example in cases where the super class is a type created with CreateSubType, or just any .NET class in general. Maybe it could also be if the super class implements a specific interface.

    _TODO_

Add a way to not have to specify __namespace__ in the class in order for reflection to work.

lostmsu:
The logic reason is that classes with `__namespace__` are .NET classes as well as Python classes, and classes without it are just Python classes. This is by design.

filmor:
@lostmsu The point is that it could be enough that the base class is a .NET class. Since we control the `__new__` in this case anyhow, this wouldn't be a huge change (a breaking one, yes, but I think the new behaviour might be less surprising).

lostmsu:
@filmor are you suggesting all new derived classes to be .NET classes by default? What would be the way to turn that off?

rmadsen-ks:
Maybe I can add that if the base class is abstract, inheriting from it without __namespace__ causes a an exception during construction: "Python.Runtime.PythonException: cannot instantiate abstract class". A similar issue happens with interfaces.

Also the python object seems to loose all memory of its type once it gets converted back into an C# object, casting it to its base type.

Both these issues could be fixed by this change.

phuongdtvd:
@rmadsen-ks  Hi can I ask if you find any workaround for the problem? I’m having the exact same problem of inheriting from an abstract base class then converting back into C# loses all info. GetType.GetFields() won’t work at all and I’m honestly stuck right now.

rmadsen-ks:
[...]

Issue #1768, opened 2022-04-13
Add Support for .NET Attributes when Creating .NET Classes 


rmadsen-ks:
### Environment

-   Pythonnet version: 3.x
-   Python version: 3.x
-   Operating System: Any
-   .NET Runtime: .Net Framework 4 / .Net6

### Details

I want to be able to specify attributes on my class defined in python. This is essential for my specific use case, but I imagine generally useful. I can come up with 3 different situations where it could be useful to add attributes:

1. Attributes on types - This can be done with an annotation
2. Attributes on methods - This can be done with an annotation
3. Attributes on properties - It seems clr.clrproperty needs to be modified to support it.

```python
### from test_clrmethod.py

#1. Attaching attributes to types:
@clr.attribute(DebuggerDisplayAttribute, "ExampleClrClass: X={X}")
class ExampleClrClass(System.Object):
    __namespace__ = "PyTest"
    def __init__(self):
        self._x = 3

    # 2. Attaching attributes to methods
    @clr.clrmethod(int, [int])
    @clr.attribute(DescriptionAttribute, "Test method")
    def test(self, x):
        return x*2

    def get_X(self):
        return self._x
    def set_X(self, value):
        self._x = value

   # 3. Attaching attributes to properties
    X = clr.clrproperty(int, get_X, set_X)\
        .attribute(DescriptionAttribute, "Gets or sets X")\
        .attribute(BrowsableAttribute, True)

    @clr.clrproperty(int)
    def Y(self):
        return self._x * 2
```

I can probably implement this myself, but before making a PR I want to know if it will be accepted or it is a bad idea.

lostmsu:
Sounds good to me. But I believe at least your variant of syntax would be extremely tricky to implement for classes (methods are fine), because Python would first create class `ExampleClrClass`, and only then apply the attribute to it. But in .NET you can't create a class and then change it by adding attributes - you'd have to create a new class, which in your example would have the same name (which might not be a problem as long as you don't use reflection to find it).

BTW, the code in question is in `CreateDerivedType` function.

Finally, for the scenario you described, why can't you just override `ToString` function?

rmadsen-ks:
@lostmsu, you are right the '@'-attribute won't work for classes. 

So maybe something like this then?

```py
 __attributes__ = [(DebuggerDisplayAttribute, "ExampleClrClass: X={X}")]
```

Your final comment I assume is referring to DebuggerDisplayAttribute, but I just used that as an example.

[...]

PR #2019, opened 2022-11-18
Object identity, Interfaces and Attributes

### LINKED ISSUE #1776: Generic virtual method in baseclass causes InvalidProgramException
### Environment

-   Pythonnet version: 3.0.0-preview2022-04-11
-   Python version: 3.10
-   Operating System: MacOS and Ubuntu 2021.4
-   .NET Runtime: .net 6

### Details

If I create a python class which inherits from a C# class that has a virtual generic method and exception is thrown while the class is being defined. Note, not during object instanciation time.

To reproduce, take the following C# class:
```cs
    public abstract class GenericVirtualMethodTest
    {
        public virtual Q VirtMethod<Q>(Q arg1)
        {
            return arg1;
        }
    }
```

Then inherit from it as in this test:

```py
def test_virtual_generic_method():
    class OverloadingSubclass(GenericVirtualMethodTest):
        __namespace__ = "test_virtual_generic_method_cls"
    obj = OverloadingSubclass()
    assert obj.VirtMethod[int](5) == 5
```

The result is the following exception (Actually a C# exception inside a python exception inside a C# exception):
```
Python.Runtime.PythonException : An attempt was made to load a program with an incorrect format.
 (0x8007000B)
  File "/Users/bruger/code/work/pythonnet/src/../tests/test_subclass.py", line 268, in test_virtual_generic_method
    class OverloadingSubclass(GenericVirtualMethodTest):
  File "<string>", line 19, in <module>
   at Python.Runtime.PythonException.ThrowLastAsClrException() in /Users/bruger/code/work/pythonnet/src/runtime/PythonException.cs:line 53
   at Python.Runtime.PythonException.ThrowIfIsNull(NewReference& ob) in /Users/bruger/code/work/pythonnet/src/runtime/PythonException.cs:line 455
   at Python.Runtime.PythonEngine.RunString(String code, BorrowedReference globals, BorrowedReference locals, RunFlagType flag) in /Users/bruger/code/work/pythonnet/src/runtime/PythonEngine.cs:line 625
   at Python.Runtime.PythonEngine.Exec(String code, PyDict globals, PyObject locals) in /Users/bruger/code/work/pythonnet/src/runtime/PythonEngine.cs:line 530
   at Python.PythonTestsRunner.PythonTestRunner.RunPythonTest(String testFile, String testName) in /Users/bruger/code/work/pythonnet/src/python_tests_runner/PythonTestRunner.cs:line 64
-----

One or more child tests had errors
  Exception doesn't have a stacktrace
```

### Proposed Fix

I understand that it is probably a bit complicated to actually override a virtual generic method in python code and maybe it is not even necesary. So maybe the best solution is to just ignore virtual methods in this case.

rmadsen-ks:
This PR contains multiple improvements:
- All classes inheriting from a .net class has their own .net type. This means they can be returned and given as arguments without loosing their original identity (and properties).
- Interfaces can be implemented. Also classes only inheriting interfaces are supported.
- Attributes can be used in multiple different flavors. 
   - Class attributes using e.g `@attribute(Browsable(False))`, or __clr_attributes__ = [Browsable(False), ...]
   - Method attributes for methods exported to .NET.
       ``` 
           @attribute(DisplayName("X"))
           @method()
           def X(self):
                 pass
       ```
   - Property attributes using `property(Double,0.0).add_attribute(DisplayName("X")`
  

Close #1776, #1774, #1768 


src/runtime/Types/MetaType.cs
```diff
@@ -79,61 +82,83 @@ public static NewReference tp_new(BorrowedReference tp, BorrowedReference args,
...
+
+                if (clsDict.HasKey("__assembly__") || clsDict.HasKey("__namespace__")
+                                                   || (ctor != null))
                 {
-                    return TypeManager.CreateSubType(name, base_type, clsDict);
+                    if (!clsDict.HasKey("__namespace__"))
+                    {
+                        clsDict["__namespace__"] =
+                            (clsDict["__module__"].ToString()).ToPython();
+                    }
```
lostmsu, line 156: Can't introduce breaking changes

src/python_tests_runner/PythonTestRunner.cs
```diff
@@ -17,6 +17,8 @@ public class PythonTestRunner
         [OneTimeSetUp]
         public void SetUp()
         {
+            Python.Runtime.Runtime.PythonDLL =
+                "C:\\Python37.2\\python37.dll";
```
lostmsu, line 21: clean up the history

src/runtime/PythonTypes/PythonTypeAttribute.cs
```diff
@@ -0,0 +1,27 @@
...
+    public string TypeName { get; }
+
+    /// <summary> Importable module name. </summary>
+    public string Module { get; }
+
+    /// <summary>
+    /// Creates a new instance of PythonTypeAttribute.
+    /// </summary>
+    /// <param name="pyTypeModule"></param>
+    /// <param name="pyTypeName"></param>
```
lostmsu, line 21: This XML doc does not have any useful information.

lostmsu:
Please, split smaller fixes into individual PRs (e.g. #1776).  I'd rather have #1774 also have its own PR as the code does not seem to be interdependent.

rmadsen-ks:
@lostmsu, ok, I understand.  I just want to make sure you agree with the intentions here before I spend a lot of work trying to separate things out. What about the support for Attributes and interfaces? Are you ok with adding something like that?


Anyway, it sounds like you want #1774 implemented in a specific way? Maybe it is easier that it is not me who develops that then? 

I'll try to take #1776 and create a separate PR for it.

lostmsu:
Should be superseded by smaller changes

PR #2026, opened 2022-11-29
1776 Generic Virtual Method Causes Invalid Program: Fix + test

### LINKED ISSUE #1776: Generic virtual method in baseclass causes InvalidProgramException

rmadsen-ks:
- If a method is virtual AND generic, it cannot be overridden by the python class. Hence the method call is deferred to the base class.
- Added a unit test which verifies this behavior is now working.

Close #1776 

### Checklist

### DIFF
--- a/src/python_tests_runner/PythonTestRunner.cs
+++ b/src/python_tests_runner/PythonTestRunner.cs
@@ -35,7 +35,6 @@ namespace Python.PythonTestsRunner
             // Add the test that you want to debug here.
             yield return new[] { "test_indexer", "test_boolean_indexer" };
             yield return new[] { "test_delegate", "test_bool_delegate" };
-            yield return new[] { "test_subclass", "test_virtual_generic_method" };
         }
 
         /// <summary>
--- a/src/python_tests_runner/PythonTestRunner.cs
+++ b/src/python_tests_runner/PythonTestRunner.cs
@@ -35,6 +35,7 @@ namespace Python.PythonTestsRunner
             // Add the test that you want to debug here.
             yield return new[] { "test_indexer", "test_boolean_indexer" };
             yield return new[] { "test_delegate", "test_bool_delegate" };
+            yield return new[] { "test_subclass", "test_virtual_generic_method" };
         }
 
         /// <summary>
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@ This document follows the conventions laid out in [Keep a CHANGELOG][].
 
 ### Fixed
 
+- Fixed error occuring when inheriting a class containing a virtual generic method.
+
 ## [3.0.1](https://github.com/pythonnet/pythonnet/releases/tag/v3.0.1) - 2022-11-03
 
 ### Added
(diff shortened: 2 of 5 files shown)


src/python_tests_runner/PythonTestRunner.cs
```diff
@@ -35,6 +35,7 @@ static IEnumerable<string[]> PythonTestCases()
             // Add the test that you want to debug here.
             yield return new[] { "test_indexer", "test_boolean_indexer" };
             yield return new[] { "test_delegate", "test_bool_delegate" };
+            yield return new[] { "test_subclass", "test_virtual_generic_method" };
```
lostmsu, line 38: I see test failures. What happens if you drop this change? I could take a look at the crash later if we create a work item for it.

PR #2028, opened 2022-11-29
1783 Implement Interface And Inherit Class

### LINKED ISSUE #1783: Cannot implement an interface that extends another interface
### Environment

-   Pythonnet version: 3.0.0a2
-   Python version: 3.8.10
-   Operating System: Ubuntu 20.04.4 LTS (docker run --rm -it ubuntu:20.04)
-   .NET Runtime: .NET Core 6.0.202

### Details

    Attempting to implement interfaces defined in .NET. Works fine for simple interfaces but I receive a weird error when the interface extends IDisposable

```python
from clr_loader import get_coreclr
rt = get_coreclr("conf.json")
from pythonnet import set_runtime
set_runtime(rt)
import clr

import System
import System.Buffers


class MyDisposable(System.IDisposable):
    __namespace__ = "MyNamespace"

    def Dispose(self) -> None:
        print("Dispose")

md = MyDisposable()
md.Dispose()


class MyBuffer(System.Buffers.IMemoryOwner):
    __namespace__ = "MyNamespace"


    def Dispose(self) -> None:
        print("Dispose")


mb = MyBuffer()
mb.Dispose()    
```

conf.json:
```json
{
  "runtimeOptions": {
    "tfm": "net6.0",
    "framework": {
      "name": "Microsoft.NETCore.App",
      "version": "6.0.4"
    }
  }
}
```

```
(venv) root@ca20f043eef5:/t# python demo.py
Dispose
  File "demo.py", line 21, in <module>
    class MyBuffer(System.Buffers.IMemoryOwner):
TypeError: Method 'Dispose' in type 'MyNamespace.MyBuffer' from assembly 'Python.Runtime.Dynamic, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null' does not have an implementation.
```

rmadsen-ks:
Added support for multiple inheritance when inheriting from one base class and/or multiple interfaces.

Added a unit test verifying that it works with a simple class and interface. This unit test would previously have failed since multiple types are in the class super class list.

Close #1783

### Checklist

### DIFF
--- a/src/runtime/Types/MetaType.cs
+++ b/src/runtime/Types/MetaType.cs
@@ -1,5 +1,8 @@
 using System;
+using System.Collections.Generic;
 using System.Diagnostics;
+using System.Linq;
+using System.Reflection;
 using System.Runtime.InteropServices;
 using System.Runtime.Serialization;
 
@@ -79,40 +82,46 @@ namespace Python.Runtime
             BorrowedReference bases = Runtime.PyTuple_GetItem(args, 1);
             BorrowedReference dict = Runtime.PyTuple_GetItem(args, 2);
 
-            // We do not support multiple inheritance, so the bases argument
-            // should be a 1-item tuple containing the type we are subtyping.
-            // That type must itself have a managed implementation. We check
-            // that by making sure its metatype is the CLR metatype.
+            // Extract interface types and base class types.
+            var interfaces = new List<Type>();
+            var baseType = new List<ClassBase>();
 
-            if (Runtime.PyTuple_Size(bases) != 1)
+            var cnt = Runtime.PyTuple_GetSize(bases);
+
+            for (uint i = 0; i < cnt; i++)
+            {
+                var base_type2 = Runtime.PyTuple_GetItem(bases, (int)i);
+                var cb2 = (ClassBase) GetManagedObject(base_type2);
+                if (cb2 != null)
+                {
+                    if (cb2.type.Valid && cb2.type.Value.IsInterface)
+                        interfaces.Add(cb2.type.Value);
+                    else baseType.Add(cb2);
+                }
+            }
+            // if the base type count is 0, there might still be interfaces to implement.
+            if (baseType.Count == 0)
+            {
+                baseType.Add(new ClassBase(typeof(object)));
+            }
+
+            // Multiple inheritance is not supported, unless the other types are interfaces
+            if (baseType.Count > 1)
             {
                 return Exceptions.RaiseTypeError("cannot use multiple inheritance with managed classes");
             }
 
-            BorrowedReference base_type = Runtime.PyTuple_GetItem(bases, 0);
-            BorrowedReference mt = Runtime.PyObject_TYPE(base_type);
-
-            if (!(mt == PyCLRMetaType || mt == Runtime.PyTypeType))
+            var cb = baseType[0];
+            try
             {
-                return Exceptions.RaiseTypeError("invalid metatype");
+                if (!cb.CanSubclass())
+                {
+                    return Exceptions.RaiseTypeError("delegates, enums and array types cannot be subclassed");
+                }
             }
-
-            // Ensure that the reflected type is appropriate for subclassing,
-            // disallowing subclassing of delegates, enums and array types.
-
-            if (GetManagedObject(base_type) is ClassBase cb)
+            catch (SerializationException)
             {
-                try
-                {
-                    if (!cb.CanSubclass())
-                    {
-                        return Exceptions.RaiseTypeError("delegates, enums and array types cannot be subclassed");
-                    }
-                }
-                catch (SerializationException)
-                {
-                    return Exceptions.RaiseTypeError($"Underlying C# Base class {cb.type} has been deleted");
-                }
+                return Exceptions.RaiseTypeError($"Underlying C# Base class {cb.type} has been deleted");
             }
 
             BorrowedReference slots = Runtime.PyDict_GetItem(dict, PyIdentifier.__slots__);
@@ -121,7 +130,8 @@ namespace Python.Runtime
                 return Exceptions.RaiseTypeError("subclasses of managed classes do not support __slots__");
             }
 
-            // If __assembly__ or __namespace__ are in the class dictionary then create
+            // If the base class has a parameterless constructor, or
+            // if __assembly__ or __namespace__ are in the class dictionary then create
             // a managed sub type.
             // This creates a new managed type that can be used from .net to call back
             // into python.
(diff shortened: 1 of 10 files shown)


src/runtime/Types/MetaType.cs
```diff
@@ -79,49 +82,56 @@ public static NewReference tp_new(BorrowedReference tp, BorrowedReference args,
...
 
-            if (Runtime.PyTuple_Size(bases) != 1)
+            var cnt = Runtime.PyTuple_GetSize(bases);
+
+            for (uint i = 0; i < cnt; i++)
             {
-                return Exceptions.RaiseTypeError("cannot use multiple inheritance with managed classes");
+                var base_type2 = Runtime.PyTuple_GetItem(bases, (int)i);
+                var cb2 = (ClassBase) GetManagedObject(base_type2);
+                if (cb2 != null)
```
lostmsu, line 95: You can use the same pattern to check for `null` as in the original code to clear compiler warning

e.g. `if (GetManagedObject(base_type2) is ClassBase cb2) ...`

src/runtime/Types/MetaType.cs
```diff
@@ -79,49 +82,56 @@ public static NewReference tp_new(BorrowedReference tp, BorrowedReference args,
...
+                {
+                    if (cb2.type.Valid && cb2.type.Value.IsInterface)
+                        interfaces.Add(cb2.type.Value);
+                    else baseType.Add(cb2);
```
lostmsu, line 99: IMHO, it is better to enforce the order to be `BaseClass, IInterface1, IInterface2`, etc.

I believe the current code has a bug down below because it uses `base_type` which is not the `BaseClass` if the order is different.

lostmsu (changes requested):

src/runtime/Types/MetaType.cs
rmadsen-ks: I have only added this functionality for types which goes through  TypeManager.CreateSubType, so it should only affect that branch.

for code flows not invoking CreateSubType, the behavior willl be exactly the same as before.

src/runtime/Types/MetaType.cs
```diff
@@ -79,49 +82,59 @@ public static NewReference tp_new(BorrowedReference tp, BorrowedReference args,
...
+            }
 
             BorrowedReference slots = Runtime.PyDict_GetItem(dict, PyIdentifier.__slots__);
             if (slots != null)
             {
                 return Exceptions.RaiseTypeError("subclasses of managed classes do not support __slots__");
             }
 
-            // If __assembly__ or __namespace__ are in the class dictionary then create
+            // If the base class has a parameterless constructor, or
```
lostmsu, line 136: > If the base class has a parameterless constructor

??? What is this for? If that's right, it seems breaking to me.

src/runtime/Types/MetaType.cs
```diff
@@ -130,10 +143,12 @@ public static NewReference tp_new(BorrowedReference tp, BorrowedReference args,
                 using var clsDict = new PyDict(dict);
                 if (clsDict.HasKey("__assembly__") || clsDict.HasKey("__namespace__"))
                 {
-                    return TypeManager.CreateSubType(name, base_type, clsDict);
+                    return TypeManager.CreateSubType(name, baseTypes[0], interfaces, clsDict);
                 }
             }
 
+            var base_type = Runtime.PyTuple_GetItem(bases, 0);
```
lostmsu, line 150: `bases` here can be empty since you removed check above.

src/runtime/Types/MetaType.cs
```diff
@@ -79,49 +82,59 @@ public static NewReference tp_new(BorrowedReference tp, BorrowedReference args,
             BorrowedReference bases = Runtime.PyTuple_GetItem(args, 1);
             BorrowedReference dict = Runtime.PyTuple_GetItem(args, 2);
 
-            // We do not support multiple inheritance, so the bases argument
-            // should be a 1-item tuple containing the type we are subtyping.
-            // That type must itself have a managed implementation. We check
-            // that by making sure its metatype is the CLR metatype.
+            // Extract interface types and base class types.
+            var interfaces = new List<Type>();
```
lostmsu, line 86: IMHO, repeating the same base interface multiple times should be an error.

And it might be, add a test.

src/runtime/Types/MetaType.cs
```diff
@@ -79,49 +82,59 @@ public static NewReference tp_new(BorrowedReference tp, BorrowedReference args,
...
+            {
+                baseTypes.Add(new ClassBase(typeof(object)));
             }
 
-            // Ensure that the reflected type is appropriate for subclassing,
-            // disallowing subclassing of delegates, enums and array types.
+            // Multiple inheritance is not supported, unless the other types are interfaces
+            if (baseTypes.Count > 1)
+            {
+                return Exceptions.RaiseTypeError("cannot use multiple inheritance with managed classes");
```
lostmsu, line 114: Since you collected multiple base types, perhaps it would be useful to list them in the error message in case user did not notice which ones are classes.

src/runtime/Types/MetaType.cs
```diff
@@ -79,49 +82,59 @@ public static NewReference tp_new(BorrowedReference tp, BorrowedReference args,
...
 
-            if (!(mt == PyCLRMetaType || mt == Runtime.PyTypeType))
+            for (nint i = 0; i < baseClassCount; i++)
             {
-                return Exceptions.RaiseTypeError("invalid metatype");
+                var baseTypeIt = Runtime.PyTuple_GetItem(bases, (int)i);
+
+                if (GetManagedObject(baseTypeIt) is ClassBase classBaseIt)
+                {
+                    if (classBaseIt.type.Valid && classBaseIt.type.Value.IsInterface)
```
lostmsu, line 100: Invalid `ClassBase` instances should also raise an error regardless of anything else.

lostmsu (changes requested):
Please, rebase before posting a new iteration.

P.S. sorry for the late response

src/runtime/Types/MetaType.cs
rmadsen-ks: How can bases be length 0? This should only be called when a C# type is subclassed, right? In this case, it should at least be length 1.

src/runtime/Types/MetaType.cs
rmadsen-ks: got it, adding exceptions.

src/runtime/Types/MetaType.cs
rmadsen-ks: Ok, I'll include this.

src/runtime/Types/MetaType.cs
```diff
@@ -79,49 +82,59 @@ public static NewReference tp_new(BorrowedReference tp, BorrowedReference args,
...
+                        interfaces.Add(classBaseIt.type.Value);
+                    else baseTypes.Add(classBaseIt);
+                }
```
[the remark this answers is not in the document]
rmadsen-ks: Added exceptions here in the latest version.

src/runtime/Types/MetaType.cs
rmadsen-ks: Test added.

rmadsen-ks:
@lostmsu,

Thanks for doing the review.

I have rebased and fixed things you mentioned.

Will you mark the conversations resolved where appropriate?

lostmsu (approved):

lostmsu:
Thank you for the contribution!

PR #2032, opened 2022-11-30
1768 .NET Attribute Support: Fix

### LINKED ISSUE #1768: Add Support for .NET Attributes when Creating .NET Classes

rmadsen-ks:
- Added support for specifying .NET attributes using __clr_attributes__. Classes, properties and methods can specify this value.

- Added support for Attribute shorthand without  the "Attribute" ending.

- Added unit test verifying the behavior

Close #1768

### Checklist

### DIFF
--- a/src/runtime/Types/ClassDerived.cs
+++ b/src/runtime/Types/ClassDerived.cs
@@ -243,6 +243,18 @@ namespace Python.Runtime
             if (py_dict != null && Runtime.PyDict_Check(py_dict))
             {
                 using var dict = new PyDict(py_dict);
+
+                // if there are any attributes, add them to the type.
+                if (dict.HasKey("__clr_attributes__"))
+                {
+                    using var attributes = new PyList(dict["__clr_attributes__"]);
+                    foreach (var attr in attributes)
+                    {
+                        var builder = GetAttributeBuilder(attr);
+                        typeBuilder.SetCustomAttribute(builder);
+                    }
+                }
+
                 using var keys = dict.Keys();
                 foreach (PyObject pyKey in keys)
                 {
@@ -502,6 +514,16 @@ namespace Python.Runtime
                 returnType,
                 argTypes.ToArray());
 
+            if (func.HasAttr("__clr_attributes__"))
+            {
+                using var attributes = new PyList(func.GetAttr("__clr_attributes__"));
+                foreach (var attr in attributes)
+                {
+                    var builder = GetAttributeBuilder(attr);
+                    methodBuilder.SetCustomAttribute(builder);
+                }
+            }
+
             ILGenerator il = methodBuilder.GetILGenerator();
 
             il.DeclareLocal(typeof(object[]));
@@ -600,6 +622,16 @@ namespace Python.Runtime
                 propertyType,
                 null);
 
+            if (func.HasAttr("__clr_attributes__"))
+            {
+                using var attributes = new PyList(func.GetAttr("__clr_attributes__"));
+                foreach (var attr in attributes)
+                {
+                    var builder = GetAttributeBuilder(attr);
+                    propertyBuilder.SetCustomAttribute(builder);
+                }
+            }
+
             if (func.HasAttr("fget"))
             {
                 using var pyfget = func.GetAttr("fget");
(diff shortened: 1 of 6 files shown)


src/runtime/Types/ClassDerived.cs
```diff
@@ -678,6 +710,127 @@ private static ModuleBuilder GetModuleBuilder(string assemblyName, string module
...
+        {
+            // this is a bit complicated because we want to support both unnamed and named arguments
+            // in C# there is a discrepancy between property setters and named argument wrt attributes
+            // in python there is no difference. But luckily there is rarely a conflict, so we can treat
+            // named arguments and named properties or fields the same way.
+            var tp = new PyTuple(attr);
+            Type attribute = tp[0].As<Type>();
+            if (typeof(Attribute).IsAssignableFrom(attribute) == false)
+            {
+                throw new Exception($"This type cannot be used as an attribute type: {attribute}");
```
lostmsu, line 727: The exception type must be specific.

Also, please add a test for this scenario to ensure it does not lead to a fatal error. You might need to raise a Python exception instead.

src/runtime/Types/ClassDerived.cs
```diff
@@ -678,6 +710,127 @@ private static ModuleBuilder GetModuleBuilder(string assemblyName, string module
...
+        /// Builds a C# attribute based on a python tuple definition (Attibute Type, *attributeArgs, ** attributeKwArgs)
+        /// </summary>
+        static CustomAttributeBuilder GetAttributeBuilder(PyObject attr)
+        {
+            // this is a bit complicated because we want to support both unnamed and named arguments
+            // in C# there is a discrepancy between property setters and named argument wrt attributes
+            // in python there is no difference. But luckily there is rarely a conflict, so we can treat
+            // named arguments and named properties or fields the same way.
+            var tp = new PyTuple(attr);
+            Type attribute = tp[0].As<Type>();
```
lostmsu, line 724: Same as below, you need to handle the scenario where the object passed here is not actually a `Type` properly.

src/runtime/Types/ClassDerived.cs
```diff
@@ -678,6 +710,127 @@ private static ModuleBuilder GetModuleBuilder(string assemblyName, string module
...
+            if (typeof(Attribute).IsAssignableFrom(attribute) == false)
+            {
+                throw new Exception($"This type cannot be used as an attribute type: {attribute}");
+            }
+
+            var args = tp[1].As<PyObject[]>();
+            var dict = new PyDict(tp[2]);
```
lostmsu, line 731: Same.

src/runtime/Types/ClassDerived.cs
```diff
@@ -678,6 +710,127 @@ private static ModuleBuilder GetModuleBuilder(string assemblyName, string module
...
+            var dict2 = new Dictionary<string, PyObject>();
```
lostmsu, line 732: Please name variables with descriptive names.

src/runtime/Types/ClassDerived.cs
```diff
@@ -678,6 +710,127 @@ private static ModuleBuilder GetModuleBuilder(string assemblyName, string module
...
+            foreach (var key in dict.Keys())
+            {
+                var k = key.As<string>();
+                dict2[k] = dict[key];
+            }
+
+            var allConstructors = attribute.GetConstructors();
+            foreach (var constructor in allConstructors)
```
lostmsu, line 740: This reimplements our overloading logic. We definitely do not want to do that. Instead the Python side should use some simple mechanism to set the fields. For example, there could be a `.with(attributes: dict[str, Any]) -> Self` method, that would set values of the fields/properties from the dictionary.

E.g. `DllImport("kernel32").with({"CharSet": Auto})` (just an example, a different way to do this could also be acceptable).

src/runtime/Types/ModuleObject.cs
```diff
@@ -132,6 +132,25 @@ public NewReference GetAttribute(string name, bool guess)
...
+                          + "   import {1}\n"
+                          + "   return ({0}, args, kwargs)\n";
+                str = string.Format(str, qname2, _namespace, name);
+                PythonEngine.Exec(str);
+                using var attributeData = PythonEngine.Eval(name + "__AttributeBuilder");
+                var pinnedAttributeData = attributeData.NewReferenceOrNull();
+                StoreAttribute(name, pinnedAttributeData.Borrow());
+                return new NewReference(pinnedAttributeData);
+            }
+
```
lostmsu, line 153: This seems overly complicated. Can't you just do what the code above does for type lookup, and just add another `StoreAttribute` for the alias?

lostmsu (changes requested):

src/runtime/Types/ModuleObject.cs
rmadsen-ks: So for e.g BrowsableAttribute, this generates a function which takes the arguments and stores it as a tuple along with the type name. That way we can reconstruct which arguments are used for creating the attribute when later generating the code.

The reason it needs to be a function is that whenever e.g `Browsable`is invoked, it needs to generate a new tuple because the arguments are not the same every time.

lostmsu:
@rmadsen-ks btw generally speaking the code down at that level should not be using `PyObject` unless strictly necessary. Instead, `BorrowedReference` and `NewReference` should be used in conjunction with `Runtime.Py...` methods.

src/runtime/Types/ClassDerived.cs
rmadsen-ks: I think that might make sense for properties. I just wanted the call to be analogous to how it is invoked in C#, since people will always be finding the C# documentation/ code samples for how those attributes are used. e.g
`
[AttributeTest("UnnamedArg", NamedArg : "x", NamedProperty="y")]
`
This would now be used like this:
```py
AttributeTest("UnnamedArg", NamedArg = "x", NamedProperty="y")
```

But you are suggesting this instead?

```py
AttributeTest("UnnamedArg", NamedArg = "x").with({"NamedProperty": "y"})
```

lostmsu: @rmadsen-ks correct. The reason is that we don't want to maintain yet another overload resolution handler. That's one of the most complicated parts of C# that we try to mirror.

rmadsen-ks: Ok, so you are not talking so much about changing these specific lines, but rather changing how the data for that is stored?

I am not sure we can use the other overload resolution handler, because that is for runtime overload resolution, right? this is compile/load time and the data we can put into those attributes is limited.

rmadsen-ks:
> @rmadsen-ks btw generally speaking the code down at that level should not be using `PyObject` unless strictly necessary. Instead, `BorrowedReference` and `NewReference` should be used in conjunction with `Runtime.Py...` methods.

Which code do you mean specifically? This code used PyDict before I arrived.. :)

src/runtime/Types/ClassDerived.cs
lostmsu: @rmadsen-ks since this is running in Python, it is runtime.

I believe you should be able to use `MethodBinder.Bind` which returns a `Binding`.

rmadsen-ks: Unless I'm completely confused, The selected overload gets compiled to IL code by System.Reflection.Emit. That is what the returned CustomAttributeBuilder is for.

lostmsu: @rmadsen-ks not sure I understand this comment.

`MethodBinder.Bind` performs basically what your loop performs: selects the method from the list of overloads, and returns it + the arguments for it as .NET would see them.

lostmsu:
Closing as the requested changes were not implemented

PR #2049, opened 2022-12-15
Added a decoder to be able to handle Func<> parameters

Romout:

It's a decoder to ease use of C# Func<> method arguments. Previously, it was necessary to cast the python lambda expression to System.Func[...] which is very cumbersome for a script-user.

Was not my intention.

### Any other comments?

The solution I used is quite complex as it needs the create wrapper types for each Func-Type-Variant. Code creation is implemented using the IL Code Generator, types are cached, generic type arguments are again converted using ToPython()

I hope my change is of some help!

### Checklist

(mongoDB error)

### DIFF
--- a/src/runtime/Codecs/FuncDecoder.cs
+++ b/src/runtime/Codecs/FuncDecoder.cs
@@ -1,0 +1,176 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Reflection.Emit;
+using System.Reflection;
+using System.Text;
+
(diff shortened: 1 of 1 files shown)


koubaa:
FWIW I attempted a function codec before - and it was mentioned in the PR that it should have used code generation to be more correct: https://github.com/pythonnet/codecs/pull/1

Romout:
I don't know why the build for 3.7 on 32bit didn't succeed. Can someone help with that?

Issue #1945, opened 2022-09-17
Deriving from .NET types in Python does not inherit constructor


lostmsu:
```python
from System import Object

class L1(Object):
  __namespace__ = 'TestNS'
  def __init__(self):
    super().__init__()
    print('42')

class L2(L1):
  __namespace__ = 'TestNS'

o = L2() # 42 is not printed
```

## Workaround
Explicitly redefine `__init__` in `L2`:

```python
class L3(L1):
  __namespace__ = 'TestNS'
  def __init__(self):
    super().__init__()

o = L3() # prints '42'
```

Issue #1787, opened 2022-05-09
Can not call inherited method on nested class, when another nested class inherits from it and defines an override.


Algoryx-NilsHjelte:
This is a very specific setup with classes and inheritance, but is exactly what happened to us and where pythonnet does not do the right thing.

It is closely related to #1420 but a slightly different configuration.

Here is code to reproduce the issue:

```cs
namespace pythonnet_testing {
  public class Creator {
    public static object CreateObject() =>
      new BaseClass.NestedClass();
  }

  public class BaseClass {


    public virtual string GetSignal() =>
      "hello";

    public class NestedClass : BaseClass {
      // NOTE: This is using the default `GetSignal` implementation
    }

    // NOTE: This class inherits from the nested class, and overrides the method in the base class
    // When this class is defined, we can no longer call `GetSignal` on instances of `NestedClass`
    // ==> `Unhandled exception. Python.Runtime.PythonException: 'NestedClass' object has no attribute 'GetSignal'`
    public class InheritFromNested : BaseClass.NestedClass {
      public override string GetSignal() =>
        "override";
    }
  }


  class Program {
    static void Main(string[] args) {
      var pyCode = @"
import pythonnet_testing
o = pythonnet_testing.Creator.CreateObject()
print(f'Method call: {o.GetSignal()}')
      ";
      Runtime.PythonDLL = "libpython3.9.dylib";
      PythonEngine.Initialize();

      using (Py.GIL()) {
        PythonEngine.Exec(pyCode);
      }

      PythonEngine.Shutdown();
    }
  }
}
```

NOTE: If I remove the `InheritFromNested` class, then the call to `o.GetSignal` will resolve the proper inherited virtual base method.

Issue #1107, opened 2020-04-08
Derived class in Python not being called by C#


jeffzi19:
Hello - this is my first post so apologies if this is the wrong forum.

I'm trying to define a base class in C# and derive from that class in python / pythonnet.  The C# class is very basic:

    public class Order
    {
        public string symbol;

        public virtual void Evaluate( Bar b )
        {
            Console.WriteLine( "from Order.Evaluate() - hello!" );
        }
    }

In python I have the specific order class defined as:

class LimitOrder(Order):

    def __init__(self):
        self.symbol = "SQ"

    def Evaluate( self, bar ):
        print( "LimitOrder.Evaluate() - yay python!" )

In C# I have an OMS class with a method: public void AddOrder( Order order )

When I pass in LimitOrder and call a method in the OMS that calls Evaluate the Order class Evaluate method is called instead of the LimitOrder Evaluate method.

I'm using python 3.6.7 and VS 2019 / .NET 4.7.2

In searching for this I'm only seeing 1 instance of a similar question being asked but no solid response.  The Evaluate() method (and entire Order class for that matter) can be abstract but when I try that I get an error creating the LimitOrder object saying I can't create an abstract class.

Any help would be appreciated.  Thanks in advance.

mhsmith:
When making a Python subclass of a C# base class, you must include a `__namespace__` attribute. For example, see https://github.com/pythonnet/pythonnet/issues/567#issuecomment-341960560.

Issue #674, opened 2018-05-18
implement a C# Interface in Python for .NET


willhyper:
### Environment

-   Pythonnet version: 2.3.0
-   Python version: 3.6.5
-   Operating System: Windows

### Details

trying to implement a C# Interface in Python for .NET

This post exemplifies my exact problem. 
https://stackoverflow.com/questions/49736531/implement-a-c-sharp-interface-in-python-for-net#

However, I found relatively few discussion and solution. Forgive me raise the issue here. :)
I am not sure if this is an intent for pythonnet not to support. If so, reasons are welcomed.
Otherwise, I find the practice (implementing an interface) is common, and so why there is no support.
My use case is, I am given a dll. The interface is defined there. I want to develop Python class that implements the interface. Without this approach viable, I foresee my developed application would then be half in C# and half in Python, which I consider ugly.

denfromufa:
@willhyper is this something what you are looking for? scroll to the end of the issue with comments how this was implemented in pywebview:

https://github.com/pythonnet/pythonnet/issues/567

willhyper:
@denfromufa Awesome! __namespace__ is the missing single line that solves my question!

mkimberling:
I was recently confused by this and had to do alot of googling to find the answer on Stack Overflow.  

It would be a great improvement to have

- an example in the docs
- an example of implementing the interface in test_interface.py

PR #2055, opened 2022-12-19
Improved Type Interop with CLR

eirannejad:

This PR addresses these issues:
- Current pythonnet does not allow subclassing from CLR Abstract classes
- Running python script with types containing `__namespace__`, more than once, would throw `Type Exists` exception
- Calling `super().VirtualMethod()` inside a python-derived virtual method, would result in an infinite loop

Due to the nature of the changes, this is a single PR instead of multiple smaller ones.

After PR merge:
- Python types can derive from CLR abstract classes. As before, abstract methods that are not implemented on the python type, will dynamically throw `NotImplemented` exception.
- Specifying `__namespace__` is not required to created a `ClassDerived` anymore. All python derived classes are now represented by `ClassDerived` in managed memory.
- `ClassDerived` now caches generated types based on namespace, type name, and chain of base classes. This avoids regenerating types or throwing `Type Already Exists` exceptions. A new format for generated type names include full name of base class and interfaces in the typename. The naming format still ends in the python type name to be backward compatible and passes the unit tests
    - **Example:** Full name of derived clr type for python type named `SubClass`, deriving from `BaseClass` and implementing `BaseInterface1` and  `BaseInterface2`:
      `Python.Runtime.Dynamic.BaseClass__BaseInterface1__BaseInterface2__main__SubClass`
- Support for one base class and multiple interfaces (Merged some ideas and code from #2028 - Potential merge conflicts)
- Derived class can call chosen constructors using `super().__init__()` pattern.

```python
class SubClass(BaseClassA):
   def __init__(self, v):
       super().__init__(v)
```
- Derived class can call base virtual methods using `super().method()` pattern.
```python
class SubClass(BaseClassA):
    def DoWorkWith(self, value):
        r = super().DoWorkWith(value) + 10
        return r
```
- Improved virtual method and getter/setter routing. Custom attributes (`OriginalMethod` and `RedirectedMethod`) are now used to mark the original and redirected virtual methods. The method name format for `OriginalMethod ` is changed to `$"_BASEVIRTUAL__{name}"` so it can handle calling original methods on base classes that does not match the name of current class. Previously this was not working.
- Other misc refactoring and improvements. No change in behavior. 

Potentially (no tests have been done to ensure these issues are resolved)
- #1945
- #1107
- #1787

### Any other comments?

...

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ This document follows the conventions laid out in [Keep a CHANGELOG][].
 
 ### Added
 
+- Added support for subclassing abstract classes ([#2055][p2055]).
+
 ### Changed
 
 ### Fixed
@@ -940,3 +942,4 @@ This version improves performance on benchmarks significantly compared to 2.3.
 [i238]: https://github.com/pythonnet/pythonnet/issues/238
 [i1481]: https://github.com/pythonnet/pythonnet/issues/1481
 [i1672]: https://github.com/pythonnet/pythonnet/pull/1672
+[p2055]: https://github.com/pythonnet/pythonnet/pull/2055
--- a/src/runtime/ClassManager.cs
+++ b/src/runtime/ClassManager.cs
@@ -185,11 +185,14 @@ namespace Python.Runtime
             else if (type == typeof(Exception) ||
                      type.IsSubclassOf(typeof(Exception)))
             {
-                impl = new ExceptionClassObject(type);
+                if (PythonDerivedType.IsPythonDerivedType(type))
+                    impl = new ExceptionClassDerivedObject(type);
+                else
+                    impl = new ExceptionClassObject(type);
             }
 
 #pragma warning disable CS0618 // Type or member is obsolete. OK for internal use.
-            else if (null != PythonDerivedType.GetPyObjField(type))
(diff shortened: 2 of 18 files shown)


eirannejad:
Hello. Would anyone help me get some information on the error on these test fails? I can not run the `dotnet test --runtime any-x64 --logger "console;verbosity=detailed" src/embed_tests/` test on my macOS machine as it throws lots of exceptions. Not sure what sort of environment it needs for testing locally

filmor:
If it fails to run locally, check the output. You are probably missing the `PYTHONNET_PYDLL` export. For me, it actually runs through on .NET Core but is *extremely* slow on Mono. That's why the respective CI jobs are timing out.

lostmsu:
Please, rebase on master if you are still willing to get this PR in.

eirannejad:
@lostmsu Sure thing. Working on getting `master` merged in

mhsmith:
As part of this work, it would be good for Python subclassing of CLR types to be properly documented (https://github.com/pythonnet/pythonnet/issues/674#issuecomment-1630858068).

eirannejad:
Closing this PR as it is too old. I have merged the 3.0.3 master into my fork so I can create new updated PRs

PR #2239, opened 2023-09-18
use enum name in repr

### LINKED ISSUE #2238: Enum __repr__ string could be improved
### Environment

-   Pythonnet version: 3.1.0
-   Python version: 3.11
-   Operating System: Windows
-   .NET Runtime: .NET FW

### Details

- Many .NET APIs use enum values as method return types or properties. In an interactive shell, it may be useful to see the enum value with `repr`. This is how IronPython seems to do things:

**IronPython 2.7.4** (2.7.0.40) on .NET 4.0.30319.42000 (64-bit)
>>> import System
>>> System.DayOfWeek.Monday
System.DayOfWeek.Monday

**Python 3.11.4** (tags/v3.11.4:d2340ef, Jun  7 2023, 05:45:37) [MSC v.1934 64 bit (AMD64)] on win32
>>> import clr; import System
>>> System.DayOfWeek.Monday
<System.DayOfWeek object at 0x0000024D5BAC7300>

koubaa:

Proposal to fix #2238

## Before the changes:

```
Python 3.11.4 (tags/v3.11.4:d2340ef, Jun  7 2023, 05:45:37) [MSC v.1934 64 bit (AMD64)] on win32
>>> import clr; import System
>>> System.DayOfWeek.Monday
<System.DayOfWeek object at 0x0000024D5BAC7300>
```

## IronPython (as a reference):

```
IronPython 2.7.4 (2.7.0.40) on .NET 4.0.30319.42000 (64-bit)
>>> import System
>>> System.DayOfWeek.Monday
System.DayOfWeek.Monday
```

## With these changes:

```
Python 3.11.4 (tags/v3.11.4:d2340ef, Jun  7 2023, 05:45:37) [MSC v.1934 64 bit (AMD64)] on win32
>>> import clr; import System
>>> System.DayOfWeek.Monday
DayOfWeek.Monday
```

### Checklist

### DIFF
--- a/src/runtime/Types/ClassObject.cs
+++ b/src/runtime/Types/ClassObject.cs
@@ -49,6 +49,26 @@ namespace Python.Runtime
 
 
         /// <summary>
+        /// given an enum, write a __repr__ string formatted in the same
+        /// way as a python repr string. Something like:
+        ///   '&lt;Color.GREEN: 2&gt;';
+        /// with a binary value for [Flags] enums
+        /// </summary>
+        /// <param name="inst">Instace of the enum object</param>
+        /// <returns></returns>
+        private static string getEnumReprString(object inst)
+        {
+            var obType = inst.GetType();
+
+            var isFlags = obType.IsFlagsEnum();
+            var intValue = Convert.ToInt32(inst);
+            var intStr = isFlags ? "0x" + intValue.ToString("X4") : intValue.ToString();
+
+            var repr = $"<{obType.Name}.{inst}: {intStr}>";
+            return repr;
+        }
+
+        /// <summary>
         /// ClassObject __repr__ implementation.
         /// </summary>
         public new static NewReference tp_repr(BorrowedReference ob)
@@ -57,13 +77,9 @@ namespace Python.Runtime
             {
                 return Exceptions.RaiseTypeError("invalid object");
             }
-
-            var inst = co.inst;
-            var obType = inst.GetType();
-            if (obType.IsEnum)
+            if (co.inst.GetType().IsEnum)
             {
-                var repr = string.Format("{0}.{1}", obType.Name, inst.ToString());
-                return Runtime.PyString_FromString(repr);
+                return Runtime.PyString_FromString(getEnumReprString(co.inst));
             }
 
             return ClassBase.tp_repr(ob);
--- a/src/runtime/Types/ClassObject.cs
+++ b/src/runtime/Types/ClassObject.cs
@@ -1,7 +1,9 @@
 using System;
 using System.Diagnostics;
+using System.Globalization;
 using System.Linq;
 using System.Reflection;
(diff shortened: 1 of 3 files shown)


lostmsu:
Can you please make the format similar to Python's `enum.Enum`:
```python
>>> from enum import Enum
>>> class Color(Enum):
...   RED = 1
...   GREEN = 2
...
>>> repr(Color.GREEN)
'<Color.GREEN: 2>'
```

koubaa:
@lostmsu Done - but I'm not really that happy with the Flags case.  `ToString()` seems to make some choices about how to represent the value. Do you mean to *only* show the hex value if its Flags, instead of trying to show both integer and name?

src/runtime/Types/ClassObject.cs
```diff
@@ -48,6 +48,43 @@ internal NewReference GetDocString()
...
+        /// with a binary value for [Flags] enums
+        /// </summary>
+        /// <param name="inst">Instace of the enum object</param>
+        /// <returns></returns>
+        private static string getEnumReprString(object inst)
+        {
+            var obType = inst.GetType();
+
+            var isFlags = obType.IsFlagsEnum();
+            var intValue = Convert.ToInt32(inst);
```
lostmsu, line 64: This will fail for `Int64` and `UInt64`-based enums. Also you might have to separately handle `UInt64` case.

src/runtime/Types/ClassObject.cs
```diff
@@ -48,6 +48,43 @@ internal NewReference GetDocString()
...
+            var intStr = isFlags ? "0x" + intValue.ToString("X4") : intValue.ToString();
```
lostmsu, line 65: `X4` is insufficient for `Int32`, `X8` is required.
IMHO, it's best to set X according to the underlying integer type. Then it will immediately be clear from the `repr` what integer type backs the enum.

lostmsu (changes requested):

koubaa:
@lostmsu done. 

The switch case isn't so pretty, but I couldn't think of a way to improve it. here's a minimal compiler explorer link with that logic if anyone wants to try and see if they can improve it:  https://godbolt.org/z/dYG6sdnKo

src/runtime/Types/ClassObject.cs
```diff
@@ -48,6 +48,82 @@ internal NewReference GetDocString()
...
 
+        /// <summary>
+        /// given an enum, write a __repr__ string formatted in the same
+        /// way as a python repr string. Something like:
+        ///   '&lt;Color.GREEN: 2&gt;';
+        /// with a binary value for [Flags] enums
+        /// </summary>
+        /// <param name="inst">Instace of the enum object</param>
+        /// <returns></returns>
+        private static string getEnumReprString(object inst)
```
lostmsu, line 59: NIT: CamelCase

PR #2249, opened 2023-09-27
Python 3.12

filmor:
- Add type offsets for 3.12 and update interop generation
- Drop unused custom incref/decref
- Add 3.12 to CI and metadata


filmor:
@lostmsu Can you have a look in particular at https://github.com/pythonnet/pythonnet/commit/1bcf21869948ee545dafdb05c5161529972e8d2f? I'm not quite sure what exactly you are testing here but it doesn't work on 3.12 anymore.

lostmsu:
Please take the commit from https://github.com/losttech/pythonnet/tree/instanceless-explicit to change the test. If it still does not raise in 3.12, keep the version guard and refer to https://github.com/python/cpython/issues/101578

filmor:
Same result. It actually does raise, but the reference check for `PyErr` fails and it doesn't convert it.

lostmsu:
On Python 3.12 I think it should raise an instance of `StopIteration`. Prior to 3.12 that would be an exception with type `StopIteration` but without instance, which is what decoder was handling. E.g. if you replace the check for the wrapper with a check for `StopIteration` instance in 3.12, that should do it.

filmor:
Well, that's not what this one is testing, though. I'll keep it deactivated for now. If the rest is fine with you, I'll merge it and cut a new version (maybe after having another look at #2245).

lostmsu:
> Well, that's not what this one is testing, though.

👍I can fix the test as I outlined above. My understanding is that 3.12 always normalizes exceptions while 3.11 and before did not.

E.g. `throw new PythonException(stopIteration, value: null, traceback: null)` would end up raising a Python error with type `StopIteration` and `None` value in 3.11 and before, while 3.12 will automatically call `StopIteration` constructor in this case.

The decoder is needed in 3.11 and before because otherwise you have no way in C# to handle `None` exception value.

PR #2260, opened 2023-10-06
Revert "Only run docs CI/main CI for corresponding changes"

filmor:
Reverts pythonnet/pythonnet#2257


filmor:
@lostmsu I was to quick, this is not correct. The documentation is generated from code for the reference docs, so any code change might lead to docs changes.

Issue #2269, opened 2023-10-18
ClassManager cache hash problem


rayfalling:
### Environment

-   Pythonnet version: 3.x
-   Python version: 3.9
-   Operating System: Windows11
-   .NET Runtime: Unity Mono

### Details

-   Describe what you were trying to get done.
 
Abstract generic classes under C# are repeatedly added to the Cache of ClassManager, which will cause a crash when  calling RestoreRuntimeData()

![image](https://github.com/pythonnet/pythonnet/assets/13343271/25e7ba83-da7e-4945-aca3-44238808bb18)

rayfalling:
Maybe the MaybeType should set custom hash function?

rayfalling:
![image](https://github.com/pythonnet/pythonnet/assets/13343271/2e164294-d1be-4a8d-84a6-881c36e73e76)
Could I use AssemblyQualifiedName to get hash?

PR #2270, opened 2023-10-18
Explicitly compare MaybeType objects by name

filmor:

As observed by @rayfalling, multiple instances of `ReflectedType` values might be created in the cache, leading to crashes during stashing.

@rayfalling, could you provide a test-case for this?

#2269 

### Any other comments?

...

### Checklist


rayfalling:
What I meet is a Unity Project with FingerGesture plugin
I will prepare the Unity Project tomorrow
It's nine o'clock in the evening now, I need to get off work first :)

rayfalling:
This is the test-case.
You can open it in Unity 2019.4.38f1
[minimalTest.zip](https://github.com/pythonnet/pythonnet/files/13039217/minimalTest.zip)

lostmsu:
Do you know circumstances under which this happens? What is the scenario when we can have two different `MaybeType` with the same name?

PR #2303, opened 2023-12-26
Improve loader robustness 

sefgit:

Ignore and print loader error and proceed with loading if possible.

There are cases where
 "import <namespace>" in python failed even when clr.AddReference() succeeded, especially when loading "manipulated (probably protected/obfuscated)" C# DLL.

unknown

### Checklist

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -86,3 +86,4 @@
 -   ([@gpetrou](https://github.com/gpetrou))
 -   Ehsan Iran-Nejad ([@eirannejad](https://github.com/eirannejad))
 -   ([@legomanww](https://github.com/legomanww))
+-   Effendi Soewono ([@sefgit](https://github.com/sefgit))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ This document follows the conventions laid out in [Keep a CHANGELOG][].
 
 ### Fixed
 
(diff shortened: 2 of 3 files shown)


filmor:
Can you give an example and a testcase for this? If we fail to load types, I'd rather have `clr.AddReference` error out entirely, and it should certainly not write to stderr.

sefgit:
> Can you give an example and a testcase for this? If we fail to load types, I'd rather have `clr.AddReference` error out entirely, and it should certainly not write to stderr.

There is no error in calling clr.AddReference().
It's just not possible to call "import <namespace>" afterwards.
I can't share the dll here publicly but I could provide you its URL thru private message if possible.

lostmsu:
Console is not a proper way to report errors. Either `AddReference` should throw or (if that can't be allowed due to backward compatibility concerns if it would fail too often) we should provide a generic DLL loading error handler (e.g. `event OnLoadError`). Dumping such errors to `Debug.WriteLine` or Python's warning module would also be acceptable, but a generic handler would still be needed.

filmor:
There are direct C-API functions to issue warnings: https://docs.python.org/3/c-api/exceptions.html#issuing-warnings

sefgit:
A verbose flag to enable/disable warning/error console print would be an alternative 
Silent failure is totally confusing amd desperating to handle with.
Maybe:
`clr.VERBOSE = True`
or even introduce verbose level ?

lostmsu:
Closing as the requested changes were not implemented.

PR #2322, opened 2024-02-15
`IComparable` and `IEquatable` implementations

lostmsu:
`IComparable` and `IEquatable` implementations for `PyInt`, `PyFloat`, and `PyString` for primitive .NET types

Let's one do `PyInt.CompareTo(42)` and similar things

N/A

### Any other comments?

Should we take GIL in these methods? Might be useful for `Array.Sort` when called from Python.

On a more general topic: should we have a way for Python to call a .NET method and have it not release the GIL for compatibility with 3rd party (to Python.NET) libraries that assume they can call back any methods they wish at any time. Because having explicit GIL acquisition in `CompareTo` to use with `Array.Sort` would kill performance. E.g. something like:

```python
with clr.HoldGIL():
  Array.Sort(dotnet_array_of_python_objects)
```

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,9 @@ This document follows the conventions laid out in [Keep a CHANGELOG][].
 
 -   Added `ToPythonAs<T>()` extension method to allow for explicit conversion using a specific type. ([#2311][i2311])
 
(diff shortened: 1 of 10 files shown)


filmor (approved):

filmor:
Regarding the GIL, maybe we could a custom thread local ref count could speed things up?

PR #2327, opened 2024-02-23
 #2240 Recursion error when doing reversed python operations on C# types

### LINKED ISSUE #2240: RecursionError when substracting System Decimal from Python Decimal
### Environment

-   Pythonnet version: Version: 3.0.1
-   Python version: 3.10
-   Operating System: Any (Windows/Linux)
-   .NET Runtime: 6.0.0

### Details

-   Trying to substract System Decimal from Python Decimal (the other way works fine)

    _TODO_

- import both system decimal and python decimal
- create a separate variable from each type
- subtract System Decimal from Python Decimal

```python
    from System import Decimal as SysDecimal
    from decimal import Decimal as PyDecimal
    sys_dec = SysDecimal(1)
    py_dec = PyDecimal(1)
    print(str(sys_dec - py_dec )) # works fine
    print(str(py_dec - sys_dec)) # crashes
```

-   error thrown
RecursionError: maximum recursion depth exceeded while calling a Python object

gertdreyer:
### What does this fix? Explain your changes.

Reverse operators were not implemented correctly for C# types which resulted in Recursion errors when embedding C# in Python. Changed the model binding process to add reverse operators for all applicable types. Added a flag to MethodBinder that signals the operation is a reversed operation and that the arguments to the operator needs to be reversed.

Added tests and an wrapper object for int without overloaded operator methods and a wrapper codec to test the reverse operations (and forward operator on the same type for safety).
...

This should resolve #2240.
...

### Any other comments?

...

### Checklist

### DIFF
--- a/src/embed_tests/TestOperator.cs
+++ b/src/embed_tests/TestOperator.cs
@@ -41,6 +41,17 @@ namespace Python.EmbeddingTest
                 _value = value;
             }
 
+            public override int GetHashCode()
+            {
+                return unchecked(65535 + _value.GetHashCode());
+            }
+
+            public override bool Equals(object obj)
+            {
+                return obj is OwnInt @object &&
+                       Num == @object.Num;
+            }
+
             public static OwnInt operator -(OwnInt p1, OwnInt p2)
             {
                 return new OwnInt(p1._value - p2._value);
@@ -125,14 +136,8 @@ namespace Python.EmbeddingTest
                 return objectType.Name == "int" && targetType == typeof(OwnInt);
             }
 
-            public bool TryDecode<T>(PyObject pyObj, out T? value)
+            public bool TryDecode<T>(PyObject pyObj, out T value)
             {
-                if (pyObj.PyType.Name != "int" || typeof(T) != typeof(OwnInt))
-                {
-                    value = default(T);
-                    return false;
-                }
-
                 value = (T)(object)new OwnInt(pyObj.As<int>());
                 return true;
             }
--- a/src/embed_tests/TestOperator.cs
+++ b/src/embed_tests/TestOperator.cs
@@ -25,9 +25,9 @@ namespace Python.EmbeddingTest
         }
 
         // Mock Integer class to test math ops on non-native dotnet types
-        public struct OwnInt
+        public readonly struct OwnInt
         {
-            private int _value;
+            private readonly int _value;
 
             public int Num => _value;
 
--- a/src/embed_tests/TestOperator.cs
+++ b/src/embed_tests/TestOperator.cs
@@ -15,12 +15,127 @@ namespace Python.EmbeddingTest
         public void SetUp()
         {
             PythonEngine.Initialize();
+            OwnIntCodec.Setup();
         }
 
         [OneTimeTearDown]
         public void Dispose()
         {
             PythonEngine.Shutdown();
+        }
+
+        // Mock Integer class to test math ops on non-native dotnet types
+        public struct OwnInt
+        {
+            private int _value;
+
+            public int Num => _value;
+
+            public OwnInt()
+            {
+                _value = 0;
+            }
+
+            public OwnInt(int value)
+            {
+                _value = value;
+            }
+
+            public static OwnInt operator -(OwnInt p1, OwnInt p2)
+            {
+                return new OwnInt(p1._value - p2._value);
+            }
+
+            public static OwnInt operator +(OwnInt p1, OwnInt p2)
+            {
+                return new OwnInt(p1._value + p2._value);
+            }
+
+            public static OwnInt operator *(OwnInt p1, OwnInt p2)
+            {
+                return new OwnInt(p1._value * p2._value);
+            }
+
+            public static OwnInt operator /(OwnInt p1, OwnInt p2)
+            {
+                return new OwnInt(p1._value / p2._value);
+            }
+
+            public static OwnInt operator %(OwnInt p1, OwnInt p2)
+            {
+                return new OwnInt(p1._value % p2._value);
+            }
+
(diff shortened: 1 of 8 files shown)


src/embed_tests/TestOperator.cs
```diff
@@ -23,6 +24,120 @@ public void Dispose()
             PythonEngine.Shutdown();
         }
 
+        // Mock Integer class to test math ops on non-native dotnet types
+        public struct OwnInt
```
lostmsu, line 28: I know this is test code, but please override `Equals` and `GetHashCode` to clear the warning.

src/embed_tests/TestOperator.cs
```diff
@@ -23,6 +24,120 @@ public void Dispose()
...
+            {
+                PyObjectConversions.RegisterDecoder(new OwnIntCodec());
+            }
+
+            public bool CanDecode(PyType objectType, Type targetType)
+            {
+                return objectType.Name == "int" && targetType == typeof(OwnInt);
+            }
+
+            public bool TryDecode<T>(PyObject pyObj, out T? value)
```
lostmsu, line 128: Please add `#nullable enable` to the file and fix the warnings or remove this nullable annotation.

src/embed_tests/TestOperator.cs
```diff
@@ -23,6 +24,120 @@ public void Dispose()
...
+            {
+                if (pyObj.PyType.Name != "int" || typeof(T) != typeof(OwnInt))
```
lostmsu, line 130: I don't think you need this check.

src/runtime/MethodBinder.cs
```diff
@@ -363,10 +371,10 @@ public MismatchedMethod(Exception exception, MethodBase mb)
                 _methods = GetMethods();
             }
 
-            return Bind(inst, args, kwargDict, _methods, matchGenerics: true);
+            return Bind(inst, args, kwargDict, _methods, matchGenerics: true, reverse_args);
         }
 
-        static Binding? Bind(BorrowedReference inst, BorrowedReference args, Dictionary<string, PyObject> kwargDict, MethodBase[] methods, bool matchGenerics)
+        private static Binding? Bind(BorrowedReference inst, BorrowedReference args, Dictionary<string, PyObject> kwargDict, MethodBase[] methods, bool matchGenerics, bool reversed = false)
```
lostmsu, line 377: naming: it is unclear what is reversed. Should be `argsReversed`

src/runtime/MethodBinder.cs
```diff
@@ -28,17 +28,22 @@ internal class MethodBinder
...
         [NonSerialized]
         public bool init = false;
+
         public const bool DefaultAllowThreads = true;
         public bool allow_threads = DefaultAllowThreads;
 
-        internal MethodBinder()
+        public bool args_reversed = false;
+
+        internal MethodBinder(bool reverse_args = false)
```
lostmsu, line 37: naming: here and in other places the parameter should be `reverseArgs`

lostmsu (commented):
Can you explain what caused the issue was (e.g. what caused the error) and how this fix addresses this scenario?

gertdreyer:
Changes made...

I traced it down to the nb_add, nb_mul, etc slots only having the forward operators defined/linked for C# types being used in Python, this somehow caused the RecursionError when any reverse operators are called as they are probably just pointing to random pointers. 

`py * cs` calls `py.__mul__(cs)` which is returns not implemented for the cs type and codecs does not get called prior to the __mul__ call which results in `cs.__rmul__(py)` being called which caused the recursion error. I do not believe that there are problems with the codec system but it is needed to test this due to being on both sides of the CLR boundary.

It seems there were only catered for forward operators when both are the different (with overloaded operators, i.e. OwnInt - Int64 and Int64 - OwnInt). Thus all the code to detect whether a call is a reverse operator tested based on calling and passed types. Thus when called with the same types it could never detect it was a reversed operation even if it were defined.  This needed some changes to the MethodBinding and ModelBinding to ensure the reversed calls are linked to the right C# operators and called with the arguments swapped.

src/runtime/MethodBinder.cs
```diff
@@ -28,17 +28,22 @@ internal class MethodBinder
...
+        public bool argsReversed = false;
+
+        internal MethodBinder(bool argsReversed = false)
         {
             list = new List<MaybeMethodBase>();
+            this.argsReversed = argsReversed;
         }
 
-        internal MethodBinder(MethodInfo mi)
+        internal MethodBinder(MethodInfo mi, bool argsReversed = false)
```
lostmsu, line 43: IMHO, you don't need this parameter in constructor. Just set the field directly when needed.

E.g. `new MethodBinder(...) { argsReversed = true }`

src/runtime/MethodBinder.cs
```diff
@@ -271,10 +276,11 @@ internal static int ArgPrecedence(Type t)
         /// <param name="inst">The Python target of the method invocation.</param>
         /// <param name="args">The Python arguments.</param>
         /// <param name="kw">The Python keyword arguments.</param>
+        /// <param name="argsReversed">Reverse arguments of methods. Used for methods such as __radd__, __rsub__, __rmod__ etc</param>
         /// <returns>A Binding if successful.  Otherwise null.</returns>
-        internal Binding? Bind(BorrowedReference inst, BorrowedReference args, BorrowedReference kw)
+        internal Binding? Bind(BorrowedReference inst, BorrowedReference args, BorrowedReference kw, bool argsReversed = false)
```
lostmsu, line 281: Do you need to pass this parameter at all? Don't you have the field?

gertdreyer: Cleaned it up a bit. I missed that... The one method is static though so we will need to still pass it there...

lostmsu (approved):

Issue #2311, opened 2024-02-01
Allow setting typed objects via PyModule.Set


dmitry-medvedev:
### Environment

-   Pythonnet version: 3.0.1
-   Python version: 3.11
-   Operating System: Windows 11
-   .NET Runtime: .NET 4.8

### Details

-   Right now, a .NET object can be passed to the Python scripting engine via PyModule.Set, which internally calls ToPythonDetectType(value)
  We'd like to pass a certain interface that the object supports, so only methods of this interface are available for Python scripting.

I suspect that adding the overloaded method PyModule.Set or ConverterExtension.ToPython that accepts the type of the object may resolve this issue.

dmitry-medvedev:
We were able to achieve the desired effect by using the following code:

```
 internal static class PythonModuleExtensions
 {
     public static void Set(this PyModule module, string name, object value, Type type)
     {
         using var pyValue = Converter.ToPython(value, type);
         SetPyValue(module, name, pyValue.Borrow());
     }

     private static void SetPyValue(PyModule module, string name, BorrowedReference value)
     {
         Check(module);
         if (name is null)
             throw new ArgumentNullException(nameof(name));

         using var pyKey = new PyString(name);

         var result = Runtime.PyObject_SetItem(module.variables, pyKey.obj, value);
         if (result < 0)
         {
             throw PythonException.ThrowLastAsClrException();
         }
     }

     private static void Check(PyModule module)
     {
         if (module.Reference.IsNull)
         {
             throw new ObjectDisposedException(nameof(PyModule));
         }
     }
 }
```

However, as some of these types/methods are internal, this is only possible if we use IgnoresAccessChecksToGenerator, which we would really like to avoid.

lostmsu:
`PyModule` inherits from `PyObject`, and `PyObject` has a `Set` method that accepts `PyObject` value.

Sounds like we just need an overload for `ToPython` that takes type parameter. I think it would be better to have both generic and non-generic variants. PR would be welcome.

dmitry-medvedev:
We've sent PR via a separate ticket:
https://github.com/pythonnet/pythonnet/pull/2323

PR #2330, opened 2024-02-27
Added `ToPythonAs<T>()` 

lostmsu:
Added `ToPythonAs<T>()`  extension method to allow for explicit conversion using a specific type

The existing `ToPython()` extension always uses current object type. There are scenarios though where objects have to be explicitly represented as their base types or interfaces. For example, `42.ToPython()` produces Python `int`, but there's currently no easy way from C# to pass `42` as `IConvertible` so that Python could call the `IConvertible` members.

The new extension can be used for this purpose: `42.ToPythonAs<IConvertible>()` will return a `PyObject` that has all `IConvertible` members.

fixes https://github.com/pythonnet/pythonnet/issues/2311

### Any other comments?

N/A

### Checklist

### DIFF
--- a/src/runtime/Converter.cs
+++ b/src/runtime/Converter.cs
@@ -133,7 +133,8 @@ namespace Python.Runtime
             if (EncodableByUser(type, value))
             {
                 var encoded = PyObjectConversions.TryEncode(value, type);
-                if (encoded != null) {
+                if (encoded != null)
+                {
                     return new NewReference(encoded);
                 }
             }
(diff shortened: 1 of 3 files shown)


src/runtime/Converter.cs
```diff
@@ -980,5 +981,11 @@ public static PyObject ToPython(this object? o)
             if (o is null) return Runtime.None;
             return Converter.ToPython(o, o.GetType()).MoveToPyObject();
         }
+
+        public static PyObject ToPythonAs<T>(this T? o)
```
filmor, line 985: `ToPython<T>` would work as well, right?

lostmsu: Not really, that would be code incompatibility, because 

```csharp
IWhoops value = new Foo();
var py = value.ToPython();
```

if we name the function `ToPython` it would be called here instead of the original one without any code modification, and the result will be different (e.g. interface representation instead of `Foo` object).

Issue #2335, opened 2024-03-06
System.NotSupportedException: BinaryFormatter serialization and deserialization are disabled within this application


LiuxinABC:
### Environment

-   Pythonnet version:3.0.3
-   Python version:381
-   Operating System:win11
-   .NET Runtime:.NET8

### Details

-   In my .NET 8 console environment, I executed a simple operation to call a Python method that returns a string. However, I encountered an error when executing PythonEngine.Shutdown().

    _TODO_

-    python&code：
import json
def return_json_string():
    data = {"message": "这是从Python返回的字符串"}
    return "{\"name\":\"Michael\"}"  # 这将会是一个字符串
-    .net&code
 public static void GetStr()
 {
     Runtime.PythonDLL = @"D:\Program Files\Python312\python312.dll"; // 根据需要更改Python版本和路径
     PythonEngine.Initialize(); // 初始化Python引擎

     using (Py.GIL()) // 确保我们在Python全局解释锁(GIL)的上下文中
     {
         dynamic pyModule = Py.Import("ReturnJson"); // 导入Python方法所在文件名
         string result = pyModule.return_json_string(); // 调用函数并获取返回值

         Console.WriteLine(result);
     }

     PythonEngine.Shutdown(); // 关闭Python引擎
 }

```error
    System.NotSupportedException: BinaryFormatter serialization and deserialization are disabled within this application. See https://aka.ms/binaryformatter for more information.
   at System.Runtime.Serialization.Formatters.Binary.BinaryFormatter.Serialize(Stream serializationStream, Object graph)
   at Python.Runtime.RuntimeData.Stash()
   at Python.Runtime.Runtime.Shutdown()
   at Python.Runtime.PythonEngine.Shutdown()
```

filmor:
This should be fixed now.

nwoolls:
@filmor is this fix available on NuGet? I'm still seeing this issue with the latest 3.1 preview version on NuGet.

Issue #2282, opened 2023-11-16
BinaryFormatter serialization and deserialization warning which throws an error in .net8


RobertClapham:
### Environment

-   Pythonnet version: 3.0.3
-   Python version: 3.9
-   Operating System: Windows 11
-   .NET Runtime: .net8 (release version 16/11/23)

### Details

-   When shutting down the the PythonEngine via a Dispose method on the impermentation class, the error occurs at the PythonEngine.Shutdown(); method and throws the error System.NotSupportedException: BinaryFormatter serialization and deserialization are disabled within this application (see trace below). The same code worked in .net7 without any issues, the error message leads to this description
- 
- https://learn.microsoft.com/en-us/dotnet/standard/serialization/binaryformatter-security-guide
- 
- Proir to the Shutdown error the pythonnet works correctly and calls python functions and returns the expected results
- 
- The resolution is to add <EnableUnsafeBinaryFormatterSerialization>true</EnableUnsafeBinaryFormatterSerialization> to the .csproj file
-     <PropertyGroup>
        <TargetFramework>net8.0</TargetFramework>
        <EnableUnsafeBinaryFormatterSerialization>true</EnableUnsafeBinaryFormatterSerialization>
       </PropertyGroup>

    _TODO_

- _logger.Log is to to trace the point of failure  
- _logger.Log(LogLevel.Error, "Shutting down Python engine"); runs 
- it crashes at the PythonEngine.Shutdown(); method 
- _logger.Log(LogLevel.Error, "Python engine shut down"); does not run

    private void Dispose(bool disposing)
    {
        _logger.Log(LogLevel.Error, "Dispose");
        if (_disposed) return;
        if (disposing)
        {
            _logger.Log(LogLevel.Error, "Shutting down Python engine");
            PythonEngine.Shutdown();
            _logger.Log(LogLevel.Error, "Python engine shut down");
        }
        _disposed = true;
    }

```python
    print('TODO')
```

-   System.NotSupportedException: BinaryFormatter serialization and deserialization are disabled within this application. See https://aka.ms/binaryformatt...

System.NotSupportedException
BinaryFormatter serialization and deserialization are disabled within this application. See https://aka.ms/binaryformatter for more information.
   at System.Runtime.Serialization.Formatters.Binary.BinaryFormatter.Serialize(Stream serializationStream, Object graph)
   at Python.Runtime.RuntimeData.Stash()
   at Python.Runtime.Runtime.Shutdown()
[...]

Issue #2221, opened 2023-08-14
SerializationException when PythonEngine.Shutdown()


muler415:
### Environment

-   Pythonnet version: 3.0.1
-   Python version: 3.8
-   Operating System: win11
-   .NET Runtime: 4.6.1

### Details

-   Describe what you were trying to get done.

    my python code listed as following:
```python
import datetime
import csv
import clr
clr.AddReference('CyberStone.Atp.Storage')
from CyberStone.Atp.Storage.Query import QueryCondition
from CyberStone.Atp.Modules.Analysis.Models import SearchResModel
from System.Collections.Generic import List
from System import String, DateTime

#define query condition
condition = QueryCondition()
condition.QueryValueVariables = List[String]()
condition.QueryValueVariables.Add('Decoder/projectValue')
condition.RowMax = 100
condition.StartTime = DateTime(2018, 8, 2)
condition.EndTime = DateTime(2018, 8, 3)

#start query database
result = dataAnalysis.GetSearchDataSet(condition)
print(result.Count) 
```

when I execute PythonEngine.Shutdown(), there is exception. do you know how to resolve it?

System.Runtime.Serialization.SerializationException
  HResult=0x8013150C
  Message=程序集“CyberStone.Atp.Modules.Analysis, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null”中的类型“CyberStone.Atp.Modules.Analysis.Models.SearchResModel”未标记为可序列化。
  Source=mscorlib
  StackTrace:
   at System.Runtime.Serialization.FormatterServices.InternalGetSerializableMembers(RuntimeType type)
   at System.Collections.Concurrent.ConcurrentDictionary`2.GetOrAdd(TKey key, Func`2 valueFactory)
   at System.Runtime.Serialization.FormatterServices.GetSerializableMembers(Type type, StreamingContext context)
   at System.Runtime.Serialization.Formatters.Binary.WriteObjectInfo.InitMemberInfo()
   at System.Runtime.Serialization.Formatters.Binary.WriteObjectInfo.InitSerialize(Type objectType, ISurrogateSelector surrogateSelector, StreamingContext context, SerObjectInfoInit serObjectInfoInit, IFormatterConverter converter, SerializationBinder binder)
   at System.Runtime.Serialization.Formatters.Binary.ObjectWriter.WriteArray(WriteObjectInfo objectInfo, NameInfo memberNameInfo, WriteObjectInfo memberObjectInfo)
[...]

PR #1892, opened 2022-07-29
Best effort serialization

BadSingleton:
### What does this implement/fix? Explain your changes.

This PR adds a "last line of defense, best effort serialization" to serialize types not marked as Serializable. Such objects, when derivable, are deserialized as derived classes with all (overridable) methods and properties overriden to throw a "Not Serialized" Exception. Fields are not initialized and may be null. Instances of classes that are not derivable are null.

### Does this close any currently open issues?

No, but it is an issue we've encountered with Unity.

### Any other comments?

This is a "best effort" to ensure no SerializationExceptions are thrown on domain unloads/reloads. This should help users deal with serialization issues as sometimes unserializable classes may be in compiled libraries and therefore unmodifiable.

### Checklist

Check all those that are applicable and complete.

-   [x] Make sure to include one or more tests for your change
-   [ ] If an enhancement PR, please create docs and at best an example
-   [x] Ensure you have signed the [.NET Foundation CLA](https://cla.dotnetfoundation.org/pythonnet/pythonnet)
-   [x] Add yourself to [`AUTHORS`](../blob/master/AUTHORS.md)
-   [ ] Updated the [`CHANGELOG`](../blob/master/CHANGELOG.md)

PR #2336, opened 2024-03-07
Expose serialization api

BadSingleton:
Adds post-serialization and pre-deserialization hooks for additional customization.

This PR supersedes #1892 , offer another workaround for issues #2335 #2221 #2282 , possibly others.

This addresses @lostmsu 's comments of exposing an API instead of adding functionality

### Any other comments?

According to Unity dev answers on the Unity forums, they will implement a ring-link architecture as a replacement for domain load/unload. I have added no tests yet in this PR considering that domain load/unload is deprecated from a .NEt perspective.
https://forum.unity.com/threads/unity-future-net-development-status.1092205/page-34#post-8918231

This PR is made in my own name, I am no longer working at the company that signed the CLA for me.

### Checklist

### DIFF
--- a/src/runtime/StateSerialization/RuntimeData.cs
+++ b/src/runtime/StateSerialization/RuntimeData.cs
@@ -54,11 +54,11 @@ namespace Python.Runtime
         /// <summary>
         /// Callback called as a last step in the serialization process
         /// </summary>
-        public static Action? PostStashHook {get; set;} = null;
+        public static Action? PostStashHook { get; set; } = null;
         /// <summary>
         /// Callback called as the first step in the deserialization process
         /// </summary>
-        public static Action? PreRestoreHook {get; set;} = null;
+        public static Action? PreRestoreHook { get; set; } = null;
         public static ICLRObjectStorer? WrappersStorer { get; set; }
 
         /// <summary>
@@ -280,16 +280,17 @@ namespace Python.Runtime
             }
         }
 
+        static readonly string serialization_key_namepsace = "pythonnet_serialization_";
         /// <summary>
-        /// Frees the pointer stored in a Python capsule and the capsule stored 
-        /// on the sys module object with the given key if it exists. 
+        /// Removes the serialization capsule from the `sys` module object.
         /// </summary>
         /// <remarks>
-        /// The memory on the capsule must have been allocated via <code>StashDataInCapsule</code>
+        /// The serialization data must have been set with <code>StashSerializationData</code>
         /// </remarks>
         /// <param name="key">The name given to the capsule on the `sys` module object</param>
-        public static void FreeCapsuleData(string key)
+        public static void FreeSerializationData(string key)
         {
+            key = serialization_key_namepsace + key;
             BorrowedReference oldCapsule = PySys_GetObject(key);
             if (!oldCapsule.IsNull)
             {
@@ -301,42 +302,51 @@ namespace Python.Runtime
         }
 
         /// <summary>
-        /// Stores the <paramref name="data"/>data parameter in a Python capsule and stores 
+        /// Stores the data in the <paramref name="stream"/> argument in a Python capsule and stores 
         /// the capsule on the `sys` module object with the name <paramref name="key"/>. 
-        /// This method allocates global memory to hold the data of the <paramref name="data"/>
-        /// parameter.
         /// </summary>
         /// <remarks>
         /// No checks on pre-existing names on the `sys` module object are made.
         /// </remarks>
         /// <param name="key">The name given to the capsule on the `sys` module object</param>
-        /// <param name="data">The data to be contained in the capsule</param>
-        public static void StashDataInCapsule(string key, byte[] data)
+        /// <param name="stream">A MemoryStream that contains the data to be placed in the capsule</param>
+        public static void StashSerializationData(string key, MemoryStream stream)
         {
+            var data = stream.GetBuffer();
             IntPtr mem = Marshal.AllocHGlobal(IntPtr.Size + data.Length);
             // store the length of the buffer first
             Marshal.WriteIntPtr(mem, (IntPtr)data.Length);
             Marshal.Copy(data, 0, mem + IntPtr.Size, data.Length);
 
+            try
+            {
             using NewReference capsule = PyCapsule_New(mem, IntPtr.Zero, IntPtr.Zero);
             int res = PySys_SetObject(key, capsule.BorrowOrThrow());
             PythonException.ThrowIfIsNotZero(res);
         }
+            catch
+            {
+                Marshal.FreeHGlobal(mem);
+            }
 
+        }
+
+        static byte[] emptyBuffer = new byte[0];
         /// <summary>
-        /// Retreives the pointer to previously stored data on a Python capsule. 
+        /// Retreives the previously stored data on a Python capsule. 
         /// Throws if the object corresponding to the <paramref name="key"/> parameter
         /// on the `sys` module object is not a capsule.
         /// </summary>
         /// <param name="key">The name given to the capsule on the `sys` module object</param>
-        /// <returns>The pointer to the data, or IntPtr.Zero if name matches the key</returns>
-        public static IntPtr GetDataFromCapsule(string key)
+        /// <returns>A MemoryStream containing the previously saved serialization data. 
+        /// The stream is empty if no name matches the key.  </returns>
+        public static MemoryStream GetSerializationData(string key)
         {
             BorrowedReference capsule = PySys_GetObject(key);
             if (capsule.IsNull)
             {
                 // nothing to do.
-                return IntPtr.Zero;
+                return new MemoryStream(emptyBuffer, writable:false);
             }
             var ptr = PyCapsule_GetPointer(capsule, IntPtr.Zero);
             if (ptr == IntPtr.Zero)
@@ -345,7 +355,10 @@ namespace Python.Runtime
                 // as a capsule's value
                 PythonException.ThrowIfIsNull(null);
             }
-            return ptr;
+            var len = (int)Marshal.ReadIntPtr(ptr);
+            byte[] buffer = new byte[len];
+            Marshal.Copy(ptr+IntPtr.Size, buffer, 0, len);
+            return new MemoryStream(buffer, writable:false);
         }
 
         internal static IFormatter CreateFormatter()
--- a/src/runtime/StateSerialization/RuntimeData.cs
+++ b/src/runtime/StateSerialization/RuntimeData.cs
@@ -17,20 +17,29 @@ namespace Python.Runtime
 {
     public static class RuntimeData
     {
-        private static Type? _formatterType;
-        public static Type? FormatterType
+
+        public delegate IFormatter FormatterFactoryDelegate();
+        private readonly static FormatterFactoryDelegate DefaultFormatter = () => new BinaryFormatter();
+        private static FormatterFactoryDelegate? _formatter { get; set; } = null;
+
+        public static FormatterFactoryDelegate Formatter
         {
-            get => _formatterType;
+            get
+            {
+                if (_formatter is null)
+                {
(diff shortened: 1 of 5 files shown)


src/runtime/StateSerialization/RuntimeData.cs
```diff
@@ -17,7 +17,27 @@ namespace Python.Runtime
 {
     public static class RuntimeData
     {
-        private static Type? _formatterType;
+
+        private readonly static Func<IFormatter> DefaultFormatter = () => new BinaryFormatter();
```
filmor, line 21: I know that this is a significant change in behaviour, but shouldn't we default to Noop here? The feature is rarley used (in particular, it's never used when .NET is embedded into Python or if the runtime is never unloaded).

BadSingleton: I think that it is a good idea, but I'm not sure it is within scope of that PR.

filmor:
@lostmsu any notes?

src/runtime/StateSerialization/RuntimeData.cs
psyfrett: I think the confusion is in the first comment "offer other fixes for issues 2335, 2221, 2282". Maybe the comment should corrected to exclude 2335 and 2282 regarding the BinaryFormatter in dotnet 8?

src/runtime/StateSerialization/RuntimeData.cs
```diff
@@ -31,6 +51,14 @@ public static class RuntimeData
...
         }
 
+        /// <summary>
+        /// Callback called as a last step in the serialization process
+        /// </summary>
+        public static Action? PostStashHook {get; set;} = null;
+        /// <summary>
+        /// Callback called as the first step in the deserialization process
+        /// </summary>
+        public static Action? PreRestoreHook {get; set;} = null;
```
lostmsu, line 61: Minor formatting: spaces around `get; set;`

src/runtime/StateSerialization/RuntimeData.cs
```diff
@@ -250,11 +280,82 @@ private static void RestoreRuntimeDataObjects(SharedObjectsState storage)
...
 
+        /// <summary>
+        /// Frees the pointer stored in a Python capsule and the capsule stored 
+        /// on the sys module object with the given key if it exists. 
+        /// </summary>
+        /// <remarks>
+        /// The memory on the capsule must have been allocated via <code>StashDataInCapsule</code>
+        /// </remarks>
+        /// <param name="key">The name given to the capsule on the `sys` module object</param>
+        public static void FreeCapsuleData(string key)
```
lostmsu, line 291: We might want to add some sort of namespace to all keys to avoid accidentally corrupting existing sys objects.

src/runtime/StateSerialization/RuntimeData.cs
```diff
@@ -250,11 +280,82 @@ private static void RestoreRuntimeDataObjects(SharedObjectsState storage)
...
+        /// parameter.
+        /// </summary>
+        /// <remarks>
+        /// No checks on pre-existing names on the `sys` module object are made.
+        /// </remarks>
+        /// <param name="key">The name given to the capsule on the `sys` module object</param>
+        /// <param name="data">The data to be contained in the capsule</param>
+        public static void StashDataInCapsule(string key, byte[] data)
+        {
+            IntPtr mem = Marshal.AllocHGlobal(IntPtr.Size + data.Length);
```
lostmsu, line 316: Same: key namespace

src/runtime/StateSerialization/RuntimeData.cs
```diff
@@ -250,11 +280,82 @@ private static void RestoreRuntimeDataObjects(SharedObjectsState storage)
...
+            // store the length of the buffer first
+            Marshal.WriteIntPtr(mem, (IntPtr)data.Length);
+            Marshal.Copy(data, 0, mem + IntPtr.Size, data.Length);
+
+            using NewReference capsule = PyCapsule_New(mem, IntPtr.Zero, IntPtr.Zero);
+            int res = PySys_SetObject(key, capsule.BorrowOrThrow());
+            PythonException.ThrowIfIsNotZero(res);
```
lostmsu, line 323: `StashDataInCapsule` does not appear to be atomic. If `PySys_SetObject` fails, `mem` leaks.

Generally, when action has multiple steps and fails in the middle, the earlier steps need to be rolled back.

src/runtime/StateSerialization/RuntimeData.cs
```diff
@@ -250,11 +280,82 @@ private static void RestoreRuntimeDataObjects(SharedObjectsState storage)
...
+        /// the capsule on the `sys` module object with the name <paramref name="key"/>. 
+        /// This method allocates global memory to hold the data of the <paramref name="data"/>
+        /// parameter.
+        /// </summary>
+        /// <remarks>
+        /// No checks on pre-existing names on the `sys` module object are made.
+        /// </remarks>
+        /// <param name="key">The name given to the capsule on the `sys` module object</param>
+        /// <param name="data">The data to be contained in the capsule</param>
+        public static void StashDataInCapsule(string key, byte[] data)
```
lostmsu, line 314: NIT: These method names expose implementation details. They should be named after their purpose.

src/runtime/StateSerialization/RuntimeData.cs
```diff
@@ -250,11 +280,82 @@ private static void RestoreRuntimeDataObjects(SharedObjectsState storage)
...
+        }
+
+        /// <summary>
+        /// Retreives the pointer to previously stored data on a Python capsule. 
+        /// Throws if the object corresponding to the <paramref name="key"/> parameter
+        /// on the `sys` module object is not a capsule.
+        /// </summary>
+        /// <param name="key">The name given to the capsule on the `sys` module object</param>
+        /// <returns>The pointer to the data, or IntPtr.Zero if name matches the key</returns>
+        public static IntPtr GetDataFromCapsule(string key)
```
lostmsu, line 333: While this works, consider providing higher-level interface.

I think something like a `MemoryStream` or `Span` would be better.

Might need to keep a `PyObject` around to ensure capsule is not destroyed while being accessed.

lostmsu: As-is there's no way for the consumer to know the pointer points to data that starts with the memory block length.

lostmsu:
Add some tests.

src/runtime/StateSerialization/RuntimeData.cs
BadSingleton: @psyfrett My bad, I incorrectly worded things here, it should have been "offers another workaround for issues ...". I edited the description.

src/runtime/StateSerialization/RuntimeData.cs
BadSingleton: > Might need to keep a PyObject around to ensure capsule is not destroyed while being accessed.

I guess this could happen if a user calls `FreeCapsuleData` before using the data. Deallocating the capsule using `FreeCapsuleData` would not deallocate the data as we do not implement capsule destructors. I feel that's more a usage/documentation problem.
Otherwise it would mean returning a copy of the data, or the capsule itself.

src/runtime/StateSerialization/RuntimeData.cs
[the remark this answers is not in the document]
lostmsu: Ideally, behavior on .NET 6 and before should be unchanged, and after we should have null-formatter.

src/runtime/StateSerialization/RuntimeData.cs
lostmsu: A minimal wrapper for the capsule type might make sense here.

BadSingleton: I have been thinking about it and having a wrapper over the capsule type does not circumvent the "calling FreeCapsuleData before using the data" scenario. The capsule does not own its data unless there is a destructor function set on it; tp_dealloc calls it if it exists (calling tp_dealloc on the capsule would leak the memory if no one frees the pointer in this scenario). Short of emitting assembly code for the destructor, I am not sure how we can accomplish the ownership of the data by the capsule.

lostmsu: It does not have to be a generic capsule type, just something for this specific use case.

src/runtime/StateSerialization/RuntimeData.cs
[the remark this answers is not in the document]
filmor: I'd be okay with changing the default to null formatting. This logic is only really useful for a very particular case that can be thoroughly documented. We could bump the minor version to emphasise the change.

src/runtime/StateSerialization/RuntimeData.cs
BadSingleton: I opted for using `MemoryStream` for both Stash and Get functions, as to give symmetry in the API.

src/runtime/StateSerialization/RuntimeData.cs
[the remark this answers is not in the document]
BadSingleton: In the end, what is the action to take in _this_ PR?

tests/domain_tests/TestRunner.cs
```diff
@@ -1151,6 +1263,11 @@ public static int Main()
         {{
             try
             {{
+                RuntimeData.FormatterFactory = () =>
```
BadSingleton, line 1266: Would it be better to limit this to the `test_serialize_unserializable_object` case?

src/runtime/StateSerialization/RuntimeData.cs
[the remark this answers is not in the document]
lostmsu: Initialize to null formatter on .NET 8 and above.

tests/domain_tests/TestRunner.cs
lostmsu: Doesn't matter IMHO.

lostmsu:
BTW, what's your interested in this functionality working?

Cause if there's none, I think we should consider just bumping version to 4.x and dropping domain reload support altogether.

BadSingleton:
> BTW, what's your interest in this functionality working

Superseding/closing the previous PR. At this point, considering Unity's public plans, it's not unreasonable to drop this. Unity has their own fork with what was in the previous PR anyway.

> Do you know anyone in Unity who stills works on it?
 I think so. Last I knew of, another team had assumed maintaintership of the package. I'll be sending a few messages see if they are still at Unity/on those projects

filmor:
@lostmsu Any more comments? I took the liberty to implement your comments on the formatter factory.

BadSingleton:
> I'll be sending a few messages see if they are still at Unity/on those projects

To close the loop on this, the people I could get in touch with are no longer at Unity.

src/runtime/StateSerialization/RuntimeData.cs
```diff
@@ -250,11 +285,95 @@ private static void RestoreRuntimeDataObjects(SharedObjectsState storage)
...
+        /// the capsule on the `sys` module object with the name <paramref name="key"/>.
+        /// </summary>
+        /// <remarks>
+        /// No checks on pre-existing names on the `sys` module object are made.
+        /// </remarks>
+        /// <param name="key">The name given to the capsule on the `sys` module object</param>
+        /// <param name="stream">A MemoryStream that contains the data to be placed in the capsule</param>
+        public static void StashSerializationData(string key, MemoryStream stream)
+        {
+            var data = stream.GetBuffer();
```
lostmsu, line 320: This might have extra data in the end

lostmsu (changes requested):

src/runtime/StateSerialization/RuntimeData.cs
BadSingleton: What do you mean by extra data? The stream container is longer than the data in it? The code is using `Length` not `Capacity`, so it should not be an issue, no?

PR #2353, opened 2024-03-31
Linting and Formatting

filmor:
- Add formatting and linting for C# files
- Format and fix lints in Python (test) code


filmor:
I must have messed something up in the C# code while trying to fix the NonCopyableAnalyzer reports. To be honest, at this point I'm not sure how much this one actually helps, seeing that patterns like `return NewReference(...)` are wrongly reported.

lostmsu:
Just doublechecking for C#: this change does the proper thing in regards to our `.editorconfig`? E.g. all settings default except the one set there?

lostmsu:
TBH, I am not sure formatting is important unless it is in some way extreme.

PR #2370, opened 2024-05-04
Use non-BOM encodings

### LINKED ISSUE #2369: Possible bug in reading zero width no-break space character
### Environment

-   Pythonnet version: 3.0.3
-   Python version: 3.10
-   Operating System: Windows 11
-   .NET Runtime: 8.0.4

### Details

I want to read a file that may contain zero width no-break space characters anywhere in the file. If I read a file using C# into Python using Python.NET, these characters get dropped if they are at the start of a line.

```python
from pythonnet import load
load("coreclr")
import clr
clr.AddReference('System')
clr.AddReference('System.Runtime')
clr.AddReference('System.IO')

import System
file_path = r"path\to\zwnbsp_test_data.txt"

result = System.IO.File.ReadAllLines(file_path)
print(str(result[1][0]))
```

Native C# code does not seem to have this problem:
```
C#
String[] lines = File.ReadAllLines(@"path\to\zwnbsp_test_data.txt");

int firstChar = char.ConvertToUtf32(lines[1], 0);
int secondChar = char.ConvertToUtf32(lines[1], 1);
Console.WriteLine(firstChar);
Console.WriteLine(secondChar);
```

[zwnbsp_test_data.txt](https://github.com/pythonnet/pythonnet/files/15194162/zwnbsp_test_data.txt)

Some screenshots to help illustrate: 
image
image

No crash

filmor:
Use non-BOM encodings for both C#->Python and Python->C#, as the byteorder is always the native one and the BOM is neither never or always used.

Fixes #2369.

### DIFF
--- a/src/embed_tests/TestPyType.cs
+++ b/src/embed_tests/TestPyType.cs
@@ -28,7 +28,7 @@ namespace Python.EmbeddingTest
             const string name = "nÁmæ";
             const string docStr = "dÁcæ";
 
-            using var doc = new StrPtr(docStr, Encoding.UTF8);
+            using var doc = new StrPtr(docStr, Encodings.UTF8);
             var spec = new TypeSpec(
                 name: name,
                 basicSize: Util.ReadInt32(Runtime.Runtime.PyBaseObjectType, TypeOffset.tp_basicsize),
--- a/src/runtime/Loader.cs
+++ b/src/runtime/Loader.cs
@@ -12,7 +12,7 @@ namespace Python.Runtime
         {
             try
             {
-                var dllPath = Encoding.UTF8.GetString((byte*)data.ToPointer(), size);
+                var dllPath = Encodings.UTF8.GetString((byte*)data.ToPointer(), size);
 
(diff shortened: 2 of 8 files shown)


lostmsu:
@filmor can you ELI5? For someone not familiar with intricacies of BOM, but aware of byte order issues.

My biggest question is if this change has any potential to introduce bugs to handling strings that actually have BOM? E.g. imagine a scenario when someone serialized and persisted something with BOM using 3.0.3, but after this change in 3.0.4 if they read it back BOM will be in their string data.

filmor:
It's the other way round. Strings that are being passed between Python and .NET are UTF16 in the respective native byte order (usually LE), without a BOM. The functions that we were using for the conversions (in particular [`PyUnicode_DecodeUTF16`](https://docs.python.org/3/c-api/unicode.html#c.PyUnicode_DecodeUTF16) and the defaulr encoding objects from `Encoding`) try to be "smart" and will interpret a leading set of FE FF or FF FE as the byte order mark, removing it from the converted string. By passing the correct endian-ness explicitly, this behaviour is disabled.

Issue #2371, opened 2024-05-05
Requesting Py_buffer with additional fields either crashes or returns struct with partially garbage values


chickenservice:
### Environment

-   Pythonnet version: 3.0.3
-   Python version: 3.12.1
-   Operating System: Windows 11
-   .NET Runtime: .NET 4.7.2, .NET 6.0

### Details

-   Describe what you were trying to get done.

#### Requesting buffer with format string crashes (both .NET 4.7.2 and NET 6.0)
Requesting a buffer with [PyBUF_FULL](https://docs.python.org/3/c-api/buffer.html#c.PyBUF_FULL) crashes the process. This seems to be caused by the string marshaling of the format field in the `Py_buffer` struct. Changing this to IntPtr and marshalling the string by hand  with e.g. `Marshal.PtrToStringAnsi` resolves the crash.

#### Requesting buffer with strides for bytearray returns garbage strides values (.NET 6.0)
The issue seems to be that the .NET marshaller creates a **copy** and not a reference of the struct if the struct is **not blittable**.  `bytearray` uses the default `PyBuffer_FillInfo` implementation from abstract.c to populate the Py_buffer struct. When requesting strides the function sets the strides field to the address of the `len` field [PyBuffer_FillInfo](https://github.com/python/cpython/blob/b6f0ab5b1cb6d779efe4867d83a60e8d66c48dee/Objects/abstract.c#L794). Upon returning from the P/Inovke call the marshaller copies the values back to the managed struct. The pointer is copied correctly but now points to garbage because it's referencing the field in the unmanaged struct created by the marshaller. From [Formatted Blittable Classes](https://learn.microsoft.com/en-us/dotnet/framework/interop/copying-and-pinning#formatted-blittable-classes):
>  When these types require marshalling, a pointer to the object in the heap is passed to the callee directly. The callee can change the contents of the memory location being referenced by the pointer.

and [Formatted Non-Blittable Classes](https://learn.microsoft.com/en-us/dotnet/framework/interop/copying-and-pinning#formatted-non-blittable-classes)
> If a non-blittable class is marshalled by reference, the callee receives a pointer to a pointer to a copy of the data structure.
...
If the [OutAttribute](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.outattribute) attribute is set, the state is always copied back to the instance on return, marshalling as necessary.

[...]

Issue #1838, opened 2022-06-27
Use Python buffer protocol when converting Python objects to .NET arrays


lostmsu:
Copying data from NumPy array to C# is a common task. Python.NET implements a simple sequence-to-array copying mechanism, which uses iterators. It is extremely inefficient for large arrays of primitive types.

Now we have a better way - using `PyBuffer` and buffer protocol. We could detect, that Python object implements the buffer protocol and use it to copy when possible, and fallback to iteration otherwise.

lostmsu:
Yep.

bruceYangxianglin:
Copying data from  C# array to Python list or Numpy array is also a common task.  Is there a more efficient way?

bjudkewitz:
Agreed, converting a python array to a C# array is a very common task, and it is currently extremely slow. It would be fantastic to make it faster. Is it currently possibly with Python (and Python.NET) only, without any C# programming?

yahav6893:
Hi! I’m looking for some guidance on how to speed up copying from Python to C#. 
Would anyone be able to point me in the right way to do that?
Thanks ahead.

chickenservice:
Hi, is someone currently working on this? If not I'd be happy to try. I've been using pythonnet at work successfully for years and would be glad to help. I'm familiar with the topic because I have already dealt with converting numpy arrays from/to C# arrays in an efficient manner. It'd be great to have this natively in pythonnet.

filmor:
@chickenservice To my knowledge, no one is currently actively working on this, so give it a shot! An implementation would be greatly appreciated!

chickenservice:
@filmor Perfect, is there anything besides the steps in CONTRIBUTING.md that I need to be aware of?

PR #2372, opened 2024-05-05
1838 Use Python buffer protocol when converting Python objects to .NET arrays

chickenservice:

If a python object exposes the buffer protocol then copy the buffer directly to a .NET array. This works for (blittable) single- or multidimensional arrays. I added tests for numpy arrays, bytearrays and python arrays to check whether this change works across different buffer protocol implementations.

My implementation currently has the following caveats:

- The Python object must implement the buffer protocol and be **C-contiguous**. This is only relevant for multidimensional arrays. I explicitly request a C-contiguous buffer because (multidimensional) .NET arrays are C-contiguous as well. This allows copying the whole chunk of memory at once. Supporting multidimensional Fortran-contiguous buffers would make the implementation more complicated though I have not thought too much about it.
- PIL-style buffers that do not store the whole buffer contiguously but through pointers via [suboffsets](https://docs.python.org/3/c-api/buffer.html#c.Py_buffer.suboffsets) are not supported. This would also need special handling.
- The .NET element type of the array must be **blittable** and **not a struct**.  Structs can be blittable if all their members are blittable. I imagine this may make the implementation more complicated because as far as I know no there's no out-of-the-box way to check whether a type is blittable. I think this is better handled with custom Codecs?
- Works for **multidimensional arrays** which `ToArray` previously didn't handle, so we weaken a precondition of the converter. However efficient copying of multidimensional arrays is highly desirable in my opinion. 
- I use `Buffer.MemoryCopy` which is marked as not CLS-compliant. If this is not acceptable I will of course find another way.
- The implementation is done directly with the C-API. I can rewrite this with `PyBuffer` if you want.

[1838 Use Python buffer protocol when converting Python objects to .NET arrays](https://github.com/pythonnet/pythonnet/issues/1838)

### Any other comments?

The test with bytearray to .NET byte array is currently failing for NET 6.0 due to the way the Py_Buffer struct is being marshalled. I opened a separate issue #2371 to address this.

### Checklist

### DIFF
--- a/src/runtime/Converter.cs
+++ b/src/runtime/Converter.cs
@@ -2,6 +2,7 @@ using System;
 using System.Collections;
 using System.Collections.Generic;
 using System.Diagnostics;
+using System.Linq;
 using System.Reflection;
 using System.Runtime.InteropServices;
 using System.Security;
@@ -880,11 +881,48 @@ namespace Python.Runtime
         /// Convert a Python value to a correctly typed managed array instance.
         /// The Python value must support the Python iterator protocol or and the
         /// items in the sequence must be convertible to the target array type.
+        /// If the Python value supports the buffer protocol the underlying buffer
+        /// will be copied directly.
         /// </summary>
         private static bool ToArray(BorrowedReference value, Type obType, out object? result, bool setError)
         {
             Type elementType = obType.GetElementType();
             result = null;
+
+            // structs can also be blittable but for the standard usecases this suffices
+            var isBlittable = elementType.IsPrimitive && elementType != typeof(char) && elementType != typeof(bool);
+           if (isBlittable && Runtime.PyObject_GetBuffer(value, out var view, (int)PyBUF.CONTIG_RO) == 0)
+            {
+                GCHandle ptr;
+                try
+                {
+                    var shape = new IntPtr[view.ndim];
+                    Marshal.Copy(view.shape, shape, 0, view.ndim);
+                    Array arr = Array.CreateInstance(elementType, shape.Select(x => (long)x).ToArray());
+                    ptr = GCHandle.Alloc(arr, GCHandleType.Pinned);
+                    var addr = ptr.AddrOfPinnedObject();
+                    unsafe
+                    {
+                        Buffer.MemoryCopy((void*)view.buf, (void*)addr, arr.Length * Marshal.SizeOf(elementType), view.len);
+                    }
+
+                    result = arr;
+                    return true;
+                }
+                finally
+                {
+                    if (ptr.IsAllocated)
+                    {
+                        ptr.Free();
+                    }
+
+                    Runtime.PyBuffer_Release(ref view);
+                }
+            }
+            else
+            {
+                Exceptions.Clear();
+            }
 
             using var IterObject = Runtime.PyObject_GetIter(value);
             if (IterObject.IsNull())
--- a/src/embed_tests/TestConverter.cs
+++ b/src/embed_tests/TestConverter.cs
@@ -2,6 +2,7 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Numerics;
+using System.Text;
 
 using NUnit.Framework;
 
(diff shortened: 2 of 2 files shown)


src/runtime/Converter.cs
```diff
@@ -880,12 +881,49 @@ private static void SetConversionError(BorrowedReference value, Type target)
...
+        /// will be copied directly.
         /// </summary>
         private static bool ToArray(BorrowedReference value, Type obType, out object? result, bool setError)
         {
             Type elementType = obType.GetElementType();
             result = null;
 
+            // structs can also be blittable but for the standard usecases this suffices
+            var isBlittable = elementType.IsPrimitive && elementType != typeof(char) && elementType != typeof(bool);
+           if (isBlittable && Runtime.PyObject_GetBuffer(value, out var view, (int)PyBUF.CONTIG_RO) == 0)
```
lostmsu, line 894: I'm not sure it is safe to assume getting buffer has the same semantics as using iterator for an arbitrary type that supports GetBuffer.

Imho, this could be made the default behavior for converting a fixed set of standard types, but for the others it might be better to expose some kind of explicit call instead of conversion. E.g. `pyObjectInst.ReadBuffer(destinationArray)`

lostmsu (changes requested):

src/runtime/Converter.cs
filmor: Isn't this caught by the primitive blitable check?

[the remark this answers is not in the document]
chickenservice: Would it be ok to match the typecode of the .NET type to the corresponding format string of the buffer, i.e. roughly like so:
```csharp      
var compatible = false;
var format = buf.format;
switch (tc)
{
    case TypeCode.Boolean:
        compatible = format == "?";
        break;
    case TypeCode.Byte:
        compatible = format == "b";
        break;
        //etc...
    default:
        break;
```

chickenservice: We could move the whole thing to after the `if (IterObject.IsNull()` check to make sure it's an iterable? Other than that as @filmor said I already allow only primitive blittable types as array element types.

lostmsu: This issue is not about element types, but about container types. Does Python guarantee, that GetBuffer()[0], GetBuffer()[1], etc is the same sequence as .iter[0], .iter[1], etc

filmor: I thought that was the point of the buffer protocol :). I'll read up more on it.

chickenservice: I'm not sure such a guarantee can be made, wouldn't it fail for multidimensional arrays? Is it even necessary? If you define the semantics of `ToArray()` as _convert a given python object to a .NET array if possible_ than the information describing [Complex arrays](https://docs.python.org/3/c-api/buffer.html#complex-arrays) (shape, strides, format, contiguity) is I believe enough to copy such a buffer correctly to an array and to check whether the user requested a matching array in the first place. By always requesting `GetBuffer(PyBUF_FULL_RO)` we can make sure to avoid converting buffers that do not give us all information. However just doing the optimization for a certain set of types to avoid any unforseen issues may be the better option anyway.

lostmsu: @chickenservice the concern is about breaking changes, especially within Python.NET 3.x, more especially within Python.NET 3.0.x.

If there's a difference in result, we should delay the change of behavior to 4.x, and in 3.x only add new functions.

IMHO, we can only immediately apply the speedups for the types that we know have the same behavior.

chickenservice: @lostmsu that makes sense. The tests cover bytearray, array.array and ndarray. From my experience these behave according to what you described and are the canonical implementations anyway. Any concerns? Are there other types you have in mind? Should we limit ndim == 1 for now?

lostmsu: If the tests do not cover multidim or the behavior in multidim is different, we should limit ndim to 1.

But again, I strongly suggest to add an explicit new method that does the right thing with buffers, and only works with buffer protocol.

chickenservice: I think I misunderstood your suggestion, to clarify:

For upcoming minor/patch release expose this as a dedicated method and remove it from `ToArray` completely.

For the next major release reevaluate whether this can/should be done implicitly in `ToArray` and accept breaking changes.

lostmsu: @chickenservice yep, that's exactly right.

vadimkantorov:
My related attempts / investigation on ways of NumPy <> C# arrays interop: 
- https://github.com/pythonnet/pythonnet/discussions/2226
- I also discovered a zero-copy way of NumPy -> C# array interop: https://gist.github.com/vadimkantorov/c9ab97aff995ea01ea7f6248ddc733bc#file-csharpfrompython-cs-L23-L41 using some sort of `System.Buffers.MemoryManager` device

lostmsu:
Closing as the requested changes were not implemented.

PR #2386, opened 2024-05-13
Fix access violation exception on shutdown

Frawak:

When nulling the GC handles on shutdown the reference count of all objects pointed to by the IntPtr in the `CLRObject.reflectedObjects` are zero. This caused an exception in some scenarios because `Runtime.PyObject_TYPE(reflectedClrObject)` is called while the reference counter is at zero.

After `TypeManager.RemoveTypes();` is called in the `Runtime.Shutdown()` method, reference count decrements to zero do not invoke `ClassBase.tp_clear` for managed objects anymore which normally is responsible for freeing GC handles and removing references from `CLRObject.reflectedObjects`. Collecting objects referenced in `CLRObject.reflectedObjects` only after leads to an unstable state in which the reference count for these object addresses is zero while still maintaining them to be used for further pseudo-cleanup. In that time, the memory could have been reclaimed already which leads to the exception.

#1977 

### Any other comments?

...

### Checklist

### DIFF
--- a/src/runtime/Runtime.cs
+++ b/src/runtime/Runtime.cs
@@ -158,6 +158,7 @@ namespace Python.Runtime
             ClassManager.Reset();
             ClassDerivedObject.Reset();
             TypeManager.Initialize();
+            CLRObject.creationBlocked = false;
             _typesInitialized = true;
 
             // Initialize modules that depend on the runtime class.
@@ -356,6 +357,7 @@ namespace Python.Runtime
                 {
                     NullGCHandles(CLRObject.reflectedObjects);
                     CLRObject.reflectedObjects.Clear();
+                    CLRObject.creationBlocked = true;
                 }
             }
             return false;
--- a/src/runtime/Runtime.cs
+++ b/src/runtime/Runtime.cs
@@ -278,7 +278,8 @@ namespace Python.Runtime
             ClearClrModules();
             RemoveClrRootModule();
 
-            TryCollectingGarbage(MaxCollectRetriesOnShutdown, forceBreakLoops: true);
+            TryCollectingGarbage(MaxCollectRetriesOnShutdown, forceBreakLoops: true,
+                                 obj: true, derived: false, buffer: false);
 
             NullGCHandles(ExtensionType.loadedExtensions);
             ClassManager.RemoveClasses();
@@ -329,7 +330,8 @@ namespace Python.Runtime
 
         const int MaxCollectRetriesOnShutdown = 20;
         internal static int _collected;
-        static bool TryCollectingGarbage(int runs, bool forceBreakLoops)
+        static bool TryCollectingGarbage(int runs, bool forceBreakLoops,
+                                         bool obj = true, bool derived = true, bool buffer = true)
         {
             if (runs <= 0) throw new ArgumentOutOfRangeException(nameof(runs));
 
@@ -342,7 +344,9 @@ namespace Python.Runtime
                     GC.Collect();
                     GC.WaitForPendingFinalizers();
                     pyCollected += PyGC_Collect();
-                    pyCollected += Finalizer.Instance.DisposeAll();
+                    pyCollected += Finalizer.Instance.DisposeAll(disposeObj: obj,
+                                                                 disposeDerived: derived,
+                                                                 disposeBuffer: buffer);
                 }
                 if (Volatile.Read(ref _collected) == 0 && pyCollected == 0)
                 {
--- a/src/runtime/Runtime.cs
+++ b/src/runtime/Runtime.cs
@@ -278,6 +278,8 @@ namespace Python.Runtime
             ClearClrModules();
             RemoveClrRootModule();
 
+            TryCollectingGarbage(MaxCollectRetriesOnShutdown, forceBreakLoops: true);
+
             NullGCHandles(ExtensionType.loadedExtensions);
             ClassManager.RemoveClasses();
             TypeManager.RemoveTypes();
@@ -295,8 +297,7 @@ namespace Python.Runtime
             PyObjectConversions.Reset();
 
             PyGC_Collect();
-            bool everythingSeemsCollected = TryCollectingGarbage(MaxCollectRetriesOnShutdown,
-                                                                 forceBreakLoops: true);
+            bool everythingSeemsCollected = TryCollectingGarbage(MaxCollectRetriesOnShutdown);
             Debug.Assert(everythingSeemsCollected);
 
             Finalizer.Shutdown();
--- a/src/runtime/Runtime.cs
+++ b/src/runtime/Runtime.cs
@@ -281,6 +281,7 @@ namespace Python.Runtime
 
             TryCollectingGarbage(MaxCollectRetriesOnShutdown, forceBreakLoops: true,
                                  obj: true, derived: false, buffer: false);
+            CLRObject.creationBlocked = true;
 
             NullGCHandles(ExtensionType.loadedExtensions);
             ClassManager.RemoveClasses();
@@ -357,7 +358,6 @@ namespace Python.Runtime
                 {
                     NullGCHandles(CLRObject.reflectedObjects);
                     CLRObject.reflectedObjects.Clear();
-                    CLRObject.creationBlocked = true;
                 }
             }
             return false;
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -38,6 +38,7 @@
 -   Dmitriy Se ([@dmitriyse](https://github.com/dmitriyse))
 -   Félix Bourbonnais ([@BadSingleton](https://github.com/BadSingleton))
 -   Florian Treurniet ([@ftreurni](https://github.com/ftreurni))
+-   Frank Witscher ([@Frawak](https://github.com/Frawak))
 -   He-chien Tsai ([@t3476](https://github.com/t3476))
 -   Inna Wiesel ([@inna-w](https://github.com/inna-w))
 -   Ivan Cronyn ([@cronan](https://github.com/cronan))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ This document follows the conventions laid out in [Keep a CHANGELOG][].
 
 -   Fixed RecursionError for reverse operators on C# operable types from python. See #2240
 -   Fixed probing for assemblies in `sys.path` failing when a path in `sys.path` has invalid characters. See #2376
+-   Fixed possible access violation exception on shutdown. See ([#1977][i1977])
 
 ## [3.0.3](https://github.com/pythonnet/pythonnet/releases/tag/v3.0.3) - 2023-10-11
 
@@ -970,3 +971,4 @@ This version improves performance on benchmarks significantly compared to 2.3.
 [i1481]: https://github.com/pythonnet/pythonnet/issues/1481
 [i1672]: https://github.com/pythonnet/pythonnet/pull/1672
 [i2311]: https://github.com/pythonnet/pythonnet/issues/2311
+[i1977]: https://github.com/pythonnet/pythonnet/issues/1977
--- a/src/runtime/Finalizer.cs
+++ b/src/runtime/Finalizer.cs
@@ -191,7 +191,7 @@ namespace Python.Runtime
             Instance.started = false;
         }
 
-        internal nint DisposeAll()
+        internal nint DisposeAll(bool disposeObj = true, bool disposeDerived = true, bool disposeBuffer = true)
         {
             if (_objQueue.IsEmpty && _derivedQueue.IsEmpty && _bufferQueue.IsEmpty)
                 return 0;
@@ -216,7 +216,7 @@ namespace Python.Runtime
 
                 try
                 {
-                    while (!_objQueue.IsEmpty)
+                    if (disposeObj) while (!_objQueue.IsEmpty)
                     {
                         if (!_objQueue.TryDequeue(out var obj))
                             continue;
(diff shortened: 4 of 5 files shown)


Frawak:
Alright. At least the embedding tests went through. I am going to take another look tomorrow.

Frawak:
The original change broke the shutdown flow for derived objects. The (hopefully) final proposition from me remains the same, only restricting the early collection run to the managed objects which would, as described above, run into the problem of the cleared `tp_clear` slot at a later point.

Frawak:
@filmor @lostmsu Could you please trigger the workflows again?

lostmsu:
This feels like a band-aid rather than proper fix. Theoretically invoking GC should not leave anything in a broken state, but types derived across .NET/Python boundaries are tricky. We should come up with a correct way to finalize them rather than adding such a band-aid, IMHO.

Frawak:
@lostmsu To make sure we are on the same page and in risk of repeating myself, I will copy some code snippets.

The shutdown method:

```csharp
            NullGCHandles(ExtensionType.loadedExtensions);
            ClassManager.RemoveClasses();
            TypeManager.RemoveTypes(); // this removes tp_clear slot
            _typesInitialized = false;

            ...

            PyGC_Collect();
            bool everythingSeemsCollected = TryCollectingGarbage(MaxCollectRetriesOnShutdown,
                                                                 forceBreakLoops: true); // this decreases the reference count
```

The tp_clear method:

```csharp
        public static int tp_clear(BorrowedReference ob)
        {
            var weakrefs = Runtime.PyObject_GetWeakRefList(ob);
            if (weakrefs != null)
            {
                Runtime.PyObject_ClearWeakRefs(ob);
            }

            if (TryFreeGCHandle(ob)) // what NullGCHandles tries to do after the fact
            {
                IntPtr addr = ob.DangerousGetAddress();
                bool deleted = CLRObject.reflectedObjects.Remove(addr); // the normal removal from hashset
                Debug.Assert(deleted);
            }

            int baseClearResult = BaseUnmanagedClear(ob);
            if (baseClearResult != 0)
            {
                return baseClearResult;
            }

            ClearObjectDict(ob);
            return 0;
        }
```

tp_clear should be responsible for freeing the GC handle and removing the address from `CLRObject.reflectedObjects` what the `forceBreakLoops` tries to do separatedly. 

```csharp
                for (int i = 0; i < 2; i++)
                {
                    GC.Collect();
                    GC.WaitForPendingFinalizers();
                    pyCollected += PyGC_Collect();
                    pyCollected += Finalizer.Instance.DisposeAll();
                }
                if (Volatile.Read(ref _collected) == 0 && pyCollected == 0)
                {
                    if (attempt + 1 == runs) return true;
                }
                else if (forceBreakLoops)
                {
                    NullGCHandles(CLRObject.reflectedObjects);
                    CLRObject.reflectedObjects.Clear();
                }
```

There are two crucial points, I hope you see are wrong in the current shutdown flow:

1. `tp_clear` is not invoked after decreasing the reference count of a respective object (referenced in `CLRObject.reflectedObjects`) to zero.
2. The reference count to all addresses in `CLRObject.reflectedObjects` is zero when calling ` NullGCHandles(CLRObject.reflectedObjects);`, i.e. handling the address in this state.

This can be seen just debugging and stepping into the code, exception or not. Hence, the state is already broken after the line `pyCollected += Finalizer.Instance.DisposeAll();`.

Is this not enough that a fix is necessary? 
An error message does not help because nothing can be done really to handle it. If you extend Python and unload pythonnet, you cannot use C# exceptions types anymore to catch them in the Python domain. This is an unhandled exception that crashes the Python process. 

Regarding your concern: 

> It is possible your scenario keeps invalid references.

What references could be problematic/left-over running this code:

```csharp
            NullGCHandles(ExtensionType.loadedExtensions);
            ClassManager.RemoveClasses();
            TypeManager.RemoveTypes();
            _typesInitialized = false;

            MetaType.Release();
            PyCLRMetaType.Dispose();
            PyCLRMetaType = null!;

            Exceptions.Shutdown();
            PythonEngine.InteropConfiguration.Dispose();
            DisposeLazyObject(clrInterop);
            DisposeLazyObject(inspect);
            DisposeLazyObject(hexCallable);
            PyObjectConversions.Reset();
```

After running:

```csharp
                    while (!_objQueue.IsEmpty)
                    {
                        if (!_objQueue.TryDequeue(out var obj))
                            continue;

                        if (obj.RuntimeRun != run)
                        {
                            HandleFinalizationException(obj.PyObj, new RuntimeShutdownException(obj.PyObj));
                            continue;
                        }

                        IntPtr copyForException = obj.PyObj;
                        Runtime.XDecref(StolenReference.Take(ref obj.PyObj));
                        collected++;
                        try
                        {
                            Runtime.CheckExceptionOccurred();
                        }
                        catch (Exception e)
                        {
                            HandleFinalizationException(obj.PyObj, e);
                        }
                    }
```

It is effectively the only change.

Frawak:
> This feels like a band-aid rather than proper fix.

Is this a feeling because of function name semantics which could be dissolved by refactoring the solution?

Effectively, my changes only run this:
```csharp
                    while (!_objQueue.IsEmpty)
                    {
                        if (!_objQueue.TryDequeue(out var obj))
                            continue;

                        if (obj.RuntimeRun != run)
                        {
                            HandleFinalizationException(obj.PyObj, new RuntimeShutdownException(obj.PyObj));
                            continue;
                        }

                        IntPtr copyForException = obj.PyObj;
                        Runtime.XDecref(StolenReference.Take(ref obj.PyObj));
                        collected++;
                        try
                        {
                            Runtime.CheckExceptionOccurred();
                        }
                        catch (Exception e)
                        {
                            HandleFinalizationException(obj.PyObj, e);
                        }
                    }
```

Before this: 

```csharp
            NullGCHandles(ExtensionType.loadedExtensions);
            ClassManager.RemoveClasses();
            TypeManager.RemoveTypes();
            _typesInitialized = false;

            MetaType.Release();
            PyCLRMetaType.Dispose();
            PyCLRMetaType = null!;

            Exceptions.Shutdown();
            PythonEngine.InteropConfiguration.Dispose();
            DisposeLazyObject(clrInterop);
            DisposeLazyObject(inspect);
            DisposeLazyObject(hexCallable);
            PyObjectConversions.Reset();
```

lostmsu:
An explanation is not enough, we need a minimal test that fails before the fix, and passes after, preferably with just a handful of allocated objects.

I am being so picky because without certainty about exactly what's going on I can not be sure that your fix does not just randomly happens to help your particular use case while breaking someone else's use case not yet covered by tests. And it introduces a little extra complexity too.

Frawak:
This is the most down grinded version of my case that I could accomplish: https://github.com/Frawak/pythonnet/tree/fix/1977-access-violation-gc-testcase/exception_scenario
Please read the README.

I hope that the exception is as reliable on other machines.

> preferably with just a handful of allocated objects
The problem with that: The exception only occurs with some internal memory management from Python (my interpretation) which is hard to reproduce in general and I do not see how it can be done with a few simple objects. The process has to be triggered by its inner logic to access the memory while pythonnet is shutting down.

I also had to add the other open-source library because while using that, the exception occurred in specific cases. But there is nothing wrong with the library and its types. They are pretty simple and no IDisposables or the like.
Trying to magically trigger this exception with default .NET library types is gambling in lottery.

And as I said: It is not about the surrounding conditions when the exception is thrown. They are not the cause. The cause can be seen with any script using .NET objects from Python and debugging into the shutdown method. It is just lucky that the exception is rare.

Frawak:
@lostmsu Are you going to look at the code example? Does the exception get thrown when you run it?

It's a process. Work with me.

lostmsu:
@Frawak I took a look at the example, just want to see if I understood it correctly: it does not appear to have either Python classes derived from .NET classes, nor .NET classes derived from Python classes? E.g. all it does is imports .NET lib, instantiates a few classes from there as temporary variables, and that's it?

Frawak:
> I took a look at the example, just want to see if I understood it correctly: it does not appear to have either Python classes derived from .NET classes, nor .NET classes derived from Python classes? E.g. all it does is imports .NET lib, instantiates a few classes from there as temporary variables, and that's it?

That is correct.

> P.S. sorry about late response, we are unpaid maintainers here, doing it at our leisure.

No need to apologize. Of course I know that. But any information (like "maybe in x weeks") is better than no information. Otherwise, I don't know if it has genuinely slipped your mind and you need a poke or you come around to it at your own time. In my experience, the former is more common.

rohanjain101:
@lostmsu could you please help take another look?

lambda:
Just wanted to report that we were consistently running into this issue in our Windows CI tests, for a case where we were using pythonnet to load a proprietary .net file format parser (since it's proprietary, can't necessarily submit it as a test case).

We switched to point to this branch of pythonnet, and the issue has gone away.

So, hopefully that might provide another data point to address "I can not be sure that your fix does not just randomly happens to help your particular use case while breaking someone else's use case not yet covered by tests," this does fix a problem for us and doesn't seem to have introduced any new ones across our test matrix (Python 3.10, 3.11, 3.12 on Debian, and 3.10 on Windows).

Please do review and ensure this branch is properly tested, but we are eagerly looking forward to the release of this fix, or some other fix to this problem.

filmor:
@lostmsu The current investigation here would be enough for me to accept the PR as is. That derived classes require special handling is not surprising, given how they are implemented.

@Frawak It would be great if you could put your test-case into a separate repository and keep it alive for a while. If you have the time to integrate it into our test-suite, even better :)

filmor (approved):

lostmsu:
I still plan to take a look.

lostmsu:
OK, I did a review, and IMHO, this fix will just leak .NET memory because the GC handles will never be released after .NET types are disabled in Python. I am not sure, but it also might mean that some critical finalizers won't run.

IMHO, what should be done is after `tp_clear` is disabled, we should replace it with a Python function that will grab the GC handle and put it somewhere the rest of the shutdown procedure can look (e.g. some global Python object, essentially `list[int]` where `int` is the value of `GCHandle`). That way we can have a variant of `TryCollectGarbage` that does not refer to `reflectedObjects` and instead releases `GCHandle` from that list.

Ideally, after `tp_clear` is disabled we should also stop putting new objects into `reflectedObjects`. Then the `TryCollectGarbage` won't need special `bool` telling it if the objects in this list are still valid.

Frawak:
I created the scenario of your first paragraph with the code snippets shown below and debugged into the code (with my changes).
I could observe that the `forceBreakLoop` (before `RemoveTypes`) causes the `LoopReference` instance to be finalized at the next `GC.Collect()`. Otherwise, the finalizer gets called when the program exits. So, I get the point behind it now.
But I do not comprehend the necessity of breaking loops after `RemoveTypes`. I tried some additional code variations and debugged through them but could not figure out how to produce a respective example.

```csharp
using System;
namespace Temp;
public class LoopReference
{
    private readonly object _P;

    public LoopReference(object p)
    {
        _P = p;
    }

    ~LoopReference()
    {
        if (_P is IDisposable p)
            p.Dispose();
    }
}
```

```csharp
PythonEngine.Initialize();

PyDict locals = new PyDict();

PythonEngine.Exec(@"
import clr
clr.AddReference(r'[the path to the DLL containing LoopReference]')
from Temp import LoopReference
class P(object):
    def set_o(self, o):
        self.o = o
p = P()
o = LoopReference(p)
", null, locals);

PythonEngine.Shutdown();
```

> Ideally, after tp_clear is disabled we should also stop putting new objects into reflectedObjects. 

Essentially, this is the biggest puzzle piece for me. How are new objects added? I do not see any follow-up code in the shutdown procedure that does that.

filmor:
Maybe you could introduce a flag that makes adding an object to `reflectedObjects` throw after Shutdown? That would at least give us a stacktrace.

Frawak:
@filmor Like so (see last commit)?

Could you start the workflows again? I ran the tests locally.

Frawak:
In one of the configurations, an obsolete test failed:

https://github.com/pythonnet/pythonnet/blob/9ebfbde35eef742e39b80d4fe2464bdece4e34be/src/embed_tests/TestFinalizer.cs#L90-L92

The failed assert is also before running the shutdown.

GBBBAS:
Can confirm this branch fixes my Access Violation Exception issue I've been getting

filmor:
@lostmsu Anything else you would like to see?

src/runtime/Runtime.cs
```diff
@@ -351,6 +357,7 @@ static bool TryCollectingGarbage(int runs, bool forceBreakLoops)
                 {
                     NullGCHandles(CLRObject.reflectedObjects);
                     CLRObject.reflectedObjects.Clear();
+                    CLRObject.creationBlocked = true;
```
lostmsu, line 360: I think this needs to be moved to the callsite instead. E.g. immediately after `TryCollectingGarbage(forceBreakLoops: true)`.

Otherwise if you have a Python object that needs to temporarily create a .NET object in its destructor (for instance write log summary to a file), its destructor will fail if it happens to be freed on the second iteration of loop breaking.

lostmsu (changes requested):
Still uneasy about this, I think it might break stuff, but I was not able to come up with a test case.

Can be merged after `creationBlocking` issue mentioned above is resolved.

lostmsu (approved):

filmor:
Nice, thank you very much for your contribution @Frawak. I'll see this weekend if there is more to do before cutting a new release.

lostmsu:
Not sure if it's related yet, but we have two crashes in CI.

Frawak:
The difference in the log of the two failed jobs compared to others is the following:
The `CollectOnShutdown` test fails which I mentioned [above](https://github.com/pythonnet/pythonnet/pull/2386#issuecomment-2214237067). The error message `Garbage should contains the collected object` indicates that the test fails before the shutdown is called, i.e. before any change. 

https://github.com/pythonnet/pythonnet/blob/9ebfbde35eef742e39b80d4fe2464bdece4e34be/src/embed_tests/TestFinalizer.cs#L99-L102

Links to the log lines:
https://github.com/pythonnet/pythonnet/actions/runs/10274663048/job/28431871059#step:10:195
https://github.com/pythonnet/pythonnet/actions/runs/10274663048/job/28431871476#step:10:199

PR #2406, opened 2024-06-30
Add null checks before binding

filmor:


src/runtime/ClassManager.cs
```diff
@@ -288,15 +288,13 @@ internal static void InitClassBase(Type type, ClassBase impl, ReflectedClrType p
...
+            mb != null &&
+            (mb.IsPublic || mb.IsFamily || mb.IsFamilyOrAssembly);
 
-        internal static bool ShouldBindField(FieldInfo fi)
-        {
-            return (fi.IsPublic || fi.IsFamily || fi.IsFamilyOrAssembly);
-        }
+        internal static bool ShouldBindField(FieldInfo fi) =>
+            fi != null &&
+            (fi.IsPublic || fi.IsFamily || fi.IsFamilyOrAssembly);
```
lostmsu, line 297: Doesn't this just mask issues? I don't remember if we have nullability annotations on, but I think if you want to add a check here, it should throw `ArgumentNullException`, and instead `null` should be handled at the call site.

filmor: It does mask issues, but in a non-failing way which is a lot better than raising during startup (which effectively renders this library useless). This is an adaptation of what fixes https://github.com/pythonnet/pythonnet/discussions/2405. Throwing an exception here is (apart from having a slightly better stacktrace) effectively the same as leaving the code unchanged.

lostmsu: I mean I suggest to check for `null` at the call site which is what https://github.com/pythonnet/pythonnet/discussions/2405#discussioncomment-9850990 does, and optionally add an argument check in `ShouldBindMethod` to throw something better than `NullReferenceException`.

Simply returning `false` for `null` inputs is error-prone.

filmor: I don't agree, but I will also not put any more work into this. If you want to propose a different solution (in code), be my guest.

lostmsu: @filmor opened https://github.com/pythonnet/pythonnet/pull/2409

filmor:
Superceded by #2409

PR #2409, opened 2024-07-01
Fix crash when event does not have `Add` method and improve message for some other internal errors

lostmsu:

1. Fixes crash on events with no [`AddMethod`](https://learn.microsoft.com/en-us/dotnet/api/system.reflection.eventinfo.addmethod)
2. Adds `InternalPythonnetException` that can tell user what .NET type caused failure if an unexpected internal error occurs

https://github.com/pythonnet/pythonnet/discussions/2405

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ This document follows the conventions laid out in [Keep a CHANGELOG][].
 ### Fixed
 
 -   Fixed RecursionError for reverse operators on C# operable types from python. See #2240
+-   Fixed crash when .NET event has no `AddMethod`
 -   Fixed probing for assemblies in `sys.path` failing when a path in `sys.path` has invalid characters. See #2376
 
 ## [3.0.3](https://github.com/pythonnet/pythonnet/releases/tag/v3.0.3) - 2023-10-11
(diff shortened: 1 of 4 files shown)


filmor:
As said, while looking mildly prettier, this still renders the library completely useless in this case. I don't think there is any valid reason for us to bail out on init just because a single type is not loaded properly.

So, sure, we can merge this, but we can not claim that it fixes the reported issue.

lostmsu:
@filmor the fix is the new null check in `ShouldBindEvent`.

The rest is to make it easier to do bug reports.

filmor:
I see. Still don't see why making `ShouldBindEvent` special in that it swallows the error instead of throwing is better than just behaving like this uniformly. I'd rather have none of these functions throw, but at least it fixes the reported issue.

lostmsu:
@filmor it does not swallow the error.

The checks on methods and fields are different. They actually check that method or field was passed, and not just `null`. In case of event the event itself must exist as well (e.g. not null), but some events apparently do not have a corresponding add method for handlers: [AddMethod property is nullable](https://learn.microsoft.com/en-us/dotnet/api/system.reflection.eventinfo.addmethod)

BTW I was unable to reproduce it with C#.

PR #2419, opened 2024-07-17
Encoders no support specifying the target encoding type

joaompneves:
When you have an object that implements multiple interfaces in an explicit way, there's no easy way in Encoders to specify which of the interface we want to encode.
This PR allows that, by introducing an extra type paramater in the TryEncode method, implementors can know understand what is the target type that should be used. Also there's a new extension method that allows specifying the target type, when the type is not. known at build time (thus we can't use the generic version of the ToPython).
...

...

### Any other comments?

...

### Checklist

### DIFF
--- a/src/runtime/Converter.cs
+++ b/src/runtime/Converter.cs
@@ -987,5 +987,11 @@ namespace Python.Runtime
             if (o is null) return Runtime.None;
             return Converter.ToPython(o, typeof(T)).MoveToPyObject();
         }
+
+        public static PyObject ToPythonAs(this object o, Type type)
+        {
+            if (o is null) return Runtime.None;
+            return Converter.ToPython(o, type).MoveToPyObject();
+        }
     }
 }
--- a/src/runtime/Codecs/EncoderGroup.cs
+++ b/src/runtime/Codecs/EncoderGroup.cs
@@ -28,13 +28,13 @@ namespace Python.Runtime.Codecs
         /// <inheritdoc />
         public bool CanEncode(Type type) => this.encoders.Any(encoder => encoder.CanEncode(type));
         /// <inheritdoc />
-        public PyObject? TryEncode(object value)
+        public PyObject? TryEncode(object value, Type type)
         {
             if (value is null) throw new ArgumentNullException(nameof(value));
 
             foreach (var encoder in this.GetEncoders(value.GetType()))
             {
-                var result = encoder.TryEncode(value);
+                var result = encoder.TryEncode(value, type);
                 if (result != null)
                 {
                     return result;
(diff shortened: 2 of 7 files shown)


lostmsu:
Can you please provide an example of how this can be used?

src/embed_tests/Python.EmbeddingTest.csproj
```diff
@@ -2,6 +2,7 @@
 
   <PropertyGroup>
     <TargetFrameworks>net472;net6.0</TargetFrameworks>
+    <RollForward>LatestMajor</RollForward>
```
joaompneves, line 5: This allows anyone that has recent versions of .net to run these tests. I can revert but I think its usefull, otherwise one needs to install .net6 which is already deprecated

joaompneves:
> Can you please provide an example of how this can be used? Even better if it is a real-world usage example.

Added a test.

src/embed_tests/Python.EmbeddingTest.csproj
lostmsu: I think this should be conditioned on not running in CI

src/runtime/Converter.cs
```diff
@@ -987,5 +987,11 @@ public static PyObject ToPythonAs<T>(this T? o)
             if (o is null) return Runtime.None;
             return Converter.ToPython(o, typeof(T)).MoveToPyObject();
         }
+
+        public static PyObject ToPythonAs(this object? o, Type type)
+        {
+            if (o is null) return Runtime.None;
+            return Converter.ToPython(o, type).MoveToPyObject();
+        }
```
lostmsu, line 995: Please forward the old call to the new one.

src/runtime/PythonTypes/PyObject.cs
```diff
@@ -136,14 +136,14 @@ public IntPtr Handle
         /// Given an arbitrary managed object, return a Python instance that
         /// reflects the managed object.
         /// </remarks>
-        public static PyObject FromManagedObject(object ob)
+        public static PyObject FromManagedObject(object ob, Type? type = null)
         {
```
lostmsu, line 140: We need a use case and a test for this change as well.

Also, you have to add a separate overload because AFAIK adding default arguments breaks binary compatibility (e.g. libs compiled prior to this change will fail with `MissingMethodException` when they'll try to call the original method)

lostmsu (changes requested):

src/embed_tests/Python.EmbeddingTest.csproj
joaompneves: Is there any config/env var which I can check?

PR #2439, opened 2024-08-22
Update documentation to reflect API changes

### LINKED ISSUE #2433: Update documentation to reflect API changes
Similar to #2187, the Wiki also contains outdated code: The [Threading](https://github.com/pythonnet/pythonnet/wiki/Threading) page is still using `PythonEngine.ImportModule()`.

There might be other pages that are outdated, like [Loading Python file in .NET and Call function](https://github.com/pythonnet/pythonnet/wiki/Loading-Python-file-in-.NET-and-Call-function) where `Py.GIL()` is not inside a `using`. I'm unsure if this is still correct.

I've also seen various code snippets where the new `Runtime.DLL` requirement is absent, most notably in the [README](https://github.com/pythonnet/pythonnet/blob/ea059ca0aeec4cdad490c1a68c783908dd879607/README.rst?plain=1#L86) and the [documentation](https://pythonnet.github.io/pythonnet/dotnet.html).

TheJanzap:
Adds the now mandatory use of setting the Python DLL in C# to the README and documentation

Fixes #2433 

### Any other comments?
Also updated the use of `PyModule.Import` in the wiki with https://github.com/pythonnet/pythonnet/wiki/Threading/aee4b778efd079911d1b771c8b59ba5bd7dd156d

### Checklist


doc/source/dotnet.rst
```diff
@@ -15,7 +15,9 @@ These classes include PyObject, PyList, PyDict, PyTuple, etc.
 At a very high level, to embed Python in your application one will need
 to:
 
--  Reference ``Python.Runtime.dll`` (e.g. via a ``PackageReference``)
+-  Reference ``Python.Runtime.dll`` (e.g. via a ``PackageReference``,
+   the ``Runtime.PythonDLL`` property or the ``PYTHONNET_PYDLL``
```
filmor, line 19: This is wrong. The `Runtime.PythonDLL` and `PYTHONNET_PYDLL` reference `libpython*.so/dylib` or `python*.dll`.

README.rst
```diff
@@ -85,6 +85,7 @@ Example
 
    static void Main(string[] args)
    {
+       Runtime.PythonDLL = "python38.dll";
```
filmor, line 88: I really don't want to give people the idea that this is the way to go. Maybe we can point them toward (or include) https://github.com/losttech/WhichPython? Maybe give a note above to set the `PYTHONNET_PYDLL` environment variable?

filmor (changes requested):

doc/source/dotnet.rst
TheJanzap: Ah, I didn't know that there's a difference. But since it's now mandatory to reference a Python DLL, I think it's a good idea to mention it here. Maybe as a separate bullet point?

README.rst
TheJanzap: Yeah, I wasn't really keen on this either, but I figured since this was one of the examples given in the paragraph above it that this would be at least acceptable.
I also thought about mentioning the env var, but I didn't want to clutter this example with a comment (at least not without the permisson of you guys)

TheJanzap:
I've noticed that `Runtime.PythonDLL` is also absent in the [C# reference](https://pythonnet.github.io/pythonnet/reference.html). Would it be enough to just remove the comment from
```rst
.. doxygennamespace:: Python::Runtime
       :members:
       :outline:
```
to enable it? Apologies, I'm new to RST and I get errors about missing `index.xml` files, so I can't compile it myself.

TheJanzap:
@filmor Is this PR still relevant?

PR #2440, opened 2024-08-22
Try to be strict around non-copyability

filmor:


src/runtime/Types/ManagedType.cs
```diff
@@ -83,8 +83,8 @@ internal static unsafe void DecrefTypeAndFree(StolenReference ob)
 
             var freePtr = Util.ReadIntPtr(type, TypeOffset.tp_free);
             Debug.Assert(freePtr != IntPtr.Zero);
-            var free = (delegate* unmanaged[Cdecl]<StolenReference, void>)freePtr;
-            free(ob);
+            var free = (delegate* unmanaged[Cdecl]<ref StolenReference, void>)freePtr;
+            free(ref ob);
```
lostmsu, line 87: I think this is wrong. It does not say anywhere but I'd expect `tp_free` to take a reference to `PyObject` (e.g. `PyObject*`), not a reference to reference to it (e.g. not a `PyObject**`).

tools/NonCopyableAnalyzer/NonCopyableAnalyzer.cs
```diff
@@ -0,0 +1,293 @@
...
+using System.Collections.Immutable;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.Diagnostics;
+using Microsoft.CodeAnalysis.Operations;
+
+namespace NonCopyable
+{
+    [DiagnosticAnalyzer(LanguageNames.CSharp)]
+    public class NonCopyableAnalyzer : DiagnosticAnalyzer
```
lostmsu, line 11: Why can't we put it into a separate repository?

Maybe use as a submodule if you don't want to reference it as a package.

src/runtime/Types/ManagedType.cs
filmor: Ah, that's probably one reason for the crashes :)

tools/NonCopyableAnalyzer/NonCopyableAnalyzer.cs
filmor: It's currently buggy. Once evrything works, we can move the changes to your repo again.

lostmsu:
Seems abandoned - requested changes not addressed.

PR #2454, opened 2024-09-19
Python 3.13

filmor:
This is crashing right now.

- [x] Investigate whether the generated type offsets are correct (investigated, but missed an error, see https://github.com/pythonnet/pythonnet/pull/2454#issuecomment-2539826002)
- [x] Decide how to deal with the removed `Get`ter for the thread state
- [x] Find out why it crashes :)


RobPasMue:
@filmor @lostmsu - is there a plan for enabling support for 3.13? I see this PR has been stale for quite a while already. Just checking what's the status :)

filmor:
All development is public. You can see here that the "simple" approach of adding new offsets (that I have indeed verified) fails. Debugging gave me additionally that it crashes in some very early dictionary access, but I don't really know how to go further.

src/runtime/Runtime.Delegates.cs
```diff
@@ -23,7 +23,14 @@ static Delegates()
...
             PyThreadState_New = (delegate* unmanaged[Cdecl]<PyInterpreterState*, PyThreadState*>)GetFunctionByName(nameof(PyThreadState_New), GetUnmanagedDll(_PythonDll));
             PyThreadState_Get = (delegate* unmanaged[Cdecl]<PyThreadState*>)GetFunctionByName(nameof(PyThreadState_Get), GetUnmanagedDll(_PythonDll));
-            _PyThreadState_UncheckedGet = (delegate* unmanaged[Cdecl]<PyThreadState*>)GetFunctionByName(nameof(_PyThreadState_UncheckedGet), GetUnmanagedDll(_PythonDll));
+            try
+            {
+                _PyThreadState_UncheckedGet = (delegate* unmanaged[Cdecl]<PyThreadState*>)GetFunctionByName(nameof(_PyThreadState_UncheckedGet), GetUnmanagedDll(_PythonDll));
+            }
+            catch (MissingMethodException)
+            {
+                // Not supported in Python 3.13 anymore
```
kunom, line 32: Isn't it that the function merely has been renamed to `PyThreadState_GetUnchecked` (see https://docs.python.org/3/c-api/init.html#c.PyThreadState_GetUnchecked).

filmor: Great, thanks! I only looked for `PyThreadState_UncheckedGet` :/

kunom:
According to `cvdump "c:\Program Files\Python313\python313.pdb"` (`cvdump` from [here](https://github.com/Microsoft/microsoft-pdb/blob/master/cvdump/cvdump.exe)), the `PyHeapTypeStruct` is stated to have a size of 928 bytes:

```
[...]
S_UDT:             0x1211, PyHeapTypeObject
[...]
0x1211 : Length = 58, Leaf = 0x1505 LF_STRUCTURE
	# members = 13,  field list type 0x1210, 
	Derivation list type 0x0000, VT shape type 0x0000
	Size = 928, class name = _heaptypeobject, unique name = .?AU_heaptypeobject@@, UDT(0x00001211)
[...]
```

However, when I look at the value of the `TypeOffset313.init` value, it is also 928. 
image

We seem to have one property too much, don't we?

filmor:
Aaah, got it, thanks @kunom!

The issue is that, unlike all other members, the new `tp_watched` and `tp_versions_used` attributes are `char` and `uint16_t`. I'll update our `geninterop` script to pick this up and release a new version latest tomorrow (CET).

I might have to release a new version of `clr-loader` as well to bump the `cffi` dependency.

kunom:
Not shure whether this still matters, but I was finally able to produce an offset listing of the full struct using [Ghidra reverse engineering](https://ghidra-sre.org/). According to that tool, the following struct members have a size different from 8:

|Name|Size|
|----|---|
|`tp_flags`|4|
|`tp_version_tag`|4|
|`tp_watched`|1 (as stated above)|
|`tp_versions_used`|2 (as stated above)|
|`getitem_version`|4|

filmor:
On my local machine, the embedding tests also pass for 3.13. I'll have to spend some more time tomorrow to get the CI going, but I'm quite confident that I can release a new version either tomorrow or this weekend.

lostmsu:
Not that I'm against 3.13 support, but this seems to fail multiple tests.

filmor:
It is not failing tests, our CI is just a bit rusty. I ran a many combinations locally and will follow up in the next week to reactivate the excluded platforms.

maged-hanna:
@filmor I'm receiving a 'Attempted to read or write protected memory' error in my C# code when interacting with Python 3.13. This suggests potential memory corruption. Could this error be due to incompatibility with Python 3.13, or is it a sign that 3.13 is not supported?  
`var m_threadState = PythonEngine.BeginAllowThreads();
using (Py.GIL()){}`

Issue #2348, opened 2024-03-22
Unhandled Exception when exiting from nPython.exe (net472)


xiazeyi:
### Environment

-   Pythonnet version: 3.0.3
-   Python version: 3.11.5
-   Operating System: Windows 10
-   .NET Runtime:

### Details

- Download source code from https://github.com/pythonnet/pythonnet/releases/tag/v3.0.3
- Build the console with VS 2022
- Set PYTHONNET_PYDLL=python311.dll (Python 3.11.5 is installed)
- Run nPython.exe in net472
- Input `quit()` and press Return to exit
  
```
C:\pythonnet-3.0.3\net472>nPython.exe
Python 3.11.5 (tags/v3.11.5:cce6ba9, Aug 24 2023, 14:38:34) [MSC v.1936 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> quit()

Unhandled Exception: System.AccessViolationException: Attempted to read or write protected memory. This is often an indication that other memory is corrupt.
   at Python.Runtime.Runtime.PyGILState_Ensure()
   at Python.Runtime.Py.GIL()
   at Python.Runtime.PythonEngine.Shutdown()

Unhandled Exception: System.InvalidOperationException: GIL must always be released, and it must be released from the same thread that acquired it.
   at Python.Runtime.Py.GILState.Finalize()
```

- Run nPython.exe in net472
- Input `Ctrl + Z` and press Return to exit

```
C:\pythonnet-3.0.3\net472>nPython.exe
Python 3.11.5 (tags/v3.11.5:cce6ba9, Aug 24 2023, 14:38:34) [MSC v.1936 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> ^Z


Unhandled Exception: System.AccessViolationException: Attempted to read or write protected memory. This is often an indication that other memory is corrupt.
   at Python.Runtime.Runtime.PyGILState_Ensure() in C:\pythonnet-3.0.3.tar\pythonnet-3.0.3\src\runtime\Runtime.cs:line 705
   at Python.Runtime.Py.GIL() in C:\pythonnet-3.0.3.tar\pythonnet-3.0.3\src\runtime\Py.cs:line 13
   at Python.Runtime.PythonEngine.Shutdown() in C:\pythonnet-3.0.3.tar\pythonnet-3.0.3\src\runtime\PythonEngine.cs:line 397
[...]

PR #2456, opened 2024-09-19
Drop console application

filmor:
I don't really see the point of this one. We could keep it as an example, but
apparently it is even buggy (#2348, probably due to mixing the embedding modes).


lostmsu:
I know we aren't supporting it, but:
1. What do you mean by "mixing the embedding modes"?
2. If a simple sample like this one has problems, are we not on fire?

filmor:
It uses `Py_Main` which no one else uses. And if you would like to run .NET code from within Python, using the `clr_loader` path is the far more stable option.

lostmsu (approved):

PR #2470, opened 2024-10-01
Use the BinaryFormatter only on .NET FW and Mono

filmor:


src/embed_tests/StateSerialization/MethodSerialization.cs
```diff
@@ -33,6 +33,10 @@ static T SerializationRoundtrip<T>(T item)
     {
         using var buf = new MemoryStream();
         var formatter = RuntimeData.CreateFormatter();
+
+        if (formatter == null)
```
lostmsu, line 37: It might be a good idea to make the whole test class `[Obsolete]` to prevent warnings

src/runtime/StateSerialization/RuntimeData.cs
```diff
@@ -82,32 +84,37 @@ static void ClearCLRData ()
...
-
             ClearCLRData();
 
-            using NewReference capsule = PyCapsule_New(mem, IntPtr.Zero, IntPtr.Zero);
-            int res = PySys_SetObject("clr_data", capsule.BorrowOrThrow());
-            PythonException.ThrowIfIsNotZero(res);
-            PostStashHook?.Invoke();
+            IFormatter? formatter = CreateFormatter();
+
+            if (formatter != null)
```
lostmsu, line 91: Can't we just exit the method if formatter is null? Would avoid polluting git history

src/runtime/StateSerialization/RuntimeData.cs
```diff
@@ -124,28 +131,32 @@ internal static void RestoreRuntimeData()
 
         private static void RestoreRuntimeDataImpl()
         {
-            PreRestoreHook?.Invoke();
-            BorrowedReference capsule = PySys_GetObject("clr_data");
-            if (capsule.IsNull)
+            IFormatter? formatter = CreateFormatter();
+
+            if (formatter != null)
```
lostmsu, line 136: same

lostmsu (approved):
Look good, but I'd like NITs to be fixed and the commit squashed

src/runtime/StateSerialization/RuntimeData.cs
filmor: Sure, will do.

filmor:
Well, without the stashing, `Shutdown`+`Initialize` cycles are broken. This is going to be a lot more work than anticipated.

lostmsu:
We could consider dropping the support for Shutdown/Initialize for .NET Core+. I don't know if there are any good scenarios for it.

src/embed_tests/StateSerialization/MethodSerialization.cs
lostmsu: Obsolete code warning that comes from using `BinaryFormatter`. You can see it on GitHub review page.

src/runtime/StateSerialization/RuntimeData.cs
```diff
@@ -124,6 +132,12 @@ internal static void RestoreRuntimeData()
 
         private static void RestoreRuntimeDataImpl()
         {
+            IFormatter? formatter = CreateFormatter();
```
lostmsu, line 135: I think coming here should be an error.

Issue #2531, opened 2024-12-17
Unimplemented or broken methods in Python wrappers to standard C# containers


DiegoBaldassarMilleuno:
### Environment

-   Pythonnet version: 3.0.5
-   Python version: 3.13
-   Operating System: MacOS Sequoia 15.1.1
-   .NET Runtime: 8.0.403

### Details

After #2530, I took the time to test from Python some of the most common C# containers, testing in particular the methods from the standard Python abc's (`Sequence`, `MutableSequence`, `Mapping`, `MutableMapping`, ...).
Here are my results, along with a few proposed fixes. Please let me know if there are any mistakes.

### Shared setup
```python
import pythonnet
pythonnet.load('coreclr')
import clr
from System.Collections.Generic import Dictionary, List, KeyValuePair
from System.Collections.Immutable import ImmutableArray, ImmutableList, ImmutableDictionary
from System import Int32, String, Object, Nullable, Array
from System.Collections.Generic import CollectionExtensions
```

## Array-like containers
```python
ArrT = Array[Int32]
ImmArrT = ImmutableArray[Int32]
ListT = List[Int32]

arr = ArrT([0, 1, 2, 3, 4])
immarr = ImmutableArray.Create[Int32](0, 1, 2, 3, 4)  # is this supposed to work, instead of using Create()?
lst = ListT(ArrT([0, 1, 2, 3, 4]))
immlst = ImmutableList.ToImmutableList[Int32](ArrT([0, 1, 2, 3, 4]))
```

### What worked everywhere
- `__iter__`
- `__reversed__`
- `__contains__`
- `__len__`
- `reverse()`
- `__getitem__` when the index is within bounds
- `index()` when the element is in the container
- `__setitem__` when the index is within bounds

### What didn't work
- `__getitem__` when the index is negative or out of bounds
```python
arr[-1]     # OK
arr[999]    # OK, raises IndexError
immarr[-1]  # System.IndexOutOfRangeException: Index was outside the bounds of the array.
lst[-1]     # System.ArgumentOutOfRangeException: Index was out of range. Must be non-negative and less than the size of the collection. (Parameter 'index')
immlst[-1]  # System.ArgumentOutOfRangeException: Specified argument was out of the range of valid values. (Parameter 'index')
```
[...]

PR #2535, opened 2024-12-20
Add tests for collection mixins (see issue #2531)

DiegoBaldassarMilleuno:
As requested in https://github.com/pythonnet/pythonnet/issues/2531#issuecomment-2548279224, submitting a draft pull request testing for compliance to `collections.abc` protocols by a few standard containers.
Currently many tests fail.

A few notes:
1. the tests that hard crash are temporarily disabled via `pytest.mark.xfail`, for convenience;
2. many requirements I added in these tests are rather pedantic (like supporting values of the wrong type in `__contains__` or requiring specific exceptions to be thrown); maintainers feel free to remove any that you believe are outside of contract.
3. I haven't checked that this works with older versions of Python, it's a quick draft

### DIFF
--- a/tests/test_collection_mixins_mapping.py
+++ b/tests/test_collection_mixins_mapping.py
@@ -1,0 +1,252 @@
+from __future__ import annotations
+
+import operator
+from typing import Any
+from collections.abc import Container
+
+import pytest
+
+import clr
+clr.AddReference('System.Collections.Immutable')
+from System import Nullable, Object, ValueType, Int32, String
+import System.Collections.Generic as Gen
+import System.Collections.Immutable as Imm
+import System.Collections.ObjectModel as OM
+
+kv_pairs_1 = ((0, "0"), (10, "10"), (20, "20"), (30, "30"))
+kv_pairs_2 = kv_pairs_1 + ((40, None), )
+kv_pairs_3 = (("0", 0), ("10", 10), ("20", 20), ("30", 30))
+kv_pairs_4 = kv_pairs_3 + (("40", None), )
+
+
+def exactly_equal(val1, val2):
+    return type(val1) is type(val2) and val1 == val2, '{0!r} != {1!r}'.format(val1, val2)
+
+
+def translate_pytype(pytype, *, nullable=False):
+    if pytype is int:
+        cstype = Int32
+    elif pytype is str:
+        cstype = String
+    else:
+        raise NotImplementedError('Unsupported type: {0!r}'.format(pytype))
+
+    if nullable and issubclass(cstype, ValueType):
+        cstype = Nullable[cstype]
+
+    return cstype
+
+
+class MappingTests:
+    ktype: type
+    vtype: type
+    nullable: bool
+    cs_ktype: type
+    cs_vtype: type
+    mapped_to_null: Container[Any]
+    dct: Any
+
+    def __init_subclass__(cls, /, values=None, **kwargs):
+        if values is not None:
+            cls.ktype,cls.vtype,cls.nullable,cls.cs_ktype,cls.cs_vtype = cls.deduce_types(values)
+            cls.mapped_to_null = tuple(key for key,val in values if val is None)
+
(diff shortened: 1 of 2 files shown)


tests/test_collection_mixins_mapping.py
```diff
@@ -0,0 +1,252 @@
...
+        if nullable:
+            vtypes.remove(type(None))
+        (vtype, ) = vtypes
+        cs_ktype = translate_pytype(ktype)
+        cs_vtype = translate_pytype(vtype, nullable=nullable)
+        return (ktype, vtype, nullable, cs_ktype, cs_vtype)
+
+    def test_len(self):
+        assert type(len(self.dct)) is int
+        assert len(self.dct) == 4 if not self.nullable else 5
```
lostmsu, line 68: This line is very hard to read

tests/test_collection_mixins_mapping.py
```diff
@@ -0,0 +1,252 @@
...
+            assert type(tpl) is tuple and len(tpl) == 2
+            key,val = tpl
+            assert exactly_equal(key, self.ktype(idx * 10))
+            exp_val = None if self.ktype(10 * idx) in self.mapped_to_null else self.vtype(10 * idx)
+            assert exactly_equal(val, exp_val)
+
+    def test_contains(self):
+        assert self.ktype(10) in self.dct
+        assert self.ktype(50) not in self.dct
+        assert 12.34 not in self.dct
```
lostmsu, line 98: We need to pay attention here because .NET semantics might be different from Python semantics. In .NET Dictionary is a ICollection of KeyValuePair, so it has different Contains based on that interface.

Basically need a 2x2 explicit asserts: 2 for key vs KVP and 2 for `in` vs `not in`

tests/test_collection_mixins_mapping.py
```diff
@@ -0,0 +1,252 @@
...
+        dct[key] = val
+        assert exactly_equal(dct[key], val)
+
+    def test_setitem_raise(self):
+        if isinstance(self.dct, Object):  # this is only relevant for CLR types
+            dct = self.get_copy()
+            with pytest.raises(Exception):
+                dct[12.34] = self.vtype(0)
+
+    @pytest.mark.xfail(reason='Known to crash', run=False)
```
lostmsu, line 135: Address that?

tests/test_collection_mixins_mapping.py
DiegoBaldassarMilleuno:
```suggestion
        expected_length = 4 if not self.nullable else 5
        assert len(self.dct) == expected_length
```
Something like this?

tests/test_collection_mixins_mapping.py
DiegoBaldassarMilleuno: @lostmsu standard Python semantics for mappings want `key in mapping` to check for keys (equivalent of `ContainsKey`) and `kvp in mapping.items()` to check for key-value pairs (as tuples, otherwise analogous to `Contains`); they also have a mapping be an iterable of its keys, so `key in mapping` is equivalent to `key in list(mapping)`.  
Like you said, this is different from .NET where an `IDictionary<TKey,TValue>` is an `ICollection<KeyValuePair<TKey,TValue>>`.  
I'm for using Python semantics when using Python operators and .NET semantics when using .NET methods, both for interoperability with existing libraries and to avoid confusion.
That said I could add tests checking that `.items()` returns an object implementing [ItemsView](https://docs.python.org/3/library/collections.abc.html#collections.abc.ItemsView) (also see [here](https://docs.python.org/3/library/collections.abc.html#collections-abstract-base-classes)), including what you said.  
Would that be ok?

tests/test_collection_mixins_mapping.py
DiegoBaldassarMilleuno: Ok, it doesn't crash anymore, will reenable this and `test_delitem_raise` next commit

PR #2543, opened 2025-01-13
Proposal: Implement binding manager to hold binding overrides

koubaa:

This adds a place to register at runtime options to customize the binding of types. This customization can be specified at a type or assembly level.

No, but it can be used as a way to close issues going forwards - especially in cases where there is disagreement about what the correct way to bind to python, or if there are legacy compatibility reasons to bind in a way that is not the default way. For example, in the `ansys-pythonnet` fork, this is proposed to be used in two ways:

1. to bind properties and methods coming from [explicit interface implementation](https://github.com/ansys/ansys-pythonnet/pull/23) in order to satisfy a requirement in one library for compatibility with legacy IronPython scripts. In IronPython, explicit interface implementations are provided in Python, even when that is not the idiomatic choice.
2. to automatically add [pep8 aliases](https://github.com/ansys/ansys-pythonnet/pull/24), allowing python code to use normal python style even when the C# library uses the normal C# style.

As this is just a proposal - I will move forward with the checklist if the maintainers of pythonnet agree with the direction.

### Example usage (from python)

```
m = clr.AddReference('MyLib')
from Python.Runtime import BindingManager, BindingOptions
binding_options = BindingOptions()
binding_options.SuppressDocs = True
BindingManager.SetBindingOptions(m, binding_options)
```

### Checklist

### DIFF
--- a/src/runtime/BindingOptions.cs
+++ b/src/runtime/BindingOptions.cs
@@ -1,0 +1,65 @@
+using System;
+using System.Collections.Generic;
+using System.Reflection;
+
+namespace Python.Runtime
+{
+    public class BindingOptions
+    {
+        private bool _SuppressDocs = false;
+        private bool _SuppressOverloads = false;
+
+        //[ModuleProperty]
+        public bool SuppressDocs
+        {
+            get { return _SuppressDocs; }
+            set { _SuppressDocs = value; }
+        }
+
+        //[ModuleProperty]
+        public bool SuppressOverloads
+        {
+            get { return _SuppressOverloads; }
+            set { _SuppressOverloads = value; }
+        }
+    }
+
+    public class BindingManager
+    {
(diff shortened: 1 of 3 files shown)


koubaa:
@filmor is this PR under consideration?

lostmsu:
We wouldn't want to have binding options as it immensely increases the test matrix: every toggle doubles the space to test.

A option to completely override the binding mechanism with deferring the responsibility to the user might be a good idea, but that would require exceptionally careful work with virtual methods.

filmor:
@koubaa I would like to take this into consideration after we have finally landed the 3.14 changes some time this week. I would like to revisit the interface, though. I'm not a huge fan of using Python.NET to configure itself (documenting the `Python.Runtime` namespace as something to use in user code), so I would rather like something that uses the `pythonnet` or `clr` (though we should eventually deprecate this one) with a pythonic API to modify the settings.

koubaa:
@filmor @lostmsu I would be open to working on this if you have any API designs you'd like to explore

PR #2590, opened 2025-05-06
Implement coerce logic for net method binding

### LINKED ISSUE #2360: PythonNET unknown kwargs being discarded
### Environment

-   Pythonnet version: 3.0.3
-   Python version: 3.11
-   .NET Runtime: .NET 8.0

### Details
When accessing a .NET function using PythonNET, if a kwarg is incorrect or doesn’t exist, it is discarded by the PythonNET execution engine

If I have a .NET function like this:

Public Function SomeFunction(name as String, age as String) as String
    Return $"{name} - {age}"
End Function
And I call it in PythonNET like this:

if __name__ == "__main__":
    name = "John"
    age = "3"
    SomeFunction(name, age=age, fake_arg="something")

The fake arg is discarded, and no error or exception is thrown. Python compiles as if the additional kwarg were not there. In this situation, normal Python would be throwing a TypeError.

This is an issue with a multiple kwargs in the methods/functions is that a user can easily misspell and result in unexpected behavior.

[PythonDemo.zip](https://github.com/pythonnet/pythonnet/files/15038606/PythonDemo.zip)

generalloki:

This allows additional checks to be performed before .NET members are called from Python scripts.  
It also enables the cancellation of a method call if it is not permitted by the application's security policies.

This allows to fix [issue #2360](https://github.com/pythonnet/pythonnet/issues/2360) by implementing a custom coercion handler.

### DIFF
--- a/src/runtime/MethodBinder.cs
+++ b/src/runtime/MethodBinder.cs
@@ -9,6 +9,12 @@ using System.Linq;
 namespace Python.Runtime
 {
     using MaybeMethodBase = MaybeMethodBase<MethodBase>;
+
+    public delegate void MethodBinderCoerceBindDelegate(
+        Dictionary<string, PyObject> arguments,
+        MethodBase[] methods,
+        ref MethodBase? foundBinding);
+
     /// <summary>
     /// A MethodBinder encapsulates information about a (possibly overloaded)
     /// managed method, and is responsible for selecting the right method given
@@ -516,7 +522,7 @@ namespace Python.Runtime
                     }
                 }
 
-                return new Binding(mi, target, margs, outs);
+                return CoerceResult(new Binding(mi, target, margs, outs));
             }
             else if (matchGenerics && isGeneric)
             {
@@ -528,7 +534,7 @@ namespace Python.Runtime
                 MethodInfo[] overloads = MatchParameters(methods, types);
                 if (overloads.Length != 0)
                 {
-                    return Bind(inst, args, kwargDict, overloads, matchGenerics: false);
+                    return CoerceResult(Bind(inst, args, kwargDict, overloads, matchGenerics: false));
                 }
             }
             if (mismatchedMethods.Count > 0)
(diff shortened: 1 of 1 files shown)


filmor:
Sorry for the late reply. Could you give a sketch of how one would fix 2360 with this?

generalloki:
We implemented such method for that:

```cs
        public static void CoerceBindHandler(
            Dictionary<string, PyObject> arguments,
            MethodBase[] methods,
            ref MethodBase foundMethod)
        {
            var fb = foundMethod;

            if (foundMethod is not null && arguments is not null)
            {
                foreach (var argName in arguments.Keys)
                {
                    if (!IsValid(argName))
                    {
                        Exceptions.SetError(Exceptions.TypeError, $"Invalid parameter name: {argName}");
                        foundMethod = null;
                        return;
                    }
                }
            }

            bool IsValid(string paramName)
            {
                foreach (ParameterInfo param in fb.GetParameters())
                {
                    if (param.Name == paramName)
                        return true;
                }

                return false;
            }
        }
```

So, if coerce logic will be added to the library we could just use it

```cs
MethodBinderEvents.CoerceBind = ScriptEngine.CoerceBindHandler;
```

in order to fix #2360

lostmsu:
What is the use case? Can you give your specific example? Why can't you just have an overload with explicit kwargs on C# side, and then route as needed?

generalloki:
I am not sure what you suggest to overload. Initially our customer reported the problem. Here is his request:

<hr/>

When accessing a .NET function using PythonNET bindings I noticed that if a kwarg is incorrect or doesn’t exist it is discarded by the PythonNET parser. For example

If I have a .NET function like this:

```vb
Public Function SomeFunction(name as String, age as String) as String
    Return $"{name} - {age}"
End Function
```

And I call it in PythonNET like this:

```py
if __name__ == "__main__":
    name = "John"
    age = "3"
    SomeFunction(name, age=age, fake_arg="something")
```

The fake arg is discarded and no error or exception is thrown. The python compiles as if the additional kwarg is not there. In this situation normal python would be throwing a TypeError.

This is an issue for us when we have multiple kwargs in our methods/functions that a user can easily misspell and result in unexpected behaviour.

<hr/>

We posted an issue about kwargs but it is not fixed from that moment, so we suggest  an improvement that will allow bind a customizable event which is fired before .net method is called. In the implementation of this event we can fix the issue and also add some other security related logic

PR #2611, opened 2025-08-09
Python 3.14

filmor:


kunom:
Some subjective feedback from my Windows machine with VS 2022:
- To get the tests compiling, I needed to replace the target framework identifiers from `net6.0` to `net8.0` in the .csproj files (that's maybe a subset of #2612).
- You forgot to reapply the [alignment hack from TypeOffset313.cs](https://github.com/pythonnet/pythonnet/blob/a21c7972809500edd6364003c37695193ac1a5d4/src/runtime/Native/TypeOffset313.cs#L85). With that applied, I get a few access memory violations less.

guywithface:
Also had to replace the target framework from net6.0 to net8.0. Also some feedback for .net 8:

- Add `#pragma warning disable SYSLIB0011` for where BinaryFormatter is being used 
- Add  `<EnableUnsafeBinaryFormatterSerialization>true</EnableUnsafeBinaryFormatterSerialization>` to all projects
- Remove NonCopyable from ref struct NewReference
- Remove NonCopyable from ref struct StolenReference
- Need to consider migrating away from using BinaryFormatter as it will be deprecated in .net9

And now getting an access memory violation for PyGC_Collect(); in Runtime.cs

filmor:
Status report: After applying the alignment "fix" (thanks!), things work for the most part. There are some crashes in explicit garbage collection, so I have disabled those tests for now.

filmor:
Also, we currently apparently have a weird inheritance order for our meta type, which leads to it using the `PyObject_GenericSetAttr` instead of `type_setattro`, which in turn triggers assertions since https://github.com/python/cpython/commit/5a1618a2c8c108b8c73aa9459b63f0dbd66b60f6

kunom:
@filmor As the CI seems to be pretty broken (`pytest: command not found`), could you please provide a list of the outstanding tasks for this PR to reach completion, as you did in the description block of [the corresponding 3.13 PR](https://github.com/pythonnet/pythonnet/pull/2454).

Things not clear to me:
- do you want to have the exluded tests re-enabled?
- do the tests failing due to missing BinarySerializer functionality block this PR?

Thanks ;-)

filmor:
There you go. There are still pending issues with the embedding tests that are being worked on in #2638, but I'd be fine with releasing a wheel if we get to the point that all Python tests run through without crashes on 3.14.

andre-romano:
Please verify pythonnet support for Python version 3.13.9 (building is not working in Windows, similarly to version 3.14)

filmor:
This is the issue that we run into with the `PyObject_GenericSetAttr` usage:
- https://github.com/cython/cython/issues/6323
- Related Python PR: https://github.com/python/cpython/pull/118454

filmor:
@lostmsu Please have a look. It will require quite a bit more work to get everything green again (CI-wise, in particular around the embedding), but with these changes, we should be able to release a wheel for Python 3.14.

goodyes:
Hi, just new on. I'need compatibility with python 3.14.3. I dont know much about githubs, merges, etc.. I only know that using nuget (vs2026) the latest version I can add of pythonnet is 3.0.5 (vb.net project 4.7.2) which has no support for python 3.14.3.
How can I get compatibility for python 3.14.3 ? thanks.

Also using a vb.net (10. net, not net framework say 4.7.2) also only allows me to nuget pythonnet 3.0.5 which not supports python 3.14.3.

Any update ?

goodyes:
Hi, after cloning repository and done the modifications said in pull 2611 (Python 3.14 #2611) . I've got folowing error:  System.AccessViolationException: 'Intento de leer o escribir en la memoria protegida. A menudo, esto indica que hay otra memoria dañada.'

goodyes:
Hi, the error is on       
  internal static PyGILState PyGILState_Ensure() => Delegates.PyGILState_Ensure();
first I must said that previously I run without locking, then after putting the Using (Py.GIL()) I still get the same error, from runtime.cs

filmor:
Merging as is now, will fix the CI in #2669

Issue #2627, opened 2025-10-09
Multiple levels of Inheritance causes AccessViolationException due to Finalize being called multiple times


rmadsen-ks:
Pythonnet version: v3.0.5  built from source

Reproduced on Windows 11. Not reproduced on MacOS.

The issue occurs during Shutdown, when all the live objects gets disposed.

I managed to reproduce it using a new unit test when I was trying to fix #2574.

PR #2628, opened 2025-10-09
2574 StackOverflow Exception On Super Method Call Fix

### LINKED ISSUE #2574: StackOverflowException on calling super().method()
### Environment
* Python.NET version: 3.0.5
* Tested on two configurations:
  * Windows 10, Python 3.12.4, .NET Framework
  * Windows 10, Python 3.12.4, .NET 9.0.201

### Details

Code to reproduce the issue is below.

.NET library code:
```csharp
namespace TestLib
{
    public class BaseClass
    {
        public virtual string Foo()
        {
            return "Foo";
        }

        public virtual string Bar()
        {
            return "Bar";
        }
    }

    public class DerivedClass : BaseClass
    {
        public override string Foo()
        {
            return "DerivedFoo";
        }
    }
}
```

Python code:
```python
from pathlib import Path

import clr

clr.AddReference(str(Path(__file__).parent / 'TestLib.dll'))

from TestLib import BaseClass, DerivedClass


class TestClass(DerivedClass):
    __namespace__ = 'CrlTest'

    def Foo(self):
        return 'Test' + super().Foo()

    def Bar(self):
        # This work fine
        # return 'Test' + DerivedClass.Bar(self)
        # This calls TestClass::Bar for some reason
        return 'Test' + super().Bar()


if __name__ == '__main__':
    stc = TestClass()
    print('Foo() -> ' + stc.Foo())
    print('Bar() -> ' + stc.Bar())
```

Python script output:
```
Foo() -> TestDerivedFoo

Process is terminated due to StackOverflowException.
```

For some reason calling `super().Bar()` results in calling `TestClass.Bar(self)` instead of `DerivedClass.Bar(self)`, which in the end results in infinite recursion. Calling `DerivedClass.Bar(self)` directly works as expected.

It seem to be related to the fact that the method you try to override in Python is not defined in the immediate parent class, since `Foo` in the example works fine.

rmadsen-ks:
~**Update: Does not actually fix #2574, but a similar problem(#2629). I'll provide a fix for 2574 as well.**~
**Update 2: Now this also fixes #2574**
This fixed an issue with a StackOverflow exception being  generated when a virtual method is being called, but a class in the middle of the inheritance tree does not override the method.

This happens because the CallVirtual method calls the method on the python object if the wrapper got generated, causing everything to be evaluated again from the top. Doing it this way, causes the the python method only to get invoked if there is an implementation on the class, which can then correctly call `super().Method()`.

For the finalizer fix, we now skip generating the call to PyFinalize if the base class is already a python class, because if that is the case the finalize method has already been added.

Additionally it simplifies how the generated code for virtual methods look. Now, a virtual wrapper method is not automatically generated. Only if the method is actually being overwritten by the class it will prompt the generation of a virtual method wrapper.

This closes #2574, #2627, #2629

### Any other comments?

...

### Checklist

### DIFF
--- a/src/runtime/Types/ClassDerived.cs
+++ b/src/runtime/Types/ClassDerived.cs
@@ -198,12 +198,15 @@ namespace Python.Runtime
 
             // Override any properties explicitly overridden in python
             var pyProperties = new HashSet<string>();
+            var dictKeys = new HashSet<string>();
             if (py_dict != null && Runtime.PyDict_Check(py_dict))
             {
                 using var dict = new PyDict(py_dict);
                 using var keys = dict.Keys();
                 foreach (PyObject pyKey in keys)
                 {
+                    var keyString = pyKey.As<string>();
+                    dictKeys.Add(keyString);
                     using var value = dict[pyKey];
                     if (value.HasAttr("_clr_property_type_"))
                     {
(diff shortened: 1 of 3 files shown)


filmor:
Thanks a lot for this. I will try my best to get the CI working again on the weekend and will "enqueue" this one for merging.

rmadsen-ks:
I have added the actual fix for #2574.

In this case, it did not find the correct method to call the base class variant of the virtual method called. Instead it decided to call the top function, causing a stack overflow exception. 

This has been fixed by using the correct name for the base class function invoke method

I think this is ready for merge, but I will continue testing it.

PR #2637, opened 2025-10-21
PyConfig initialization for venv-support

filmor:


filmor:
We'll instead use PyInitConfig from 3.14 onwards.

PR #2665, opened 2025-12-05
Fix clr_loader version parsing for .NET 10.*

zhangys2:
The `clr_loader` library has a bug where it parses .NET versions using string slicing `[:3]`. This works for single-digit major versions (e.g., `9.0.0` becomes `9.0`), but fails for double-digit major versions like `10.0.0`, resulting in an invalid TFM `net10.` and version `10..0`.

This change adds a monkeypatch in `pythonnet/__init__.py` to override the `tfm` and `floor_version` properties of `DotnetCoreRuntimeSpec` with logic that correctly handles double-digit versions by splitting the version string. This ensures support for .NET 10+.

### DIFF
--- a/pythonnet/__init__.py
+++ b/pythonnet/__init__.py
@@ -4,6 +4,22 @@ import sys
 from pathlib import Path
 from typing import Dict, Optional, Union, Any
 import clr_loader
+from clr_loader.util.runtime_spec import DotnetCoreRuntimeSpec
+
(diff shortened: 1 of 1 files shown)


filmor:
1. [`clr_loader`](https://github.com/pythonnet/clr-loader) is part of the same org. If there is something to fix, fix it in there directly, please.
2. In fact, this has already been patched in 0.2.8, are you experiencing issues with that version?

PR #2669, opened 2025-12-09
CI Improvements

filmor:
Notes on Mono on macOS: The active fork at winehq *should* work on arm, but there are no binaries, yet. I don't intend to change that for the CI. Thus, we will only run .NET Core tests on macos arm for now. Alternatively, we use Homebrew, but we should extend it with some caching and set an env variable s.t. clr-loader can find it.

/edit: Still no binaries outside of homebrew, so we will go with that instead.


filmor:
Merging as is, will follow up on the remaining issues in a separate PR.

Issue #2222, opened 2023-08-16
Cannot import pandas in pythonnet


muler415:
### Environment

- Pythonnet version: 3.0.1
- Python version: 3.8
- Operating System: win11
- .NET Runtime: 4.6.1

### Details

-   Describe what you were trying to get done.

  my python script only have one line - 
  
```python
import pandas as pd
```
when I exec and exception is thrown 
partially initialized module 'pyarrow' has no attribute '__version__' (most likely due to a circular import);

please help.

Issue #2205, opened 2023-08-01
Initialize() crash from .NET Framework application


anovik:
### Environment

-   Pythonnet version: 3.0.1
-   Python version: 3.10
-   Operating System: windows 11/10
-   .NET Runtime: NET Framework 4.6.1

### Details

I am trying to call Python code from .NET Framework app but on some machines PythonEngine.Initialize() crashes.

```
Runtime.PythonDLL = "C:\\Python\\Python310\\python310.dll";
PythonEngine.Initialize();
```

Exception:

System.TypeInitializationException: 'The type initializer for 'Delegates' threw an exception.'

There are two inner exceptions:

DllNotFoundException: Could not load C:\Python\Python310\python310.dll.
Win32Exception: %1 is not a valid Win32 application

The path is correct and the dll exists.

![image](https://github.com/pythonnet/pythonnet/assets/8667411/76731411-e0ee-4975-b868-1a4e597cb8f7)

filmor:
.NET Framework 4.6 is not supported, use at least 4.7.2, please. Even if it was, the issue hints that your .NET Framework application is compiled for a different bitness (.NET defaults to 32bit).

PR #2701, opened 2026-03-27
Support .NET Framework 4.6.1

Metadorius:

- Bundles all required by `net461` assemblies
- Adds an AssemblyResolver to resolve them, because something prevents that and `.config` is not really accessible
- Adds a PyInstaller hook to copy all those files properly, overshadowing the contrib repo hook which only copied `Python.Runtime.dll`

Closes #2695 

### Any other comments?

Requires merge of https://github.com/pythonnet/clr-loader/pull/114

### Checklist


Metadorius:
@filmor tested this with my end users, it indeed works just fine

Metadorius:
Are any tests expected for this PR?

Metadorius:
My bad, I should've checked the behavior on .NET Core too. @filmor could you please approve the workflows to run again?

Metadorius:
Bump, just need tests to be approved to run.

Metadorius:
Is there any way I can help speed up the review/merge?

Metadorius:
@filmor sorry for bothering, any ETA on the merge?

filmor:
Two issues:

1. The new csproj is missing from the manifest, I'll fix that.
2. This increases the wheel size substantially. Seeing that you seem to be the only one who needs this, would it work for you if our distributed wheels were still built directly and you could build your own wheel with an environment variable for internal distribution? Do you even need the wheels?

Metadorius:
1. Thanks for the help, overlooked this one.
2. The wheels are a preferred choice. If it's a critical thing, I can of course point `uv` to git, however, it's more trouble than just pre-built wheels, so I'd be happy if it was available on pypi. For example, `uv` caching caused issues with tag-based versioning when I was building the packages from source and did a partial fork, which cost me some hours figuring out what the issue was. I am sure there are other hidden nuances like this one.

   (I was sure that there's an ability to ship "variants" of a wheel, but turns out it was just extra packages via `package[extra]` syntax. Not sure if it's possible to have somehting like `pythonnet[net461]`, it looks like not).

   #2205 #2221 #2222 #2425 seem all to be mentioning 4.6.1 in environment and are past 3.0. Likely not every user submits an issue, yet alone a PR to fix the issue, so I don't think I am the only one; there are probably other users who would benefit from this.

   As for the size increase - I checked, and the size seems to be increased from 0.2 MB to 1.6 MB. I feel like on an absolute scale even after increase that's negligible in 2026. Popular packages like numpy, pandas, cryptography are all multple megabytes worth, like 2-10 MB, and I've never heard this to be a problem.

filmor:
Well, I don't want to join the large-package-size cohort if not necessary. Wheel variants don't exist, yet (there is an open PEP on this), otherwise that would have been the preferred way.

I think I will simply add a Windows wheel build that includes the .NET 4.6, then at least we don't carry all of the shim DLLs around on other platforms.

@lostmsu, what is your take on this?

Metadorius:
I feel like I am missing something, is 1.6 MB really that large? to me it looks like an increase from microscopic to just small 🤔

That works too, of course, just wondering if there's really a point in keeping it separate.

doc/source/python.rst
```diff
@@ -45,7 +45,8 @@ Mono (``mono``)
 
 .NET Framework (``netfx``)
     Default on Windows and also only supported there. Must be at least version
-    4.6.1, with 4.7.2 or later recommended.
+    4.6.1, with 4.7.2 or later recommended. For .NET 4.6 support, the wheel has
+    to be built with the environment variable `PYTHONNET_BUILD_NET46_SUPPORT=1`.
```
Metadorius, line 49: seems to be outdated now, judging by `uv build --wheel -C="--global-option=--net46-support"` below

Issue #72, opened 2015-04-16
Add support for DLR types


filmor:
Using [this](http://stackoverflow.com/questions/5306018/how-to-call-dynamicobject-trygetmember-directly) it's possible to implement all required functionality in a separate DLR meta-class. I've already tried to do this in Python but sadly the .NET types (or their representation in Python-space) are not patchable.

filmor:
Indeed I have now found a way around this, check out this gist: https://gist.github.com/filmor/58f5099f96570c64b1d3

The implementation of the member access is done by basically translating the code that the C# compiler produces when accessing dynamic properties to Python.

Note: Strictly speaking this does *not* implement full support, as it only helps classes that derive from `DynamicObject`, not those that only implement `IDynamicMetaObjectProvider`.

denfromufa:
@filmor what is the use case for exposing DLR types in python?

filmor:
My particular use case is a configuration library that allows for constructs like:

    string proxy = c.proxy.http;

where `c` is a dynamic object of which the members are implicitly created by reading from the config. It allows us to pass partial configuration around easily and reduces string-littering.

denfromufa:
@filmor  can this be used to access dynamic attributes of COM objects? if yes, can you provide minimal example?

denfromufa:
@filmor here is my first attempt at using com objects from pythonnet:

https://gist.github.com/denfromufa/ec559b5af41060c5ac318f7f59d8b415#file-excel_interop_vsto-ipynb

denfromufa:
@filmor how would you add support for IDynamicMetaObjectProvider?

denfromufa:
This looks really powerful library:

https://github.com/ekonbenefits/dynamitey/wiki/UsageReallyLateBinding

PR #2706, opened 2026-04-18
Implement support for DLR get/set

filmor:
Implements #72 and should thus fix #2696.

@spkl Could you try out this branch?

### DIFF
--- a/src/runtime/Types/ClassBase.cs
+++ b/src/runtime/Types/ClassBase.cs
@@ -2,6 +2,7 @@ using System;
 using System.Collections;
 using System.Collections.Generic;
 using System.Diagnostics;
+using System.Dynamic;
 using System.Linq;
 using System.Reflection;
 using System.Runtime.InteropServices;
@@ -22,6 +23,10 @@ namespace Python.Runtime
     [Serializable]
     internal class ClassBase : ManagedType, IDeserializationCallback
     {
+        static readonly DynamicObjectMemberAccessor dynamicMemberAccessor = new();
+
+        internal static void Reset() => dynamicMemberAccessor.Clear();
+
         [NonSerialized]
         internal List<string> dotNetMembers = new();
         internal Indexer? indexer;
@@ -603,6 +608,105 @@ namespace Python.Runtime
             => type.GetMethods(BindingFlags.Public | BindingFlags.Instance)
                 .Where(m => m.Name == "__call__");
 
+        static NewReference tp_getattro_dlr(BorrowedReference ob, BorrowedReference key)
+        {
+            var attr = Runtime.PyObject_GenericGetAttr(ob, key);
+            if (!attr.IsNull())
+            {
+                return attr;
+            }
+
+            // Only run the DLR binder if the error was AttributeError, otherwise preserve the original error
+            if (Runtime.PyErr_ExceptionMatches(Exceptions.AttributeError) == 0)
+            {
+                return default;
+            }
+
+            if (!Runtime.PyString_Check(key))
+            {
+                return default;
+            }
+
+            if (GetManagedObject(ob) is not CLRObject co)
+            {
+                return default;
+            }
+
+            // Slot registration already guarantees this type supports DLR
+            var dynamicObject = (IDynamicMetaObjectProvider)co.inst;
+
+            string? memberName = Runtime.GetManagedString(key);
+            if (memberName is null)
+            {
+                return default;
+            }
+
+            if (!dynamicMemberAccessor.TryGetMember(dynamicObject, memberName, out object? value))
+            {
+                return default;
+            }
+
+            // Clear the lingering AttributeError
+            Runtime.PyErr_Clear();
+
+            using var pyValue = value.ToPython();
+            return pyValue.NewReferenceOrNull();
+        }
+
+        static int tp_setattro_dlr(BorrowedReference ob, BorrowedReference key, BorrowedReference val)
+        {
+            int result = Runtime.PyObject_GenericSetAttr(ob, key, val);
+            if (result == 0)
+            {
+                return 0;
+            }
+
+            // Preserve non-attribute errors exactly as they are.
+            if (Runtime.PyErr_ExceptionMatches(Exceptions.AttributeError) == 0)
+            {
+                return -1;
+            }
+
+            // Deletion fallback is intentionally not handled by DLR binder yet.
+            if (val == null)
+            {
+                return -1;
+            }
+
+            if (!Runtime.PyString_Check(key))
+            {
+                return -1;
+            }
+
+            if (GetManagedObject(ob) is not CLRObject co)
+            {
+                return -1;
+            }
+
+            // Slot registration already guarantees this type supports DLR.
+            var dynamicObject = (IDynamicMetaObjectProvider)co.inst;
+
+            string? memberName = Runtime.GetManagedString(key);
+            if (memberName is null)
+            {
+                return -1;
+            }
+
+            if (!Converter.ToManaged(val, typeof(object), out object? managedValue, true))
+            {
+                return -1;
+            }
+
+            if (!dynamicMemberAccessor.TrySetMember(dynamicObject, memberName, managedValue))
+            {
+                return -1;
+            }
+
+            // Clear the lingering AttributeError
+            Runtime.PyErr_Clear();
+            return 0;
+        }
+
         public virtual void InitializeSlots(BorrowedReference pyType, SlotsHolder slotsHolder)
         {
             if (!this.type.Valid) return;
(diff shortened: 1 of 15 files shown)


spkl:
@filmor If I understand this correctly, one wouldn't even need the dynamic.py script anymore for DynamicObject access to work, correct? That seems great :)
To test this in our setup, can you point me to a wheel that I can install, or provide documentation how to build it?

filmor:
@spkl Yes, the dynamic support would be integrated. To build a wheel, the simplest way is to get `uv` and the .NET SDK installed on your machine, check out the repo and run `uv build`. That builds the wheel in `dist/`.

spkl:
OK so I did a little digging and found out that I can `pip install .` in the root directory of the pythonnet repository to install this version into my venv.
Edit: Is this wrong? Should I try again with uv?

Getting the value of dynamic properties seems to work, but I encountered a problem when setting dynamic properties. This is what I observed when running a script like this:
```python
x = obj.MyProp
obj.MyProp = None
y = obj.MyProp
```
For line 1, `TryGetMember` is called for name "MyProp" and `x` will correctly be assigned the value returned from there.
For line 2, `TrySetMember` is NOT called.
For line 3, `TryGetMember` is NOT called, `y` will be assigned the value from line 2 (`None`), but that value is unknown to the .NET implementation.

src/runtime/Types/ClassBase.cs
```diff
@@ -603,6 +608,105 @@ static IEnumerable<MethodInfo> GetCallImplementations(Type type)
...
+                return default;
+            }
+
+            // Slot registration already guarantees this type supports DLR
+            var dynamicObject = (IDynamicMetaObjectProvider)co.inst;
+
+            string? memberName = Runtime.GetManagedString(key);
+            if (memberName is null)
+            {
+                return default;
```
lostmsu, line 641: These return defaults are a little concerning. We are basically hiding errors (which I am not sure can even happen in the first place).

src/runtime/Types/ClassBase.cs
```diff
@@ -603,6 +608,105 @@ static IEnumerable<MethodInfo> GetCallImplementations(Type type)
...
+                return 0;
+            }
+
+            // Preserve non-attribute errors exactly as they are.
+            if (Runtime.PyErr_ExceptionMatches(Exceptions.AttributeError) == 0)
+            {
+                return -1;
+            }
+
+            // Deletion fallback is intentionally not handled by DLR binder yet.
```
lostmsu, line 670: "yet" here means we can't fix it until 4.0 as it would be breaking change

src/runtime/Types/ClassBase.cs
```diff
@@ -603,6 +608,105 @@ static IEnumerable<MethodInfo> GetCallImplementations(Type type)
...
+            if (val == null)
```
lostmsu, line 671: I don't remember if BorrowedReference == null is implemented. It might be better to explicitly use IsNull

src/runtime/Types/ClassBase.cs
```diff
@@ -603,6 +608,105 @@ static IEnumerable<MethodInfo> GetCallImplementations(Type type)
...
+            }
+
+            // Slot registration already guarantees this type supports DLR.
+            var dynamicObject = (IDynamicMetaObjectProvider)co.inst;
+
+            string? memberName = Runtime.GetManagedString(key);
+            if (memberName is null)
+            {
+                return -1;
+            }
```
lostmsu, line 693: Same as above, silent error masking

src/runtime/Types/ClassBase.cs
```diff
@@ -612,6 +716,12 @@ public virtual void InitializeSlots(BorrowedReference pyType, SlotsHolder slotsH
                 TypeManager.InitializeSlotIfEmpty(pyType, TypeOffset.tp_call, new Interop.BBB_N(tp_call_impl), slotsHolder);
             }
 
+            if (typeof(IDynamicMetaObjectProvider).IsAssignableFrom(this.type.Value))
+            {
+                TypeManager.InitializeSlotIfEmpty(pyType, TypeOffset.tp_getattro, new Interop.BB_N(tp_getattro_dlr), slotsHolder);
+                TypeManager.InitializeSlotIfEmpty(pyType, TypeOffset.tp_setattro, new Interop.BBB_I32(tp_setattro_dlr), slotsHolder);
```
lostmsu, line 722: Are there any conflicts possible?

src/runtime/Types/ClassBase.cs
filmor: We are not. If we arrive at any of these, we have already gotten an error from the initial Python getattr. Since we are not clearing it until the very end, this is what Python is going to see. I'll add some more tests for that.

src/runtime/Types/ClassBase.cs
filmor: TBH, I'm seriously leaning towards just striking out our claim that we do semantic versioning. We have _two_ sets of behaviours that we "have" to keep stable, and it blocks _so many_ improvements we could do in the name of eternal compatibility with fringe use-cases. DLR is not supported _at all_ right now from the Python side, so with this line of argument, we can't proceed with this PR either without bumping the major version.

src/runtime/Types/ClassBase.cs
filmor: What kind of conflicts? Either `tp_g/setattro` are set before-hand, then we leave them alone (but don't support DLR member access) or they are not and we get the "new" behaviour.

src/runtime/Types/ClassBase.cs
lostmsu: There's no need to remove semantic versioning claim. If you want this, we can just bump the major version.

src/runtime/Types/ClassBase.cs
filmor: Switched to `.IsNull`

filmor:
@spkl Could you test again whether this works for you?

src/runtime/Types/ClassBase.cs
filmor: Yeah, I don't want to bump the major version all the time. Also, on this branch in particular, all code would still work unchanged.

The _only_ possible breakage I see is

- deriving in Python from a .NET DLR-enabled class and then relying on all properties to still live purely in Python (not sure why one would ever want to do that)
- relying on non-existence of properties of a DLR object in Python that were set in .NET

The changes I had to do to enable 3.14 support are likely much more impactful than this... (see breaking compatibility with the older DLR hack).

spkl:
Will do ASAP and let you know

spkl:
OK @filmor, this is pretty great. With the current version from this branch and with all previous workarounds removed, all of our test cases pass. 🎉

Performance is also improved, which I could see in our two dedicated performance tests. These tests don't strictly measure pythonnet performance, but that is a significant part of them. Numbers are normalized to fractions, lower is better.

|Variant|Test 1|Test 2|
|-|-|-|
|Python 3.12, pythonnet 3.0.5 + our own adaptation of filmors old DynamicObject workaround|1.0|1.0|
|Python 3.14, pythonnet 3.1.0rc0 + our own (problematic) workaround using codecs|1.18|1.04|
|Python 3.14, pythonnet with DLR support (this branch)|0.77|0.96|

There is one significant behavioral difference that I observed: With past solutions, there was always an AttributeError when setting an attribute for which TrySetMember returns false (= which does not exist). With this branch, the attribute is now set on the Python side without any error. This can be problematic, because it can mask errors / typos in scripts and make troubleshooting harder. I don't know if this was intended. If it is, we can probably find a different solution.

I tried to find a workaround for that behavior and discovered something else where I'm not sure if it's correct: When throwing an exception in TrySetMember, the process crashes with a .NET exception, so there is no Python traceback to pinpoint the error location. In contrast, when throwing an exception in Try**Get**Member, it comes up as an AttributeError with Python traceback output.

Because we support running in different environments and Python versions, we probably need to implement automatic switching between DynamicObject workaround and first-class DLR support in pythonnet. Is there a way to find out if the currently loaded pythonnet module has DLR support? An information about the version would also work, but pythonnet currently has no `__version__` attribute.

filmor:
I'll try to handle these cases, definitely not intentional :)

If you need to get a package's version in your code, please use `importlib.metadata.version()`.

greateggsgreg:
#2718 should fix the `TryGetMember` issue highlighted above.

alexchandel:
You want to rebase this and retest with the `TryGetMember` fix?

filmor:
@spkl Can you please test out the latest version?
@greateggsgreg I stole your pending commit that includes the caching, I hope that's fine :)
@lostmsu One last round of review? Once this is through, I would merge this and publish it as 3.1.0-rc2, we can discuss on a separate issue (after I update the Changelog to reflect the actual changes) whether we want to release it as 4.0. I don't want to delay the release any further :)

spkl:
@filmor All good from my side 👍

PR #2719, opened 2026-05-09
Fix MethodBinding/OverloadMapper memory leak (#691)

### LINKED ISSUE #691: Memory Leak on MethodBinding for generic method

greateggsgreg:
Fixes #691.

## Cause

`MethodBinding` and `OverloadMapper` hold a `PyObject target` but didn't release it on `tp_clear`, so the underlying CLR instance waited on the .NET finalizer chain to drop the refcount. They also shared the same C# `PyObject` instance across `mp_subscript` / `Overloads` paths, so disposing one wrapper corrupted the others.

## Fix

- `ExtensionType`: add virtual `OnClear()` hook called from `tp_clear`.
- `MethodBinding` / `OverloadMapper`: override `OnClear` to dispose `target`. (`targetType` left alone — disposing it broke unrelated subclass tests.)
- Each sharing site now passes `new PyObject(self.target.Reference)` so each wrapper owns its own INCREF'd reference.

## Tests

The three existing `*_does_not_leak_memory` tests cover the three sharing sites but their 0.9 MB/iter threshold was too loose — master was leaking ~600 KB/iter and still passing. Tightened to 0.1 MB/iter (104 KB).

## Verification (Python 3.14 GIL, linux-aarch64)

| | .NET 8 | .NET 10 |
|---|---|---|
| master | **FAIL** 765 KB/iter | **FAIL** 572-647 KB/iter |
| this PR | PASS -0.6 KB/iter | PASS -0.5 KB/iter |

### DIFF
--- a/src/runtime/Types/ExtensionType.cs
+++ b/src/runtime/Types/ExtensionType.cs
@@ -84,8 +84,18 @@ namespace Python.Runtime
             DecrefTypeAndFree(lastRef.Steal());
         }
 
+        /// <summary>
+        /// Called during tp_clear before the GCHandle is released.
+        /// Override to eagerly dispose Python object references (PyObject fields)
+        /// held by the subclass, preventing the multi-hop .NET finalizer chain
+        /// from delaying Python-side refcount decrements.
+        /// </summary>
+        protected virtual void OnClear() { }
+
         public static int tp_clear(BorrowedReference ob)
         {
+            (GetManagedObject(ob) as ExtensionType)?.OnClear();
+
             var weakrefs = Runtime.PyObject_GetWeakRefList(ob);
             if (weakrefs != null)
             {
--- a/src/runtime/PythonTypes/PyObject.cs
+++ b/src/runtime/PythonTypes/PyObject.cs
@@ -93,6 +93,12 @@ namespace Python.Runtime
             Finalizer.Instance.ThrottledCollect();
         }
 
+        /// <summary>
+        /// Create a new PyObject instance of this object, bumping the reference
+        /// count.
+        /// </summary>
+        public PyObject NewReference() => new(this);
+
         // Ensure that encapsulated Python object is decref'ed appropriately
         // when the managed wrapper is garbage-collected.
         ~PyObject()
--- a/src/runtime/PythonTypes/PyType.cs
+++ b/src/runtime/PythonTypes/PyType.cs
@@ -35,6 +35,12 @@ namespace Python.Runtime
                 throw new ArgumentException("object is not a type");
         }
 
+        /// <summary>
+        /// Create a new PyType instance of this object, bumping the reference
+        /// count.
+        /// </summary>
+        public new PyType NewReference() => new(this);
+
         protected PyType(SerializationInfo info, StreamingContext context) : base(info, context) { }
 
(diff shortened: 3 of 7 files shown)


lostmsu:
If the objects are not shared anymore and are always owned by `MethodBinding` the `NewReference` is a better type to hold it than `PyObject`.

src/runtime/Types/ExtensionType.cs
```diff
@@ -84,8 +84,18 @@ public unsafe static void tp_dealloc(NewReference lastRef)
...
         }
 
+        /// <summary>
+        /// Called during tp_clear before the GCHandle is released.
+        /// Override to eagerly dispose Python object references (PyObject fields)
+        /// held by the subclass, preventing the multi-hop .NET finalizer chain
+        /// from delaying Python-side refcount decrements.
+        /// </summary>
+        protected virtual void OnClear() { }
+
```
lostmsu, line 94: Is there a reason to not have `ExtensionType` be `IDisposable` instead of exposing `OnClear`?

greateggsgreg: I think `tp_clear` and `Dispose` have different contracts. `tp_clear` releases references to break GC cycles but the Python object can still be reachable afterward, whereas `Dispose` should mean that the instance is dead. `ExtensionType` instances are owned by Python's GC, not by .NET callers — there's no `using` site and no one calls `Dispose()` on them. `OnClear` makes it clear that it's a hook fired from `tp_clear`.

greateggsgreg:
`NewReference` is declared as a `ref struct`, so it can't be a class field. The closest owned type is `PyObject`. Unless I'm missing something? We'd also have to swap out a few of the callsites in MethodBinding that expect a `PyObject`

Ownership is still explicit: `OnClear` disposes it, the field has no external readers, and every sharing site INCREFs into its own Python object.

filmor:
@greateggsgreg I took the liberty of pushing my notes as a commit :)

- There was one missing location for the ownership transfer of the passed types
- Now that this is solved, we can `Dispose` the type object as well
- I introduced helper functions for the refcount bump

If these are fine with you, I'd go ahead and merge this once the CI is through.

greateggsgreg:
Awesome, looks good. Thank you!

Issue #2720, opened 2026-05-09
Python 3.14 free-threaded support


greateggsgreg:
Tracking issue for free-threaded Python (`Py_GIL_DISABLED`) support. Companion to the umbrella 3.14 tracker #2610.

## Why FT is different from a normal GIL build

- `PyObject_HEAD` is 16 bytes larger; the refcount is split across `ob_ref_local` + `ob_ref_shared` and must be read via `Py_REFCNT` (a real exported symbol on 3.14+).
- Build detection: `sys._is_gil_enabled()` (added in 3.13) returns `False` on FT.
- The GIL has been silently serialising every pythonnet static cache, lock-free counter, finalizer-thread interaction, and `Reflection.Emit` use site. All of those need explicit synchronisation under FT.

## Hazards in pythonnet (resolved by PR #2721)

| Area | What broke | How it was fixed |
|---|---|---|
| Refcount / ABI | Single-offset read | `Py_REFCNT` P/Invoke + `ObjectHeadOffset = 16` for FT |
| Type-creation race | Duplicate `cache.Add`; partial-type visibility | Two-cache (`cache` + `_inProgressCache`) under `_cacheCreateLock` |
| GCHandle in `tp_clear`/`tp_dealloc` | Double-free under main-thread + finalizer-thread race | `Interlocked.Exchange` on the slot |
| `Reflection.Emit` | Concurrent `DefineType` corrupts IL / throws "Duplicate type name" | Lock both `CreateDerivedType` and `DelegateManager.GetDispatcher` |
| Static collections | Plain `Dictionary`/`HashSet`/`List` thread-safety issues | `ConcurrentDictionary` where possible; `Interlocked`/`Volatile` for single-cell state; locks for nested mutation and ordered list semantics |
| Finalizer-thread / `Py_Finalize` interaction | Stale `ob_ref_local` reads after teardown crash the process | `Runtime._Py_IsFinalizing()` guards on all decref-from-finalizer paths |
| Chained `ClassDerived.Finalize` IL | Nested Python-derived classes with `__namespace__` chained through emitted `Finalize` methods, queuing the same `__pyobj__` twice → `PyObject_GC_Del` on freed memory | Skip the base-Finalize chain when the base is itself pythonnet-emitted (`IPythonDerivedType`) |

See PR #2721 for the per-file detail.

## References

- PR: #2721
- Umbrella 3.14 tracker: #2610
- CPython PEP 703: https://peps.python.org/pep-0703/
[...]

Issue #2610, opened 2025-08-05
Add support for Python 3.14


freakboy3742:
Python.net's `requires-python` places a hard upper pin on `<3.14`. 

Python 3.14 is nearing release; RC1 is out, which indicates that ABI stability now exists.

It would be desirable for Python.net to support the upcoming Python release.

EliaOndacs:
hey, hello, as of right now i'm writing this python 3.14 released, and i unexpectedly ran into this problem of pythonnet not supporting python 3.14, so i'm running into this problem too, (runtime error while trying to run a python script that is using pythonnet), i wonder how long it takes to update it for python 3.14 considering its release-candidates and pre-releases have been out for a while

filmor:
Any help is appreciated. There is a branch you can test (https://github.com/pythonnet/pythonnet/pull/2611), but our CI is currently thoroughly broken and I haven't found the time to fix it, yet.

freakboy3742:
@filmor I'm unfortunately not in a position to help directly - but if this is a situation that can be fixed by throwing money at someone who *does* have the skills to investigate the problem, I *might* be able to help. If that's the case, let me know.

NasirNS45:
Any update on this? Ran into the same issue while installing awsebcli

filmor:
Work is happening in #2611. It works in general, but there are at least two crashes that I (or someone who wants to dive in) have to investigate.

andre-romano:
Pythonnet building is crashing in Windows env for Python 3.13.9 as well (similarly to version 3.14)

filmor:
It is not.

1. Please don't cross post. If you have an issue, open one, don't comment in two unrelated threads.
2. Please add more data than "does not work". We are well aware that compiling with .NET >6 is problematic, and there is work going on to fix this (or rather, it is fixed on master but not all tests pass). It is unnecessary to compile Python.NET if you don't want to set up the outdated .NET 6 SDK, you can just use the precompiled Nuget or wheel packages.

Houbein:
Installing pythonnet under Python 3.14 now just installs 3.0.0rc6, the last version not to enforce a max Python version. 
This happens quietly, not informing the user that they are installing a very old version. 
Please make it a priority to fix this.

filmor:
3.1.0rc0 is out now with the Python 3.14 support merged. Please test and report back. I still have not found time to fix the CI completely, which is a prerequisite for me to cut this as a full release.

freakboy3742:
[...]

PR #2721, opened 2026-05-09
Python 3.14 free-threaded support

greateggsgreg:
First step toward Python 3.14 free-threaded (`Py_GIL_DISABLED`) support, tracked under #2720 (and the umbrella #2610).

After this PR, pythonnet runs the full pytest suite on `python3.14t` against both .NET 8 and .NET 10 on Linux. Existing GIL builds (3.11 - 3.14) are unaffected.

### Free-threaded `PyObject` layout + refcount

FT `PyObject` is 16 bytes longer (`ob_tid` + `ob_flags` + `ob_mutex` + `ob_gc_bits` + `ob_ref_local` + `ob_ref_shared` replaces `ob_refcnt`) and the refcount is split, so there is no single offset to read.

- `ABI.Initialize` detects the build via `sys._is_gil_enabled()` and sets `ObjectHeadOffset = 16` on FT, shifting the generated `TypeOffset*` values to absolute `PyHeapTypeObject` offsets.
- `Runtime.Refcount` prefers a P/Invoke into `Py_REFCNT` (a real symbol on CPython 3.14+, required on FT). Older Pythons fall back to the existing direct read.

### Atomic type creation in `ReflectedClrType.GetOrCreate` / `TypeManager.GetType`

The classic check-then-act on `ClassManager.cache` and `TypeManager.cache` raced two ways under FT: duplicate `cache.Add` (`ArgumentException`), and partial-type visibility from outside-the-lock readers observing a half-initialised type during recursive class init. Two-cache design:

- `cache` — only fully-initialised types; outside-the-lock fast path is safe.
- `_inProgressCache` — partial types visible only to the lock-holding builder, so self-referential definitions resolve via the reentrant lock.

Serialisation snapshots remain `Dictionary<,>` on the wire for binary compatibility.

### Atomic GCHandle ownership in `tp_clear` / `tp_dealloc`

`subtype_clear` (main thread) and the .NET finalizer thread can both reach a CLR-bound Python object's GCHandle slot. Previous read-then-zero in `ManagedType.TryFreeGCHandle` and `ClassDerivedObject.tp_dealloc` (strong→weak swap) let two threads observe the same handle and double-free it. Both paths now use `Interlocked.Exchange` to atomically claim the slot.

### `Reflection.Emit` is not thread-safe

Concurrent `TypeBuilder` / `ModuleBuilder` operations corrupt the IL stream or throw `"Duplicate type name within an assembly"`. Two emit-and-bake sequences are now serialised:

- `ClassDerivedObject.CreateDerivedType` (Python subclasses of CLR types) on `_buildersLock`.
- `DelegateManager.GetDispatcher` (Python callables → CLR delegates) on a new `_emitLock`.

### Concurrent collections / atomics on shared state

Plain `Dictionary<,>`, `HashSet<>`, and counters that the GIL implicitly serialised tear under FT.

- `ConcurrentDictionary<,>`: `ExtensionType.loadedExtensions`, `CLRObject.reflectedObjects`, `ClassManager.cache`, `TypeManager.cache`, `ModuleObject.cache` / `allNames`, `Interop.delegateTypes`, `ClassBase.ClearVisited`, `InternString._string2interns` / `_intern2strings`.
- `Interlocked` + `Volatile`: `Finalizer._throttled` (counter), `Runtime.run` (epoch), `PyBuffer.disposedValue` (dispose flag — finalizer + explicit `Dispose` no longer double-free `_view`).
- Lock-protected (where nested mutation or list-with-equality semantics rule out lock-free containers): `GenericUtil.mapping` (nested `Dictionary`/`List`), `PythonEngine.ShutdownHandlers`, `Runtime._pyRefs`.
- `volatile bool`: `Runtime._isInitialized` / `_typesInitialized`, `PythonEngine.initialized`, `Finalizer.started`.

### Guard the .NET-finalizer / Python-shutdown interaction

The .NET finalizer thread can dispatch `Py_DecRef` concurrently with `Py_Finalize`; a stale `ob_ref_local` read after teardown crashes the process. `Runtime._Py_IsFinalizing()` guards on `Finalizer.ThrottledCollect`, `PyObject`'s finalizer, and the `Refcount > 0` `Debug.Assert` in `XDecref` / `AddFinalizedObject` (kept on GIL, skipped on FT).

### Tests + CI

New tests in `tests/test_thread.py` cover the refcount/ABI invariants, the `ConcurrentDictionary` caches, atomic GCHandle ownership, the two `Reflection.Emit` locks, and the `ShutdownHandlers` lock. Some tests are marked  `@freethreaded_only` because the GIL-build code path triggers a separate pre-existing pythonnet crash under high-contention CLR allocation that's reproducible on master and out of scope for this PR.

`.github/workflows/main.yml` adds `3.14t` to the Linux/macOS matrix and skips Mono on 3.14t (`clr-loader`'s mono backend isn't yet validated against free-threaded Python).

## Verification

Linux aarch64, full pytest suite (`tests/`, `--ignore=tests/domain_tests`):

| | .NET 8 | .NET 10 |
|---|---|---|
| 3.11 - 3.14 | 478 passed, 11 skipped | 478 passed, 11 skipped |
| 3.14t       | 484 passed, 5 skipped  | 484 passed, 5 skipped  |

(`11 skipped` on GIL = 5 pre-existing + 6 FT-only tests; `5 skipped` on FT = the 5 pre-existing skips.)

3.13t is not in the matrix because `cffi <2.0` does not support free-threaded 3.13 and `clr-loader` requires `cffi`.


filmor:
Thanks a lot for your work here. Just as a note: Since I would like to finally get a release out after the DLR PR is solved, I will not be able to review this in the next week.

PR #2722, opened 2026-05-11
Alternative attempt at fixing memleak by being explicit about copies

filmor:


src/runtime/PythonTypes/PyObject.cs
```diff
@@ -93,6 +93,8 @@ internal PyObject(in StolenReference reference)
             Finalizer.Instance.ThrottledCollect();
         }
 
+        public PyObject Copy() => new(this);
```
lostmsu: The name of this function is too generic and therefore misleading

filmor: How is it misleading? This is exactly what it does (at least in Python terms). The current behaviour, being able to freely pass the object around, is the misleading thing IMO.

lostmsu: It does not make a copy of an object. It makes a separate living reference to it.

filmor: `.NewReference()`?

